PR #6, opened 2019-12-27
Configuere checksum & homebrew tasks

aalmiray:
This PR has the following updates
- simplify the `createChecksum` task by
  - using FQN for its task type
  - adding explicit dependsOn in task configuration
  - using `distZip.destinationDir` as `outputDir`
- simplify `homebrew` task by
  - adding explicit dependsOn in task configuration
  - define an input.property on the project's version (reduces up-to-date misses)
  - delay read of sha256 file until the very last moment, once filename has been written

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.crypto.checksum.Checksum
+
 plugins {
     id "distribution"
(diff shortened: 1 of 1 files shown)


maxandersen:
It works! excellent! thanks @aalmiray

PR #24, opened 2020-01-16
Adding devfile and 'Developer Workspace' button in readme for trying the project on Hosted Che / che.openshift.io

ibuziuk:
Adding devfile and 'Developer Workspace' button in readme for trying the project on Hosted Che / che.openshift.io

Try the PR on Hosted Che:

[Contribute](https://che.openshift.io/f?url=https://github.com/ibuziuk/jbang/tree/devfile)

jbang

image

### DIFF
--- a/devfile.yaml
+++ b/devfile.yaml
@@ -1,0 +1,40 @@
+---
+apiVersion: 1.0.0
+metadata:
+  name: jbang
+components:
+  -
+    type: chePlugin
+    id: redhat/java/latest
+  -
+    type: dockerimage
+    alias: gradle
+    image: quay.io/eclipse/che-java11-gradle:nightly
+    env:
+      - name: GRADLE_USER_HOME
+        value: /home/gradle/.gradle
+      - name: JAVA_OPTS
+        value: "-XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
+          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90
+          -Dsun.zip.disableMemoryMapping=true -Xms20m -Djava.security.egd=file:/dev/./urandom"
+      - name: JAVA_TOOL_OPTIONS
+        value: "-XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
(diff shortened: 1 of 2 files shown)


devfile.yaml
```diff
@@ -0,0 +1,45 @@
+---
+apiVersion: 1.0.0
+metadata:
+  name: jbang
+projects:
+  -
+    name: jbang
+    source:
+      type: git
+      location: "https://github.com/maxandersen/jbang.git"
```
maxandersen: this means forks wont use their own repo for edits - can we avoid that ?

devfile.yaml
```diff
@@ -0,0 +1,45 @@
...
+    id: redhat/java/latest
+  -
+    type: dockerimage
+    alias: gradle
+    image: quay.io/eclipse/che-java11-gradle:nightly
+    env:
+      - name: GRADLE_USER_HOME
+        value: /home/gradle/.gradle
+      - name: JAVA_OPTS
+        value: "-XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
```
maxandersen: how can I avoid having to specify this in every devfile ? I don't want to maintain this...

devfile.yaml
ibuziuk: I suppose the fork low you describe is not supported
@skabashnyuk  could you please confirm?
But in any case having a "Try it on Hosted Che" button for the original project is the step forward IMO.

devfile.yaml
ibuziuk: I do not think it is possible atm - one need to explicitly specify dedicated env vars

devfile.yaml
ibuziuk: fixed. Starting from 7.8.0 upstream version it is possible not to specify the project

maxandersen: awesome! do you know when this is available on che.openshift.io ?

ibuziuk: @maxandersen it is already available, you can try via the button in the PR description

maxandersen:
sorry for the delay on trying this out (been bandwidth challenged the last month :)

Here is what I get when I click the button in this PR - a 401 error:

maxandersen:
on second click it tells me my account needs activating.

I click activate and now its looking like its booting che.

so something in the auth check flow isn't right as it shouldn't have let me in the first time ;)

maxandersen:
merged this.

PR #28, opened 2020-01-16
Property draft in build.gradle is of type boolean

brunoborges:


maxandersen:
this is already fixed in master afaics - was just there briefly.

PR #33, opened 2020-01-17
Fresh Samples

brunoborges:
Also a better folder structure for the `examples` directory.


build.gradle
```diff
@@ -154,7 +154,7 @@ githubRelease {
 	// set ORG_GRADLE_PROJECT_github_token
 	token = project.hasProperty('github_token') ? getProperty('github_token') : "unknown_github_token"
 	owner = "maxandersen"
-	draft = "true"
+	draft = true
```
maxandersen: possible for you to rebase with current master? this is fixed there.

maxandersen:
thanks for the sample contribution - for now I would like to keep the dir flat to make it easier to import. if we get many more samples then lets look at it.

PR #37, opened 2020-01-20
Update jetty.jsh

brunoborges:


maxandersen:
var makes it only work on java 10...but hey; people should use newer java anyway ;)

brunoborges:
@maxandersen you should set Java 11 as minimum requirement! Afterall, we want people to move to a post-Java8 era. :-)

maxandersen:
well the great thing is that jbang works equally fine on all of these things and just like I made effort to have a package for windows on choco (still pending) I went the extra and made jbang core runs on java 8 so it will run basically everywhere but let you write apps for *any* Java version...:)

PR #42, opened 2020-01-24
Fixing the name of Netbeans to NetBeans

geertjanw:


maxandersen:
that was fast :) Here I thought I had been a good buy and remember to prefix it with Apache.

Live and learn - thanks for the fix.

PR #60, opened 2020-02-19
example using webdriver

aaronwalker:
Very basic Selenium WebDriver example using WebDriverManager https://github.com/bonigarcia/webdrivermanager


maxandersen:
That’s an awesome example - thank you @aaronwalker !

PR #66, opened 2020-02-25
inetTest - measure time taken in ms to resolve localhostname.

johnaohara:
Forked idea from: https://github.com/thoeni/inetTester

example script to measure time to resolve local hostname. Incorrectly configured `/etc/hosts` on Linux can cause slow JVM startup for server frameworks.


maxandersen:
cool - excellent!

Issue #50, opened 2020-02-08
warn users if javac version is higher than java


maxandersen:
[some users](https://twitter.com/mojavelinux/status/1225945104700936192) might have a setup where java is of a lower version than javac meaning they will get a classfile major version error.

We could run java -version / javac -version and just check that java numbers are same or bigger than javac

AtharvaThorve:
I have forked and cloned the repository and I wanted to know a little more about this issue like when should the user be warned about the issue mentioned here

maxandersen:
when we do a build we are anyway spending some seconds to run a build so in the background (to not slow things down) we could run the check.

And just print out that warning if/when mismatch is found.

AtharvaThorve:
after writing my code and running ./gradlew build install i am getting error in spotlessJava 
so i ran gradle spotlessApply but it gives an error stating 'Only Project build scripts can contain plugins {} blocks'

Need some help resolving this issue.

maxandersen:
are you using the gradle wrapper ? i.e. ./gradlew instead of gradle ?  if not then it could be you are picking up older version of gradle ?

AtharvaThorve:
I know this is a dumb doubt but i successfully built the gradle but whatever changes i have made are not being executed whenever i run a program using jbang.
I've been working on it for past couple of hours...need some guidance here

maxandersen:
Do you have the right jbang in your path ? If using locally built one you need the one in build/install folder

AtharvaThorve:
I installed it using sdkman[linux ubuntu].
Its path is /home/atharva/.sdkman/candidates/jbang/current/bin/jbang

AtharvaThorve:
I have created a new pull request which is working fine on my local pc but failed the test...what to do next?

maxandersen:
lets discuss on #50

maxandersen:
closing as complexity not worth the gain.

PR #69, opened 2020-02-29
javac and java version check added

AtharvaThorve:
Issue #50 resolved. Whenever user uses jbang to run a program while the jar file is building the version of java and javac are checked and if they are different then user is warned.

### DIFF
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -307,7 +307,6 @@ public class Main implements Callable<Integer> {
 					}
 				}
 			}
-			// System.out.println(output);
 			String[] jcVersion = output.toString().split(" ");
 			for (String string : jcVersion) {
 				if (string.contains("1")) {
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -263,6 +263,42 @@ public class Main implements Callable<Integer> {
 		}
 	}
 
+	static boolean checkVersions() {
+		String version = System.getProperty("java.version");
+		if (version.startsWith("1.")) {
+			version = version.substring(2, 3);
+		} else {
+			int dot = version.indexOf(".");
+			if (dot != -1) {
+				version = version.substring(0, dot);
+			}
+		}
+		int javaVersion = Integer.parseInt(version);
+		int javacVersion = 0;
+		try {
+			Process p = Runtime.getRuntime().exec("javac -version");
+			StringBuilder output = new StringBuilder();
+			BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+			String line;
+			while ((line = reader.readLine()) != null) {
+				output.append(line + "\n");
+			}
+			String[] s = output.toString().split(" ");
+			if (s[1].startsWith("1.")) {
+				s[1] = s[1].substring(2, 3);
+			} else {
+				int dot = s[1].indexOf(".");
+				if (dot != -1) {
+					s[1] = s[1].substring(0, dot);
+				}
+			}
+			javacVersion = Integer.parseInt(s[1]);
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
+		return (javaVersion == javacVersion);
+	}
+
 	// build with javac and then jar... todo: split up in more testable chunks
 	static void build(Script script, Main m) throws IOException {
 		File baseDir = new File(Settings.getCacheDir(), "jars");
@@ -284,6 +320,9 @@ public class Main implements Callable<Integer> {
 			optionList.addAll(Arrays.asList(script.backingFile.getPath()));
 
 			Util.info("Building jar...");
+			if (!checkVersions()) {
+				Util.warnMsg("Java and Javac versions are different!!");
+			}
 			if (m.verbose)
 				Util.info("compile: " + String.join(" ", optionList));
 			Process process = new ProcessBuilder(optionList).inheritIO().start();
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -296,7 +296,7 @@ public class Main implements Callable<Integer> {
 					output.append(line);
 				}
 			}
-
+			System.out.println(output);
 			String[] jcVersion = output.toString().split(" ");
 			for (String string : jcVersion) {
 				if (string.contains("1")) {
@@ -306,6 +306,7 @@ public class Main implements Callable<Integer> {
 		} catch (IOException e) {
 			Util.warnMsg(e.toString());
 		}
+		System.out.println(javaVersion + " " + javacVersion);
 		return (javaVersion.equals(javacVersion));
 	}
 
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -80,6 +80,9 @@ public class Main implements Callable<Integer> {
 	@Option(names = "--completion", help = true, description = "Output auto-completion script for bash/zsh.\nUsage: source <(jbang --completion)")
 	boolean completionRequested;
 
+	@Option(names = { "-D" }, description = "set a system property")
+	Map<String, String> properties = new HashMap<>();
+
 	public int completion() throws IOException {
 		String script = AutoComplete.bash(
 				spec.name(),
@@ -550,14 +553,19 @@ public class Main implements Callable<Integer> {
 		}
 	}
 
-	String generateArgs(List<String> args) {
+	String generateArgs(List<String> args, Map<String, String> properties) {
 
 		String buf = "String[] args = { " +
 				args.stream()
 					.map(s -> '"' + StringEscapeUtils.escapeJava(s) + '"')
 					.collect(Collectors.joining(", "))
 				+
-				" }";
+				" }" +
+				(properties.isEmpty() ? "" : "\n") +
+				properties	.entrySet()
+							.stream()
+							.map(x -> "System.setProperty(\"" + x.getKey() + "\",\"" + x.getValue() + "\");")
+							.collect(Collectors.joining("\n"));
 		return buf;
 	}
 
@@ -569,6 +577,7 @@ public class Main implements Callable<Integer> {
 
 		String javacmd = resolveInJavaHome("java");
 		if (script.backingFile.getName().endsWith(".jsh")) {
+
 			javacmd = resolveInJavaHome("jshell");
 			if (!classpath.trim().isEmpty()) {
 				optionalArgs.add("--class-path=" + classpath);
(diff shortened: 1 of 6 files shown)


src/main/java/dk/xam/jbang/Main.java
```diff
@@ -260,6 +260,42 @@ public Integer call() throws IOException {
...
+			BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+			String line;
+			while ((line = reader.readLine()) != null) {
+				output.append(line + "\n");
+			}
+			String[] s = output.toString().split(" ");
+			if (s[1].startsWith("1.")) {
+				s[1] = s[1].substring(2, 3);
+			} else {
+				int dot = s[1].indexOf(".");
```
maxandersen: in the checks you can see the following stacktrace:

```
java.lang.ArrayIndexOutOfBoundsException: 1
	at dk.xam.jbang.Main.checkVersions(Main.java:287)
	at dk.xam.jbang.Main.build(Main.java:323)
```

so this check you are doing is not working on whatever version string github action linux is running with.

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -260,6 +260,42 @@ public Integer call() throws IOException {
...
+		} else {
+			int dot = version.indexOf(".");
+			if (dot != -1) {
+				version = version.substring(0, dot);
+			}
+		}
+		int javaVersion = Integer.parseInt(version);
+		int javacVersion = 0;
+		try {
+			Process p = Runtime.getRuntime().exec("javac -version");
```
maxandersen: you are not executing this the same way as the other code in jbang calls out for javac - find that and reuse the logic so you are not actually calling some *other* javac.

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -260,6 +260,42 @@ public Integer call() throws IOException {
 		}
 	}
 
+	static boolean checkVersions() {
+		String version = System.getProperty("java.version");
```
maxandersen: look for supportsModules() - here we are calling out and check if java supports modules by looking at -version string - maybe we can reuse/share that logic so we avoid calling out twice ?

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -260,6 +260,42 @@ public Integer call() throws IOException {
...
+				s[1] = s[1].substring(2, 3);
+			} else {
+				int dot = s[1].indexOf(".");
+				if (dot != -1) {
+					s[1] = s[1].substring(0, dot);
+				}
+			}
+			javacVersion = Integer.parseInt(s[1]);
+		} catch (IOException e) {
+			e.printStackTrace();
```
maxandersen: we should not just print stacktraces - we use info/warn with a e as argument so printing is automatically handled.

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -260,6 +260,42 @@ public Integer call() throws IOException {
...
+		}
+		return (javaVersion == javacVersion);
```
maxandersen: here you are comparing the java version jbang is running - with the javacVersion found in path. That is not the correct comparison.

what needs to be tested is if the java used for building and javac used for compiling are compatible.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: I have changed this part to print e using Util.warnMsg.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: Yes i am going through this part of inheritIO in process builder where javac is called.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: I did not understand this part a bit.
Should i look for modules that let us retrieve version of java being used?

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: In jbang javac is called using the ProcessBuilder object but whenever i use inheritIO method and then start method on it the version of javac is displayed but its value is not readable using a bufferedreader.
Can you tell me what am i doing wrong?

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: I am sorry to be repetitive but I am having a hard time figuring out how to resolve this.

src/main/java/dk/xam/jbang/Main.java
maxandersen: I mean, look at how `resolveInJavaHome()` are used to find the proper executable and handle the differences when running on windows. In your code you directly run `javac` which won't necessarily be the same that `resolveInJavaHome()` will find.

src/main/java/dk/xam/jbang/Main.java
maxandersen: I mean, that in `supportModules()` there is already a call out to `javac` - optimally we would only want to call out to get the java version *once* at most.

src/main/java/dk/xam/jbang/Main.java
maxandersen: you haven't pushed that code to this PR yet have you ?

src/main/java/dk/xam/jbang/Main.java
maxandersen: you can't just take the java path from the vm running `jbang` you need to call out to `java -version ` similar as you do for `javac`.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: No I have not I thought I'll implement the told changes and then push it.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: Okay so I have to just pass `"javac -version"` string to `resolveInJavaHome()` and then execute it using `Runtime.getRuntime.exec()` to get the correct version depending upon the OS.

AtharvaThorve: When I implemented the above part even then I am required to use the split method to handle the output from the above part but using split and accessing the index 1 value lead to the exception.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: Thanks a lot for this tip i implemented the call for `java -version` the same way for `javac -version`

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: So to find the version for both `javac` and `java` I should use `supportModules()` method?
Also there is no call for `javac` but for `java` there is.

maxandersen: yes sorry - I meant for `java`

maxandersen:
for the testsuite first try and run it locally with a setup where java and javac versions are supposed to be the same - does it pass for you there ? if no - then there must be something wrong with the comparisons.

if yes, then you'll need to figure out what java versions are being used in github actions that are triggering your code - it shouldn't if the github action java setup is consistent.

To explore github actions you can enable running github actions on your own fork on  https://github.com/AtharvaThorve/jbang/actions and you can then modify the .github/workflows/* if you need to or simply push changes with some system.out.println's  to print out info you need to figure out what is going on.

AtharvaThorve:
The testsuite executed perfectly with no errors on my local setup on ubuntu.

AtharvaThorve:
The testsuite executed perfectly on my local setup.

AtharvaThorve:
I tried using println statements but I am still not able to resolve it.
What should i do next?

AtharvaThorve:
Since all the checks have been passed so now what is next?

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -263,6 +263,61 @@ public Integer call() throws IOException {
...
+	static boolean checkVersions() {
+		String javaVersion = "";
+		String javacVersion = "";
+		try {
+			String x = resolveInJavaHome("java");
+			if (x.contains(" ")) {
+				x = '"' + x + '"';
+			}
+			javaVersion = getJavaVersion(x);
+			Process p = Runtime.getRuntime().exec(resolveInJavaHome("javac -version"));
```
maxandersen: this doesn't look right - "-version" is a argument and should be added as such rather than part of the command?

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -282,8 +337,10 @@ static void build(Script script, Main m) throws IOException {
 			}
 			optionList.addAll(Arrays.asList("-d", tmpJarDir.getAbsolutePath()));
 			optionList.addAll(Arrays.asList(script.backingFile.getPath()));
-
 			Util.info("Building jar...");
+			if (!checkVersions()) {
+				Util.warnMsg("java and javac versions are different!!");
```
maxandersen: you should include the version info here.

AtharvaThorve: Yes i have changed the structure of the method a little bit to print out the version info too.

src/main/java/dk/xam/jbang/Main.java
AtharvaThorve: I swithced to using `ProcessBuilder` to run the process rather than `Runtime.getRuntime` and passed it a list of arguments.

AtharvaThorve:
A FileNotFoundException occured when running the tests.
I am sorry for being repititive but what should I do?

maxandersen:
you can download the test report in upper right corner of the ci view. Its under "artifacts" in upper right corner on https://github.com/maxandersen/jbang/pull/69/checks?check_run_id=490100848

AtharvaThorve:
I went through the test report and the build failed because it could not download file so server threw a 403 error
`java.io.FileNotFoundException: No file to download. Server replied HTTP code: 403
`
It was not able to download the file in the url `https://gitlab.com/maxandersen/jbang-gitlab/-/raw/master/helloworld.java`

maxandersen:
403 indicates something was forbidden/not authorized - meaning it did communicate with gitlab but for some reason it decided the machine/user accesing did not have rights to access.

Since this does not fail normally I would say to just rerun the PR (you can push a change and it will run again) and see if it was just a temporary glitch.

AtharvaThorve:
Yeah it was a temporary glitch as all checks passed now.
Thank you for the suggestion.

AtharvaThorve:
Heyy @maxandersen should I close this PR?

maxandersen:
hi @AtharvaThorve no, just keep it open. I'm hesitant mainly to integrate it since I haven't heard other have the issue yet and thinking about it after seeing the needed implementation i'm not sure the added complexitiy is worth the very few times it will happen. 

but keeping it open in case others have the issue and comments.

AtharvaThorve:
Okay, thank you @maxandersen

maxandersen:
i think for now i'm going to close this one - the PR is fine and good but the complexity for very little gain makes it not worth integrating at this time. Thanks for the effort!

PR #73, opened 2020-03-04
Support for "aliases"

quintesse:
With this PR you can do something like:

 - `jbang --alias=printprops https://github.com/quintesse/jbang-scripts/blob/master/system_properties.java`

Script is compiled and executed and prints output.

 - `jbang printprops`

Script prints output

If it some point you want to get rid of the alias you can do:

 - `jbang --remove-alias printprops`

Let me know what you think @maxandersen ! :-)

### DIFF
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -159,6 +159,10 @@ public class Main implements Callable<Integer> {
 			"-n", "--native" }, description = "Build via native-image and run", defaultValue = "false")
 	boolean nativeImage;
 
+	@Option(names = { "-a",
+			"--alias" }, description = "Stores the current script URL/path as an alias with the given name.")
+	String alias;
+
 	public int completion() throws IOException {
 		String script = AutoComplete.bash(
 				spec.name(),
@@ -314,6 +318,10 @@ public class Main implements Callable<Integer> {
 					info("run: " + cmdline);
 				}
 				out.println(cmdline);
+
+				if (alias != null) {
+					Settings.addAlias(alias, scriptOrFile);
+				}
 			}
 		}
 		return 0;
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -907,7 +907,7 @@ public class Main implements Callable<Integer> {
 		if (scriptFile == null) {
 			// Not found as such, so let's check the aliases
 			if (Settings.getAliases().containsKey(scriptResource)) {
-				scriptFile = getScriptFile(Settings.getAliases().getProperty(scriptResource));
+				scriptFile = getScriptFile(Settings.getAliases().get(scriptResource).resource);
 			}
 		}
 
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -163,6 +163,9 @@ public class Main implements Callable<Integer> {
 			"--alias" }, description = "Stores the current script URL/path as an alias with the given name.")
 	String alias;
 
+	@Option(names = { "--remove-alias" }, description = "Removes the alias with the given name.")
+	boolean removeAlias;
+
 	public int completion() throws IOException {
 		String script = AutoComplete.bash(
 				spec.name(),
@@ -224,6 +227,8 @@ public class Main implements Callable<Integer> {
 			return 0; // quit(0);
 		} else if (completionRequested) {
 			return completion();
+		} else if (removeAlias) {
+			Settings.removeAlias(scriptOrFile);
 		}
 
 		if (insecure) {
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -907,7 +907,7 @@ public class Main implements Callable<Integer> {
 		if (scriptFile == null) {
 			// Not found as such, so let's check the aliases
 			if (Settings.getAliases().containsKey(scriptResource)) {
-				scriptFile = getScriptFile(Settings.getAliases().get(scriptResource).resource);
+				scriptFile = getScriptFile(Settings.getAliases().get(scriptResource).scriptRef);
 			}
 		}
 
--- a/src/main/java/dk/xam/jbang/Main.java
+++ b/src/main/java/dk/xam/jbang/Main.java
@@ -890,7 +890,49 @@ public class Main implements Callable<Integer> {
 	}
 
 	static Script prepareScript(String scriptResource) throws IOException {
-		File scriptFile = null;
+		File scriptFile = getScriptFile(scriptResource);
+		if (scriptFile == null) {
+			// Not found as such, so let's check the aliases
+			if (Settings.getAliases().containsKey(scriptResource)) {
+				scriptFile = getScriptFile(Settings.getAliases().getProperty(scriptResource));
+			}
+		}
+
+		// Support URLs as script files
+		/*
+		 * if(scriptResource.startsWith("http://")||scriptResource.startsWith("https://"
+		 * )) { scriptFile = fetchFromURL(scriptResource)
+		 *
+		 * includeContext = URI(scriptResource.run { substring(lastIndexOf('/') + 1) })
+		 * }
+		 *
+		 * // Support for support process substitution and direct script arguments
+		 * if(scriptFile==null&&!scriptResource.endsWith(".kts")&&!scriptResource.
+		 * endsWith(".kt")) { val scriptText = if (File(scriptResource).canRead()) {
+		 * File(scriptResource).readText().trim() } else { // the last resort is to
+		 * assume the input to be a java program scriptResource.trim() }
+		 *
+		 * scriptFile = createTmpScript(scriptText) }
+		 */
+		// just proceed if the script file is a regular file at this point
+		if (scriptFile == null || !scriptFile.canRead()) {
+			throw new IllegalArgumentException("Could not read script argument " + scriptResource);
+		}
+
+		// note script file must be not null at this point
+
+		Script s = null;
+		try {
+			s = new Script(scriptFile);
+			s.setOriginal(new File(scriptResource));
+		} catch (FileNotFoundException e) {
+			throw new ExitException(1, e);
+		}
+		return s;
+	}
+
+	private static File getScriptFile(String scriptResource) throws IOException {
+		File scriptFile;
 
 		// we need to keep track of the scripts dir or the working dir in case of stdin
 		// script to correctly resolve includes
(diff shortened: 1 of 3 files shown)


quintesse:
Btw, I tried to add a `--list-aliases` option, but I don't know PicoCLI enough to know how I can add an option that "circumvents" the normal argument validation. All the examples I see (like --help and --version) seem to be hard-coded special cases.

maxandersen:
I get the idea but whats the advantage over this versus using your shell's alias feature ? 

ie. `alias printprops=jbang https://github.com/quintesse/jbang-scripts/blob/master/system_properties.java`

quintesse:
Well, for starters there´s no alias feature on Windows and on other OSes people would have to figure out how to use it, which most likely means manually adding some arcane commands to some hidden config file. This way you can simply give the user a set of instructions: "just type this followed by that and in the future you only have to type `jbang fubar`! Cool huh?". The same instructions taking into account all the different OSes and shells would be a lot harder. So in the end it's just more user-friendly this way :-)

Another idea for future extension using this code is to get to a point where you could have repositories of useful commands that you could easily add. Eg. something like `jbang --repo-add https://github.com/quintesse/jbang-scripts` which would read https://github.com/quintesse/jbang-scripts/blob/master/jbang-aliases and make those commands available.

maxandersen:
hmm - you make a good argument; let me think about it...i'm mainly concerned how it interacts with  `jbang --edit <alias>`  and similar...

quintesse:
Cool, if you want me to make any changes just let me know. And no worries if you decide not to merge it. Just trying to come up with way to make JBang even more useful than it already is hehe

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -83,6 +83,13 @@ boolean debug() {
 	@Option(names = { "-D" }, description = "set a system property")
 	Map<String, String> properties = new HashMap<>();
 
+	@Option(names = { "-a",
+			"--alias" }, description = "Stores the current script URL as an alias with the given name.")
```
maxandersen, line 87: i assume file based locations works too ?

maxandersen:
okey; lets get this in. can you rebase it ?

quintesse:
Done!

maxandersen:
thanks @quintesse, spotted two things: 

- alias is not just a url (i.e. `jbang --alias=example examples/helloworld.java`) works thus the description talking just about url should be updated
- `jbang --clear-cache` nukes `~/.jbang/aliases` should that not survive it? maybe add a `--clear-aliases` option ?

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -216,6 +223,9 @@ private Integer doCall() throws IOException {
 			return 0; // quit(0);
 		} else if (completionRequested) {
 			return completion();
+		} else if (removeAlias) {
+			Settings.removeAlias(scriptOrFile);
+			return 0; // quit(0);
```
maxandersen, line 228: you can do `jbang --clear-cache --init=test.java --edit-live myfile.java`  thus shouldn't `return 0`; here as no issue in doing `jbang --remove-alias='example' other.java`

quintesse:
> alias is not just a url (i.e. jbang --alias=example examples/helloworld.java) works thus the description talking just about url should be updated

Ok

> jbang --clear-cache nukes ~/.jbang/aliases should that not survive it? maybe add a --clear-aliases option ?

My counter-question would be: why is `--clear-cache` removing everything in `~/.jbang`? :-)
Perhaps we should have a dedicated `~/.jbang/cache` folder which gets removed.
Not seeing much use for a `--clear-aliases` to be honest. Imagine using it by accident when you've got a 100+ aliases you painstakingly gathered ;-)

src/main/java/dk/xam/jbang/Main.java
quintesse: Ok

quintesse: (not that a really agree, but I guess that's the issue when using flags as a kind of command)

quintesse:
I was looking at adding a patch for moving the cache to `~/.jbang/cache`, which in itself is simple enough of course. But I'm not sure how you'd want to treat the `JBANG_CACHE_DIR` env var. Rename it to simply `JBANG_DIR` or `JBANG_CONFIG_DIR` and just have the `cache` folder always as a sub-folder of that? Or have 2 different env vars and potentially have the cache in a different folder from the config? (I lean towards the latter)

src/main/java/dk/xam/jbang/Main.java
maxandersen: yes, it has advantages and disadvantages for sure....i often ponder if should move to command layout instead but i can't my head around  that `jbang <script>` would be need to do `jbang run <script>` 

maybe for jbang2 one day ...

maxandersen:
introduce JBANG_DIR works for me, keeping JBANG_CACHE_DIR (`$JBANG_DIR/cache` by default)

src/main/java/dk/xam/jbang/Main.java
quintesse: Couldn't that be made a "shortcut" of some sort? That way you could perhaps do `jbang [run] <url | path>`. It would probably need a bit of custom handling where if no command is found it's assumed to be `run`.

quintesse:
Done adding the new folder + env var

quintesse:
Another thing. Now that Gson is a dependency anyway, do you perhaps want me to turn the `aliases` file into an `aliases.json`? That way we would perhaps be a little more "future proof" if we want to extend things a bit. (For example add a description to an alias, so you can remind yourself what `jget=https://github.com/quintesse/jbang-scripts/blob/master/jget.java` does exactly)

quintesse:
Implemented my last suggestion as a separate commit to this PR. I like it better that way and it's more in line with some other ideas I still have ;-)

src/main/java/dk/xam/jbang/Main.java
maxandersen: yeah, you could but it would be `[run] <url|path|alias|...>`  thus not exactly super clear.

quintesse: True, imagine someone having created an alias named `run` ;-)

Surely if we sleep on it some more we'll come up with a good idea

quintesse: One possible solution might be to not allow aliases in the shortcut version of the jbang command. So you'd get:

```
jbang <url|path>
jbang run <url|path|alias>
```

(and if you have a file called `run` you'll have to do `jbang ./run` to make the shortcut version work)

quintesse: Well, now that I think about it, you would still be able to use aliases with the shortcut version as well as long as there is no ambiguity. So running an alias called "run" is still possible using `jbang run run`.

You could say that the shortcut version is nice for humans typing in the command, while advising to use `jbang run` in scripts because there'll never be any ambiguity in that case. (Not even when installing a newer version of JBang with more commands)

maxandersen: you are right! wanna create a separate issue to consider what a proper command structure would be before we declare jbang 1.0 ? :)

quintesse: #94

maxandersen:
@all-contributors please add @quintesse  for code

src/main/java/dk/xam/jbang/Settings.java
```diff
@@ -66,4 +87,61 @@ public static void clearCache() {
...
+					Gson parser = new Gson();
+					aliasInfo = parser.fromJson(in, AliasInfo.class);
+				} catch (IOException e) {
+					// Ignore errors
+				}
+			} else {
+				aliasInfo = new AliasInfo();
+			}
+		}
+		return aliasInfo.aliases;
```
maxandersen, line 121: when I try this out I consistentl gets:

```
jbang printprops
	at dk.xam.jbang.Settings.getAliases(Settings.java:121)
	at dk.xam.jbang.Main.prepareScript(Main.java:917)
	at dk.xam.jbang.Main.doCall(Main.java:308)
	at dk.xam.jbang.Main.call(Main.java:365)
	at dk.xam.jbang.Main.call(Main.java:77)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1783)
	at picocli.CommandLine.access$900(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2150)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2144)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2108)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:1975)
	at picocli.CommandLine.execute(CommandLine.java:1904)
	at dk.xam.jbang.Main.main(Main.java:190)
```

maxandersen: happens when aliases.json file is empty

src/main/java/dk/xam/jbang/Settings.java
```diff
@@ -66,4 +87,61 @@ public static void clearCache() {
...
+			getAliases().put(name, new Alias(resource));
+			Gson parser = new GsonBuilder().setPrettyPrinting().create();
+			parser.toJson(aliasInfo, out);
+		} catch (IOException ex) {
+			Util.warnMsg("Unable to add alias: " + ex.getMessage());
+		}
+	}
+
+	public static void removeAlias(String name) {
+		if (aliasInfo.aliases.containsKey(name)) {
```
maxandersen, line 136: this can be null

src/main/java/dk/xam/jbang/Settings.java
```diff
@@ -66,4 +87,61 @@ public static void clearCache() {
...
+	public static void setupJBangDir() {
+		// create JBang configuration dir if it does not yet exist
+		JBANG_DIR.toFile().mkdirs();
+	}
+
+	static class Alias {
+		final String resource;
+
+		Alias(String resource) {
+			this.resource = resource;
```
maxandersen, line 100: this is not  a resource per say, lets call it `scriptRef` or similar.

maxandersen:
the code doing gson parsing is too flakey for me being ok to commit it. maybe add some tests ?

issues seen:
- when aliases.json is empty: nullpointerexception
- when errors but syntax valid in aliases.jason: another differnet nullpointer exception

nice feature discovered but seem to fail on second level:

```
jbang --alias printprop <url>
jbang --alias p2 printprop
jbang p2
```

actually works, but

```
jbang --alias p3 p2
```

fails with: 

```
java.lang.IllegalArgumentException: Could not read script argument p2
	at dk.xam.jbang.Main.prepareScript(Main.java:940)
	at dk.xam.jbang.Main.doCall(Main.java:308)
	at dk.xam.jbang.Main.call(Main.java:365)
	at dk.xam.jbang.Main.call(Main.java:77)
```

either disallow it or handle it better.

maxandersen:
@quintesse will you be able to rebase/fix this based on the items found above ^ ?

quintesse:
Hi, problems should be fixed now.

Don'-t know if throwing an ExitException when I detect that the user tries to make an alias of an alias is the correct way. It prints a message but the execution still continues.

No tests yet.

maxandersen:
merged!

PR #74, opened 2020-03-05
Minor change in the way the smee example stays running

quintesse:
It's a really minor change and not important at all but
I just think it's a cleaner way of waiting around than
writing an endless loop with waits.


maxandersen:
thanks! :)

PR #77, opened 2020-03-05
Added `--naive` option to trust all SSL certificates

quintesse:


maxandersen:
used --insecure instead and now in master

quintesse:
Sure, the name was more a bit of fun, I expected you to change it ;-)

PR #100, opened 2020-05-30
Hide non-relevant files in vscode

fbricon:
to reduce clutter

* Before:
Screen Shot 2020-05-30 at 2 24 01 AM

* After:
Screen Shot 2020-05-30 at 2 21 58 AM


maxandersen:
okey - makes sense. thanks!

PR #107, opened 2020-05-30
docs: remove incorrect header markup

manikmagar:
Fix header rendering


maxandersen:
thanks - bad cut of temporary warning.

Issue #113, opened 2020-05-31
have url fetch protected by trusted domains white/blacklist


maxandersen:
to protect against @gunnarmorling rickrolls we should add notion of trusted domains.

vscode has a nice simple scheme for opening links we could probably reuse.

see https://github.com/microsoft/vscode/blob/master/src/vs/workbench/contrib/url/test/browser/trustedDomains.test.ts
and https://github.com/microsoft/vscode/blob/master/src/vs/workbench/contrib/url/browser/trustedDomainsValidator.ts

﻿

maxandersen:
implemented in 0.27

PR #117, opened 2020-06-01
Add support for trusted sources

### LINKED ISSUE #116: basic error info not printed unless --verbose
in exit handler we no longer print actual info error message unless the exceptoin has a underlying cause. should just always print.﻿

maxandersen:
Why:

 * @gunnarmorling rickroll https://twitter.com/gunnarmorling/status/1267033700031778816
   made it clear it would be good if could add basic protection to not run arbitrary code
   on the internet.

 * fix bug where error info not printed
 * before fetching url the url gets validated against trusted domains.
 * trusted domains are a list of patterns, if a match is found it is allowed;
   otherwise not.
 * if url not found jbang will print out instructions on how to add trust for that url.

Fixes #116, #113


gunnarmorling:
So "trusted domain" made me first think it's about, well, domains, only.
Whereas it seems I also can decide to trust a specific sub-path only, like
https://gist.github.com/gunnarmorling/*, correct? If so, "trusted sources"
or similar might perhaps be less confusing?

>

maxandersen:
trusted domain or trusted sites seem to be what vscode and browsers used so i used that but you have a point though...

maxandersen:
there you go @gunnarmorling  :)

gunnarmorling:
Nice! :thumbs_up:

PR #120, opened 2020-06-06
(fix) run as non-root user

### LINKED ISSUE #119: Support to run jbang-action non-root user
The jbang-action right not run as root user this prevents it from running many Kubenretes distributions like OpenShift.

https://github.com/jbangdev/jbang-action/issues/5

kameshsampath:
The update allows jbang container to run as arbitary non root user. This makes it safe and comptaible to be run on OpenShift.

Fixes #119


maxandersen:
Thanks @kameshsampath - looks sane. I'll add it and get a few other fixes for the docker image in (--trust broke it ;/ and then release it next.

@all-contributors please add kameshsampath for code

maxandersen:
@all-contributors please add @kameshsampath for code

PR #131, opened 2020-06-17
Make sure we properly quote the command on Windows

### LINKED ISSUE #129: Specifying --trust=* is too hard on windows
I have been trying in several ways to specify --trust=* on Windows without success. Often strange values get inserted and once that even resulted in an illegal JSON file that broke jbang completely (it would throw parse exceptions straight away).

In the end I had to insert a dummy trust value and then edit the resulting file and change it to *.

quintesse:
Fixes #129

### DIFF
--- a/src/main/scripts/jbang.bat
+++ b/src/main/scripts/jbang.bat
@@ -16,6 +16,6 @@ set JBANG_FILE="$1"
 
 rem run it using command substitution to have just the user process once jbang is done
 rem eval "exec $(${JAVA_EXEC} -classpath ${jarPath} dk.xam.jbang.Main "$@")"
-FOR /F "tokens=*" %%a in ('%JAVA_EXEC% %JBANG_JAVA_OPTIONS% -classpath %jarPath% dk.xam.jbang.Main %*') do SET OUTPUT=%%a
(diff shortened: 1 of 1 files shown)


maxandersen:
why is it we don't want to have globbing on windows like we have it on linux ?

quintesse:
Ok, this would disable globbing entirely true, but you have to think there's is no shell globbing on Windows. So you'll have to tell people how to escape those globbing characters because normal users won't have any idea how to do that.

maxandersen:
im actually puzzled what part of that line in the batch file does the globbing ...as you said - globbing isn't normal on windows :)

quintesse:
It's the `FOR /F` (where F stands for File).

maxandersen:
the "for /f" is just about iterating the result ....it shouldn't be affecting the expansion inside the string..

maxandersen:
okey so after spending waay too long on getting karate running both locally and in github actions its very clear we got a problem on windows.

see https://github.com/jbangdev/jbang/issues/157

thus we should get this fixed ;)

now that karate.java is running - can you add some tests for the quote situations this code is fixing ? As i'm sure we'll end up breaking them again if we dont add tests for it.

quintesse:
Closing this PR because it's not the right solution (I tried preventing the glob-expansion the FOR does, for example, and there just doesn't seem a way to do that. So you could never do `jbang echo.java \*` and just have it print an asterisk). See https://github.com/jbangdev/jbang/issues/157#issuecomment-653906179

PR #138, opened 2020-06-20
Quote JAVA_HOME for non english Windows.

dem2k:
JAVA_HOME on non english Windows System is usually on Path with Spaces inside like "C:\Program Files\OpenJDK\", so we need to quote them.


maxandersen:
as a Dane I feel bad I missed that one thus far :)

Thank you!

Issue #94, opened 2020-05-26
Move CLI to proper command structure


quintesse:
Right now there are a variety of actions the jbang command is capable of performing by passing flags on the command line. But this leads to a situation where you can have combinations of flags that don't really make a lot of sense.

It would be nice to move to a proper command structure where each action would be its own (sub) command of the jbang main command. Eg. `jbang edit` vs `jbang --edit` and `jbang alias remove` instead of `jbang --remove-alias`.

The only special case we have to consider is the most often used one where the user just wants to run some code. The requirement is that this should still be possible without having to use a command (eg. `run`) by using some kind of shorthand feature. (eg. `jbang path/to/file.java` and `jbang run path/to/file.java` would be considered equivalent)

Exceptions to this are allowed to resolve ambiguities (eg. trying to execute a file called "run" by typing `jbang run`) in which case requiring the command is considered okay (eg, you'd have to type `jbang run run`).

Suggested commands:

 - **jbang** *[-d] [-Dkey=value]... [-o] <path|url|alias>*
 - **jbang run** *[-d] [-Dkey=value]... [-o] <path|url|alias>*
 - **jbang init** *<path|url|alias>*
 - **jbang edit** *[--live[=< editor >]] <path|url|alias>*
 - **jbang alias add** *<name> <path|url>*
 - **jbang alias remove** *<name>*
 - **jbang alias list**
 - **jbang clear-cache** (cache clear?)
 - **jbang completion**
 - **jbang version**

maxandersen:
the one "combination" command I use constantly are `jbang --init=cli --edit-live=idea <path>`

wondering how to best do this in this structure ?

maxandersen:
maybe allow `jbang init --kind=cli --edit-live=idea <path>` while still having `jbang edit ...` ?

quintesse:
Or one, yet to be named, command that can do both, a bit like it is now, eg. `jbang xxxx --init=cli --edit-live=code <path>`. Possible idea for the name `project`, `file`, `script`. Something sufficiently vague :-)

PR #139, opened 2020-06-22
WIP: Creating command structure

quintesse:
For #94


quintesse:
Still far from ready, just sharing progress so far. Haven't yet tested if everything still works. Tests don't compile yet. No special case for the "run" command yet.

quintesse:
All features are implemented now. Some extra tests were added. Put everything through its paces but not extensively so. Let me know what you think @maxandersen

maxandersen:
can you do a `gradle spotlessApply`  and push - then you should get longer in the CI test.

maxandersen:
first impressions is that over all looks good - haven't reviewed code yet; just the new "UX".

Found some gotchas though:

- [ ] `jbang completions` doesn't seem to work. With 0.31 I get completions that helps for files and -- commad line arguments. This version it just completes on any file.
- [ ] `jbang script` just feels..awkward. I think I would prefer doing `jbang init` and `jbang edit` as separate commands and just have a "shortcut" from init to specify `--edit` ...but maybe `jbang init --template=cli max.java; jbang edit --live max.java` is fine? 
- [ ] `jbang clear-cache` should be `jbang cache clear` i think - i.e. `jbang cache prune` would be useful at some point to delete just older builds and not all...
- [ ] `jbang alias` and `jbang trust` should proably both have `list`, `add`, `clear` and actually show the alias/trustee (can be done in follow up PR)

quintesse:
* I'm not sure about the completions, isn't that something that's entirely handled by Picocli? Never tried it (not in the old version nor in this one), I just checked it generated _something_
 * script -> init/edit. Sure.
 * yeah, I thought about that too, but I wasn't sure you'd want a subcommand with just a single subsubcommand. Consider it done.
 * same here, it was my idea to do it for trust too, but I wasn't entirely sure how to do the deletion of trust. Having to put the exact trust domain seemed a little user-unfriendly so I just decided not to add it yet. But no problem in at least adding something that works.

quintesse:
For the code I'd actually split up all the subcommand in separate files. But again I wasn't entirely sure how you'd want to structure it. Just all in the same package as Main? Or make a `cli` package for example?

maxandersen:
> For the code I'd actually split up all the subcommand in separate files. But again I wasn't entirely sure how you'd want to structure it. Just all in the same package as Main? Or make a `cli` package for example?

I'll need some sleep before answering that :)

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -81,259 +79,377 @@
...
+	public static void main(String... args) {
+		CommandLine cli = getCommandLine();
+		args = handleDefaultRun(cli.getCommandSpec(), args);
+		int exitcode = cli.execute(args);
+		System.exit(exitcode);
+	}
 
-	@Option(names = { "--interactive" }, description = "activate interactive mode")
-	boolean interactive;
+	private static String[] handleDefaultRun(CommandSpec spec, String[] args) {
```
maxandersen, line 140: is this the way to deal with defaulting the run command ? I could imagine this is what is messing with completion handling not working or that this injecting of "run" is not recognized by the completor?

quintesse: I'm not sure, but I can't imagine right now how that could cause problems? The completion system doesn't know about this so it should just generate completion options for what's defined, right? I suspect the problem will remain even if we simply comment this code out.

src/main/java/dk/xam/jbang/cli/JbangAlias.java
```diff
@@ -0,0 +1,54 @@
+package dk.xam.jbang.cli;
+
+import dk.xam.jbang.Settings;
+
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "alias", description = "Manage aliases.", subcommands = { JbangAliasAdd.class,
+		JbangAliasList.class,
```
maxandersen, line 8: instead of subclasses per subcommand i think it would be simpler to use methods as subcommands - see https://picocli.info/#_subcommands_as_methods

that would make this class simpler/more readable.

src/main/java/dk/xam/jbang/Main.java
```diff
@@ -1,143 +1,21 @@
...
-
-			return e.getStatus();
-		}
-	}
-
-	public abstract Integer doCall() throws IOException;
-}
-
-@Command(name = "jbang", footer = "\nCopyright: 2020 Max Rydahl Andersen, License: MIT\nWebsite: https://github.com/jbangdev/jbang", description = "Compiles and runs .java/.jsh scripts.", subcommands = {
-		JbangRun.class, JbangEdit.class, JbangInit.class, JbangAlias.class, JbangTrust.class, JbangCache.class,
```
maxandersen: To much bang here - lets just do `Run`, `Edit`, `Init`, `Alias`, `Trust` and `Cache`

they are already in jbang.cli. package.

src/main/java/dk/xam/jbang/cli/JbangAlias.java
quintesse: I can try that, but I think the gains will be very small, a couple of lines at most.

src/main/java/dk/xam/jbang/Main.java
quintesse: Sure, although you can never have enough bang for your buck :-D

quintesse: Done

src/main/java/dk/xam/jbang/cli/JbangAlias.java
quintesse: Done

maxandersen: i find that much nicer - 3 less classes :)

maxandersen:
this looks good - my only nitpick is that `jbang --init=cli hello.java` is now `jbang init --template=cli hello.java` but I reckon thats a loss for the win of cleaner UX and api elsewhere.

maxandersen:
merged - i'm sure I or you will find things that can be improved but didn't want to leave it hanging too long blocking other improvements.

Next up - fix the docs :)

quintesse:
> my only nitpick is that jbang --init=cli hello.java is now jbang init --template=cli hello.java

I think we should at least add a single char shortcut for the option: `jbang init -t cli hello.java`

maxandersen:
its `-t` would be decent comprimise :)

PR #143, opened 2020-06-26
Updated README for new command structure

quintesse:


maxandersen:
the new structure just reads better. good move. thanks for updating the docs!

maxandersen:
might help if I remembered to merge before closing  - now done ;)

PR #146, opened 2020-06-26
Extend aliases to store arguments and properties

quintesse:
You can now specify arguments and properties with aliases, just like with the `run` command. If you run an alias that has arguments and/or properties without specifying arguments and/or properties yourself then those arguments/properties will be used.

(depends on #144)

### DIFF
--- a/src/main/java/dk/xam/jbang/Script.java
+++ b/src/main/java/dk/xam/jbang/Script.java
@@ -52,22 +52,34 @@ public class Script {
 	// true if in this run a jar was build/created
 	private boolean createdJar;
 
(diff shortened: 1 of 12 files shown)


quintesse:
What do you think about this @maxandersen ? I think the feature is pretty cool, although I had to change `Script` to make it work without having to change even more changes.

It also means I've made it possible again to base aliases on aliases. This way you can have for example a "base" alias to make it easier to deal with a complex URL, while you could then create another alias to pass arguments.

maxandersen:
okey :) 

would be good to add to integration tests but i think i'm going to move them to karate this weekend thus lets just add that there.

quintesse:
Yay! :-)

PR #151, opened 2020-06-30
Remove last occurrences of xam

quintesse:


maxandersen:
thanks! wonder how the tests actually passed when main class was wrong ;)

quintesse:
Because you specify it explicitly in the script: https://github.com/jbangdev/jbang/blob/master/src/main/scripts/jbang#L42

PR #155, opened 2020-07-02
Fixes #154 Incorrect handling of arguments with space/metachars

### LINKED ISSUE #154: Incorrect handling of arguments with space/metachars
If you run JBang with single quoted arguments to prevent special processing taking place then things go wrong once the first processing phase hsa finished and the actual Java commands get executed.

Eg the command `jbang echo.java 'foo *' bar` should result in `echo.java` being called with two arguments: `foo *` and `bar`, but instead it receives those 2 plus the entire list of iles that happen to be in the current directory.

This is because in the final command that gets executed the arguments are not quoted at all. This problem also exists for any other meta-characters like redirection or variables etc etc

On Linux/Mac the suggestion is to single quote *all* arguments passed by the user to prevent any extra special processing taking place (that should already have been handled by the user's shell when they started JBang).

On Windows it's a bit more complicated but can be summarized by:

 1. Double quote all arguments with `^"` (not simply `"`)
 2. Then within the string escape all metacharacters (`(`, `)`, `%`, `!`, `^`, `"`, `<`, `>`, `&`, `|`) using `^` (including `^` itself `^^`)
 3. Except for double quotes which need to be escaped like this: `\^"`

quintesse:
This seems to work. Not sure if there are any other special chars I might have missed, but I've Googled the special chars for each environment (shell vs CMD) and added all other chars on my keyboard just to be sure.

On Linux even multi-line arguments work correctly. (On Windows that's not really possible)

We might need to add an integration test for this though.

### DIFF
--- a/src/main/java/dev/jbang/Settings.java
+++ b/src/main/java/dev/jbang/Settings.java
@@ -25,7 +25,7 @@ public class Settings {
 	public static final String TRUSTED_SOURCES_JSON = "trusted-sources.json";
 	public static final String DEPENDENCY_CACHE_TXT = "dependency_cache.txt";
 
-	final public static String CP_SEPARATOR = Util.isWindows() ? ";" : ":";
+	final public static String CP_SEPARATOR = File.pathSeparator;
 
 	private static TrustedSources trustedSources;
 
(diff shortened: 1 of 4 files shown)


maxandersen:
merged with karate test for echo testing

PR #159, opened 2020-07-06
Minor refactor in usage of `getAlias()`

quintesse:
Hi @maxandersen , could you merge this? I saw you fixed an issue with the usage of `getAlias()`, but it was actually a problem in `getAlias()` itself.


quintesse:
Thanks! :-)

PR #160, opened 2020-07-07
Fix edit

ssaarinen:
Noticed that Init command used deprecated version of edit on console output.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,6 +55,7 @@ tasks.withType(AbstractArchiveTask) {
 }
 
 spotless {
+	lineEndings 'UNIX'
 	format 'misc', {
 		target '**/*.gradle', '**/*.md', '**/.gitignore'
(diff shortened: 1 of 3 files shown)


ssaarinen:
This sees weird as Spotless fails on Alias.java that I did not change. It did so on my Win machine as well but spotlessApply did not cause any changes but after that spotlessCheck succeeded. Had to force spotless to use UNIX EOL though as otherwise spotlessApply would have changed those.

maxandersen:
thanks @ssaarinen ! 

@all-contributors please add ssaarinen for bug

maxandersen:
had to revert the lineendings setting as that is causing conflicts in build. added gitattributes to force use of LF instead of CRLF.

ssaarinen:
I think that's better as spotless has support for it as well 👍

Issue #45, opened 2020-01-26
run Maven artifacts directly


quintesse:
Explore the possibility to do something like `jbang org.groupid:artifactid` or `jbang org.groupid:artifactid org.groupid.artifactid.MyMainClass` where the first argument would be the name of a Maven artifact to run.

The first form would be for a Maven artifact that has a `Main-Class` set in their `MANIFEST`, the second form would be for those that haven't.

@maxandersen suggested to use the form  `jbang mvn://org.groupid:artifactid` which would work too (although we could/should perhaps drop the 2 slashes)

maxandersen:
reason for mvn:// would be its then a url which has special treatment anyway.

ultimately the logic would be something like this:

```
jbang <file>
if file.contents().startsWith("PK") {
   metadata = findMetadatafromurl(filepath); // here if mvn:// then use resolver to fetch dependencies as it’s a mvn location
   if metadata==null
      metadata = findMetadataInJar(filepath); // use what might be embedded in the jar pom.xml etc.
  lanch(filepath, metadata);
} else {
  doAsUsual();
}
```

then this could be used on a .jar directly.

maxandersen:
thinking more of this it could just be:

`jbang org.groupid:artifactid:2.1` that is unique enough pattern to recognize as artifact.

for no main better to have it be:
`jbang org.groupid:artifactid:2.1#org.mymain.App` or similar. # is problematic for shells as its a comment thus maybe some other character

quintesse:
We could perhaps do: `jbang org.groupid:artifactid:2.1/org.mymain.App`. Having several ´:´s is probably still enough to distinguish it from a URL.

quintesse:
Btw, do you think there would be a way to leave out the version so it would use the "latest"? Or is that a concept that doesn't exist in Maven?

maxandersen:
There are ways to look it up. Not sure which shrink-wraps support but we can figure ways around thst

maxandersen:
in mvn 3 they removed the notion of latest and release as considered bad for repeatable builds. 

But mechanism to calculate it is still there used in mvn versions plugin (see https://www.baeldung.com/com maven-dependency-latest-version) we should be able to mimick whatever it does. 

I assume it's fetching metadata.xml and finding the highest versionin there.

maxandersen:
considering #148 it would make sense that `jbang -jar <jar ref>` would be way to invoke this.

[...]

PR #163, opened 2020-07-08
First attempt on supporting launches jar from file or GAV.

maxandersen:
Basically just says "if script ends in jar run with jar" and
when resolving sarguments GAV's gets honored


maxandersen:
relates to #45

src/main/java/dev/jbang/DependencyUtil.java
```diff
@@ -42,15 +44,23 @@
...
-	 * 
+	 *
 	 * @param deps
 	 * @param repos
 	 * @param loggingEnabled
 	 * @return string with resolved classpath
 	 */
 	public String resolveDependencies(List<String> deps, List<MavenRepo> repos,
-			boolean offline, boolean loggingEnabled) {
+			boolean offline, boolean loggingEnabled, boolean transivity) {
```
quintesse, line 63: I think it's "transitivity" :-)

maxandersen: fixed

maxandersen:
the changes wasn't that big and it doesn't break existing functionallity thus merged.

quintesse:
Yay!

PR #167, opened 2020-07-09
The `alias.json` file now uses kebab naming

quintesse:
Older files will still be read and be automatically updated next time
a change has been made


quintesse:
GSON actually supports "alternative field names", so was able to support the new names without any changes to the code. yay!

maxandersen:
Awesome

PR #173, opened 2020-07-12
Now quoting all arguments generated by Jbang before passing on to bash/cmd

### LINKED ISSUE #171: Path error stops jbang finding parameter main class
On Windows 10 Powershell I get an error saying it can't find the main class to load. This looks like a failure to cope with spaces in the path:
```
PS C:\WINDOWS\system32> jbang "C:\systemsincode\Java\jbang\jbangHelloWorld.java"
Could Not Find C:\Users\Matthew O'Surname\.jbang\31237.tmp
Error: Could not find or load main class O'Surname\.jbang\cache\jars\jbangHelloWorld.java.29e25d772ac7cc3c03799c4ab6cc6061d8d0999846df1db70f534a5a69c94866.jar
Caused by: java.lang.ClassNotFoundException: O'Surname\.jbang\cache\jars\jbangHelloWorld.java.29e25d772ac7cc3c03799c4ab6cc6061d8d0999846df1db70f534a5a69c94866.jar
```
Even thought it exists in the right location:
image

quintesse:
Fixes #171. Quoting is now also a bit smarter, only quoting when necessary.

@maxandersen I tested this with the following command:

```
$ JBANG_CACHE_DIR=/tmp/jbang\ cache\ test ./build/install/jbang/bin/jbang echo.java test
```

But I'm not sure how to add that to Karate, so we're still missing a test for this.

### DIFF
--- a/itests/quoting-win.feature
+++ b/itests/quoting-win.feature
@@ -3,14 +3,10 @@ Feature: jshell
 
 Background:
 * if (javaversion == 8) karate.abort()
-* if (windows) karate.abort()
-
-Scenario: piping code via stdin
-When command('echo \'System.out.println(\"Hello World\")\' | jbang -')
-Then match out == "Hello World\n"
+* if (!windows) karate.abort()
 
 Scenario: check quotes are kept when wrapped with quotes
-When command('jbang echo.java \'foo *\'')
+When command('jbang echo.java "foo *"')
 Then match out == "0:foo *\n"
 
 Scenario: check expansion does happen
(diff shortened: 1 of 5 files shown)


maxandersen:
You can pass in env map to command to set that

quintesse:
> You can pass in env map to command to set that

You have any example for that?

And creating (and cleaning up) the temp folder?

maxandersen:
> You have any example for that?

https://github.com/jbangdev/jbang/blob/master/itests/dependencies.feature#L5

basically: `command('jbang classpath_log.java', { JBANG_CACHE_DIR: scratch + "/cache"})`

maxandersen:
> And creating (and cleaning up) the temp folder?

atm. it just call's createTempFile ...but haven't yet added `deleteOnExit()` call...I guess we could/should :)

quintesse:
Tests added and they pass so this should be safe to merge.

quintesse:
Nope, hold that. Still problems on WIndows.

quintesse:
Ok, tested it on Windows too now, so if the CI build finishes correctly things should be good to merge.

PR #175, opened 2020-07-12
Fix npe if test failed

dem2k:


maxandersen:
i'm not seeing the NPE  - any way to reproduce?

PR #176, opened 2020-07-12
Fix watchservice take

dem2k:
watchservice.take methode never comes back. the process should be interrupted hard.


maxandersen:
educate me here - I can see its not great; but what is actually failing for you ?

dem2k:
I can show you better:  https://www.dropbox.com/sh/7u7x3a11b5aysee/AADX3LkbS213GYean6DG_ASIa?dl=0

maxandersen:
great videos - now it makes sense. yeah - needs fixing ;)

maxandersen:
looks good! merging!

maxandersen:
Had to revert this as it assumes the edit process runs forever. that is not the case when you are using launchers as used by code, idea or emacs -nw

PR #177, opened 2020-07-12
Add build command + move insecure to base. Fixes #62

### LINKED ISSUE #62: add a --no-run option to prep cache

maxandersen:
Was tinkering with #62 by adding a super naive "build" command
that is exactly the same as Run but just don't do run part.

When I noticed --insecure used for downloading was only applied to Run and
no other commands...should get that moved more central.

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseScriptCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseScriptCommand.java
@@ -2,10 +2,17 @@ package dev.jbang.cli;
 
 import static dev.jbang.Util.swizzleURL;
 
-import java.io.*;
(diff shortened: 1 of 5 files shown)


quintesse:
Just move it to `BaseScriptCommand`, then it will apply to any commands that deal with Scripts (as the name so expertly implies :-) )

quintesse:
Can we perhaps wait with merging this until we merge #178 if possible? There's been quite a bit of changes in the build code, while in this PR you've basically only moved it, right? I think it will be easier to merge/redo this part last than the other way around.

maxandersen:
yeah i'm fine holding back on this one - its a trivial change.

Issue #133, opened 2020-06-19
have jbang download (adopt)jdk for end-user script to use


maxandersen:
if jbang could detect missing java runtime and possibly be guided by a `// JAVA_VERSION 14+` in script and download the matching adoptopen JDK for the current platform.

should still honor PATH and JAVA_HOME first (with option for overriding)  

/cc @quintesse @galderz

maxandersen:
here is experiment using adopt java client:

```
//usr/bin/env jbang "$0" "$@" ; exit $?

//DEPS net.adoptopenjdk:net.adoptopenjdk.v3.api:0.3.3
//DEPS net.adoptopenjdk:net.adoptopenjdk.v3.vanilla:0.3.3


import net.adoptopenjdk.v3.api.AOV3Architecture;
import net.adoptopenjdk.v3.api.AOV3Exception;
import net.adoptopenjdk.v3.api.AOV3HeapSize;
import net.adoptopenjdk.v3.api.AOV3ImageKind;
import net.adoptopenjdk.v3.api.AOV3JVMImplementation;
import net.adoptopenjdk.v3.api.AOV3OperatingSystem;
import net.adoptopenjdk.v3.api.AOV3ReleaseKind;
import net.adoptopenjdk.v3.api.AOV3Vendor;
import net.adoptopenjdk.v3.vanilla.AOV3Clients;

import java.io.IOException;
import java.math.BigInteger;
import java.util.Optional;

class adoptopenjdk {

    public static void main(String... args) throws Exception {
        var clients = new AOV3Clients();
        try (var client = clients.createClient()) {
            var result = client.binaryForLatest(e -> System.out.println(e),
                    AOV3Architecture.of(arch()),
                    BigInteger.valueOf(11),
                    AOV3HeapSize.NORMAL,
                    AOV3ImageKind.JDK,
                    AOV3JVMImplementation.HOTSPOT,
                    AOV3OperatingSystem.of(os()),
                    AOV3ReleaseKind.GENERAL_AVAILABILITY,
                    AOV3Vendor.ADOPT_OPENJDK, Optional.of("jdk")).execute();
            System.out.println(result);
        }
    }

    static String arch() {
        String arch = System.getProperty("os.arch");

        if(arch.contains("32")) {
            return "x32";
        } else if(arch.contains("64")) {
            return "x64";
        } else {
            return arch;
        }

    }
    static String os() {
        String os = System.getProperty("os.name").toLowerCase();

        if (os.contains("win")) {
[...]

Issue #83, opened 2020-04-16
j'bang should recompile class if script is rerun using a different JDK version


fbricon:
I initially ran my jbang script with Java 11. Much later on, I had to switch the default Java version in my terminal, to Java 8. When rerunning, j'bang fails because it's trying to rerun the already compiled class:
```
❯ jbang https://gist.githubusercontent.com/fbricon/4028e0e2dda01acd0e7bd0f3e3c2ed63/raw/e06c5553387e5d357f9d38426a9c3c2289104425/GenerateChangelog.java https://github.com/redhat-developer/vscode-java/milestone/73 https://github.com/eclipse/eclipse.jdt.ls/milestone/72
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: GenerateChangelog has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:495)
```

Hopefully I could fix it with `jbang --clear-cache`, but maybe the path of the cached class could be encoded with the JVM or java version used to compile it, to avoid that issue.

maxandersen:
hmm - here I thought once you go java 11 you never go back ? :) 

not sure there is a reliable thing to key on here.....but yeah; would be nice to handle more cleanly.

maxandersen:
just because java 11 was used to compile does not mean java 8 can't run it...

[...]

PR #178, opened 2020-07-14
Downloading of JDKs

quintesse:
First stab at implementing #133 

Current state is that you can pass a `--version` ro `jbang run` with the major number of the Java version you want to use. It will then check the version of the JDK pointed at by `JAVA_HOME` or on the path by running `javac -version` (*) and parsing the output. If the version is lower than the requested version or no locally installed JDK was found or its version was lower than 8, the ZIP/TAR.GZ for the requested JDK will be downloaded from AdoptOpenJdk and unpacked in the `JBANG_CACHE_DIR/jdks/<version>/` folder.

Also, if no local JDK (of at least version 8) was found and no version was requested version 11 will be downloaded and unpacked. (This might seem useless because how can we be running Jbang if there's no Java, but perhaps we can compile Jbang to native code in the future using Graal/Quarkus, who knows)

(*) Running `javac` because then we know it's a JDK and not just a JRE. It's also easier to parse its output.


src/main/java/dev/jbang/cli/Run.java
```diff
@@ -33,6 +30,10 @@
 			"--main" }, description = "Main class to use when running. Used primarily for running jar's.")
 	String main;
 
+	@CommandLine.Option(names = { "-v",
```
maxandersen, line 33: I kinda always felt the version used should be coming from the script, i.e. `//JAVA_VERSION 11+` - and it should be able to say 11+ meaning if you have 14 that is fine. If you say 11 it must be 11. that kinda thing.

This flag would be an override of that setting - and as such I would say it should be named something like `-j/--java 11`

src/main/java/dev/jbang/JdkManager.java
```diff
@@ -0,0 +1,91 @@
...
+		Path jdkDir = getJdkPath(version);
+		UnpackUtil.unpack(jdkPkg, jdkDir);
+		return jdkDir;
+	}
+
+	private static Path getJdkPath(int version) {
+		return Settings.getCacheDir().resolve("jdks").resolve(Integer.toString(version));
+	}
+
+	private static Integer determineJavaVersion() {
```
maxandersen, line 55: we got ModularClassPath.supportsModules() that does something similar - wanna check if we can unify/reuse this over there to not have conflicting/multiple calls to "java -version"

maxandersen:
good one - i would just say that I wouldn't want users having to specify `--version 11` at call time (should be optional) and instead have the script declare what it needs (similar to `//DEPS`) i.e. make `//JAVA_VERSION ` a thing.

Oh, and as soon as you have `--version` (or `-j/-java`) then https://github.com/jbangdev/jbang/issues/83 becomes even more relevant/interesting.

src/main/java/dev/jbang/cli/Run.java
quintesse: Yes, I agree, the setting was just the quickest to add, I needed something to test with :-)

And being able to say "X+" had occurred to me too, but I went for the easier implementation first. Now that I know you want it that way I'll change it.

`-j/--java` sure.

src/main/java/dev/jbang/JdkManager.java
quintesse: Oh really? Didn't know that, never looked at that part of the code. Will do.

quintesse:
Btw, how about adding an `jbang install` command to install JDKs? That way we can also have a `--force` option that deletes the current version and installs a new one. (That way if you do `jbang install --force 14` you will know that you have the latest version. Wdyt?

quintesse:
Btw2: You want to call it `//JAVA_VERSION` or just plain `//JAVA` or something?

Btw3: It seems that the CI server is blocking requests to/downloads from the AdoptOpenJdk domain :-/

maxandersen:
> Btw2: You want to call it `//JAVA_VERSION` or just plain `//JAVA` or something?

can't think of a good reason right now why `//JAVA 11+` shouldn't be enough. thus lets go with `//JAVA` for now...;)

maxandersen:
> Btw3: It seems that the CI server is blocking requests to/downloads from the AdoptOpenJdk domain :-/

bummer :/

maxandersen:
> Btw, how about adding an `jbang install` command to install JDKs?

not convinced. starting to smell a lot like a `sdkman` replacement ... but it might be necessary. add it but as hidden command for now ?

quintesse:
> not convinced. starting to smell a lot like a sdkman replacement ... but it might be necessary. add it but as hidden command for now ?

I definitely understand. But the other option would be to add more (and more) options to the `run` command to force it to re-download a JDK.

Just like we could add an option to force it to re-compile or re-download the script. But in fact it's somewhat similar to adding the `build` command, we could add a `--force` there as well, to make sure things get re-compiled.

Just rambling a bit here, it's not that I have a really clear idea how best to handle this.

quintesse:
Added support for `//JAVA_VERSION`, yes I know I suggested `//JAVA` but the existing parser can't distinguish between `//PREFIX` and `//PREFIX_LONG` and I didn't want to mess with that code right now. If we really prefer `//JAVA` we just need to make that code a bit smarter. But we can also use `//RELEASE` or `//VERSION` perhaps.

Anyway, this PR is now feature-complete I think.

(Being able to bootstrap Jbang itself without having Java installed is for another PR)

We now only need tests.

maxandersen:
hmm...surprised the code can't handle `//JAVA` - we should fix that ;) ...is it anything different than just require there is a whitespace after the prefix ?

quintesse:
That's what I thought as well, but there are options like `//CDS` that don't have any arguments or people might be using tabs instead of spaces so the code has to be slightly more complex. And I still wasn't sure if you really wanted `//JAVA` or not. If you're decided then I'll of course change it!

quintesse:
Updated to use `//JAVA`

src/main/java/dev/jbang/UnpackUtil.java
```diff
@@ -0,0 +1,153 @@
...
+					if (mode != 0) {
+						Set<PosixFilePermission> permissions = PosixFilePermissionSupport.toPosixFilePermissions(mode);
+						Files.setPosixFilePermissions(entry, permissions);
+					}
+				}
+			}
+		}
+	}
+
+	public static void untargz(Path targz, Path outputDir, boolean stripRootFolder) throws IOException {
```
maxandersen, line 67: did you write these methods now or copied from somewhere ? if copied lets make sure we get link to their origins (can be just)

maxandersen:
for "fun" I added this to a script: `//JAVA 145` and it did this:

```
jbang hello.java
Downloading JDK 145...
Hello World
```

pretty sure there is some missing error handling/fallback info message missing ;)

maxandersen:
and doing this:

```
❯ jbang --java 14 hello.java
Downloading JDK 14...
Installing JDK 14...
[jbang] Building jar...
java.io.IOException: Cannot run program "/Users/max/.jbang/cache/jdks/14/bin/javac": error=2, No such file or directory
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1128)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
	at dev.jbang.cli.Run.build(Run.java:165)
	at dev.jbang.cli.Run.doCall(Run.java:84)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:46)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1853)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2255)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2249)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2213)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2080)
	at picocli.CommandLine.execute(CommandLine.java:1978)
Caused by: java.io.IOException: error=2, No such file or directory
	at java.base/java.lang.ProcessImpl.forkAndExec(Native Method)
	at java.base/java.lang.ProcessImpl.<init>(ProcessImpl.java:340)
	at java.base/java.lang.ProcessImpl.start(ProcessImpl.java:271)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1107)
```

maxandersen:
cause of that last error is that the layout of a the jdk downloaded is an osx app not a traditional jdk..which is surprising.

```
tree -L 3 /Users/max/.jbang/cache/jdks/14
/Users/max/.jbang/cache/jdks/14
└── Contents
    ├── Home
    │   ├── bin
    │   ├── conf
    │   ├── include
    │   ├── jmods
    │   ├── legal
    │   ├── lib
    │   ├── man
    │   └── release
    ├── Info.plist
    └── MacOS
        └── libjli.dylib

10 directories, 3 files
```

src/main/java/dev/jbang/JdkManager.java
```diff
@@ -0,0 +1,75 @@
...
+			return Collections.emptySet();
+		}
+	}
+
+	public static boolean isInstalledJdk(int version) {
+		return Files.isDirectory(getJdkPath(version));
+	}
+
+	private static Path getJdkPath(int version) {
+		return getJdksPath().resolve(Integer.toString(version));
```
maxandersen, line 69: i'm wondering if we should make it `<cache>/jdk/<version>/default/` or similar for now? in case we want to support `//JAVA 11.0.4` notion and then don't have a folder to store that in ?

maxandersen:
does look like we'll need to add osx to the testcase runs as it will be different between linux/windows :)

maxandersen:
note, sdkman openjdk jdks doesn't have this layout thus  i'm wondering if we should keep it or just unzip Contents/Home directly in case of osx platform ? then it would look like a linux/windows style jdk.

src/main/java/dev/jbang/UnpackUtil.java
quintesse: It's copied from code we wrote for the Launcher (which , if I remember correctly comes partly from blog posts))

quintesse:
> note, sdkman openjdk jdks doesn't have this layout thus i'm wondering if we should keep it or just unzip Contents/Home directly in case of osx platform ? then it would look like a linux/windows style jdk.

That might be the simplest solution perhaps, yes.

src/main/java/dev/jbang/JdkManager.java
quintesse: Not sure, with what intention exactly? In case we want multiple v14 JDKs installed? Couldn't we than just switch to naming the directories with their full versions? Choosing the correct folder could easily be done in code.

quintesse:
> for "fun" I added this to a script: `//JAVA 145` and it did this:
> 
> ```
> jbang hello.java
> Downloading JDK 145...
> Hello World
> ```
> 
> pretty sure there is some missing error handling/fallback info message missing ;)

Yeah, that's this part of the code:

https://github.com/jbangdev/jbang/pull/178/files#diff-fb5e4f98d4678cd214b6b3686b78fddaR430

```
		} catch (IOException e) {
			// Unable to find/install Java. Should we tell the user or just hope for the
			// best?
		}
```

quintesse:
@maxandersen can you try if it works now on a Mac?

maxandersen:
> @maxandersen can you try if it works now on a Mac?

it doesn't - investigating.

src/main/java/dev/jbang/Util.java
```diff
@@ -76,8 +82,36 @@ static public void writeString(Path toPath, String scriptText) throws IOExceptio
...
+		}
+	}
+
 	public static boolean isWindows() {
-		return System.getProperty("os.name").toLowerCase().contains("windows");
+		return getOS() == OS.windows;
+	}
+
+	public static boolean isMac() {
+		return getOS() == OS.windows;
```
maxandersen, line 114: this aint right ;)

maxandersen:
```suggestion
		return getOS() == OS.mac;
```

fixing that then download for mac works

maxandersen:
fixin the mac bug above then calling "jbang --java 14 hello.java" downloads and runs....but doing it at second time - the jdk14 is downloaded again...

src/main/java/dev/jbang/JdkManager.java
maxandersen: you would scan the folder for "latest" 14 version for the default version?

src/main/java/dev/jbang/Script.java
```diff
@@ -183,6 +185,21 @@ public boolean enableCDS() {
 		return !collectRawOptions("CDS").isEmpty();
 	}
 
+	public String javaVersion() {
+		List<String> opts = collectRawOptions("JAVA");
+		if (!opts.isEmpty()) {
+			// If there are multiple //JAVA_VERSIONs we'll use the last one
```
maxandersen, line 191:
```suggestion
			// If there are multiple //JAVAs we'll use the last one
```

why not the first one ? thats the logic for others if I remember right?

src/main/java/dev/jbang/Script.java
```diff
@@ -330,6 +356,10 @@ public void createJarFile(File path, File output) throws IOException {
 		if (mainclass != null) {
 			manifest.getMainAttributes().put(Attributes.Name.MAIN_CLASS, mainclass);
 		}
+		if (buildJdk > 0) {
+			String val = buildJdk >= 9 ? Integer.toString(buildJdk) : "1." + buildJdk;
```
maxandersen, line 360: why do we care about 1.x here now ? shouldn't it still be 8 instead 1.8 ?

maxandersen:
> fixin the mac bug above then calling "jbang --java 14 hello.java" downloads and runs....but doing it at second time - the jdk14 is downloaded again...

couldn't reproduce this reliably. assuming user error for now.

maxandersen:
should `jbang --java 13+ hello.java` when I have java 14 installed not use java 14? right now it seem to download the exact version listed.

maxandersen:
btw. I did a `jbang --java 13 hello.java` and press ctrl+c ...then on second run I get:

```
 jbang --verbose --java 13+ hello.java
Downloading JDK 13...
Installing JDK 13...
Unable to download or install JDK: null
[jbang] run: java -classpath :/Users/max/.jbang/cache/jars/hello.java.fd8089dc3b98d346a3cab6a694d7bc836f15297338438e33ae67adbcdc8f3cf0.jar hello
Hello World
```

 we are missing some checksum/sanity checking

src/main/java/dev/jbang/cli/Jdk.java
```diff
@@ -0,0 +1,49 @@
...
+		} else {
+			System.err.println("JDK " + version + " is already installed");
+		}
+		return CommandLine.ExitCode.SOFTWARE;
+	}
+
+	@CommandLine.Command(name = "list", description = "Lists installed JDKs.")
+	public Integer list() throws IOException {
+		PrintWriter out = spec.commandLine().getOut();
+		JdkManager.listInstalledJdks().forEach(jdk -> out.println(jdk));
```
maxandersen, line 35: should use info. it needs to goto err to not conflict with wrapper script

src/main/java/dev/jbang/JdkManager.java
```diff
@@ -0,0 +1,75 @@
...
+	public static Path getInstalledJdk(int version) {
+		Path jdkDir = getJdkPath(version);
+		if (!Files.isDirectory(jdkDir)) {
+			jdkDir = downloadAndInstallJdk(version, false);
+		}
+		return jdkDir;
+	}
+
+	public static Path downloadAndInstallJdk(int version, boolean updateCache) {
+		System.err.println("Downloading JDK " + version + "...");
```
maxandersen, line 34: use the info/warn methods

src/main/java/dev/jbang/JdkManager.java
```diff
@@ -0,0 +1,75 @@
...
+		String url = String.format(JDK_DOWNLOAD_URL, version, Util.getOS().name(), Util.getArch().name());
+		Path jdkDir = getJdkPath(version);
+		try {
+			Path jdkPkg = Util.downloadAndCacheFile(url, updateCache);
+			System.err.println("Installing JDK " + version + "...");
+			UnpackUtil.unpack(jdkPkg, jdkDir);
+			return jdkDir;
+		} catch (IOException e) {
+			Util.deleteFolder(jdkDir, true);
+			System.err.println("Unable to download or install JDK: " + e.getMessage());
```
maxandersen, line 44: use error it will handle verbose showing of exception too

maxandersen:
I fixed the mac issue, rebased and merged to master. 

Issues identified/commented on I opened separate issues for we can fix seperatley.

src/main/java/dev/jbang/JdkManager.java
quintesse: Well, what is "the default version"? Either you let people decide (which might mean creating a `current` link, but at what level? Just the `current` global Java version or the current 14 version and have possibly multiple "current"s?) or you indeed scan for the latest,

Personally, I wouldn't mind taking on SDKMAN a bit here and let people define a single "current" version, eg: `jbang sdk default 14.0.2`, we add a `jbang jdk home` that returns the path to the current version (which uses a link so will always be `~/.jbang/cache/jdk/current`). People can use that to add `export PATH=$(jbang jdk home):$PATH` to their `.bashrc` files.

src/main/java/dev/jbang/Util.java
quintesse: Oops hehe

src/main/java/dev/jbang/Script.java
quintesse: Is it? I didn't realize that. It just seemed more logical that way, but I don't mind one way or the other.

src/main/java/dev/jbang/Script.java
quintesse: AFAIK Java 8 is still numbered 1.8 when you ask for it's version, right?

quintesse: (This is an "official" field used by Maven for example, so I'm following what it does)

quintesse:
> couldn't reproduce this reliably. assuming user error for now.

You might have run the tests in between? The unit tests wipe the cache folder (which is something we should fix because it bugs the heck out of me :-) )

quintesse:
> we are missing some checksum/sanity checking

I did recently add something for this in one of my PRs, not sure if it was this one.

src/main/java/dev/jbang/cli/Jdk.java
quintesse: Done

src/main/java/dev/jbang/JdkManager.java
quintesse: Done

src/main/java/dev/jbang/Script.java
quintesse: Conversation moved to Issue #183

PR #179, opened 2020-07-16
Implemented implicit catalogs

### LINKED ISSUE #172: Extend Catalog system with implicit catalogs
Right now Catalogs need to be explicitly added. This issue is for introducing "implicit catalogs" where, if the user tries to run `some-alias@some-catalog` and the "some-catalog" is not defined Jbang will look in some predefined places to see if that catalog exists.

Where to look? Well `some-catalog` can either be an `organization/repository` pair which will look for a `jbang-catalog.json` file in the root of that repository on either GitHub, GitLab or BitBucket. Or it can simply be the `organization` part and the repository will then be assumed to be "jbang-catalog".

To prevent having to go to the process of looking up repositories each time we should add the first successful lookup as an explicit Catalog. (So if `some-alias@some-catalog` does not exist, but looking up `some-catalog/jbang-catalog/jbang-catalog.sjon` on Github gives a result we should basically perform a `jbang catalog add some-catalog https://github.com/some-catalog/jbang-catalog/jbang-catalog.json` so any references to `some-catalog` will immediately resolve to the correct URL in the future)

quintesse:
Fixes #172


quintesse:
Really cool to just be able to do `jbang echo@quintesse foo` without the need for any configuration :-)

quintesse:
No (it)tests have been added because it all has to do with being able to access external resources. We should probably first get the `jbangdev/jbang-catalog` set up and use that (known to be available) resource in the tests.

Still, it should be safe to merge :-)

maxandersen:
> No (it)tests have been added because it all has to do with being able to access external resources. We should probably first get the `jbangdev/jbang-catalog` set up and use that (known to be available) resource in the tests.

https://github.com/jbangdev/jbang-catalog exists now - lets get that created/setup and add tests to this PR before merging :)

quintesse:
> jbangdev/jbang-catalog exists now 

Great!

Ok, one thing, I was thinking that perhaps we should support a bit more structure to those jbang-catalog repositories (you mentioned that as well IIRC). And think about that before we start setting things up in a way that's hard to change afterwards.

One thing is to be able to select sub-folders within the repo, but keep it as minimalistic as possible so the repo references by default are as simple as possible, meaning it would be great if `jbang alias@jbangdev` were still the default in most cases.

As a format we could do something like this: `alias@repo#folder`.

(I considered simply doing `alias@repo/folder` but then it becomes difficult to know what is a repo name, branch name or folder)

One thing I saw from other projects that have inline catalogs is the concept of `stable`, `unstable` and `experimental`. Now I wouldn't want to impose any specific system, but imagine that by default we look both in the root of the repo and in the `stable` subfolder. That way people can still come up with a bunch of other categories which they could specify manually.

So with this idea the following would all be equivalent:

 - `jbang run alias@implicit`
 - `jbang run alias@implicit/jbang-catalog`
 - `jbang run alias@implicit/jbang-catalog/master`
 - `jbang run alias@implicit#stable`
 - `jbang run alias@implicit/jbang-catalog#stable`
 - `jbang run alias@implicit/jbang-catalog/master#stable`

So imagine that we use the `stable` folder in `jbangdev/jbang-catalog` for out stable scripts, but for the tests we could for example put an `ci-test-foo.java` script in the `ci` subfolder with its own catalog and be able to run `ci-test-foo@jbangdev#ci`.

It might seem complicated but remember that in most cases people will just be doing:

 - `jbang run alias@implicit`

or perhaps:

 - `jbang run alias@implicit#experimental`

maxandersen:
i would say for the usecase of experimental/stable/etc. using the support for branches seems much more relevant imo.

I'm struggling to see the issue/usecase beyond that though...got links to those other catalogs that decided to use subfolders  instead of branches ? 

p.s. use of # won't work as its a comment in bash/zshs...

quintesse:
Out of experience with the Launcher (which also uses a catalog repo) I can say that we tried branches (in our case for development, staging and production) and moved away from that again. It makes management of promoting versions from dev->staging->prod harder because you have to do it cross-branch so you lose the connection (they'd end up as separate unrelated commits).

And as an example you can look at IBM's Appsody Stacks, something we actually started looking at using when still working on the Launcher: https://appsody.dev/stacks , docs: https://appsody.dev/docs/stacks/publish/ and their Git repo: https://github.com/appsody/stacks

quintesse:
And I always forget about the `#`! So perhaps `~` could be used.

maxandersen:
thus your proposed syntax is something like:

`<aliasname>@<org/user>(/<reponame>(/branch|sha1)?)(~path)`

and then path must have jbang-catalog.json ?

That could probably work.

I'm wondering if we should look for `.jbang-catalog` instead to make it fit #180 ?

maxandersen:
hmm - that was a bad merge...trying to fix ;)

maxandersen:
i reverted my attempt on merging this with master - probably simplest if you do the rebase/merge as I struggled figur e out which method to keep/delete ;)

quintesse:
> thus your proposed syntax is something like:
> 
> `<aliasname>@<org/user>(/<reponame>(/branch|sha1)?)(~path)`

Yes, although I don't think we can use the sha1 format.
(You have to remember that we determine the path to the scripts relative to the path to the catalog file, so a path with an sha1 in it is not very useful to determine where to find those scripts. Or do you see a use-case for this?)

> and then path must have jbang-catalog.json ?
> 
> That could probably work.

Indeed.

> 
> I'm wondering if we should look for `.jbang-catalog` instead to make it fit #180 ?

Dunno, not convinced yet of the naming you suggest there :-)

maxandersen:
About sha1 - What is /master and /branch if not just symbolic named sha1's?

quintesse:
> About sha1 - What is /master and /branch if not just symbolic named sha1's?

Hmm true, shit I already threw that code away ><

quintesse:
Ok, the IDE had it still in its local history :-)

PR #189, opened 2020-07-20
Make sure we really delete JDK folder if download or install fails

quintesse:
Doing this because I think it's important to at least try to clean up, even in case of a low memory error or something like that.


maxandersen:
this ain't good imo. 

if you want it to be clean then unzip to a temporary location/different name which you always delete/ignore on consecutive runs.

quintesse:
This won't apply cleanly anyway anymore so I'll close this, but can you tell me why you don't like this?
IMO it's always better to try to cleanup _now_ instead of next time.
I do agree that first unpacking to a temporary directory is a good idea though.

PR #195, opened 2020-07-21
Make JDK installs safer using more "atomic" operations

quintesse:


quintesse:
This should be safe to merge @maxandersen , it should make installing JDKs much "safer", meaning errors during the process of downloading/installing shouldn't leave us with a unrecoverable state.

We could even go a step further in the future by first trying to run `java -version` in the temporary folder and only rename it to its final name when that succeeds.

PR #198, opened 2020-07-22
WIP: Implemented bootstrapping of Jbang when Java is not available

### LINKED ISSUE #197: Make Jbang able to bootstrap itself without Java
This is about being able to download and run Jbang without having Java installed, The idea is that if no Java is installed we'd download it from AdoptOpenJDK similar to #133.

This issue is not exactly the same as #135 because that goes a step further with Jbang itself possibly not being installed. But this issue would probably make solving that issue easier.

Possible solutions:

 1. Make the current Linux/Mac/Windows Jbang bootstrap scripts download Java and install according to the same system introduced by #133 
 2. Create a native bootstrap script (Quarkus or Graal native for example)
 3. Make Jbang entirely native (also using Quarkus or Graal native)

(The problem with 1. is that doing on that for Windows will be far from trivial)

quintesse:
Fixes #197

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -1,35 +1,91 @@
 #!/usr/bin/env bash
 
+# The Java version to install when it's not installed on the system yet
+javaVersion=11
+
 absolute_path() {
-    # if the given path to the jbang launcher is absolute (i.e. it is either starting with / or a
-    # 'letter:/' when using gitbash on windows) it is returned unchanged, otherwise we construct an absolute path
-    [[ $1 = /* ]] || [[ $1 =~ ^[A-z]:/ ]] && echo "$1" || echo "$PWD/${1#./}"
+  # if the given path to the jbang launcher is absolute (i.e. it is either starting with / or a
+  # 'letter:/' when using gitbash on windows) it is returned unchanged, otherwise we construct an absolute path
+  [[ $1 = /* ]] || [[ $1 =~ ^[A-z]:/ ]] && echo "$1" || echo "$PWD/${1#./}"
 }
 
 resolve_symlink() {
-    if [[ $OSTYPE != darwin* ]]; then minusFarg="-f"; fi
-    sym_resolved=$(readlink ${minusFarg} $1)
+  if [[ $OSTYPE != darwin* ]]; then minusFarg="-f"; fi
+  sym_resolved=$(readlink ${minusFarg} $1)
 
-    if [[ -n $sym_resolved ]]; then
-        echo $sym_resolved
-    else
-        echo $1
-    fi
+  if [[ -n $sym_resolved ]]; then
+    echo $sym_resolved
+  else
+    echo $1
+  fi
 }
+
+case "$(uname -s)" in
+  Linux*)   os=linux;;
+  Darwin*)  os=mac;;
+  CYGWIN*)  os=windows;;
+  MINGW*)   os=windows;;
+  *)        echo "Unsupported Operating System"; exit 1;;
+esac
+
+case "$(uname -p)" in
+  *64*)     arch=x64;;
+  unknown)
+    case "$(uname -s)" in
+      *64*)   arch=x64;;
+      *)      arch=x32;;
+    esac
+    ;;
+  *)        arch=x32;;
+esac
 
 abs_jbang_path=$(resolve_symlink $(absolute_path $0))
 
 ## when mingw/git bash or cygwin fall out to just running the bat file.
-if [[ $(uname) == CYGWIN* ]] || [[ $(uname) == MINGW* ]]; then
-$(dirname $abs_jbang_path)/jbang.cmd $*
-exit $?
+if [[ $os == windows ]]; then
+  $(dirname $abs_jbang_path)/jbang.cmd $*
+  exit $?
 fi
 
 ## resolve application jar path from script location
 jarPath=$(dirname $abs_jbang_path)/jbang.jar
 
+if [[ -z "$JBANG_DIR" ]]; then JBDIR="$HOME/.jbang"; else JBDIR="$JBANG_DIR"; fi
+if [[ -z "$JBANG_CACHE_DIR" ]]; then TDIR="$JBDIR/cache"; else TDIR="$JBANG_CACHE_DIR"; fi
+
+mkdir -p "$TDIR/jdks"
+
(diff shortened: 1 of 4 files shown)


quintesse:
Right now it works on Linux.
It should work on Mac (can you test @maxandersen ?)
It still needs some work for Windows.

quintesse:
Windows now works as well, but the cmd file is fugly!

quintesse:
So this is actually a pretty doable solution, the scripts haven't even become _that_ much more complex. But the scripts are pretty dumb of course, so if someone who has no Java installed runs `jbang -j 13 some-script-url` or  `jbang url-to-java13-script` then we'll first download 11 only to discover we really need 13 and download that too. But I don't see an easy solution for that.

maxandersen:
testing on osx i get this:

```
jbang --version
Downloading JDK 11...
Installing JDK 11...
tar: Error opening archive: Unrecognized archive format
mv: rename /Users/max/.jbang/cache/jdks/11.tmp/Contents/Home/* to /Users/max/.jbang/cache/jdks/11.tmp/*: No such file or directory
/Users/max/code/personal/jbang/build/install/jbang/bin/jbang: line 98: /Users/max/.jbang/cache/jdks/11/bin/java: No such file or directory
```

src/main/scripts/jbang
```diff
@@ -1,35 +1,91 @@
...
 
+case "$(uname -s)" in
+  Linux*)   os=linux;;
+  Darwin*)  os=mac;;
+  CYGWIN*)  os=windows;;
+  MINGW*)   os=windows;;
+  *)        echo "Unsupported Operating System"; exit 1;;
+esac
+
+case "$(uname -p)" in
```
maxandersen, line 31: on os `uname -m` give the arch...

currently this detects 32 bit as it returns i386.

src/main/scripts/jbang
```diff
@@ -1,35 +1,91 @@
...
-if [[ -z "$JAVA_HOME" ]]; then JAVA_EXEC="java"; else JAVA_EXEC="$JAVA_HOME/bin/java"; fi
+if [[ -z "$JAVA_HOME" ]]; then
+  if [ -x "$(command -v javac)" ]; then
+    JAVA_EXEC="java";
+  else
+    if [ ! -d "$TDIR/jdks/$javaVersion" ]; then
+      echo "Downloading JDK $javaVersion..."
+      url="https://api.adoptopenjdk.net/v3/binary/latest/$javaVersion/ga/$os/$arch/jdk/hotspot/normal/adoptopenjdk"
+      if [ -x "$(command -v curl)" ]; then
+        curl -sL -o "$TDIR/bootstrap-jdk.tgz" $url
```
maxandersen, line 67: i would put a --progress-bar here or similar sas te wait can be quite long ;/

src/main/scripts/jbang
```diff
@@ -1,35 +1,91 @@
...
+      else
+        echo "Error: curl or wget not found, cannot download JDK"
+        exit 1
+      fi
+      echo "Installing JDK $javaVersion..."
+      rm -rf "$TDIR/jdks/$javaVersion.tmp/"
+      mkdir -p "$TDIR/jdks/$javaVersion.tmp"
+      tar xf "$TDIR/bootstrap-jdk.tgz" -C "$TDIR/jdks/$javaVersion.tmp" --strip-components=1
+      if [[ $os == mac ]]; then
+        mv "$TDIR/jdks/$javaVersion.tmp/Contents/Home/*" "$TDIR/jdks/$javaVersion.tmp/"
```
maxandersen, line 79: the quotes here disable expansion so it doesan't match anything.

maxandersen:
```suggestion
        mv "$TDIR/jdks/$javaVersion.tmp/Contents/Home/"* "$TDIR/jdks/$javaVersion.tmp/"
```

src/main/scripts/jbang
quintesse: Is this something you would be able to fix/implement?

src/main/scripts/jbang
quintesse: We could, but I was trying to keep the experience the same. Wget and Curl have different ways of showing progress and the progress bar on Windows makes the download a magnitude slower, it's insane.

maxandersen: mkay ;/

src/main/scripts/jbang.cmd
```diff
@@ -20,20 +20,45 @@ set OUTPUT=
...
+        powershell -NonInteractive -Command "Move-Item %%d\* !TDIR!\jdks\%javaVersion%.tmp"
+	  )
+      ren "%TDIR%\jdks\%javaVersion%.tmp" "%javaVersion%"
+    )
+    set JAVA_HOME=%TDIR%\jdks\%javaVersion%
+    set JAVA_EXEC=!JAVA_HOME!\bin\java.exe
+  )
+) else (
+  set JAVA_EXEC="%JAVA_HOME%\bin\java.exe"
+)
```
maxandersen, line 51: testing on windows i get: "INFO: Could not find files for the given pattern(s)." somewhere in this block (windows won't tell which)

src/main/scripts/jbang
quintesse: Until we find a nicer option we can perhaps add a warning, eg: "Downloading JDK 11. Be patient, this might take several minutes..."

src/main/scripts/jbang
quintesse: Ok fixed

src/main/scripts/jbang.cmd
quintesse: This should work now.

src/main/scripts/jbang
maxandersen: yeah - that wuold help.

src/main/scripts/jbang
maxandersen: I can; but checking on ubuntu `uname -m` works too ....any reason to use uname -p?

maxandersen: never mind; adjusted the code to fallthrough various combiations of `uname -` so it will work more places. just pushed the commit - let me know wyt

src/main/scripts/jbang.cmd
maxandersen: it seems to do ;)

PR #202, opened 2020-07-24
Show stacktrace when implicit download attempts fail and verbose is on

quintesse:
Using this for debugging now, but perhaps it's useful to merge into master?


dem2k:
if verbose is off and exception is thrown so no message at all is shown. not?

quintesse:
> if verbose is off and exception is thrown so no message at all is shown. not?

Yes, it's completely normal for that method to fail, the user doesn't even need to be awate in general. But if you really want to know what's going on verbose mode will tell you.

maxandersen:
merged, but not fund of verbose == debug thus for now calling it verboseMsg... we could make it proper debug if we added --verbose=debug or something.

quintesse:
Yeah, I had the same doubt ;-)

Issue #203, opened 2020-07-24
Allow user to select the "vendor" when installing JDKs


quintesse:
AdoptOpenJDK also maintains "upstream" builds of JDKs, not just their own. It would be nice if the user could select that.

Things to figure out: should Jbang "know" which vender/versions are available? Should it manage them separately and allow the user to not only select versions but also vendors?

The other option is to just see them as equals and only use the vendor information at download time.

quintesse:
Implemented. Closing.

PR #204, opened 2020-07-24
First stab at allowing Vendor selection.

quintesse:
This is a very simplistic implementation that allows the user to select
the vendor at install time by defining a `JBANG_JDK_VENDOR` env var.

For now it at least makes it possible for those people that really want this.
We should see if this , somewhat hidden, option warrants to be turned into a full-fledged feature.
But that will be a lot more work.

See #203

### DIFF
--- a/src/main/java/dev/jbang/JdkManager.java
+++ b/src/main/java/dev/jbang/JdkManager.java
@@ -12,7 +12,7 @@ import java.util.stream.Collectors;
 import picocli.CommandLine;
 
 public class JdkManager {
-	private static final String JDK_DOWNLOAD_URL = "https://api.adoptopenjdk.net/v3/binary/latest/%d/ga/%s/%s/jdk/hotspot/normal/adoptopenjdk";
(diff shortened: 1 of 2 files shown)


quintesse:
Right now you can't have a single version of two different vendors installed. There's just one version 11, or 14 or NN. You could switch from one to the other by running `JBANG_JDK_VENDOR=xxxx jbang jdk install NN`, but that's all.

quintesse:
Also, the selection of "openjdk" versions is very limited (only 8 and 11 are available and only for 64 bit platforms and no Mac support), so it doesn't seem like a good option to make that vendor the default for any platform.

But it might be a nice option for some people that really really want the TCK certified OpenJDK versions.

maxandersen:
and just to remember next time I forget - reason to not add openjdk as option is that it is limited to java 8 and 11, no osx and no 32 bit builds - meaning the places we actually can enable this will be minimal and add unnecessary complexity - at least at this stage.

PR #205, opened 2020-07-24
Fixes problem with implicit catalogs not working

quintesse:


quintesse:
It's so weird that this change causes these seemingly completely unrelated tests to fail on Windows.
The problem is that those tests try to wipe the cache folder and on Windows they are prevented from doing so because something is still holding a lock somewhere,

quintesse:
Even weirder:

image

We somehow created a folder that we don't have access to??

quintesse:
Wtf?

image

While supposedly having full control?

image

quintesse:
Ok, not even as an Administrator can I access that folder 🤔

quintesse:
Ok, fixed now.

maxandersen:
What was the cause ? A missing close of a file/stream or ?

quintesse:
THe root cause was the fact that when a user passes a script ref, which can be either a file path or a URL I basically just do:

```
File f - new File(ref);
if (f.exists() {
   // is file
} else {
   // is URL
}
```

This works fine on Linux/Mac because there are hardly any characters that are considered illegal so we never noticed that there wasn't a try-catch around that code. But passing a URL to that on Windows immediately results in an error because `http:` is not a proper drive name.

But once that was fixed a different problem popped up with the tests. Some tests pass through that code and left locks open. When other tests then tried to delete those locked files/folders things broke. Again, nothing to worry about on Linux/Mac but fatal on Windows.

This PR fixes both problems.

quintesse:
Oops, wrong button.

PR #206, opened 2020-07-24
Implements different cache types that can be cleared separately.

### LINKED ISSUE #191: jbang cache clear should allow controlling which parts gets nuked
like:

`jbang cache clear` is the big default nuke (implicitly set --all to true)
`jbang cache clear --url --jdk` nukes only url and jdk cache
`jbang cache clear --all --no-jdk`  nukes everything but jdk.

quintesse:
Fixes #191

### DIFF
--- a/src/main/java/dev/jbang/JdkManager.java
+++ b/src/main/java/dev/jbang/JdkManager.java
@@ -70,6 +70,26 @@ public class JdkManager {
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


quintesse:
Rebased.

maxandersen:
merged with karate tests, fix to have it return 0 on exit and support for --no-xxx options

Issue #219, opened 2020-07-28
Make Jbang work just with the startup script (zero install)


quintesse:
Meaning that if no .jar is found the latest version will be downloaded and unpacked in $JBANG_CACHE_DIR.

quintesse:
Was it your intention to close this? The PR hasn't been merged yet afterall...

quintesse:
Oh wait, you did... but without actually merging the PR? Weird.

maxandersen:
Did it manually as it was marked as draft

quintesse:
Ok, that was just because we were still waiting for those DNS names :-)

quintesse:
Btw, if we can agree on which ones we're going to use I can already update those scripts... the suggestion was to use `ps.jbang.dev` and `sh.jbang..dev`, right? Do you agree with that?

maxandersen:
I needed the files to be there to do the setup. 

Yes those are the urls I'm setting up. 

Note: they will be redirects so need to follow them. Ie. Curl -L

quintesse:
Here it is, let me know what you think: https://github.com/jbangdev/jbang/pull/395

PR #220, opened 2020-07-28
Be able to use Jbang while only having the script itself (zero-install)

quintesse:
See #219

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gradle" # See documentation for possible values
+    directory: "/" # Location of package manifests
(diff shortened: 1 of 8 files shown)


maxandersen:
What's the status now we have the stable URL?

quintesse:
Well, in itself it's done. I just don't know what to do with the PowerShell version. It's works perfectly fine but if there isn't a way to pass arguments it's not very useful. So not sure if we should just commit it as-is and hope we come up will some brilliant solution in the future or if we should just remove the PS version. (Btw, running the PS version when you are in the PS doesn't seem slow at all, on the contrary).

maxandersen:
didn't the approaches we discussed on chat allow passing in arguments via PS ?

quintesse:
Ah, this one, right?

`iex "& { $(iwr https://raw.githubusercontent.com/quintesse/jbang/zero_install/src/main/scripts/jbang.ps1) } version"`

So now I remember. The reason this was still marked WiP is because I wanted to have the official `sh.jbang.dev` and `ps.jbang.dev` URLs working so I could add them to the comments in the beginning of the scripts :-)

maxandersen:
Ah! Let me pull on DNS support again :)

quintesse:
Manually merged. Closing.

PR #221, opened 2020-07-28
Remove rogue backtick preventing link from working

headius:
This is good as a link, since it links to example source. However there's a
rogue backtick preventing it from linking to the right place (it appends a
"%60" to the URL). Removing it makes the link work properly.


maxandersen:
@all-contrib please add @headius for @content

maxandersen:
@all-contributors please add @headius for content

PR #228, opened 2020-07-29
WiP: Improve Jbang startup performance

quintesse:
These 2 commits result in about a 17% improvement. It's something , but it's not enough.

Even just asking Jbang for its version 10 times takes almost 7s on my system:

```
$ time zsh -c "repeat 10 ./build/install/jbang/bin/jbang version"
6,72s user 0,51s system 208% cpu 3,475 total
```

(this is with these 2 commits, which actually resulted in a 33% reduction)

### DIFF
--- a/src/main/java/dev/jbang/JavaUtil.java
+++ b/src/main/java/dev/jbang/JavaUtil.java
@@ -4,6 +4,8 @@ import static java.lang.System.getenv;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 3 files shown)


quintesse:
Thanks @maxandersen ... but it said WIP for a reason ;-)

maxandersen:
Revert ?

quintesse:
Hehe no no. I had just wanted to more more things, but what's there works and was tested.

PR #229, opened 2020-07-29
Fixed aliases not picking up pre-defined arguments

quintesse:
Very small fix that prevented pre-defined arguments (and properties) from working


maxandersen:
add a test?

quintesse:
It has a test?

maxandersen:
and this is why I shouldn't rave merge at 2 am.

PR #230, opened 2020-07-31
On untrusted runs ask user action (run once + various trust)

maxandersen:
Until now we've aborted on untrusted runs. That is safe, but it is not great a user experience.

Main reason it was not added is that since jbang are run with console() returning null
as we are in a nested batch script we couldn't reliably ask for input. Plus it is not
great blocking on such question forever.

This commit fixes it by
   - asking user wether he wants to run once or add one of three level of thrust
   - not relying on console but read directly System.in
   - add timeout for the reading (currently 10 seconds)

### DIFF
--- a/src/main/java/dev/jbang/ConsoleInput.java
+++ b/src/main/java/dev/jbang/ConsoleInput.java
@@ -1,0 +1,43 @@
+package dev.jbang;
+
+import java.util.concurrent.*;
+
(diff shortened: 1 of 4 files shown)


maxandersen:
wdyt @quintesse ?

maxandersen:
merged to master

PR #238, opened 2020-08-13
Use Docker ADD to avoid public key errors

gastaldi:
`curl: (6) Could not resolve host: github.com`


gastaldi:
Hm, looks like ADD or COPY is discouraged in this case: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

Let me see if I can find a better alternative

PR #253, opened 2020-08-23
Fix for quoting, including spaces, on Windows

### LINKED ISSUE #245: jbang doesn't work if path to java contains spaces (Windows)
jbang doesn't work if path to java binary contains spaces.

```
$ jbang https://twitter.com/maxandersen/status/1266904846239752192 twitter!
Der Befehl ""C:\Program" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
```
My Java-Location: `C:\Program Files\AdoptOpenJDK\jdk-11.0.8.10-hotspot\bin`
(Sorry, for pasting in German)

quintesse:
Fixes #245 

So the fix for the spaces was simply the fact that spaces were not being escaped.

But the fact that the IT tests didn't work (anymore) was because we had turned on delayed expansion in the batch file. Turns out that having `!` in your command line on WIndows gives all kinds of trouble when you also have delayed expansion turned on. The `!` is basically unescapable so you need to turn delayed expansion off again just before running the jbang java command.

### DIFF
--- a/itests/quoting-win.feature
+++ b/itests/quoting-win.feature
@@ -1,8 +1,6 @@
-@java9orhigher
-Feature: jshell
+Feature: quoting for windows
 
 Background:
-* if (javaversion == 8) karate.abort()
 * if (!windows) karate.abort()
 
 Scenario: check quotes are kept when wrapped with quotes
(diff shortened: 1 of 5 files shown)


maxandersen:
the failure here is because it actually requires higher than java 8 to test as it uses jshell.

maxandersen:
fix here does seem to fix space issue - yay!

PR #263, opened 2020-08-27
Fixes completion command example

aaronwalker:


maxandersen:
thanks @aaronwalker and welcome to the future :)

PR #267, opened 2020-08-28
Added options `--cp` and `--deps`

### LINKED ISSUE #169: Add option `--class-path`, `--cp` to `run` for adding dependencies
We talked about this before @maxandersen and you said you can use the `JAVA_OPTS` for that, which is correct, but I just came up with a use-case which is not easily doable in that way:

When the user wants to add dependencies on Maven artifacts.

I thought of this when getting one of those typical warnings at the startup of some artifact where it says that it's missing a Logger library. And I wondered how you could add that artifact when you do something like `jbang run com.example:my-artifact:1.2`. Well, what about:

`jbang run --cp com.example:runtime-impl:1.2 com.example:my-artifact:1.2`

(We could even permit URLs now that we have the infrastructure in place for references to GAVs and remote JARs)

Edit: I was looking for a better example than the logging, and I think for example JDBC driver artifacts are a typical example of something the user has to add themselves at runtime.

Edi2: Related issue: https://github.com/jbangdev/jbang/issues/165

quintesse:
Fixes #169 

This allows for adding additional classpath entries. Either by referring
to directories with classes or JAR files (`--cp`) or by referring to
Maven GAVs or remote JAR files (`--deps`)

### DIFF
--- a/src/main/java/dev/jbang/Script.java
+++ b/src/main/java/dev/jbang/Script.java
@@ -57,6 +57,9 @@ public class Script {
 
 	private List<String> arguments;
 
+	private List<String> additionalDeps = Collections.emptyList();
+	private List<String> additionalClasspaths = Collections.emptyList();
+
 	private Map<String, String> properties;
 	private List<MavenRepo> repositories;
 
@@ -97,7 +100,7 @@ public class Script {
 
 	public List<String> collectDependencies() {
 		if (forJar()) { // if a .jar then we don't try parse it for dependencies.
-			return Collections.emptyList();
+			return additionalDeps;
 		}
 		// early/eager init to property resolution will work.
 		new Detector().detect(new Properties(), Collections.emptyList());
@@ -112,11 +115,13 @@ public class Script {
 			p.putAll(properties);
 		}
 
-		List<String> dependencies = getLines()	.stream()
+		Stream<String> depStream = getLines()	.stream()
 												.filter(it -> isDependDeclare(it))
 												.flatMap(it -> extractDependencies(it))
-												.map(it -> PropertiesValueResolver.replaceProperties(it, p))
-												.collect(Collectors.toList());
+												.map(it -> PropertiesValueResolver.replaceProperties(it, p));
+
+		List<String> dependencies = Stream	.concat(depStream, additionalDeps.stream())
+											.collect(Collectors.toList());
 
 		return dependencies;
 	}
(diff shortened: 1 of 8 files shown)


src/main/java/dev/jbang/cli/Run.java
```diff
@@ -76,6 +76,12 @@ boolean debug() {
 			"-n", "--native" }, description = "Build via native-image and run", defaultValue = "false")
 	boolean nativeImage;
 
+	@CommandLine.Option(names = { "--deps" }, description = "Add additional dependencies.")
+	List<String> dependencies;
+
+	@CommandLine.Option(names = { "--cp" }, description = "Add class path entries.")
```
maxandersen, line 82:
```suggestion
	@CommandLine.Option(names = { "--class-path", "-classpath", "-cp",  }, description = "Add class path entries.")
```

I haven't tested if picocli actually support the single dash for long names; so please verify - but the idea here is to have the same names as standard `javac` (related to #148)

src/main/java/dev/jbang/Script.java
```diff
@@ -112,11 +115,13 @@ public Script(String script, List<String> arguments, Map<String, String> propert
...
 
-		List<String> dependencies = getLines()	.stream()
+		Stream<String> depStream = getLines()	.stream()
 												.filter(it -> isDependDeclare(it))
 												.flatMap(it -> extractDependencies(it))
-												.map(it -> PropertiesValueResolver.replaceProperties(it, p))
-												.collect(Collectors.toList());
+												.map(it -> PropertiesValueResolver.replaceProperties(it, p));
+
+		List<String> dependencies = Stream	.concat(depStream, additionalDeps.stream())
```
maxandersen, line 123: this would make the additional deps get added *last*, right ? 

I would think that they should be first to enable overriding/affecting sequence of loading of resources.

maxandersen (changes requested):
a few comments; most important one is that these deps and cps are specified on command line and thus should be considered first imo.

src/main/java/dev/jbang/cli/Run.java
quintesse: It doesn't:

> Picocli supports POSIX clustered short options: one or more single-character options without option-arguments, followed by at most one option with an option-argument, can be grouped behind one '-' delimiter.

I also think jbang's interface is way to different already to be a drop-in replacement. So having `--cp` instead of `-cp` doesn't feel too bad IMO.

If that's really what you want I think it would be better to come up with a special command (`jbang java`?) that does completely customized parsing and copies the `java` command as much as possible.

src/main/java/dev/jbang/Script.java
quintesse: Aha, yes of course, in my head the later ones would override the earlier ones, but in a class path that doesn't make sense.

src/main/java/dev/jbang/cli/Run.java
maxandersen: if not easily feasible to get `-cp` then its fine; close enough is fine :)

maxandersen:
I've rebased with latest master, added --class-path and flipped the sequence.

thanks!

quintesse:
Ok! Sorry I didn't have time yesterday to do it myself.

PR #275, opened 2020-09-01
Update 'Developer workspace' badge image

sunix:
Using https://www.eclipse.org/che/contribute.svg
See eclipse/che#17758


maxandersen:
Merged. But can't we put in a redirect ?

sunix:
> Merged.

Thanks!

> But can't we put in a redirect ?

Probably, I tried to discuss for a better transition but it didn't work out

PR #280, opened 2020-09-05
fix docs typos and asciidoctor styling

abelsromero:
I read throught de contributing and readme docs to fix some typos and asciidoctor formatting.
I have avoided any major text changes other than some commas and minnor changes.

Also:
* Corrected kubectl plugins automatic casing example based on a test run in my machine. The docs say it generates  `KubectlMyPlugin` from `kubectl-myplugin` but it generates ` `KubectlMyplugin`. Not sure if maybe is better to use `kubectl-my-plugin` and keep `KubectlMyplugin` output?

There are two things which I had no clear how to fix:

1. The example in *Using `.jsh` for `jshell`* is not clear to me. Specially about how to pass properties.
Maybe should be replaced with something like
  ```java
  System.out.println("Hello " + (args.length>0?args[0]:"World")); // <.>
  System.out.println(System.getProperty("key"); // <.>
  ```
And be accompany by the input and ouput ?
  ```
  jbang -Dkey=value World
  ```
  ```
  Hello World
  value
  ```

2. Return param in section *Build Integration [EXPERIMENTAL]* is missing some text.

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -1,7 +1,5 @@
-
 = j'bang - Having fun with Java scripting
-:toc:
-:toc-placement!:
+:toc: macro
 ifndef::env-github[:icons: font]
 ifdef::env-github[]
 :caution-caption: :fire:
@@ -62,7 +60,7 @@ toc::[]
 * Compiled jar and Dependency resolution caching
 * (EXPERIMENTAL) native-image generation (`--native`)
 * Launch with debug enabled for instant debugging from your favorite IDE
-* Transparent launch of JavaFX Applications on Java 8 and higher.
+* Transparent launch of JavaFX Applications on Java 8 and higher
 * Can be used for writing plugins to other cli's like `kubectl`
 * Init templates to get started easily (`jbang init -t cli hello.java`)
 * Generate gradle and IDE config with dependencies for easy editing in your favorite IDE (`jbang edit myfile.java`)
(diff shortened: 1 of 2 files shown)


readme.adoc
```diff
@@ -639,69 +635,68 @@ See `link:examples/jfx.java[]` and `link:examples/jfxtiles.java[]` for examples
...
 
-If jbang finds a java executable using `JAVA_HOME` or `'PATH` which satisfies the stated java version jbang will use it.
+If jbang finds a java executable using `JAVA_HOME` or `PATH` which satisfies the stated java version jbang will use it.
 If no such version is found it will automatically download and install it into jbang's cache (`~/.jbang/cache/jdks` by default).
 
 Examples:
 
-`//JAVA 11` will force use of Java 11.
-`//JAVA 13+` will require at least java 13. Java 13 or higher will be sued.
+  //JAVA 11  will force use of Java 11.
```
maxandersen, line 696: do you really mean to have the "will force use of java 11" in code style rather than text ?

maxandersen:
> 1. The example in _Using `.jsh` for `jshell`_ is not clear to me. Specially about how to pass properties.
>    Maybe should be replaced with something like

you are right - not sure what I was thinking writing the original one. can you add that fix to this pr?

maxandersen:
> 1. Return param in section _Build Integration [EXPERIMENTAL]_ is missing some text.

its evolving - it will change quite a few times more thus it will be changed/refreshed soon.

maxandersen:
> * `Not sure if maybe is better to use `kubectl-my-plugin`and keep`KubectlMyplugin` output?

yeah - that would sense. Good catch 👍

maxandersen:
btw. thanks for fantastic grammar check - i'm amazed how many mistakes you found :) 

i'll keep it open for a few days in case you want to add those additional fixes you identified.

readme.adoc
abelsromero: You are right, I will fix it.
What I saw is that it was written in a single line. Is still correct to asume it the sentences should be in 2 lines?

abelsromero:
> > 1. Return param in section _Build Integration [EXPERIMENTAL]_ is missing some text.
> 
> its evolving - it will change quite a few times more thus it will be changed/refreshed soon.

Totally understand but the text is

```
Map<String, Object> map of returns; special keys are "native-image" which is a and "files" to
*          return native-image to be run and list of files to get written to the output directory
```

Something seems to be missing after `"native-image" which is a `, and the rest is not clear to me.

readme.adoc
maxandersen: yes, two lines are fine.

maxandersen:
native-image is a path to the generated binary/native-image

maxandersen:
@all-contributors please add @abelsromero for docs

PR #290, opened 2020-09-09
Install wrapper

quintesse:
This copies the startup files and the JAR file to either the current directory or the specified one.

### DIFF
--- a/itests/karate-config.js
+++ b/itests/karate-config.js
@@ -20,8 +20,8 @@ function fn() {
         // set JBANG_REPO to not mess with users own ~/.m2
         sep = java.io.File.separator;
 
-        !('JBANG_REPO' in env) && (env.JBANG_REPO = sc + sep + "m2")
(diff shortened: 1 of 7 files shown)


quintesse:
One thing, this leaves the original names intact. I'm not sure if we should rename them to `jbangw` and `jbang.cmd` instead. It would at least be consistent with what everybody else is doing. Wdyt?

maxandersen (changes requested):
this looks good - could you add a basic karate test of install and run that ./jbang from that directory ?

maxandersen:
> One thing, this leaves the original names intact. I'm not sure if we should rename them to `jbangw` and `jbang.cmd` instead. It would at least be consistent with what everybody else is doing. Wdyt?

you mean `jbangw` and `jbangw.cmd` ? no - i don't think thats valid in this case as its literally still jbang...its just installed where it picks up jar differently.

quintesse:
Tests added!

PR #298, opened 2020-09-11
Fixes #283 "javac not found" error

### LINKED ISSUE #283: javac fails badly when just a jre installed and it does not honor //JAVA or --java
i installed fedora workstation 32. it comes with jre 1.8 (sigh) by default which confuses jbang ;) 

it errors hard as no javac present; that would be kinda okey if it wasn't so that when using `jbang --java 11 properties@jbangdev` then it still fails with javac missing albeit java 11 just got downloaded.

something mixed up with used paths or so.

quintesse:
That was a lot of work for a single-word fix! :roll_eyes: 

For #283

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -114,7 +114,7 @@ if [[ -z "$JAVA_EXEC" ]]; then
       # Activate the downloaded JDK giving it its proper name
       mv "$TDIR/jdks/$javaVersion.tmp" "$TDIR/jdks/$javaVersion"
     fi
-    JAVA_HOME="$TDIR/jdks/$javaVersion"
+    export JAVA_HOME="$TDIR/jdks/$javaVersion"
     JAVA_EXEC="$JAVA_HOME/bin/java"
   fi
 fi


quintesse:
@maxandersen not sure how to add a test for this, any ideas?

maxandersen (approved):
you could copy jdk to folder, remove javac and set the path for that ?

quintesse:
> you could copy jdk to folder, remove javac and set the path for that ?

But then it would never work, so that would not work for testing the fix, right?

maxandersen:
Wdym? Thats the issue I had. Java is there but no javac...if your fix fixes it it should downloadjava 11 and run

quintesse:
That can't have been the problem because Jbang _explicitely_ checks for the existence of `javac`, the fix is because it wasn't correctly told where to find the downloaded version.

quintesse:
But anyway, I thought we weren't downloading JDKs in out tests because they can't be run on the CI server?

quintesse:
Let me explain myself better:

I _know_ the problem isn't having a JRE installed. The problem occurs exactly the same on a system that has NO Java installed.
The problem is the one I fixed: that if you have no `javac` (for what reason whatsoever) the Jbang startup script would download a JDK and it would set the `JAVA_HOME`, _but_ ... it wouldn't `export` it, so the Jbang Java code would go to it's fallback mode and just assume that `javac` was on the PATH.

So the way this can easily be tested (which is what I did to reproduce this problem) is to make sure that a) `JAVA_HOME` is not set and b) no Java is available on the PATH.

My problem was that I have no idea how to that with Karate so I didn't write a test for it.

Issue #180, opened 2020-07-17
support "nearest" .jbang-catalog notion 


maxandersen:
Imagine this:

```
$ git clone https://github.com/quarkusio/awesomeproject
$ cd awesomeproject
$ jbang --verbose someautomationscript
Found alias `someautomationscript` in .jbang-catalog.
Running somescript.java
....
```

or 

```
$ cd somewhere
$ jbang --verbose asciidoctor
Found alias `asciidoctor` in `~/.jbang/jbang-catalog`
Running https://...
..
$ jbang --verbose someautomationscript
Found alias `someautomationscript` in .jbang-catalog.
Running somescript.java
....
```

idea is to support localized/nearest in parent alias's that you can use and have default configured in your folders / git projects.

...you can do some crazy things with this when you start thinking about it :)

This might want us to reconsider how the alias format is defined but I think this places nice into the idea of making jbang scripts super easy to consume and use.

could even work well with a  `./jbangw` (wrapper :) to "bootstrap"...

/cc @quintesse

quintesse:
Sounds like a neat idea.

quintesse:
Btw, personally I'd go for `.bang-catalog.json` if you _really_ want to make it hidden. At least it will be somewhat easier to edit. But why do you want to hide the file?

PR #299, opened 2020-09-11
feat: Implemented local catalogs

quintesse:
See #180

Jbang will now look for unqualified aliases in the following order:

1. `./jbang-catalog.json` (non-hidden file!)
2. `./.jbang/jbang-catalog.json` (hidden folder!)
3. The same for any parent folder, recursively upwards
4. And finally `$HOME/.jbang/jbang-catalog.json`

Any relative paths used in those catalog files will be relative to the "base folder", so for (1) it would be `.`, but for (2) as well! The same for any parent folders. And for (4) it would be `$HOME`.

### DIFF
--- a/jbang-catalog.json
+++ b/jbang-catalog.json
@@ -12,14 +12,14 @@
     "vertx": {
       "script-ref": "vertx.java"
     },
-    "classpath_example": {
+    "classpath_example_java": {
+      "script-ref": "classpath_example.java"
+    },
+    "classpath_example_jsh": {
       "script-ref": "classpath_example.jsh"
     },
     "hibernate": {
       "script-ref": "hibernate.java"
-    },
-    "example": {
-      "script-ref": "example"
     },
     "docker": {
       "script-ref": "docker.java"
@@ -45,12 +45,6 @@
     "webdriver": {
       "script-ref": "webdriver.java"
     },
-    "jetty": {
-      "script-ref": "jetty.jsh"
-    },
-    "scripts": {
-      "script-ref": "scripts"
-    },
     "lang": {
       "script-ref": "lang.java"
     },
@@ -64,10 +58,13 @@
       "script-ref": "asciidoc.java"
     },
     "helloworld": {
-      "script-ref": "helloworld.jsh"
+      "script-ref": "helloworld.java"
     },
     "urldep": {
       "script-ref": "urldep.java"
+    },
+    "ilc": {
+      "script-ref": "ilc.java"
     },
     "imap": {
       "script-ref": "imap.java"
@@ -77,6 +74,9 @@
     },
     "asciidoctor": {
       "script-ref": "asciidoctor.java"
+    },
+    "resource": {
+      "script-ref": "resource.java"
     },
     "plantuml": {
       "script-ref": "plantuml.java"
@@ -116,8 +116,11 @@
     },
     "junit-run": {
       "script-ref": "junit-run.java"
+    },
+    "jetty": {
+      "script-ref": "jetty.jsh"
     }
   },
   "base-ref": "examples",
-  "description": "\u0027JBang example scripts\u0027"
+  "description": "Jbang example scripts"
 }
--- a/src/main/java/dev/jbang/AliasUtil.java
+++ b/src/main/java/dev/jbang/AliasUtil.java
@@ -512,7 +512,7 @@ public class AliasUtil {
 				if (as != null) {
 					aliases = as;
 					// Validate the result (Gson can't do this)
-					check(aliases.aliases != null, "Missing required attribute 'aliases'");
+					check(aliases.aliases != null, "Missing required attribute 'aliases' in " + catalogPath);
 					for (String aliasName : aliases.aliases.keySet()) {
 						Alias alias = aliases.aliases.get(aliasName);
 						alias.aliases = aliases;
--- a/src/main/java/dev/jbang/AliasUtil.java
+++ b/src/main/java/dev/jbang/AliasUtil.java
@@ -5,6 +5,7 @@ import java.io.Reader;
 import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.*;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
@@ -18,6 +19,7 @@ import picocli.CommandLine;
 
 public class AliasUtil {
 	public static final String JBANG_CATALOG_JSON = "jbang-catalog.json";
+	public static final String JBANG_DOT_DIR = ".jbang";
 
 	private static final String GITHUB_URL = "https://github.com/";
 	private static final String GITLAB_URL = "https://gitlab.com/";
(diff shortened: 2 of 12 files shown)


maxandersen:
Why not just relative to where the catalog JSON file is ?

zakkak:
> Why not just relative to where the catalog JSON file is ?

I think that would be more intuitive.

jbang-catalog.json
```diff
@@ -12,15 +12,15 @@
     "vertx": {
```
maxandersen, line 12: we really need to move examples out to jbang-examples but thats for a separate issue.

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -47,31 +94,50 @@ static void printAliases(PrintWriter out, String catalogName, AliasUtil.Aliases
 						.stream()
 						.sorted()
```
maxandersen, line 95: should we add a "<path catalog>" somewhere in this print out (at least when --verbose) to let users grok which catalogs are being consulted.

src/main/java/dev/jbang/AliasUtil.java
```diff
@@ -260,49 +423,107 @@ public static Aliases getCatalogAliasesByRef(String catalogRef, boolean updateCa
...
+			catalogCache.put(catalogPath, aliases);
+		} else {
+			aliases = catalogCache.get(catalogPath);
+		}
+		return aliases;
+	}
+
+	static private Path scriptParent(Path script) {
+		Path parent = script.getParent();
+		if (parent.getFileName().toString().equals(JBANG_DOT_DIR)) {
```
maxandersen, line 506: I still think we should have the actual folder of the jbang-catalog being what paths are resolved against. It just feels weird other wise.

can you point to a test/usecase that breaks/doesn't work for this ?

src/main/java/dev/jbang/AliasUtil.java
```diff
@@ -18,33 +19,85 @@
...
+					baseRef += ref.substring(2);
+				} else {
+					baseRef += ref;
+				}
+				ref = baseRef;
+			}
+			if (!isAbsoluteRef(ref)) {
+				Path script = scriptParent(aliases.catalogFile).resolve(ref);
+				if (script.startsWith(cwd) || cwd.startsWith(script)
+						|| !cwd.relativize(script).startsWith("../../..")) {
```
maxandersen, line 73: what does this do ? that smells as some very specific assumption?

maxandersen (commented):
the relative path is not right to me; otherwise the feature seem to works well.

would be nice to have list print out which catalog are being consulted/included.

maxandersen (approved):
so lets get those changes figured out - its definitely a powerful feature!

src/main/java/dev/jbang/AliasUtil.java
quintesse: IT's not a specific assumption so much but more that trying to relativize  paths should come with some kind of limit, otherwise it can easily generate paths like: `../../../../../a/b/c/d/e/f.java` which is exceedingly ugly. So I just made a choice to say that 2 levels of relative paths is okay, more than that we just switch to absolute paths.

src/main/java/dev/jbang/cli/Alias.java
quintesse: I was actually planning on adding a `--show-origin` or `--show-location` flag to the `alias list` command (it's what `git config` does as well). But unfortunately the way I implemented this I can't give a log of how a single alias is found. (All aliases are flattened first before being looked up).

maxandersen: At least knowing the visited catalogs would be helpful. Having it per alias can be future thing if ever asked :)

src/main/java/dev/jbang/AliasUtil.java
zakkak: BTW, If I am not mistaken, this will not work on windows.

quintesse: It's `Path`, not `String`, it will take proper path separators into account (it's the beauty of using `Path`, you don't have to worry about stuff like that)

zakkak: Nice, thanks for the explanation @quintesse :)

maxandersen:
bug?

```
> mkdir .jbang
> touch .jbang/jbang-catalog.json
❯ jbang alias add qs examples/progress.java
❯ jbang qs
[jbang] java.lang.IllegalArgumentException: Could not read script argument qs
```

that shouldn't fail should it ?

maxandersen:
added commit to get better context when parse error.

quintesse:
> that shouldn't fail should it ?

Well, the code to auto-relativize the path to a script wasn't added yet, so it expected you to explicitly put the path relative to the catalog file :-)

I just added that code, so it should work as expected now.

quintesse:
Ok, added some tests as well. If the checks pass amd you don't find more issues this would be ok to merge for me.

quintesse:
Don't merge yet...

quintesse:
Ok, added a better error msg when aliases can't be properly resolved.

maxandersen:
sorry, but with latest I'm now getting this:

```
❯ jbang alias add qs examples/progress.java
❯ jbang qs
[jbang] java.lang.IllegalArgumentException: Alias qs from /Users/max/code/personal/jbang/jbang-catalog.json failed to resolve examples/progress.java
```
and /Users/max/code/personal/jbang/examples/progress.java does exist.

maxandersen (changes requested):
bug still there in different from - that needs fixing.

maxandersen (changes requested):
bug still there in different from - that needs fixing.

quintesse:
I'm not able to reproduce this :-/

maxandersen:
debugging it and actual error is that `examples/progress.java` gets resolved to `examples/examples/progress.java`. checking why that happens.

maxandersen:
okey, so the issue seem to be that it adds the `examples/progress.java` to the catalog but does not take into account its baseref of `examples`.

maxandersen:
I tried modifyingtests in TestAliasNearest.java to have a test (but wont be able to complete it) that has this kind of layout:

```java
static final String baseref = "{\n" +
			"  \"aliases\": {\n" +
			"    \"baseref1\": {\n" +
			"      \"script-ref\": \"baseref1ref\"\n" +
			"    },\n" +
			"    \"dotlocal2\": {\n" +
			"      \"script-ref\": \"dotlocal2inbaseref\"\n" +
			"    },\n" +
			"    \"parent3\": {\n" +
			"      \"script-ref\": \"parent3inbaseref\"\n" +
			"    },\n" +
			"    \"global4\": {\n" +
			"      \"script-ref\": \"global4inbaseref\"\n" +
			"    }\n" +
			"  },\n" +
			"  \"base-ref\": \"test\"\n" +
			"}";
```

notice the base-ref being "test" so everything inside it is relative to "test".

Now put that file in dir outside "test" and things goes weird when you start adding to it.

quintesse:
Aha, yes, very likely that the baseRef isn't taken into account. I'll fix that as soon as I have some time.

quintesse:
Hmm why are the tests failing? :thinking: 
They Work On My Machine(tm)

quintesse:
The report doesn't give much useful information either.

Could you perhaps test on your machine @maxandersen ? Just to see if it something in my environment or more the CI's environment.

maxandersen:
Will try when back later

maxandersen:
the tests only fail in github or using https://github.com/nektos/act
can't figure out why except I can see the scriptRef is wrong when loading catalog.

i've committed a better toString and enbabling of test output in gradle. the latter we should remove before merging though.

quintesse:
Yay! Finally!

maxandersen:
awesome - now I get this though (maybe not new bug ? )

```
jbang --verbose alias add -f=jbang-catalog.json test examples/lang.java
[jbang] Reading aliases from jbang-catalog.json
	at dev.jbang.AliasUtil$Aliases.getScriptBase(AliasUtil.java:126)
	at dev.jbang.AliasUtil.addAlias(AliasUtil.java:391)
	at dev.jbang.cli.AliasAdd.doCall(Alias.java:63)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:67)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:13)
[jbang] If you believe this a bug in jbang open issue at https://github.com/jbangdev/jbang/issues
```

maxandersen:
another weird one:

notice hoave gasearch@jbangdev exists but can't be resolved/found:

```
❯ jbang alias list jbangdev
env@jbangdev = Dump table of Environment Variables
gavsearch@jbangdev = Search search.maven.org for maven artifacts.
hello@jbangdev = Script that says hello back for each argument
properties@jbangdev = Dump table of System properties
❯ jbang gavsearch@jbangdev
[jbang] java.lang.IllegalArgumentException: Alias gavsearch@jbangdev from /Users/max/.jbang/cache/urls/22276b0e71b9dc0e466e9158ad103a5d9175aaadaed34eeb3bb6c8b25b80ddde/jbang-catalog.json failed to resolve gavsearch.java
❯ jbang --verbose gavsearch@jbangdev
[jbang] Downloaded catalog from https://github.com/jbangdev/jbang-catalog/blob/master/jbang-catalog.json
[jbang] Reading aliases from /Users/max/.jbang/cache/urls/22276b0e71b9dc0e466e9158ad103a5d9175aaadaed34eeb3bb6c8b25b80ddde/jbang-catalog.json
[jbang] Reading aliases from /Users/max/.jbang/jbang-catalog.json
java.lang.IllegalArgumentException: Alias gavsearch@jbangdev from /Users/max/.jbang/cache/urls/22276b0e71b9dc0e466e9158ad103a5d9175aaadaed34eeb3bb6c8b25b80ddde/jbang-catalog.json failed to resolve gavsearch.java
	at dev.jbang.cli.BaseScriptCommand.prepareScript(BaseScriptCommand.java:108)
	at dev.jbang.cli.Run.doCall(Run.java:44)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:67)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:13)
[jbang] If you believe this a bug in jbang open issue at https://github.com/jbangdev/jbang/issues
```

quintesse:
Ok, those issues are fixed now and I added some tests for them.

quintesse:
Again Yay! :-)

maxandersen:
verified the two last issues I saw don't happen now - yay! going to sqaush and merge!

Issue #271, opened 2020-08-30
make jbang edit --open=code --live quarkusapp.java work


maxandersen:
at the moment you can't open IDe without using --live; that could be fixed having --open as separate option allowing you to do :

`jbang edit --open app.java` (generate project and open ide without generating anything)

`jbang edit --live app.java` (would not open IDE but just be listening for changes, user manually opens IDE)
﻿
`jbang edit --open=idea --live app.java` (would open IDE and be in live mode)

maxandersen:
this was done in 0.46+

PR #314, opened 2020-09-15
feat(edit): add `--open=[editor] and make --live a boolean

maxandersen:
Currently you could only open editor automatically when using --live and often you don't need live as deps not changing. This change makes --live a boolean toggle and --open can now be used to open project in editor uniformly on all platforms.

BREAKING CHANGE: --live=[editor] will no longer work. Use `--open=[editor] --live` instead.

I'm not happy having to break users using --live but without an answer to 
https://groups.google.com/u/1/g/picocli/c/__tziJIw3lE I dont see a good way to fix it.

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -727,16 +727,24 @@ code `jbang edit helloworld.java`
 
 If you add additional dependencies to your file just re-run the edit command and the relevant files will be regenerated with the updated dependencies.
 
+Above does require using a shell that allows for variable evaluation, if you are on i.e. Windows then you might prefer using:
+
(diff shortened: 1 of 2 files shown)


maxandersen:
fixes to #271

quintesse (approved):
Seems like a good idea. And I don't mind the change in API that much because I doubt it's something that people will use much in scripts, so it's mostly a case about the user having to learn the new way.

gastaldi (approved):
LGTM. I think this is a nice feature and worth the break

PR #320, opened 2020-09-17
feat: Add ability for integration classes to return JVM args and main classes

stuartwdouglas:


stuartwdouglas:
Requires https://github.com/quarkusio/quarkus/pull/12155 for best results

maxandersen:
looks very similar to my own branch for this thus merging this one.

one caveat with the args is that they are not persisted when using jar from cache but I'll fix that in a follow up (idea is I add these "build time" arguments into a manifest field to use on reruns)

Issue #39, opened 2020-01-21
allow for additional source files


maxandersen:
in some cases libraries just need full public classes and with default jbang you can only have one.

would be interesting to allow for something like:

```
//FRIENDS Person.java Order.java
```

that would add these to the java compile source files list. This would also be possible to work with url based by simply fetching them relative to url location.

could even allow for 
```
//FRIENDS *.java
```

but that would NOT work in the url case so i'm hesitant to include eit.

maxandersen:
The separation is there because of what you suggested in the chat - to allow adding .java without having them compiled :)

quintesse:
AHa, didn't remember that :-)

quintesse:
Btw, we could perhaps do the following:

 - `//FILES <file>` for resource files. They also need to be able to be renamed/moved, eg:
 - `//FILES <asfile>=<file>` so we can do `//ADD META-INF/services/some.spi.definition=spi-file`
 - `//SOURCES <source-file>` adds the selected file to the compilation phase
 - `//DEPS <source-file>` which means the targeted source file get built (if needed) and its JAR file gets added to the class path

quintesse:
Although `//DEPS <source-file>` might need a different name. I chose it because it's basically another way of adding a JAR, but perhaps it's difficult to distinguish between direct references to JAR files and references to source files that fist have to be compiled. Of course perhaps looking at the file name would be enough. Wdyt?

maxandersen:
`//FILES <file>` is trivial in latest master and immediately useful.
`//FILES <file>=<mountpoint>` I understand the value but it makes `jbang edit` much more complicated...thus it might get added later once we worked with basic files first and see how far we can push it; we might not need it.
`//SOURCES source-file` +100; the current code is thought *not* ready for this...doable but requires unwinding things.
`//DEPS source-file` ...this is outside the scope of this issue, discussed in #266

maxandersen:
master now has support for:

```
//FILES <file> for resource files. They also need to be able to be renamed/moved, eg:
//FILES <asfile>=<file> so we can do //ADD META-INF/services/some.spi.definition=spi-file
```

quintesse:
Ok, although my assumption was to have the "destination name" be the first one? A bit like if it was an alias (`alias list=ls -als`). Doesn't that seem slightly more intuitive to you?

maxandersen:
Copy from to
Mv from to

Was my thinking...

maxandersen:
[...]

PR #323, opened 2020-09-19
feat: Initial support for multiple sources

maxandersen:
Thie lets you use `//SOURCES <path>` with some restrictions:

1) the reference sources are not processed for //DEPS or any other jbang support (will come)
2) jbang edit does not undestand it yet.
3) No **/* matching. Thus each file must be explicit and relative to the file that refer to them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ jbang-action
 out
 node_modules
 package-lock.json
+*.jfr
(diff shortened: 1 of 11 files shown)


maxandersen:
related to #39

maxandersen:
realized we could get some basic working mulitple source by just adding any //SOURCES as additional files.

wanna find way to solve the `edit` feature - thats trickier.

maxandersen:
Jbang edit works now. At least for files in same directory.

maxandersen:
Jbang edit works now. At least for files in same directory.

maxandersen:
added test for one/two classes.

PR #327, opened 2020-09-20
feat: Added `--quiet` option

quintesse:


maxandersen:
So default is --quit --verbose ?

What output is it that you are trying suppress ?

quintesse:
What do you mean the default? 🤔 
I'm trying to supress _all_ output with `--quiet`, but it's not the default, you'd have to specify the option.

maxandersen:
I mean what happens if you specify both --quiet and --verbose ...that would be the default behavior as it's not verbose nor quiet :)

quintesse:
No, the last one wins

src/main/java/dev/jbang/cli/Jbang.java
```diff
@@ -47,6 +47,12 @@ void setVerbose(boolean verbose) {
 		Util.setVerbose(verbose);
 	}
 
+	@Option(names = { "--quiet",
+			"-q" }, description = "jbang will not show any unnecessary messages.", scope = ScopeType.INHERIT)
```
maxandersen, line 51:
```suggestion
			"-q" }, description = "jbang will be quiet, only print when error occurs.", scope = ScopeType.INHERIT)
```

maxandersen: also, shouldn't --verbose and --quiet be mutual exclusive ? 

see https://picocli.info/#_mutually_exclusive_options on how to do that in picoclil

maxandersen: also, just "--quiet" as "--verbose" doesn't have "-v"

src/main/java/dev/jbang/cli/Jbang.java
```diff
@@ -62,13 +68,10 @@ public static CommandLine getCommandLine() {
 		@Override
 		public int handleExecutionException(Exception ex, CommandLine commandLine, CommandLine.ParseResult parseResult)
 				throws Exception {
+			Util.errorMsg(null, ex);
 			if (Util.isVerbose()) {
-				ex.printStackTrace();
```
maxandersen, line 66: why remove this ?

src/main/java/dev/jbang/cli/BaseCommand.java
```diff
@@ -61,19 +64,13 @@ boolean isVerbose() {
...
 
+	boolean isQuiet() {
+		return Util.isQuiet();
+	}
+
 	@Override
 	public Integer call() throws IOException {
-		try {
-			return doCall();
-		} catch (ExitException e) {
```
maxandersen, line 68: where is this handled if removed from here?

src/main/java/dev/jbang/cli/Jbang.java
quintesse: ok

src/main/java/dev/jbang/cli/BaseCommand.java
quintesse: In the ExitHandler, it contained the exact same code

src/main/java/dev/jbang/cli/Jbang.java
quintesse: Again because `errorMsg()` already does the same thing. Thought it best to remove code duplication

src/main/java/dev/jbang/cli/BaseCommand.java
maxandersen: ah true...

quintesse:
Updated PR with your suggestions

quintesse:
@maxandersen could you perhaps manually restart  the build? There seems to have been some hickup. Because the tests passed before and now, without any code change they fail. Seeing the failed test is something that downloads dependencies perhaps there was some kind of network problem.

maxandersen:
you don't have rerun option ? i wonder what permission is needed for that.

anyways, kicked of a re-run.

quintesse:
Hmm still problems (but different ones *sigh*).

And no, no "rerun" button (or "run" for that matter :-) )

quintesse:
Weird, it's some problem with PicoCli and those exclusive options.

If I type something like `jbang --verbose alias list` it works, but if the `--verbose` is in any other position, eg `jbang alias --verbose list` it results in an NPE.

quintesse:
Ok, found the solution: https://stackoverflow.com/a/58577453

quintesse:
Well... yes it works now, but if you type something like: `jbang --verbose alias list --quiet` PicoCli will happily allow it...

quintesse:
I've opened an issue for it: https://github.com/remkop/picocli/issues/1182

quintesse:
Shall I re-introduce the previous hack where setting verbose unsets quiet and vise-versa? At least until PicoCli has been fixed?

maxandersen:
+1 @quintesse on doing that or try his suggested possible workaruond on the picocli issue ?

quintesse:
I can try, but I don't actually think that his work-around does what we want. AFAIKS that would introduce separate verbose and quiet option on each subcommand and I don't see how they would then be mutually exclusive between them... but I might be misunderstanding the docs.

maxandersen:
easy previous hack is fine then too ;)

quintesse:
Done. Seems we have "an interesting use case" ;-)

maxandersen:
merged as it looked good - i did notice thought that ouptut for SPI's were not kept quiet...not sure if we can...but separate concern/issue.

PR #329, opened 2020-09-20
fix: made downloading more robust

quintesse:
`JdkManager.downloadAndInstall()` now restores better after an error.
`Util.downloadFile()` is now equally robust as the JDK downloader.

I noticed that interrupting a download could leave the cache in an unusable state unless first cleaned.
The JDK downloader was already very careful about first downloading to a temp location and only committing when the download finished correctly, but now the file downloader does this as well.


maxandersen:
i'll need to roll this one back - when we have multiple sources we use the initial cacheurl directory as a placeholder for the other resources.

PR #330, opened 2020-09-20
fix: `jdk uninstall` now firsts checks if JDK exists

quintesse:
Before, if the JDK didn't exist, it would first nicely download the JDK
before deling it ;-)


maxandersen:
that was a useful way to test the download feature extra much ;)

PR #336, opened 2020-09-21
refactor: Now using special exit code to allow printing to stdout

quintesse:
Before we would either print as much output to stderr, which is not
in line with the Rule of Least Surprise, or we had to print the output
as part of a shell command like `echo` or `cat`. Now we have a special
exit code that will tell the startup scripts to either execute Jbang's
output or to simply display it.

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -136,6 +136,8 @@ output=$(CLICOLOR_FORCE=1 ${JAVA_EXEC} ${JBANG_JAVA_OPTIONS} -classpath ${jarPat
 err=$?
 if [ $err -eq 0 ]; then
   eval "exec $output"
+elif [ $err -eq 3 ]; then
+  echo "$output"
 else
   exit $err
 fi
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -134,10 +134,9 @@ export JBANG_FILE="$1"
 ## run it using command substitution to have just the user process once jbang is done
 output=$(CLICOLOR_FORCE=1 ${JAVA_EXEC} ${JBANG_JAVA_OPTIONS} -classpath ${jarPath} dev.jbang.Main "$@")
 err=$?
-if [ $err -eq 0 ]; then
+if [ $err -eq 255 ]; then
   eval "exec $output"
-elif [ $err -eq 3 ]; then
-  echo "$output"
 else
+  echo "$output"
   exit $err
 fi
--- a/src/main/scripts/jbang.cmd
+++ b/src/main/scripts/jbang.cmd
@@ -87,18 +87,21 @@ SETLOCAL DISABLEDELAYEDEXPANSION
 set ERROR=%ERRORLEVEL%
 rem catch errorlevel straight after; rem or FOR /F swallow would have swallowed the errorlevel
 
-if %ERROR% NEQ 0 (
-  del ""%tmpfile%""
+if %ERROR% EQU 0 (
+  rem read generated java command by jang, delete temporary file and execute.
+  for %%A in ("%tmpfile%") do for /f "usebackq delims=" %%B in (%%A) do (
+    set "OUTPUT=%%B"
+    goto :break
+  )
+:break
+  del "%tmpfile%"
+  %OUTPUT%
+  exit /b %ERRORLEVEL%
+) else if %ERROR% EQU 3 (
+  type "%tmpfile%"
+  del "%tmpfile%"
+  exit /b 0
+) else (
+  del "%tmpfile%"
   exit /b %ERROR%
 )
-
-rem read generated java command by jang, delete temporary file and execute.
-for %%A in ("%tmpfile%") do for /f "usebackq delims=" %%B in (%%A) do (
-  set "OUTPUT=%%B"
-  goto :break
-)
-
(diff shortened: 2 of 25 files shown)


quintesse:
Works on Linux, let's see if the tests pass on Windows too.

maxandersen:
I can see how this is advantageous for output commands line alias list for sure. 

I assume for run which triggers build we still need to use current approach as we will be printing about build AND the run command, correct ?

Which means this is like a one extra kind of output. We have auxiliary output (what build prints to err today), error output (should always go to err), command output (Java cmd to execute on out) and then real output ie. alias output)

Which isn't a bad thing just to verify my understanding.

quintesse:
If you put it that way there's  indeed an extra kind of output. Before that what you call "real output" had to be done using an `echo` or `cat` for the "command output" (and `cat` is not available on Windows, so we would have had to come up with some other solution if we'd ever had the need for that).

quintesse:
Btw, still making some minor adjustments, don't merge yet

quintesse:
Ok, ready to merge

maxandersen:
did you do it for all commands ? i.e. `jbang | less` seem to still send on stderr?

src/main/scripts/jbang
```diff
@@ -131,6 +131,8 @@ output=$(CLICOLOR_FORCE=1 ${JAVA_EXEC} ${JBANG_JAVA_OPTIONS} -classpath ${jarPat
 err=$?
 if [ $err -eq 0 ]; then
   eval "exec $output"
+elif [ $err -eq 3 ]; then
+  echo "$output"
```
maxandersen, line 135: okey, now i recall my concern with this one. 

This means that jbang won't be able to have an error *and* print to the console..at least the exit code can't be used to designate what should happen...

should we possibly just say that if err is 0 then we run; but if any other code we print ?

src/main/scripts/jbang.cmd
```diff
@@ -82,18 +82,21 @@ SETLOCAL DISABLEDELAYEDEXPANSION
...
+    goto :break
+  )
+:break
+  del "%tmpfile%"
+  %OUTPUT%
+  exit /b %ERRORLEVEL%
+) else if %ERROR% EQU 3 (
+  type "%tmpfile%"
+  del "%tmpfile%"
+  exit /b 0
```
maxandersen, line 98: same as above - consider if we can just execute on error 0 and anything else we print AND return that exit code (not just 0)

src/main/scripts/jbang
quintesse: But isn't that the situation we have right now anyway?
Currently if you have an error any `echo My Text` commands won't be executed either.

Besides I don't think we want "printing output" to always return non-zero return codes, that way they'd always be considered errors by the shell/caller.  (Perhaps not a huge problem but "not nice").

Btw, before I actually printed the output _also_ in case of any other error codes (and returned that error code, while still returning 0 when the error code was the special one (3)) but I removed that.... and now I can't remember anymore why... 🤔

quintesse:
> did you do it for all commands ? i.e. jbang | less seem to still send on stderr?

I did it only for those commands that *actually generate output* as part of their "raison d'etre". To me usage and "info" weren't that :-)
But I see that other commands (git, ls, etc) actually do print their help to stdout, so yes, perhaps we should too.
Not sure where that's handled though, because I guess you set PicoCLi's `out` to stderr somewhere? I wonder if that means we'd have to selectively set it either to stdout or stderr depending on the situation 🤔

src/main/scripts/jbang
quintesse: Ah yes, I remember. It's becuase with any other error code you can't know if the output was generated by a command that was trying to print some text to the user or if it was the build/run/??? command trying to generate code to execute.

Perhaps we should designate a range of error codes for "exec" errors and a range for "print" errors? Eg:

 - 0 = OK, execute
 - 1-127 = ERROR (don't print captured stdout)
 - 128 = OK, print, return 0
 - 128-255 = ERROR (but still print captured stdout)

quintesse: Still, we'd also have to think if there's any "rational" use-case for printing captured output in case of errors. Do we have (or foresee) any commands where you could get an error mindway? And if so would it be important to show an unfinished result after the error was already printed?

maxandersen: > any "rational" use-case for printing captured output in case of errors. Do we have (or foresee) any commands where you could get an error mindway?

lets say `jbang alias list` prints while it iterates and then errors on accessing one of the catalogs...wouldn't it make sense to still get the ones it could access before it errors ?

maxandersen:
can you rebase this one ?

src/main/scripts/jbang
maxandersen: @quintesse ^

quintesse:
See if you're okay with this @maxandersen , I changed it around (as per one of my earlier suggestions) where a single special exit code (255) now means: "execute the output" while _all_ other exit codes just print the output and exit with the same code. That way simplifies the exit code handling in the startup scripts. I think it's nicer this way. Wdyt?

maxandersen:
I like this approach and I can't think of something wrong with it thus approving then we'll see what happens :)

PR #339, opened 2020-09-21
feat: Added `jdk default`, `jdk home` and `jdk java-env` commands 

quintesse:
These commands were added to facilitate the use of the Jbang's JDKs
by the user. This way it's not necessary to install additional JDKs
if we want to use Java and have previously let Jbang install a JDK.
`jdk default` set the given JDK as "current".
`jdk home` returns the path of the current or the given JDK.
`jdk java-env` returns the shell commands to make the current or
the given JDK usable in the user's shell.

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -712,8 +712,8 @@ For now this is required for `kubectl` plugin but not `git`. https://github.com/
 
 If your script requires a specific or minimal version of Java you can use `//JAVA <version>(+)`.
 
-If jbang finds a java executable using `JAVA_HOME` or `PATH` which satisfies the stated java version jbang will use it.
-If no such version is found it will automatically download and install it into jbang's cache (`~/.jbang/cache/jdks` by default).
+If Jbang finds a java executable using `JAVA_HOME` or `PATH` which satisfies the stated java version jbang will use it.
+If no such version is found it will automatically download and install it.
 
 Examples:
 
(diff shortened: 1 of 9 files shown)


quintesse:
The expand a bit on the description:

`jbang jdk default VERSION` creates the (sym)link `~/.jbang/currentjdk` and points it to `~/.jbang/cache/jdks/VERSION`.
From that moment on that JDK is considered the "current" one.

`jbang jdk home VERSION` will return the path to that JDK. If VERSION is not passed it will return `~/.jbang/currentjdk`.

`jbang jdk java-env` will return the following text:

```
export PATH="~/.jbang/currentjdk/bin:$PATH"
export JAVA_HOME="~/.jbang/currentjdk"
# Run this command to configure your shell:
# eval $(jbang jdk java-env)
```

This last one can easily be added to the user's `.bashrc` to provide them with a simple SDKMAN-like functionality (albeit only for Java JDKs).

maxandersen:
Can you do env in a way that can work On windows?

quintesse:
@maxandersen not that I know of. On Windows you'd have to mess around with system settings and such. And there's AFAIK no way of saying: execute the (multi-line) output of this program. You'd probably have to write it to a .bat file and execute that.
On Linux/Mac it isn't automatic either, a user would still have to add that `eval` to their `.bashrc`, so in the same way we could explain to the user how to add that folder to their PATH on Windows.

quintesse:
So the `java-env` command could a) at least show legal Windows commands (so you could copy&paste) and b) give some Windows-oriented explanation on how to set things up. At a later date we can perhaps figure out how to install things for the user automatically.

quintesse:
Done :point_up:

maxandersen:
yeah I was just thinking that jbang jdk java-env could mutate the setup on windows as you suggest.

quintesse:
Yeah, but that's a bit of complexity I wasn't willing to take on right now ;-)
This PR at least gives people a way to use the JDKs we download by setting up something in their system once. It's not as easy as we'd like yet but it's also not overly complex.

maxandersen:
understood - just trying to make sure we could eventually do it. i think we can - but yeah; not trivial.

maxandersen:
please include update to docs and then we should get it in.

I'm not a fan of "competing" with other package managers but it also feels wasteful to only have the jbang jdks availble to jbang.

quintesse:
I agree, on both accounts :-)
It's perhaps not likely I'll use that feature myself much because I'm already use SDKMAN and I have a need for more types of JDKs besides just the AdoptOpenJdk ones. And that's probably the case for a lot of Java devs, But for the ones that are not (pure) Java devs this might be very useful.

maxandersen:
was playing around and got to a bit of a conundrum:

```
❯ jbang jdk default
[jbang] No default JDK set, use 'jbang jdk default' to set one.
```

okey...now what ? :)

maxandersen:
and when I explicitly set a version:

```
❯ jbang jdk default 11
[jbang] Downloading JDK 11. Be patient, this can take several minutes...
[jbang] Installing JDK 11...
[jbang] Default JDK set to 11
[jbang] Default JDK set to 11
```

double jdk set to 11 - can we make that just 1 ?

quintesse:
> okey...now what ? :)

I had hoped there was no need to be so specific but I see there is :-)
Changed it to "No default JDK set, use 'jbang jdk default <version>' to set one."

quintesse:
> can we make that just 1 ?

Done

PR #340, opened 2020-09-21
Adds JBANG_DEFAULT_JAVA_VERSION details to readme

abelsromero:
Adds small description of JBANG_DEFAULT_JAVA_VERSION usage to readme.

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -1,4 +1,6 @@
 = j'bang - Having fun with Java scripting
+:idprefix:
+:idseparator: -
 :toc: macro
 ifndef::env-github[:icons: font]
 ifdef::env-github[]
@@ -8,7 +10,7 @@ ifdef::env-github[]
 :tip-caption: :bulb:
 :warning-caption: :warning:
 endif::[]
-	
+
 image:https://img.shields.io/github/release/jbangdev/jbang.svg[Release,link=https://github.com/jbangdev/jbang/releases]
 image:https://github.com/jbangdev/jbang/workflows/ci-build/badge.svg[Build Status,link=https://github.com/jbangdev/jbang/actions]
 image:https://www.eclipse.org/che/contribute.svg[Che, link=https://che.openshift.io/f?url=https://github.com/jbangdev/jbang]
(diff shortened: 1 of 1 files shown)


quintesse:
Well thing is that's actually not correct :-)
The usage of `JBANG_DEFAULT_JAVA_VERSION` is **very** specific, it's *only* used when *no* Java is installed on the system and Jbang needs to first download and install Java to be able to run itself. By default it will use version 11, but you can change that using that env var. In all other cases this env var has no effect (for example if you already have a Java 14 on your PATH, that will be used, regardless of what you set the env var to).
The option `--java` on the other hand will *always* make sure that the given Java version is used, anything else is considered a failure.

quintesse:
So this env var is highly technical. Of course it needs to be documented, but I'd hide at far away in some section for advanced rarely used cases.

abelsromero:
If I understand correctly, is mostly aimed at CI environments then?

quintesse:
> If I understand correctly, is mostly aimed at CI environments then?

Yeah I guess so, or in other situations where Jbang is used inside scripts where you want to make sure that the minimum available version is something other than 11.

abelsromero:
Based on your explanation and some testing I think I got it. I reworded the explanation to mention it as "bootstrap" version  wdyt?
I was not aware you can run jbang in one jdk and the script in another :exploding_head: 

FUN FACT: you can do `jbang jdk uninstall 11` and it will install, list jdk 1, and uninstall it in the same command.

quintesse:
> FUN FACT: you can do jbang jdk uninstall 11 and it will install, list, and uninstall it in the same command.

Yeah :-)

That's been fixed already, just waiting to be merged. But thanks for the heads up!

quintesse:
> I was not aware you can run jbang in one jdk and the script in another

Unfortunately it's really hard to do parsing of options in bash scripts and much harder on Windows. So instead of trying to figure out what JDK the user wants to use we just guess that the one they have installed on your system is the one they're most likely to use. When we guess wrong we switch over the moment we execute the user's script.

abelsromero:
> That's been fixed already, just waiting to be merged. But thanks for the heads up!

Don't consider that a bug in fact

maxandersen:
@all-contributors please add @abelsromero for doc

PR #343, opened 2020-09-21
feat: edit now uses cmd.exe / sh -c to launch editor

maxandersen:
This should allow usage of aliases (define in .profile etc.) on linux
and on windows `code` will find `code.cmd`.

Downsie is that you won't get an error if the editor is not found.

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -1,11 +1,18 @@
 package dev.jbang.cli;
 
 import static dev.jbang.Settings.CP_SEPARATOR;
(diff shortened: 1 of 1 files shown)


maxandersen:
we got about 3 or so issues asking why `jbang edit --open=code` does not work on windows. With this fix it does but no longer prints error if editor command is not found.

Kinda feel this PR is the lesser of the two evils.

quintesse (approved):
LGTM

Btw, if you want to show a message if the editor doesn't exist you could always first test if it exists, right? You can just loop over the entries in the PATH and see if the command exists there.

maxandersen:
rebased and merged.

PR #348, opened 2020-09-23
fix: Improved`--quiet` option

quintesse:
It will now also silence the dependency resolving.


quintesse:
Not sure if the shrinkwrap API does logging of itself. I see that it's possible to set a logger for the `EmbeddedMaven` class, but I haven't seen anything similar for `Maven` (doesn't mean it doesn't exist though, just could't find anything). In my simple test nothing appeared on the console though, so hopefully it works.

PR #354, opened 2020-09-24
make version less zip/tar for stable latest download url

maxandersen:


build.gradle
```diff
@@ -92,6 +92,31 @@ distributions {
 			}
 		}
 	}
+	latest {
+		contents {
```
maxandersen, line 96: @aalmiray any way to refer to the main contents to avoid duplicating the same definition ?

build.gradle
```diff
@@ -156,7 +181,7 @@ processTestResources.dependsOn copyTestResources
 
 task createChecksum(type: Checksum) {
 	dependsOn(assembleDist)
-	files = distZip.outputs.files
+	files = distZip.outputs.files + latestDistZip.outputs.files + distTar.outputs.files + latestDistTar.outputs.files
```
maxandersen, line 184: there got to be a better way for this than a explicit list of dist artifacts...

build.gradle
maxandersen: figured it out by using shared spec

Issue #373, opened 2020-09-30
support --javaagent=<file|GAV|jbangscript>[=options]


maxandersen:
to allow usecases like:

`java -javaagent:com.github.phillip-kruger.julset:julset:master-38ae41f7b5-1 -Djul.com.github.phillipkruger.myapp=FINEST -jar com.github.phillip-kruger.julset:julset-test:master-38ae41f7b5-1` 

or even more "crazy":

`java -javaagent:julset.java -Djul.com.github.phillipkruger.myapp=FINEST myapp.java`

the latter would require #130 or similar to work

phillip-kruger:
Second option would be cool !!

maxandersen:
this is all there now since #377 was fixed. thus closing

PR #377, opened 2020-10-01
support --javaagent:xyz.java

maxandersen:
everything is in place, javageent gets built if needed  *EXCEPT* I now realize I need a way to set
premain-class and agent-class in the manifest.

I kinda think we should have a //JAVAAGENT option which if present will search the java class for premain( and agentmain(
and set the respective class name.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,7 @@ dependencies {
 	implementation 'org.jboss.shrinkwrap.resolver:shrinkwrap-resolver-api:3.1.4'
 	implementation 'org.jboss.shrinkwrap.resolver:shrinkwrap-resolver-impl-maven:3.1.4'
 	implementation "org.slf4j:slf4j-nop:1.7.25"
+	implementation "org.jboss:jandex:2.2.1.Final"
 	//implementation 'org.apache.maven:maven-aether-provider:3.0.5'
 	//implementation 'com.google.guava:guava:28.2-jre'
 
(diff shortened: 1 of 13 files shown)


maxandersen:
will eventually fix #373

maxandersen:
fyi https://docs.oracle.com/javase/9/docs/api/java/lang/instrument/package-summary.html has the various manifests an agent can set.

was thinking something like (in worst case)

//JAVAAGENT Can-Redefine-Classes Can-Retransform-Classes 

`Can-Redefine-Classes: true`
`Can-Retransform-Classes: true`
`Premain-Class` set if `premain(` exists
`Agent-Class` set if `agentmain(` exists in source
`boot-class-path` will be used for transitive dependencies

launcher-agent-class usecase not covered/handled. could be added in future.

quintesse (commented):
Well I don't see anything wrong with it at least :-)

maxandersen:
so one step closer, this works now:

```
jbang -Djul.JULTest=FINEST --javaagent=JULAgent.java JULTest.java World
```

even url fetch works

```
jbang --verbose -Djul.JULTest=FINEST --javaagent=https://github.com/jbangdev/jbang/blob/agents/itests/JULAgent.java JULTest.java World
```

maxandersen:
things still missing:

- passing of options.
- custom values for Can-* options
- multiple agents

maxandersen:
passing of options.
custom values for Can-* options
multiple agents

now all possible.

maxandersen:
merged with fix commit mesage.

PR #378, opened 2020-10-03
fix(doc): Use adoc based code listing, not md based

rsvoboda:
Use adoc based code listing in .adoc files, not md based

An example of diff
 - https://github.com/jbangdev/jbang/blob/master/readme.adoc#declare-dependencies
 - https://github.com/rsvoboda/jbang/blob/ascii.vs.md/readme.adoc#declare-dependencies

GH handles md style in .adoc quite well, but the will be issue once .adoc files get processed by other tools.


maxandersen:
thanks! nice catch!

maxandersen:
@all-contributors please add @rsvoboda  for doc

PR #395, opened 2020-10-07
docs: added explanation on how to use startup scripts remotely

quintesse:


maxandersen:
Pending on ps/sh domains being setup

abelsromero:
Question...does this install and run, or just run creating only ".jbang" dir?

quintesse:
@abelsromero it's complicated ;-)

From a user's perspective it doesn't really install. You won't have a new "jbang" command that you can use after having run that command. It's for running Jbang on-the-fly, so to speak.

From a technical point of view we do actually download and copy Jbang to the user's HOME directory. So if you'd add `$HOME/.jbang/cache/jbang/jbang/bin` to your `PATH` yourself you'd basically have installed Jbang. Most likely at some future time we'd make that an actual option, eg. you'd be able to do something like:

```
curl -Ls https://sh.jbang.dev | bash -s - install
```

and from then on you'd be able to just run `jbang` from your shell.

abelsromero:
> @abelsromero it's complicated ;-)

If it was, your explanation made it really simple.
Should this be mentioned as way of running in the readme also?

quintesse:
@abelsromero yes, this should definitely go into the readme (without all the technical stuff). You know how it is, we first make the features and then we remember we should write docs as well ;-)

maxandersen:
merging but known issues found to figure out: https://github.com/jbangdev/jbang/issues/396 and https://github.com/jbangdev/jbang/issues/397

Issue #396, opened 2020-10-07
is remote jbang runner caching correct?


maxandersen:
tried running `curl -Ls https://sh.jbang.dev | bash -s - helloworld.java`﻿and a few things stood out:

1) it says "Downloading/Installing JBang..." even when I have jbang installed - is that expected ?
   (I could see reasons on why wanting to do both...but if it must I think it would be nice it says *where* it download/installs, i.e. the jbang cache ..)

2) ~/.jbang/jbang.tar is left in the directory - is that expected ? if yes, should it not be in ~/.jbang/cache/url/<url sha>/ ?  

3) jbang install is put in `~/.jbang/jars/jbang` I had expected something like `~/.jbang/jbang/jbang-0.47.1/*` or `~/.jbang/jbang/jbang-remote/*` or similar
   having it inside jars seems wrong - i.e. `jbang cache clear` would uninstall jbang...

quintesse:
1. Yeah, right it doesn't know about any jbang's not installed by itself
2. We could def download it into `url`
3. Well...
  a. with the newest version-less download there's actually no way to know what version we're using.
  b. `jbang-remote`... well we could name it that, but why? :-)
  c. actually because it's in the root of the cache `cache clear` doesn't actually delete it. But I'm not sure that's okay, why shouldn't clearing the cache (at least when using `-all`) remove something that was automatically downloaded from the internet? We are not actually "installing" anything.

ad 3c) just playing a bit of devil's advocate. I'm not 100% sure what the best behaviour would be. For example JDKs are also removed when you clear the cache... and if you've set a default one and added it to your PATH that will def cause problems. The alternative could be that we consider JDKs and Jbangs to not be throw-away and install them outside the cache. It would be easy enough to do that.

maxandersen:
I think treat jbang  like a jdk would make most sense - its not a url nor jar ...shouldn't be mixed up with jbangs own app cache.

about `jbang-remote` - just the worst name my brain could think of to indicate this is the jbang we will use for this remote notion ;)

btw. 3.a. see https://github.com/jbangdev/jbang/issues/388 which came up when @aalmiray implementing gradle plugin and had similar needs of deetcting current version.

quintesse:
> I think treat jbang like a jdk would make most sense 

Well, but that's how it's done right now, isn't it? JDKs go into `cache/jdk` and Jbang goes into `cache/jbang`.. or am I not understanding you?

> this is the jbang we will use for this remote notion

[...]

PR #403, opened 2020-10-08
refactor: The Jbang zero-install cache is now an official category

quintesse:
NB: Needs to be merged after #402

refactor: The Jbang zero-install cache is now an official category

Instead of putting the Jbang zero-install download in `cache/urls`
it's now put in its own `jbang` directory. This means it won't be
deleted when doing a normal `cache clean`, just like the JDKs.
(It will still be deleted when explicitely mentioned or when using
`cache clear -all`).

See #396

### DIFF
--- a/src/main/java/dev/jbang/cli/Cache.java
+++ b/src/main/java/dev/jbang/cli/Cache.java
@@ -19,6 +19,8 @@ public class Cache {
 			@CommandLine.Option(names = {
 					"--jar" }, description = "clear JAR cache only", negatable = true) Boolean jars,
 			@CommandLine.Option(names = {
+					"--jbang" }, description = "clear Jbang zero-install cache only", negatable = true) Boolean jbangs,
+			@CommandLine.Option(names = {
 					"--jdk" }, description = "clear JDK cache only", negatable = true) Boolean jdks,
 			@CommandLine.Option(names = {
 					"--project" }, description = "clear temporary projects cache only", negatable = true) Boolean projects,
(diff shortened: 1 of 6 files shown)


maxandersen:
Integration test failed for jbang cache.

quintesse:
Updated

src/main/java/dev/jbang/cli/Cache.java
```diff
@@ -34,6 +36,7 @@ public Integer clear(
 			classes.addAll(Arrays.asList(Settings.CacheClass.values()));
 		} else if (urls == null
 				&& jars == null
+				&& jbang == null
```
maxandersen, line 39: Minor bit but technically it's jbangs not jbang right ?

quintesse: Well, no, technically it's just a single one. That's why I used singular, although it goes against my ingrained love for symmetry 😂

quintesse: If you want I can make it plural of course.

maxandersen: symmetry ftw :) please update and i'll happily merge :)

quintesse: Done

quintesse: Hmm not yet

quintesse: Ok, now :-)

PR #404, opened 2020-10-08
docs: Some additional instructions for zero-install with PowerShell

quintesse:
The instruction now show 2 alternatives, one is shorter but might
be somewhat less flexible while the other is somewht more verbose
but has the interesting property that you can store it in a variable
and re-use it.


maxandersen:
Man- I now know more
Power shell than is healthy :)

PR #409, opened 2020-10-08
fix: GH-374 Default alias name when creating a new alias

### LINKED ISSUE #374: alias add -f directory should create jbang-catalog.json in that directory
would be nice to just go:

`jbang alias -f . xyz path/to/xyz.java`

instead of having to spell out jbang-catalog.json.

right now the above gives: `[jbang] [WARN] Unable to add alias: /Users/max/code/camel-visual/.: Is a directory`

chrsblck:
Fixes #374

### DIFF
--- a/src/main/java/dev/jbang/cli/Alias.java
+++ b/src/main/java/dev/jbang/cli/Alias.java
@@ -1,7 +1,9 @@
 package dev.jbang.cli;
 
 import java.io.PrintStream;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -29,6 +31,9 @@ abstract class BaseAliasCommand extends BaseCommand {
 		if (global) {
 			return Settings.getAliasesFile();
 		} else {
+			if (Files.isDirectory(catalogFile)) {
+				catalogFile = Paths.get(catalogFile.toString(), AliasUtil.JBANG_CATALOG_JSON);
+			}
 			return catalogFile;
 		}
 	}
--- a/src/main/java/dev/jbang/cli/Alias.java
+++ b/src/main/java/dev/jbang/cli/Alias.java
@@ -32,7 +32,7 @@ abstract class BaseAliasCommand extends BaseCommand {
 			return Settings.getAliasesFile();
 		} else {
 			if (Files.isDirectory(catalogFile)) {
-				catalogFile = Paths.get(catalogFile.toString(), AliasUtil.JBANG_CATALOG_JSON);
+				return Paths.get(catalogFile.toString(), AliasUtil.JBANG_CATALOG_JSON);
 			}
 			return catalogFile;
 		}
--- a/src/main/java/dev/jbang/AliasUtil.java
+++ b/src/main/java/dev/jbang/AliasUtil.java
@@ -368,6 +368,9 @@ public class AliasUtil {
 			cwd = getCwd();
 		}
 		catalog = cwd.resolve(catalog);
+		if (Files.isDirectory(catalog)) {
+			catalog = Paths.get(catalog.toString(), JBANG_CATALOG_JSON);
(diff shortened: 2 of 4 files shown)


chrsblck:
Shoot, reading through the `Semantic PR` docs, sorry :(

maxandersen:
fix looks right - can you add a test to verify it so we don't accidentally break it ? :) 

`itests/catalog.feature` or `src/test/java/dev/jbang/cli/TestCatalog.java` would be options for places to add that.

chrsblck:
Actually, there's a bug here, I think. I'll add 1 or 2 tests, good call!

chrsblck:
Yeah, I did introduce a bug by not resolving the default path before fetching any existing aliases. I've fixed this and added unit tests to cover this case and the case where the file doesn't exist.

That said, do you think this is the right test file (`TestCatalog.java`)?

quintesse (changes requested):
Ok, so although this is indeed correct, I'd prefer it a bit differently.

First of all you've now only solved it for `alias add`, but the other alias commands also support the `-f` option which means that now `alias add` has a feature the other commands don't have. So my first suggestions is to make it work for all of them.

And second, you solved this inside the API, which although perfectly correct, is not where I'd put this. I'd rather have that the API doesn't have any "tricks" to make things easier for the user, but that this would be handled by the code that "interfaces" with the user. In this case the commands in the `cli` package.

And in fact, there is a perfect spot for doing this: https://github.com/jbangdev/jbang/blob/master/src/main/java/dev/jbang/cli/Alias.java#L32 , which is where the `-f` option is handled for all alias commands.

chrsblck:
@quintesse Makes sense to me, thanks for the review. I'll fix this up tonight.

chrsblck:
@quintesse or @maxandersen should I move the tests? Possibly to `TestAlias.java`

quintesse:
Yeah, that's probably a good idea. TestAlias is for tests related to the alias command. (Although honestly it's all a bit mixed)

chrsblck:
> Yeah, that's probably a good idea. TestAlias is for tests related to the alias command. (Although honestly it's all a bit mixed)

Yeah, it wasn't totally clear to me if that was the right thing to do. Thanks!

chrsblck:
@quintesse Fixed. Ready for another review when you have a chance.

I'll squash the changes once you sign off 😄

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -29,6 +31,9 @@ protected Path getCatalog() {
 		if (global) {
 			return Settings.getAliasesFile();
 		} else {
+			if (Files.isDirectory(catalogFile)) {
+				catalogFile = Paths.get(catalogFile.toString(), AliasUtil.JBANG_CATALOG_JSON);
```
maxandersen, line 35: should it not just return the `Paths.get(...)` result and not change the `catalogFile` ?

quintesse: Indeed :-)

chrsblck: Fixed :)

quintesse (approved):
LGTM

maxandersen:
Thanks @chrsblck !

@all-contributors please add @chrsblck for code

maxandersen:
oh missed this - but for next time GH-374 is jira style issue reference, in github just use #374 and best to just have it in body of commit message as otherwise linking doesn't work that great.

chrsblck:
OK, will do. (yes, we use jira + github at work, so looked right to me 😄 )

maxandersen:
ha, turns out GH-374 works too...did not know that - I guess github got tired of jira syntax failing ;)

chrsblck:
Haha, yeah, I saw it in their docs somewhere... Or maybe not, I could have seen it here https://stackoverflow.com/a/6742691/1246282.

Either way, I won't use that for non jira tagging. Thanks for the pointer!

PR #411, opened 2020-10-09
Remove interactive flag from sh

### LINKED ISSUE #407: Jbang 0.50.0 hangs when edit --live --open is used
```
~/workspace/quarkiverse/jbang-catalog (master ✘)✭ ᐅ jbang edit --live --open=idea updatecatalog.java
[jbang] Running `sh -i -c idea /home/ggastald/.jbang/cache/projects/updatecatalog.java_jbang_45ac532f09b7c638f171c762ead860b7d1933b7eaaeb4883d89306c1e4614fde/updatecatalog`
[jbang] Watching for changes in /home/ggastald/workspace/quarkiverse/jbang-catalog
^C^C^C^C^C
```

gastaldi:
This makes the CLI impossible to quit in Fedora 32.

Fixes #407

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -71,7 +71,7 @@ public class Edit extends BaseScriptCommand {
 				if (isWindows()) {
 					cmd = new String[] { "cmd", "/c", optionList.stream().collect(Collectors.joining(" ")) };
 				} else {
-					cmd = new String[] { "sh", "-i", "-c", optionList.stream().collect(Collectors.joining(" ")) };
+					cmd = new String[] { "sh", "-c", optionList.stream().collect(Collectors.joining(" ")) };
 				}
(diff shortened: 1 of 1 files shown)


maxandersen:
Hmm. This would also mean your shell config files won't be read.

gastaldi:
It is only used to start the IDE and worked great in 0.47.1, so I don't think that's worse than  freezing the UI :)

maxandersen:
I just wonder why I doesn't freeze OS X. 

And if we remove -i we might as well just run it directly.

gastaldi:
Can you confirm this bug using a clean Fedora VM? I don't know if this is caused by my environment

maxandersen:
it has something to do with the shell -i process being wiried into input stream of the parent process but even when I tried avoiding it it keeps happening on fedora. 

so for now i'm just going to accept this fix as working ctrl+c is more important than ability to call interactive aliases...users can still set up aliases just need to be aware its done in non-interactive shell setup files.

PR #412, opened 2020-10-09
fix: Jbang won't print extra newline when no output was generated

quintesse:


maxandersen:
can you add a test so we don't accidentally break this in future ?

quintesse:
> can you add a test so we don't accidentally break this in future ?

Actually, I don't know how. It seems that Karate first strips the captured output before testing it, so even though the output contains a newline if I compare it with an empty string it still says it matches.

quintesse:
Ok, test added!

maxandersen:
thankayou!

PR #413, opened 2020-10-09
build: add version.txt to .zip/.tar and version.txt to release assets

maxandersen:
Thanks for submitting your Pull Request!

Please delete this text, and add description of the topic solved by this PR.

To make releases as automated as possible we use conventional commits formats.
Thus commits and pull-requests titles should before merge follow the Conventional Commits spec.

Details in https://github.com/zeke/semantic-pull-requests

If in doubt then open the pull-request and we'll help you - Thank you!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,6 +76,12 @@ spotless {
 	}
 }
 
+task versionTxt() {
+	doLast {
+		new File(project.buildDir, "tmp/version.txt").text = project.version
+	}
+}
+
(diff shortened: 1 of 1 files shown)


aalmiray:
Yes, though I may suggest to geneerate the `versions.txt` inside `build/tmp/version` instead of `build/versions.txt` otherwise you run the risk of invalidating caches more often than not.

maxandersen:
updated to have 'tmp' and merged

PR #414, opened 2020-10-10
fix: stop using -jar

### LINKED ISSUE #388: add version.txt to see what latest refer to

maxandersen:
When running jars like org.eclipse.jgit:org.eclipse.jgit.pgm:5.9.0.202009080501-r we
    got noclassdefounderrors since we need to put the dependencies in the classpath, but
    with `-jar` java ignores the classpath.
    
    Thus we need to simulate a -jar run instead.

note: this is technically a breaking change but i can not imagine anyone using -jar that worked before being affected by this change.


quintesse (approved):
LGTM

PR #435, opened 2020-10-13
fix: support nested multiple sources and gist with specific filenames

tivrfoa:
Hi,

I got the error:

```java
jbang https://gist.github.com/joshlong/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java
[jbang] https://gist.github.com/joshlong/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java is not from a trusted source thus not running it automatically.

If you trust the url to be safe to run you can do one of the following:
0) Trust once: Add no trust, just run this time
1) Trust this url in future:
    jbang trust add https://gist.github.com/joshlong/


Any other response will result in exit.

[jbang] Type in your choice (0 or 1) and hit enter. Times out after 10 seconds.
0
[jbang] Building jar...
[jbang] compile: /usr/lib/jvm/java-11-openjdk-amd64/bin/javac -d /home/leandro/.jbang/cache/jars/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java.3944bb03a8dc5c1467b8a3f97c11695c8dcb1c3affd48c204df5cca1dc266dd7 /home/leandro/.jbang/cache/urls/ec2194b86aa3eeabbbcc6d534daf4a621b390a13760fd9f3bb01da805404351f/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java
error: invalid flag: /home/leandro/.jbang/cache/urls/ec2194b86aa3eeabbbcc6d534daf4a621b390a13760fd9f3bb01da805404351f/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java
Usage: javac <options> <source files>
use --help for a list of possible options
dev.jbang.ExitException: Error during compile
	at dev.jbang.cli.BaseBuildCommand.build(BaseBuildCommand.java:173)
	at dev.jbang.cli.Run.prepareArtifacts(Run.java:79)
	at dev.jbang.cli.Run.doCall(Run.java:68)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:80)
```

Then I thought it would be cool if it worked, because many Gists have more than one file.
So this change fix this error.

tests:
```java
leandro:distributions$ ./jbang-0.51.1.2/bin/jbang https://gist.github.com/tivrfoa/503c88fb5123b1000c37a3f2832d4773#file-file1-java
[jbang] Building jar...
I'm file 1
leandro:distributions$ ./jbang-0.51.1.2/bin/jbang https://gist.github.com/tivrfoa/503c88fb5123b1000c37a3f2832d4773#file-file2-java
[jbang] Building jar...
I'm file 2
leandro:distributions$ ./jbang-0.51.1.2/bin/jbang https://gist.github.com/tivrfoa/503c88fb5123b1000c37a3f2832d4773#file-file3-java
Test jsh extension
I'm file 3
```


tivrfoa:
I need to change to handle:
. dash: file-name.java
. hash: file#name.java (although really?! =))

quintesse:
Actually a _really_ cool feature for multi-file Gists would be that you could use the `//SOURCES` feature to refer from one file to the other :-)

But that would of course mean more work. Like retrieving all files, putting them together and then compiling the one referenced explicitly in the URL. If that file would have `//SOURCES` or `//FILES` tags it would pick them up automatically.

tivrfoa:
> Actually a _really_ cool feature for multi-file Gists would be that you could use the `//SOURCES` feature to refer from one file to the other :-)
> 
> But that would of course mean more work. Like retrieving all files, putting them together and then compiling the one referenced explicitly in the URL. If that file would have `//SOURCES` (and/or `//FILES`) tags it would pick them up automatically.

I think it's an interesting feature for the backlog. =)

tivrfoa:
> I need to change to handle:
> . dash: file-name.java
> . hash: file#name.java (although really?! =))

I think there's no way to handle hash `#` in file name, because Github replaces it with a dash '-' when creating the url, eg:
https://gist.github.com/tivrfoa/4ff85540b425756a8476d8c62d5b7f7b#file-hash-test-java

tivrfoa:
> Actually a _really_ cool feature for multi-file Gists would be that you could use the `//SOURCES` feature to refer from one file to the other :-)
> 
> But that would of course mean more work. Like retrieving all files, putting them together and then compiling the one referenced explicitly in the URL. If that file would have `//SOURCES` (and/or `//FILES`) tags it would pick them up automatically.

I'll try to implement it.

maxandersen:
Having multi file gists work would definitely be a nice touch.  We got basics working for http urls already so shouldn't be that hard for gists. 

Famous last words :)

maxandersen (approved):
Approach looks fair and will be nice when it works. With experience I do know we absolutely need some tests included in here as otherwise we'll accidentally break it.

p.s. about multi file support then the current mechanism "just works" when using the raw urls so if the swizzling of gist with #'s will end up with the raw url then I think that you will get that for free.

tivrfoa:
> Approach looks fair and will be nice when it works. With experience I do know we absolutely need some tests included in here as otherwise we'll accidentally break it.
> 

Sorry for not adding the tests. I was happy that it worked and then sent the PR for initial review. I didn't create the tests because I had to work, but I'll do it.
I was expecting some change request and not the approval. I still need to squash the commits. :)

tivrfoa:
> Approach looks fair and will be nice when it works. With experience I do know we absolutely need some tests included in here as otherwise we'll accidentally break it.
> 
I added the tests, but it would be better if it was your profile instead of mine.

> p.s. about multi file support then the current mechanism "just works" when using the raw urls so if the swizzling of gist with #'s will end up with the raw url then I think that you will get that for free.

Nice! It works indeed. =)
```shell
jbang https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java
[jbang] Building jar...
Testing JBang multiple sources.
two
```

And with this PR it also works specifying the filename!!! =D
```shell
./jbang/bin/jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6#file-one-java
Testing JBang multiple sources.
two
```

And without the filename it does not work (same behavior as current JBang version):
```shell
jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6
[jbang] Building jar...
/home/leandro/.jbang/cache/urls/7ed417368f048491e09a7b4704d57d4ff121bfd5a9e50eee655e17834a179cd1/two.java:6: error: class one is public, should be declared in a file named one.java
public class one {
       ^
/home/leandro/.jbang/cache/urls/81cee824925121103883e8eb848b5c88f3a60f506627d1fb0511e63b4f186fa9/one.java:10: error: cannot find symbol
    	out.println(two.class.getName());
    	            ^
  symbol:   class two
  location: class one
2 errors
```

So the only question now is whether it's worth make a change so it works without the filename.
As with this PR we already make it work specifying the entry point, I don't think the benefit outweighs the extra complexity in the code, because we would have to parse until we find `public static void main`.
But if you think this would be nice to have, then I think it is better to be done in a different PR.

src/main/java/dev/jbang/Util.java
```diff
@@ -531,36 +531,63 @@ public static String getStableID(String input) {
...
+												.endsWith(".jsh"))
+									.findFirst();
+			}
+
+			if (file.isPresent()) {
+				Util.verboseMsg("looking at " + file);
+				return (String) file.get().getValue().getOrDefault("raw_url", url);
 			} else {
-				// Util.info("nothing worked!");
+				Util.verboseMsg("nothing worked!");
```
maxandersen, line 590: This message doesn't help user much - either give some context or remove completly.

src/main/java/dev/jbang/Util.java
```diff
@@ -531,36 +531,63 @@ public static String getStableID(String input) {
...
 					"https://api.github.com/gists/${gistid}");
-			// Util.info("looking at " + gistapi);
+
+			Util.verboseMsg("looking at " + gistapi);
 			String strdata = null;
 			try {
 				strdata = readStringFromURL(gistapi);
 			} catch (IOException e) {
-				// Util.info("error " + e);
+				Util.verboseMsg("error " + e);
```
maxandersen, line 553: if you want to include it provide some context for the user, like Util.verboseMsg("Error when extracting file from gist url: ",e);

src/main/java/dev/jbang/Util.java
```diff
@@ -531,36 +531,63 @@ public static String getStableID(String input) {
...
-																										.endsWith(
-																												".java")
-																								|| e.getKey()
-																									.endsWith(".jsh"))
-																						.findFirst();
-
-			if (first.isPresent()) {
-				// Util.info("looking at " + first);
-				return (String) first.get().getValue().getOrDefault("raw_url", url);
+			Util.verboseMsg("found " + gist.files);
```
maxandersen, line 561: same as before, if putting verbose message in then add some context to be informative/useful.

src/main/java/dev/jbang/Util.java
```diff
@@ -531,36 +531,63 @@ public static String getStableID(String input) {
...
+									.filter(e -> e	.getKey()
+													.endsWith(
+															".java")
+											|| e.getKey()
+												.endsWith(".jsh"))
+									.findFirst();
+			}
+
+			if (file.isPresent()) {
+				Util.verboseMsg("looking at " + file);
```
maxandersen, line 587: not informative. provide context or remove.

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -560,6 +560,25 @@ void testFetchFromGist(@TempDir Path dir) throws IOException {
 		assertEquals(x.getFileName().toString(), "checklabeler.java");
 	}
 
+	@Test
+	void testFetchFromGistWithFilename(@TempDir Path dir) throws IOException {
+
+		String u = Util.swizzleURL("https://gist.github.com/tivrfoa/503c88fb5123b1000c37a3f2832d4773#file-file2-java");
```
maxandersen, line 566: test in your account is fine - if/when we need to change it we would need to fork it anyway; so doesn't matter as long as you don't delete it :)

maxandersen (changes requested):
i pushed the wrong button last - did mean to ask request changes! 

change still looks good and now tests are there - just spotted the verbose logging that needs some context or be removed.

maxandersen:
> So the only question now is whether it's worth make a change so it works without the filename.
> As with this PR we already make it work specifying the entry point, I don't think the benefit outweighs the extra complexity in the code, because we would have to parse until we find `public static void main`.

I'm actually a bit surprised it fails this way using the gist url only - it apparently manages to download it and when I run it after having use raw url I actually have it working.

Thus something is "off" around the caching/url swizzling.

```
# clear cache to avoid sideffects
jbang cache clear
# this new line fails as if two.java is saved under one.java...
jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  
# this uses raw and works as expected
jbang https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java 
# now this is exactly the same as before but now works - as if we must be using the cache read/write inconsistently
jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  # <-- now this works!
```

I agree this is probably for a another PR :)

tivrfoa:
> > So the only question now is whether it's worth make a change so it works without the filename.
> > As with this PR we already make it work specifying the entry point, I don't think the benefit outweighs the extra complexity in the code, because we would have to parse until we find `public static void main`.
> 
> I'm actually a bit surprised it fails this way using the gist url only - it apparently manages to download it and when I run it after having use raw url I actually have it working.
> 
> Thus something is "off" around the caching/url swizzling.
> 
> ```
> # clear cache to avoid sideffects
> jbang cache clear
> # this new line fails as if two.java is saved under one.java...
> jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  
> # this uses raw and works as expected
> jbang https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java 
> # now this is exactly the same as before but now works - as if we must be using the cache read/write inconsistently
> jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  # <-- now this works!
> ```
> 
> I agree this is probably for a another PR :)

Actually this PR

tivrfoa:
> And with this PR it also works specifying the filename!!! =D
> 
> ```shell
> ./jbang/bin/jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6#file-one-java
> Testing JBang multiple sources.
> two
> ```

No, it doesn't work. It only works if the the `raw` url is run before, as you noted ... :/

I'll investigate what's going on.

tivrfoa:
> I'm actually a bit surprised it fails this way using the gist url only - it apparently manages to download it and when I run it after having use raw url I actually have it working.
> 
> Thus something is "off" around the caching/url swizzling.
> 
> ```
> # clear cache to avoid sideffects
> jbang cache clear
> # this new line fails as if two.java is saved under one.java...
> jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  
> # this uses raw and works as expected
> jbang https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java 
> # now this is exactly the same as before but now works - as if we must be using the cache read/write inconsistently
> jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  # <-- now this works!
> ```
> 
> I agree this is probably for a another PR :)

So, it works after the running `raw` because of the existing jar file. If the jar is deleted, then the error happens.

When running 'raw', then the url is:
https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java
And the file `two.java` exists in the same path. That is why `raw` works.

But the crazy thing is that I really don't know where **36f5073272840b4434d50a60d3deb8285f8d0e84** comes from.

When running:
https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6#file-one-java

then the raw url is:
https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/04ee65ebc47cc63a3809c968f80c9fe2cef269dd/one.java

BUT the there is no `two.java` in this path (actually there is, but it's the same content of one.java)
https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/04ee65ebc47cc63a3809c968f80c9fe2cef269dd/two.java

So to fix that, instead of:
```java
URI thingToFetch = includeContext.resolve(resource);
```
We should get the `raw_url` from here:
https://api.github.com/gists/8e6ea001f168fd4ef14763ceca3e5ab6

I'll try to fix that later today.

tivrfoa:
>https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java
> And the file `two.java` exists in the same path. That is why `raw` works.
> 
> But the crazy thing is that I really don't know where **36f5073272840b4434d50a60d3deb8285f8d0e84** comes from.
> 

It's not that I don't know where it comes from: it's the url you get when you click the 'raw' button on the gist page.

What I meant to say is that I don't understand why it's different from the API raw_url ...

tivrfoa:
> > So the only question now is whether it's worth make a change so it works without the filename.
> > As with this PR we already make it work specifying the entry point, I don't think the benefit outweighs the extra complexity in the code, because we would have to parse until we find `public static void main`.
> 
> I'm actually a bit surprised it fails this way using the gist url only - it apparently manages to download it and when I run it after having use raw url I actually have it working.
> 
> Thus something is "off" around the caching/url swizzling.
> 
> ```
> # clear cache to avoid sideffects
> jbang cache clear
> # this new line fails as if two.java is saved under one.java...
> jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  
> # this uses raw and works as expected
> jbang https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/36f5073272840b4434d50a60d3deb8285f8d0e84/one.java 
> # now this is exactly the same as before but now works - as if we must be using the cache read/write inconsistently
> jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6  # <-- now this works!
> ```
> 
> I agree this is probably for a another PR :)

So as I explained in the other posts the problem is that for Github we can't assume it will be the same path/domain for the other sources as `ScriptResource.fetchIfNeeded` does.

I would be happy to work in a solution to that if you want, in another PR.

Please let me know if there is something else for this PR.

ps: I accidentally close the commit. I read "Cancel comment", but it's "Close with comment". =)

tivrfoa:
So, when the Gist API is read it already contains the content of the files!!
That's really helpful. That's why I made this last change so you can take a look.

As we already have the contents, we only need to cache the files.
It also handles the sources.

I couldn't many tests because I hit Github API limit! Oo =D

Some tests now fail and I don't know why. Does Gradle run tests in parallel?
If I build with tests disabled, then run the tests manually they pass.

```shell
./jbang/bin/jbang https://gist.github.com/tivrfoa/503c88fb5123b1000c37a3f2832d4773#file-file2-java
[jbang] Building jar...
I'm file 2

leandro:distributions$ ./jbang/bin/jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6
[jbang] Building jar...
Testing JBang multiple sources.
two

leandro:distributions$ ./jbang/bin/jbang https://twitter.com/maxandersen/status/1266329490927616001
  _   _          _   _               _   _                             
 | | | |   ___  | | | |   ___       (_) | |__     __ _   _ __     __ _ 
 | |_| |  / _ \ | | | |  / _ \      | | | '_ \   / _` | | '_ \   / _` |
 |  _  | |  __/ | | | | | (_) |     | | | |_) | | (_| | | | | | | (_| |
 |_| |_|  \___| |_| |_|  \___/     _/ | |_.__/   \__,_| |_| |_|  \__, |
                                  |__/                           |___/
```

tivrfoa:
```shell
$ ./jbang/bin/jbang https://gist.github.com/tivrfoa/24b68952bf8977c988e7168645cd9173
[jbang] [ERROR] Gist has more than one class with main method. You must specify which class to run, eg: https://gist.github.com/tivrfoa/24b68952bf8977c988e7168645cd9173/one.java
```

tivrfoa:
```shell
$ ./jbang/bin/jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6
Testing JBang multiple sources.
two
I'm two =)
wow =DDD
wow =DDD
```

maxandersen:
> I couldn't many tests because I hit Github API limit! Oo =D

yeah - its quite easy to exhaust github gist api limits when running anonymously. We really need to implement #168 to avoid the issue.

maxandersen:
> Some tests now fail and I don't know why. Does Gradle run tests in parallel?

I also see tests fail and I haven't digged into them but they definitely seem to be caused by some change in the url detection/check that makes it report it can't read the scripts.

Let me know if you want to digg in and step through to see what happens or let me know and I can try it out.

For now there is too many failures to this "just" being gist api limits.

maxandersen:
I took a quick look - something fishy is definitely going on. `scriptResource.getFile()` returns a cache url location for files that are directly available on disk - so somehow it thinks even local files are urls, skips "downloading" but does set the file to be a location in the cache that does not exist as there was nothing to download.

Why that happens...no idea (yet).

tivrfoa:
> > Some tests now fail and I don't know why. Does Gradle run tests in parallel?
> 
> I also see tests fail and I haven't digged into them but they definitely seem to be caused by some change in the url detection/check that makes it report it can't read the scripts.
> 
> Let me know if you want to digg in and step through to see what happens or let me know and I can try it out.
> 
> For now there is too many failures to this "just" being gist api limits.

I will try to find out. I couldn't debug the tests in VSCode, but at least I can debug the application.

If you look the changes, the only tests that were supposed to be affected are for gists.
There are two things that maybe are causing side effects:
1. GistFile.isGist is `static`. I don't know if values of static fields are preserved and thus cause problems;
2. I created the method `getUrlCache` in Util because it's used in many places:
```diff
-			String urlHash = Util.getStableID(scriptURL);
-			File urlCache = Settings.getCacheDir(Settings.CacheClass.urls).resolve(urlHash).toFile();
+			File urlCache = Util.getUrlCache(scriptURL).toFile();
```

I can't test now. I'll test later.

I would ask you if you just could build without the tests and try manually some test that failed.

maxandersen:
> * GistFile.isGist is `static`. I don't know if values of static fields are preserved and thus cause problems;

that most definitely is a problem. by being static its basically a global variable. You'll need to fix that before any of this will work reliably.

src/main/java/dev/jbang/GistFile.java
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class GistFile {
+	public static boolean isGist;
```
maxandersen, line 12: we shouldn't be keeping state that is specific to one file globally. this will definitely cause problems - especially during tests or when we have multiple gist references.

src/main/java/dev/jbang/Util.java
```diff
@@ -527,40 +581,133 @@ public static String getStableID(String input) {
...
+				fileName = anchor[1];
+				for (int i = 2; i < anchor.length - 1; ++i)
+					fileName += "-" + anchor[i];
+			}
+			String gistapi = pathPlusAnchor[0].replaceFirst(
+					"^https://gist.github.com/(([a-zA-Z0-9]*)/)?(?<gistid>[a-zA-Z0-9]*)$",
 					"https://api.github.com/gists/${gistid}");
-			// Util.info("looking at " + gistapi);
+
+			Util.verboseMsg("looking at " + gistapi);
```
maxandersen, line 669: please add context or remove the verbose logging.

tivrfoa: I removed this one.

src/main/java/dev/jbang/GistFile.java
tivrfoa: I removed the `static`s that I added.
I needed to create an instance variable in `ScriptResource`:
```java
private List<Path> resolvedSourcePaths;
```
Basically if it's not `null` then there's no need to search sources.

And the good thing for gists is that it now handles `//SOURCES` in all files.
You can try this:
./jbang/bin/jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6

`one.java` references `two.java` that references `t3.java`

maxandersen: is that only working for gist or you made it work for all resources ? 

that was on my todo list thus great if you made that work ;)

maxandersen:
@tivrfoa is this ready for review again ? if it is just press "re-request review" then i'll test this again.

src/main/java/dev/jbang/GistFile.java
tivrfoa: Currently only gists (when starts with 'https://gist.github.com') are making use of that, but I think other resources will be able to use it too.

tivrfoa:
> "re-request review"

Press? I can't find any button to press xD

But it's ready for review. : ]

tivrfoa:
> > "re-request review"
> 
> Press? I can't find any button to press xD
> 
I found it 😄

src/main/java/dev/jbang/cli/BaseScriptCommand.java
```diff
@@ -303,13 +309,20 @@ private static ScriptResource fetchFromURL(String scriptURL) {
...
-			Path path = Util.downloadFileSwizzled(scriptURL, urlCache);
-
-			return new ScriptResource(scriptURL, urlCache, path.toFile());
+			if (Util.isGistURL(scriptURL)) {
+				GistCollection gistCollection = Util.extractFileFromGistWithSources(scriptURL);
+				File urlCache = Util.getUrlCache(gistCollection.getMainURL()).toFile();
+				Path path = Util.getPathToFile(gistCollection.getMainURL(), urlCache);
+				ScriptResource scriptResource = new ScriptResource(gistCollection.getMainURL(), urlCache,
+						path.toFile());
+				scriptResource.setResolvedSourcePaths(gistCollection.getResolvedSourcePathsAsList());
```
maxandersen, line 318: okey so this whole block works but not a fan of `//SOURCES` now have two different behaviours on a set a file between if you have it in a gist vs having it on the filesystem.

Do you want to give it a try to extend this to be more generic or we change this to only work based on the central file driving `//SOURCES` ?

src/main/java/dev/jbang/cli/BaseScriptCommand.java
```diff
@@ -303,13 +309,20 @@ private static ScriptResource fetchFromURL(String scriptURL) {
...
 
-			scriptURL = swizzleURL(scriptURL);
-
-			String urlHash = Util.getStableID(scriptURL);
-			File urlCache = Settings.getCacheDir(Settings.CacheClass.urls).resolve(urlHash).toFile();
-			Path path = Util.downloadFileSwizzled(scriptURL, urlCache);
-
-			return new ScriptResource(scriptURL, urlCache, path.toFile());
+			if (Util.isGistURL(scriptURL)) {
+				GistCollection gistCollection = Util.extractFileFromGistWithSources(scriptURL);
```
maxandersen, line 313: is there a reason why you don't put this logic into swizzleUrl() ?

src/main/java/dev/jbang/cli/BaseScriptCommand.java
tivrfoa: It's because the focus of this PR was to solve scripts in Gists.
I didn't want to change other places as it would make the PR bigger and harder to review.

I was thinking about making `//SOURCES` work for multiple files for the the cases, but it's probably better in another PR.
We could then even apply convention over configuration:
  - If imported classes are in the expected path, then maybe `//SOURCES` would not even be needed.

Then I would not need `//SOURCES` here:
https://github.com/tivrfoa/flatbuffers-java-rust/blob/main/java/SampleBinary.java

```java
//SOURCES MyGame/Sample/Color.java
//SOURCES MyGame/Sample/Equipment.java
//SOURCES MyGame/Sample/Monster.java
//SOURCES MyGame/Sample/Vec3.java
//SOURCES MyGame/Sample/Weapon.java
```

maxandersen: Wdym expected ? If there is no //SOURCE what would you use to know which files to look for ?

maxandersen: Btw. I'm fine with making this pr just about gist multi file support. 

I just then would like to see the behaviour be at least consistent.

src/main/java/dev/jbang/cli/BaseScriptCommand.java
tivrfoa: `swizzleUrl()` returns a String, but I need more than that, otherwise I would lose the information of what was already done, like source files path.

The way jbang works today without this PR, it gets the URL and then tries to download the files, but that's redundant work.

What does this PR solve?
  1. Make Gist work when there's a `#` in the URL, eg: jbang https://gist.github.com/joshlong/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java
  2. Make Gist work for multiple files, eg (3 files): jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6
  3. Fix: it considered the first file as the main file, but that's not necessarily the case. I check which file contains the `main` method.
  4. Fix: it redundantly downloaded the files, but that's not necessary for Gist because the content of all files are already returned by the Github API.

src/main/java/dev/jbang/cli/BaseScriptCommand.java
tivrfoa: > Wdym expected ? If there is no //SOURCE what would you use to know which files to look for ?

For example Color.java is in the *"expected"* path **relative** to the `main` class (`SampleBinary.java`):

https://github.com/tivrfoa/flatbuffers-java-rust/blob/main/java/MyGame/Sample/Color.java

maxandersen: Yes so how would you know that file is there if not with a //SoURCES ref?

maxandersen: Btw. I do want to add support for //SOURCES **/*.java but thats not configuration by convention which I dont See s way to do cleanly - curious to what approach you are considering ?

tivrfoa: > Yes so how would you know that file is there if not with a //SoURCES ref?

Beyond the current way of checking `//SOURCES`, I would check the imports if they exist in the "expected" path.

It may seem that would be to much work and inefficient, but I think it will be simple, because I would only check once for each prefix.

So in my example:
```java
import MyGame.Sample.Weapon;
import java.nio.ByteBuffer;
import java.io.FileOutputStream;
import java.nio.channels.FileChannel;
```
I would check MyGame, then MyGame.Sample, then MyGame.Sample.Weapon.

Then `java` would be checked only once for `java.nio.ByteBuffer`, and would mark that the relative path does not exist.

maxandersen: Okey, so based on imports. That just have quite a few edge cases. I would go for //SoURCEs **/*.java instead. Checking every import will be quite slow.

src/main/java/dev/jbang/cli/BaseScriptCommand.java
maxandersen: Yes, so 1, 3 and 4 I'm fine with. its 2 that makes things inconsistent. can you disable that and make PR for it working generically ? 

btw. for 4, a gist will not necessarily have *all* files as you can technically have a ref between two gists or just any url thus multiple requests is needed (which can then be cached) - at least when fully implemented.

tivrfoa: > Yes, so 1, 3 and 4 I'm fine with. its 2 that makes things inconsistent. can you disable that and make PR for it working generically ?

By *inconsistent* you mean that multiple sources was only working for Gists?
I changed the code. I think it's now working for everything.

> 
> btw. for 4, a gist will not necessarily have _all_ files as you can technically have a ref between two gists or just any url thus multiple requests is needed (which can then be cached) - at least when fully implemented.

I didn't know you could have url in //SOURCES. That's interesting.
My new change handles that.

Now the followings also work:
  1. Gist work when there's a # in the URL, eg: jbang https://gist.github.com/joshlong/d8768b697a73b62bf42d8b74ec43b760#file-greetingsserver-java
  2. Multiple sources in Gist: jbang https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6
  3. Multiple sources in files one -> two -> three:
```
./jbang/bin/jbang --verbose ~/dev/java/jbang/one.java
[jbang] Building jar...
[jbang] compile: /usr/lib/jvm/java-11-openjdk-amd64/bin/javac -d /home/leandro/.jbang/cache/jars/one.java.1fc5dcbe49ea9d83719ece10a2fb33b09d91a76cb883e660a4df77b03a8e3c89 /home/leandro/dev/java/jbang/one.java /home/leandro/dev/java/jbang/two.java /home/leandro/dev/java/jbang/three.java
[jbang] run: /usr/lib/jvm/java-11-openjdk-amd64/bin/java -classpath /home/leandro/.jbang/cache/jars/one.java.1fc5dcbe49ea9d83719ece10a2fb33b09d91a76cb883e660a4df77b03a8e3c89.jar: one
two constructor
three constructor
```

I changed the test `multiSourcesHttp` so now `two` references (//SOURCES) `three`.

One thing that already worked but needs a test: `testFetchFromGistReferencingAnotherURL`
I tried to add a test for Gist Referencing URL in //SOURCE, but it doesn't work because of the "trust" checking.
But it's working if you run: jbang https://gist.github.com/tivrfoa/39a0dee0ef32a75a064fe9c59c2bd68a

tivrfoa: Hi @maxandersen ,

Now the PR is much simpler.

I was able to simplify it by analyzing the Gist API, eg:
https://api.github.com/gists/8e6ea001f168fd4ef14763ceca3e5ab6

So the problem was that JBang sees:<br>
raw_url	ht<span>tp</span>s://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/**5d4fb921c76c686bc96a8bf4993304c2d95dcbab**/one.java

And then tries the same path for other files, eg:
ht<span>tp</span>s://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/**5d4fb921c76c686bc96a8bf4993304c2d95dcbab**/two.java

But this path **does not exist.** The path for `two.java` is:
raw_url	ht<span>tp</span>s://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/**dfc4ac2bf9ad11cd8db5f2b151162bcd3b11efda**/two.java

But then I found out that if we use history[0].version, then it's the same path for all files! :smiley:

So I basically had to make a change in swizzleURL for Gist: `extractFileFromGist`

```diff
+				String[] tmp = entry.getValue().get("raw_url").split("/raw/");
+				String prefix = tmp[0] + "/raw/";
+				String suffix = tmp[1].split("/")[1];
+				String mostRecentVersionRawUrl = prefix + gist.history[0].version + "/" + suffix;


	static class Gist {
		Map<String, Map<String, String>> files;
+		History[] history;
+	}
+
+	static class History {
+		String version;
+	}
```

tivrfoa: I still need to handle more use cases:

  1. Local file referencing a url
```
leandro:distributions$ ./jbang/bin/jbang --verbose ~/dev/java/jbang/Main.java 
[jbang] [ERROR] /home/leandro/dev/java/jbang/https:/gist.github.com/tivrfoa/bb5deb269de39eb8fca9636dd3c9f123#file-gsonhelper-java
java.nio.file.NoSuchFileException: /home/leandro/dev/java/jbang/https:/gist.github.com/tivrfoa/bb5deb269de39eb8fca9636dd3c9f123#file-gsonhelper-java
```
  2. url referencing url with sources and deps
```
leandro:distributions$ ./jbang/bin/jbang --verbose https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
[jbang] Gist url api: https://api.github.com/gists/475c4add1b3df035d13820cb756dacfc
[jbang] Gist url api: https://api.github.com/gists/bb5deb269de39eb8fca9636dd3c9f123
[jbang] [ERROR] Could not download Albums relatively to https://gist.githubusercontent.com/tivrfoa/475c4add1b3df035d13820cb756dacfc/raw/0500521b3906d849293d8893a8ee34e7c76f4209/Main.java
```

maxandersen:
You are on fire :)￼

Ready to review or more coming ?

tivrfoa:
> You are on fire :)￼
> 
> Ready to review or more coming ?

Please review.

But as I said in my last comment, those cases don't work yet.
I'll try to make it work later.
It's something to do with `fetchIfNeeded` that will probably have to change to make things work recursively.

But I think this PR is good enough now. :grin:

src/main/java/dev/jbang/cli/BaseScriptCommand.java
maxandersen: > But then I found out that if we use history[0].version, then it's the same path for all files! 😃

interesting, do you get the right version of the files then ?

tivrfoa: > > But then I found out that if we use history[0].version, then it's the same path for all files! 😃
> 
> interesting, do you get the right version of the files then ?

Yes. history[0] is the most recent.
You can check here for example:
https://api.github.com/gists/8e6ea001f168fd4ef14763ceca3e5ab6

src/main/java/dev/jbang/Util.java
```diff
@@ -669,4 +705,45 @@ private static boolean createHardLink(Path src, Path target) {
 		return true;
 	}
 
+	public static Path getUrlCache(String fileURL) {
+		String urlHash = getStableID(fileURL);
+		return Settings.getCacheDir(Settings.CacheClass.urls).resolve(urlHash);
+	}
+
+	public static Path getPathToFile(String scriptURL, File baseDir) {
```
maxandersen, line 713: this method does not seem to be used?

src/main/java/dev/jbang/cli/BaseScriptCommand.java
```diff
@@ -153,10 +157,19 @@ public static Script prepareScript(String scriptResource, List<String> arguments
 			s.setAdditionalClasspaths(classpaths);
 
 			List<Path> resolvedSourcePaths = new ArrayList<>();
-			for (FileRef collectSource : s.collectSources()) {
```
maxandersen, line 156: why move away from using FileRef ? where FileRef then resolves the relative location ?

src/main/java/dev/jbang/Util.java
```diff
@@ -669,4 +705,45 @@ private static boolean createHardLink(Path src, Path target) {
...
+		// ]{1,}void[ ]{1,}main[ ]{0,}\()
+		// TODO create regular expression
+		return content.contains("public static void main(");
+	}
+
+	public static boolean isGistURL(String scriptURL) {
+		return scriptURL.startsWith("https://gist.github.com/");
+	}
+
+	public static List<String> collectSources(List<String> lines) {
```
maxandersen, line 729: this and the other collectSources() feels like should be on Script class?

maxandersen (changes requested):
I'm going to take it for a spin but initial impressions looks good - very similar to what I originally thought it would be.

added a few comments and question about why dropping FileRef

src/main/java/dev/jbang/Script.java
```diff
@@ -629,4 +629,8 @@ public String getJavaAgentOption() {
 	public List<KeyValue> getAgentOptions() {
 		return agentOptions;
 	}
+
+	public String getScript() {
```
maxandersen, line 633: not fan of exposing this - should be accessing what we need from this instance so we can agressively cache results rather than rerun them.

src/main/java/dev/jbang/cli/BaseScriptCommand.java
tivrfoa: The `Util.collectSources` returns `List<String>` instead `List<FileRef>` because I didn't see the need to convert to FileRef only to then `collectSource.getDestination()`

But I may be missing something. I see that `toFileRef` has a split:
```java
String[] split = fileReference.split("=");
```
So probably my change can break some case that is not in the test set.
If you could send me one test to add please.

I'll change Util.collectSources to return `List<FileRef>` as it's done in `Script.collectSources` to avoid any problems, ok?

src/main/java/dev/jbang/Util.java
tivrfoa: > this and the other collectSources() feels like should be on Script class?

I moved them to the Script class.

src/main/java/dev/jbang/Script.java
tivrfoa: So my last changes were made to be possible to search for `//SOURCES` in multiple files.
`Script.collectSources` only works for the main file (entry point), that's why I created Util.collectSources(String content) so it can work recursively. But then I needed to pass Script.script to this new method, that's why I "exposed" it. =)

But now I realized that for the initial file I can still call `Script.collectSources`, so I removed `getScript` and now it's not exposed anymore. :]

src/main/java/dev/jbang/Util.java
tivrfoa: Removed.

tivrfoa:
I was able to make the following work!!! =D
I'll not send in this PR. It's better in a follow-up.

```java
distributions$ ./jbang/bin/jbang --verbose https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
[jbang] Gist url api: https://api.github.com/gists/475c4add1b3df035d13820cb756dacfc
[jbang] Gist url api: https://api.github.com/gists/bb5deb269de39eb8fca9636dd3c9f123
[jbang] Deleting /home/leandro/.jbang/cache/urls/25cae352dc346e7bd92a95c2d16f07be54e8e7c4be17ed55d0c5ac0b126cdf85.old
[jbang] Deleting /home/leandro/.jbang/cache/urls/c7277338d3f6d5673ac46de9dcf03afcf51138f800a0aef5d37e158f48910983.old
[jbang] Deleting /home/leandro/.jbang/cache/urls/69547f1befd4cdfdeee4f807ed3b60880da1f6a0d9a3b0e8cc6fa4e4c2d6e00f.old
[jbang] Deleting /home/leandro/.jbang/cache/urls/7cc01d00bdfc173712a94063891e9e5a07c867fb1f2bce53638abd6274d502c4.old
[jbang] System Java version 11 incompatible, using Jbang managed version 15
[jbang] Building jar...
[jbang] compile: /home/leandro/.jbang/cache/jdks/15/bin/javac -classpath /home/leandro/.m2/repository/com/google/code/gson/gson/2.8.6/gson-2.8.6.jar -d /home/leandro/.jbang/cache/jars/Main.java.36b65c1692d9c328adb25d3b8e80258428f6ab044ae739026c00fc170fd9e3ae /home/leandro/.jbang/cache/urls/c8f10411c43b3df4a8a1c375f5e380db65847c4c6b0442ae7d63932e52aab98b/Main.java /home/leandro/.jbang/cache/urls/25cae352dc346e7bd92a95c2d16f07be54e8e7c4be17ed55d0c5ac0b126cdf85/GsonHelper.java /home/leandro/.jbang/cache/urls/c7277338d3f6d5673ac46de9dcf03afcf51138f800a0aef5d37e158f48910983/Albums.java /home/leandro/.jbang/cache/urls/69547f1befd4cdfdeee4f807ed3b60880da1f6a0d9a3b0e8cc6fa4e4c2d6e00f/Dataset.java /home/leandro/.jbang/cache/urls/7cc01d00bdfc173712a94063891e9e5a07c867fb1f2bce53638abd6274d502c4/AlbumImages.java
[jbang] System Java version 11 incompatible, using Jbang managed version 15
[jbang] run: /home/leandro/.jbang/cache/jdks/15/bin/java -classpath /home/leandro/.jbang/cache/jars/Main.java.36b65c1692d9c328adb25d3b8e80258428f6ab044ae739026c00fc170fd9e3ae.jar:/home/leandro/.m2/repository/com/google/code/gson/gson/2.8.6/gson-2.8.6.jar Main
Testing //SOURCES url, where url also contains //SOURCES and //DEPS
Free Music Archive - Albums
Album 1
[AlbumImages@26f0a63f]
```

maxandersen:
> I was able to make the following work!!! =D

at first I couldn't grok what was different but it is that you fetch deps recursively too. 

I think I'm going to merge this as its functionally good; but I still do believe we should end up with treating sources as Scripts not as raw strings - especially when we start doing //DEPS....for now we can live with that.

src/main/java/dev/jbang/Script.java
maxandersen: I've made a commit that moves the static methods back to Util. 

I'm still a bit eery that we are collecting sources by iterating raw source strings rather than build up list of Script (or some other class) so we at least could attach other info to them - like which deps they have.

But I think we can do that in a follow up when we need it.

maxandersen:
Okey I squashed and rebased to keep history a bit clean - this is in! thanks alot @tivrfoa 

@all-contributors please add @tivrfoa  for code

maxandersen:
> ```java
> distributions$ ./jbang/bin/jbang --verbose https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
> ```

one thing I realized after merging this PR is that urls as used in //SOURCES in the above example we probably should discourage or disable as there are not a way for the user to declare which path it should be at..it only works "by chance" when they are in default package.

tivrfoa:
> > ```java
> > distributions$ ./jbang/bin/jbang --verbose https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
> > ```
> 
> one thing I realized after merging this PR is that urls as used in //SOURCES in the above example we probably should discourage or disable as there are not a way for the user to declare which path it should be at..it only works "by chance" when they are in default package.

I don't understand what you mean. This works:
```
jbang https://gist.github.com/tivrfoa/510a309ac341fda6a93100a2b620ab2d
```

One point about multiple files in a Gist:
When there are more than one file it searches the first one with `main` method and returns it.
Although currently it's a very basic check:
```java
public static boolean hasMainMethod(String content) {
    return content.contains("public static void main(");
}
```
I am working one a regular expression, but I am having trouble to make it work. xD
```java
public static final Pattern patternMainMethod = Pattern.compile(
			"(.*\\n*\\s*)(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*");
```
It passed all tests but I don't know why it's returning false for `one.java` here: https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6

```java
		assertTrue(Util.hasMainMethod("public static void main("));
		assertTrue(Util.hasMainMethod("public static void main(String"));
		assertTrue(Util.hasMainMethod("    public static void main(String"));
		assertTrue(Util.hasMainMethod("class Main {public static void main("));
		assertTrue(Util.hasMainMethod("class Main { public static void main("));
		assertTrue(Util.hasMainMethod("class Main {\tpublic static void main("));
		assertTrue(Util.hasMainMethod("class Main {\npublic static void main("));
		assertTrue(Util.hasMainMethod("class Main {\n\n    public static void main("));
		assertTrue(Util.hasMainMethod("       public   static   void   main  ("));
		assertTrue(Util.hasMainMethod("class Main {static public void main("));
		assertTrue(Util.hasMainMethod("       static   public   void   main  ( "));
		assertTrue(Util.hasMainMethod("}public static void main("));
		assertTrue(Util.hasMainMethod(";public static void main("));

		assertFalse(Util.hasMainMethod("class Main {public void main("));
		assertFalse(Util.hasMainMethod("class Main {static void main("));
		assertFalse(Util.hasMainMethod("public String helloWorld("));
```

I will try to make it work later.

tivrfoa:
> ```java
> ```java
> public static final Pattern patternMainMethod = Pattern.compile(
> 			"(.*\\n*\\s*)(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*");
> ```
> 
> 
> It passed all tests but I don't know why it's returning false for one.java here: https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6
> ```

I think I found a solution, just add `\\n*$` to the end of the pattern:
```java
public static final Pattern patternMainMethod = Pattern.compile(
			"^(.*\\s*)(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*\\n*$");
```

I'll test later.

tivrfoa:
> ```java
> ```java
> distributions$ ./jbang/bin/jbang --verbose https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
> ```
> 
> 
> 
> one thing I realized after merging this PR is that urls as used in //SOURCES in the above example we probably should discourage or disable as there are not a way for the user to declare which path it should be at..it only works "by chance" when they are in default package.
> 
> I don't understand what you mean. This works:
> jbang https://gist.github.com/tivrfoa/510a309ac341fda6a93100a2b620ab2d
> ```

Ops. I see what you mean. I think the packages are resolved automatically. I'll make some tests later.

src/main/java/dev/jbang/Script.java
tivrfoa: Store in some kind of data structure maybe it is a better approach. Script could have a list of ScriptResource or some new class (Source?)

I thought about that but it would be a much complex change, touching many places ...

But the good thing is that I think this change is not really needed now, because once we get all `//SOURCES` recursively, their `Path`s will be stored in `Script.resolvedSources`. So we only need to loop through them to get `//DEPS` and `//FILES`

maxandersen:
Ah true. Source location doesn't matter. Javac gets it the right place.

PR #436, opened 2020-10-13
fix: Default alias name when creating a new alias

### LINKED ISSUE #374: alias add -f directory should create jbang-catalog.json in that directory

chrsblck:
fixes #374

### DIFF
--- a/src/main/java/dev/jbang/cli/Alias.java
+++ b/src/main/java/dev/jbang/cli/Alias.java
@@ -1,7 +1,9 @@
 package dev.jbang.cli;
 
 import java.io.PrintStream;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


chrsblck:
@quintesse @maxandersen This is ready for review, thanks!

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -28,6 +30,8 @@
 	protected Path getCatalog() {
 		if (global) {
 			return Settings.getAliasesFile();
+		} else if (catalogFile != null && Files.isDirectory(catalogFile)) {
```
quintesse, line 33: Almost, we still need to check for the `catalogFile/.jbang/jbang-catalog.json` file as well. Also remember that in case that the file doesn't exist we still need to return `catalogFile/jbang-catalog.json` so `alias add` knows what file to.

quintesse (changes requested):

src/main/java/dev/jbang/cli/Alias.java
chrsblck: Oops 🤦

chrsblck: @quintesse Fixed.

quintesse (approved):
Looks good!

chrsblck:
Let me know if you want me to squash the changes, or maybe the GitHub `squash and merge` is enough

maxandersen:
perfect! thanks @chrsblck !

Issue #406, opened 2020-10-08
Implement local catalog indices


quintesse:
Just as we implemented local alias catalogs, with the concept of looking up an alias in the "nearest" catalog, we now want to extend that to the catalog index '~/.jbang/catalogs.json' as well.

The original idea was:

1. rename the file to `jbang-catalogs.json` for consistency
2. stay backward compatible with the old '~/.jbang/catalogs.json' (for now)
3. factor out the generic "find nearest" code so we don't duplicate it
4. add options to the catalog commands to explicitly set the catalogs file to use (like we do for alias)
5. split out the implicit catalogs to a different file (eg '~/.jbang/jbang-catalogs-implicit.json')

But the new idea is to:

1. store catalogs together with aliases in the same `jbang-catalog.json` file
2. either stay backward compatible with the old '~/.jbang/catalogs.json' or import it on first use
3. not needed anymore
4. stays the same
5. stays the same

Edit: updated description to better reflect the latest ideas

maxandersen:
What would that file in #5 actually be used for ? Avoid iterating GitHub, girl AB, bitbucket locations ?

quintesse:
Yeah, like I explained earlier, it could easily take 6-89 requests in worst case scenarios. That seems a bit excessive, right?
It's also easier to fit into the current system. Doing it without that file would require a more extensive rewrite.

maxandersen:
so... `~/.jbang/cache/implicit-catalogs.json` ?

PR #437, opened 2020-10-13
feat: Implement local catalog indices

quintesse:
Implements the features mentioned in #406 

Still needs
 - [x] tests
 - [ ] backward compatibility with the old catalogs.json file
 - [x] some docs

### DIFF
--- a/itests/catalog.feature
+++ b/itests/catalog.feature
@@ -1,30 +1,29 @@
 Feature: catalog testing
 
 Scenario: java catalog list
-  When command('jbang catalog add tc jbang-catalog.json')
+  When command('jbang catalog add --global tc jbang-catalog.json')
   And command('jbang catalog list')
   Then match out contains "tc = JBang test scripts"
 
 Scenario: add catalog and run catalog named reference
-  When command('jbang catalog add tc jbang-catalog.json')
+  When command('jbang catalog add --global tc jbang-catalog.json')
   Then command('jbang echo@tc tako')
   Then match out == "0:tako\n"
 
 Scenario: add catalog and remove
-  When command('jbang catalog add tc jbang-catalog.json')
+  When command('jbang catalog add --global tc jbang-catalog.json')
   Then command('jbang catalog remove tc')
   Then command('jbang echo@tc tako')
   Then match err contains "Unknown catalog 'tc'"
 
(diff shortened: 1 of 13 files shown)


itests/catalog.feature
```diff
@@ -1,30 +1,29 @@
 Feature: catalog testing
 
 Scenario: java catalog list
-  When command('jbang catalog add tc jbang-catalog.json')
+  When command('jbang catalog add --global tc jbang-catalog.json')
```
maxandersen, line 4: why is --global now needed here ?

maxandersen: ah to avoid any conflicts with jbang catalogs in jbang itself.

maxandersen:
btw. just using this for a few minutes and we definitely want this :) 

```
jbang catalog add jbangdev jbangdev
jbang gavsearch
```

yay!

itests/catalog.feature
quintesse: Yeah :-)

It would probably be nice if tests were run from a clean CWD but that would probably mess up the tests that need extra files.

quintesse:
So the only thing missing is the "backward compatibility with the old catalogs.json file"... and I'm actually not sure how to handle this. First I'd thought to just include it in the catalog lookup and although that works it makes it almost an official "feature" of the lookup, not really like a nice path towards deprecation. THe other way would be to check for the file at startup, move over any entries to the new file and then delete the old one. THe final option is to just ignore it and have a one-time pain point of people losing the catalogs they had defined (if any). Wdyt @maxandersen ?

maxandersen:
Those catalogs did nothing besides providing Alias and caching ?

quintesse:
> Those catalogs did nothing besides providing Alias and caching ?

Indeed. Users will only notice it if they ever _explicitly_ defined a catalog alias and were using it like `jbang run myscript@mycatalog`. With the code as it is now they'd have to re-add the catalog alias.

maxandersen:
If easy to do then import them and delete but if too tedious to do right then print info/warning.

quintesse:
@maxandersen you can merge this when the tests pass

maxandersen:
i couldn't get it to fail and code looked plain so merged ;)

PR #440, opened 2020-10-14
fix: add readme.md to edit fixes #210

### LINKED ISSUE #210: add readme.md on jbang edit
explain how project work in readme.md on jbang edit .
﻿

maxandersen:
idea is to give users a bit of a context and at least intellij opens up readme.md automatically.


maxandersen:
apparently it is README.md according to https://stackoverflow.com/questions/61966528/automatically-open-files-on-intellij-webstorm-startup which I know I have seen but can't actually get to work right now.

in any case, made it README.md for more likely to work and to have a readme ;)

Issue #363, opened 2020-09-27
Provide some samples or documentation


grumpyf0x48:
Apart from ```jbang help``` and the examples of jbang use given in the readme, it would be great to have:

1. A documentation - Picocli generated ?
2. A cheat sheet usable with https://github.com/cheat/cheat

The second one is very efficient, as you document only the most frequent usages of a command, so you find very quickly the command you need.

maxandersen:
Agreed. Would you be interested taking a stab on making any these two happen ?

maxandersen:
Btw. By documentation in #1 I assume you refer to man pages or similar ?

grumpyf0x48:
> Agreed. Would you be interested taking a stab on making any these two happen ?

Yes, i'm interested on starting a cheat sheet.

The idea i have is

- create a new file jbang in a new folder cheat of the jbang repository
- when the commands and flags of jbang will not change too much (maybe after the 1.0 release), then submit this cheat to the community ones (https://github.com/cheat/cheatsheets)

maxandersen:
Sounds like a good start. 

The command structure is fairly stable at this stage so I wouldn't be too concerned submitting to cheat Repo sooner rather than later. 

But let's get it started first and see where it goes.

grumpyf0x48:
> Btw. By documentation in #1 I assume you refer to man pages or similar ?

Yes man page is an option for Linux users but maybe not for Windows ones.

I was more thinking of the abilility to generate documentation from Picocli annotations, it could help to keep documentation up to date.

maxandersen:
There is https://picocli.info/#_generate_man_page_documentation but I haven't explored how it works and if worth it. 

PR's/suggestions welcome as man page for jbang would make sense imo.

grumpyf0x48:
> There is https://picocli.info/#_generate_man_page_documentation but I haven't explored how it works and if worth it.
> 
> PR's/suggestions welcome as man page for jbang would make sense imo.

Will take a look at that next week.

maxandersen:
just saw https://cli.github.com/manual/ - that is a sweet way of documenting a cli. wonder if we could style the docs generated with similar navigation/style ?

abelsromero:
> wonder if we could style the docs generated with similar navigation/style ?

If you are just talking about aggregating docs with a lateral menu. Antora could be an option.
[...]

PR #444, opened 2020-10-15
Provide some samples or documentation #363

grumpyf0x48:
Include manpages in distribution.zip.

manpage generated by picocli-codegen are added to a manpage folder in distribution.zip.

After build, you can see what it looks like with:

```console
$ man build/docs/manpage/jbang.1
```

i would have liked to zip each file before inclusion in distribution.zip but it looks like i'm too new to Gradle for that yet :-(


maxandersen:
Not a compute r now so can’t test it but is there html version it can generate ?

grumpyf0x48:
> Not a compute r now so can’t test it but is there html version it can generate ?

For instance i was into adding the manpages.

But yes the picocli generator can generate HTML5 pages as shown in the example.

maxandersen:
Thanks for getting it to work - my attempt failed!

so this seem to generate a man page *per* command/subcommands, 31 man pages in total. That seems excessive and not really user friendly ...like - when I do `man ./jbang.1` I get very little info and no option to navigate/discover the actual docs.

I think a html page published on jbang.dev would be better approach than this.

wdyt ?

grumpyf0x48:
> Thanks for getting it to work - my attempt failed!
> 
> so this seem to generate a man page _per_ command/subcommands, 31 man pages in total. That seems excessive and not really user friendly ...like - when I do `man ./jbang.1` I get very little info and no option to navigate/discover the actual docs.
> 
> I think a html page published on jbang.dev would be better approach than this.
> 
> wdyt ?

Yes the `ManPageGenerator` generates one asciidoc file per 'final' command and then we get one manpage per asciidoc file. I've searched a way of having less files but could not find it so far: i think about asking to Picocli folks about it.

Having an html file with all commands, subcommands and options on jbang.dev would be great, but if we use the same generator, we'll end up with many html pages as well. Maybe we need to find another generator.

maxandersen:
Maybe just create a Uber asciidoc that include all adoc so it All ends up in one?

grumpyf0x48:
> Maybe just create a Uber asciidoc that include all adoc so it All ends up in one?

I've modified the build to have:

- one uber manpage
- one manpage per sub-command

I've used asciidoc templates in the repository (which could probably be generated automatically) and we now end up with 13 manpages files instead of 31.

build.gradle
```diff
@@ -277,6 +283,25 @@ task tag(type: Exec) {
...
+	main 'picocli.codegen.docgen.manpage.ManPageGenerator'
+	args mainClassName, "--outdir=${buildDir}/generated-picocli-docs"
+}
+
+asciidoctor {
+	dependsOn(generateManpageAsciiDoc)
+	sourceDir = file("${projectDir}/misc/doc-templates")
+	outputDir = file("${buildDir}/docs")
+	logDocuments = true
+	backends 'manpage'
```
maxandersen, line 302:
```suggestion
	backends = ['html5', 'manpage']
```

then we also get the html5 version.

build.gradle
```diff
@@ -305,6 +330,7 @@ versioner {
 	}
 }
 
+build.dependsOn(asciidoctor)
```
maxandersen, line 333:
```suggestion
//build.dependsOn(asciidoctor)
```

its quite slow and still not fully releasable thus I would keep `asciidoctor` as an independent task.

build.gradle
```diff
@@ -11,6 +12,7 @@ plugins {
 	id 'de.fuerstenau.buildconfig' version '1.1.8'
 	id "com.dorongold.task-tree" version "1.5"
 	id "com.geoffgranum.gradle-conventional-changelog" version "0.3.1"
+	id "org.asciidoctor.convert" version "2.4.0"
```
maxandersen, line 15: this generates quite the few warnings - tried move it to latest `org.asciidoctor.jvm.convert` but then the whole include setup can't find their files.

build.gradle
```diff
@@ -93,6 +96,9 @@ def commonSpec = project.copySpec {
 	from(project.buildDir) {
 		include 'tmp/version.txt'
 	}
+	from("${buildDir}/docs/manpage") {
```
maxandersen, line 99: I would do the inclusion in a separate PR as it also needs enablement in the various install scripts as otherwise its just "unnecessary bagage"

maxandersen:
> * one uber manpage
> * one manpage per sub-command

this is definitely better.

But it still feels a bit "weak" if you enable html generation you get a nice single page but no links at all which surprised me.

I'm wondering if we are hitting the limits of picocli codegenerator and better if we extended it to generate a bit more useful output ?

build.gradle
grumpyf0x48: ok

build.gradle
grumpyf0x48: I removed html5 because the result was not totally ok: many pages, no links between them.

build.gradle
grumpyf0x48: ok

build.gradle
grumpyf0x48: it is working now.

grumpyf0x48:
Will we do something with this MR or we close it until Picocli handles properly sub-commands ?

maxandersen:
I think in its current form its not of that great value - I've picked out the asciidcotor setup so that is in master now. Did you see the discussion at https://github.com/remkop/picocli/issues/1227 ?

Thanks for it - I'll close it for now as I i think the generated format and needed maintanence needed of auxilary files are too "weak" to be of real value at this time.

@all-contributors please add @grumpyf0x48  for infra

grumpyf0x48:
> I think in its current form its not of that great value - I've picked out the asciidcotor setup so that is in master now. Did you see the discussion at https://github.com/remkop/picocli/issues/1227 ?
> 
> Thanks for it - I'll close it for now as I i think the generated format and needed maintanence needed of auxilary files are too "weak" to be of real value at this time.
> 
> @all-contributors please add @grumpyf0x48  for infra

I've seen the discussion but nothing of what i have tried for jbang manpages can be reused in  picocli.

To me, even if not perfect, it would have been better than nothing to have the manpages like this, which look very similar to what gît manpages are - without the links between commands.

maxandersen:
for me manpages biggest issue is they are hard to browse - so I really want that solved before committing to maintain the external adoc files.

Issue #452, opened 2020-10-18
JBang start script gives an error when used with MSYS2


kjjaeger:
I get this error when trying to run JBang on Windows using MSYS2 (https://www.msys2.org/) 
Unsupported Operating System: MSYS_NT-10.0-19041

I was able to solve the issue by editing the start script src/main/scripts/jbang replacing line 49
  CYGWIN*|MINGW*)
with
  CYGWIN*|MINGW*|MSYS*)

I may be making this into a pull request shortly.

maxandersen:
this was merged

PR #454, opened 2020-10-18
fix: allow MSYS2 to pass OS check in jbang script

kjjaeger:
Using MSYS2 in Windows causes uname -s returns a string like: MSYS_NT-10.0-19041
Added MSYS* to case statement so that using MSYS2 is detected as Windows.
This is a fix for the issue I created here: https://github.com/jbangdev/jbang/issues/452


maxandersen:
thank you @kjjaeger !

@all-contributors please add @kjjaeger for code

PR #457, opened 2020-10-20
feat: Added support downloading JDK if not found on Raspberry Pi

karstentorp:
Added Raspberry Pi armv7l architecture (arm) to the list of known architecture for download.
Raspberry Pi OS come without java installed, and with this addition to the architecture "list" jbang can auto download java.
I have tested with helloworld.java from the example dir on my Pi with fresh Raspberry Pi OS.

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -58,6 +58,8 @@ case "$(uname -m)" in
     arch=x64;;
   aarch64)
     arch=aarch64;;
+  armv7l)
+    arch=arm;;
   *)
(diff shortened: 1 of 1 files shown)


maxandersen:
@karstentorp i've triggered a re-run as the first failure seem to have been caused by error in github action infrastructure; not your code. The semantic pull request you can fix by changing the commit message to start with "feat:" to indicate its a new feature (albeit could also be a fix but i think its fine as a feature) - but not a big deal as I can squash on merge and do the rewording if need be.

PR #468, opened 2020-10-29
fix: better check for String contains main method

tivrfoa:
Hi,

This patch fix/enhance the check that a String contains `main` method.

It is probably not exhausting, but it is much better than what I did before.

### DIFF
--- a/src/main/java/dev/jbang/Util.java
+++ b/src/main/java/dev/jbang/Util.java
@@ -47,6 +47,10 @@ public class Util {
 
 	public static final String SOURCES_COMMENT_PREFIX = "//SOURCES ";
 
+	public static final Pattern patternMainMethod = Pattern.compile(
+			"^.*(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*",
+			Pattern.MULTILINE);
+
 	private static boolean verbose;
 	private static boolean quiet;
 
@@ -715,7 +719,7 @@ public class Util {
 	}
 
 	public static boolean hasMainMethod(String content) {
-		return content.contains("public static void main(");
+		return patternMainMethod.matcher(content).find();
 	}
 
 	public static boolean isGistURL(String scriptURL) {
--- a/src/test/java/dev/jbang/TestUtil.java
+++ b/src/test/java/dev/jbang/TestUtil.java
@@ -1,7 +1,55 @@
 package dev.jbang;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
 public class TestUtil {
 	public static void clearSettingsCaches() {
 		AliasUtil.catalogCache.clear();
+	}
+
+	@Test
+	void testHasMainMethod() {
+
+		// it's a false positive but it doesn't matter, because the code
+		// would not compile anyway.
+		assertTrue(Util.hasMainMethod("WOWpublic static void main("));
+
+		// this is an edge case: when it's inside a String ...
+		// But I don't think we need to handle it now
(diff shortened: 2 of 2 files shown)


quintesse:
That is indeed a lot better. Although does this match keyword being on different lines? Not main people do it but I _have_ seen things like:

```java
public static
void main(...) { ... }
```

It might be that the current regexp already matches this, but perhaps some kind of multi-line option needs to be set? I'm never sure how that works.

tivrfoa:
> That is indeed a lot better. Although does this match keyword being on different lines? Not many people do it but I _have_ seen things like:
> 
> ```java
> public static
> void main(...) { ... }
> ```
> 
> It might be that the current regexp already matches this, but perhaps some kind of multi-line option needs to be set? I'm never sure how that works.

Hi Tako. It works. Here you can test:

ps. I spent a lot of time analyzing the pattern because I was using `matches()` instead of `find()` ... :expressionless: 😁 

```java
import java.util.regex.*;
import java.nio.charset.Charset;
import java.nio.file.*;

public class Util {
	public static final Pattern patternMainMethod = Pattern.compile(
			"^.*(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*",
			Pattern.MULTILINE);

	public static boolean hasMainMethod(String content) {
		return patternMainMethod.matcher(content).find();
	}

	public static void main(String[] args) throws Exception {
		
		String content = new String(Files.readAllBytes(Paths.get("MultiLineMainMethod.java")), Charset.defaultCharset());
		assertTrue(Util.hasMainMethod(content));
		
		assertTrue(Util.hasMainMethod("public static void main("));
		assertTrue(Util.hasMainMethod("public static void main(String"));
		assertTrue(Util.hasMainMethod("    public static void main(String"));
		assertTrue(Util.hasMainMethod("class Main {public static void main(String...\n"));
		assertTrue(Util.hasMainMethod("class Main { public static void main("));
		assertTrue(Util.hasMainMethod("class Main {\tpublic static void main("));
		assertTrue(Util.hasMainMethod("class Main {\npublic static void main("));
		assertTrue(Util.hasMainMethod("class Main {\n\n    public static void main("));
		assertTrue(Util.hasMainMethod("       public   static   void   main  ("));
		assertTrue(Util.hasMainMethod("class Main {static public void main("));
		assertTrue(Util.hasMainMethod("       static   public   void   main  ( "));
		assertTrue(Util.hasMainMethod("}public static void main("));
		assertTrue(Util.hasMainMethod(";public static void main("));

		assertFalse(Util.hasMainMethod("class Main {public void main("));
		assertFalse(Util.hasMainMethod("class Main {static void main("));
		assertFalse(Util.hasMainMethod("public String helloWorld("));
	}
	
	public static boolean assertFalse(boolean b) {
		if (b) { throw new RuntimeException("Not false"); }
		return true;
	}
	public static boolean assertTrue(boolean b) {
		if (!b) { throw new RuntimeException("Not true"); }
		return true;
	}
}
```

```java
class MultiLineMainMethod {
    
    public static
	void
	main(String
	args
	[]) {
    	System.out.println("ops");
    }
}
```

tivrfoa:
> That is indeed a lot better. Although does this match keyword being on different lines? Not many people do it but I _have_ seen things like:
> 
> ```java
> public static
> void main(...) { ... }
> ```
> 
> It might be that the current regexp already matches this, but perhaps some kind of multi-line option needs to be set? I'm never sure how that works.

@quintesse you can test using JBang 😄 

jbang https://gist.github.com/tivrfoa/b45c0c7f811ef410348590587a5797ca

maxandersen:
looks good - covers more edge cases, still fail for some but those were also failures before. merging!

quintesse:
> Hi Tako. It works. Here you can test:

Cool, nice work :-)

PR #472, opened 2020-11-01
initial support for //SOURCE <pattern>

maxandersen:
This is adding basic (& naive) `//SOURCES **/*.java` support.

This is draft for now for a few reasons:

1) https is broken - needs fixing to track proper source.
2) file scanning is done even when the pattern is just a file ref, i.e. `//SOURCE file.java`﻿ should just fetch the file ; not do any ifo.


maxandersen:
with latest update 1 and 2 above is fixed.

We need to decide how jbang should happen when running via http with **/*.java here i'm leaning towards to print a warning as it might just still work.

maxandersen:
> We need to decide how jbang should happen when running via http with **/*.java here i'm leaning towards to print a warning as it might just still work.

added warning for this case.

PR #469, opened 2020-10-29
feat: search //SOURCES and //DEPS in multiple files

tivrfoa:
Hi!

This PR is not ready. I just coded something to illustrate some things that need to be solved to be able to search `//SOURCES` and `//DEPS` in multiple files recursively.

Originally JBang was just for one file (right? =), but with `//SOURCES` now it can be a script composed of multiple files.
So now we need to also search those new files for `//SOURCES`, `//DEPS` and possibly other things (maybe `//FILES`? I still need to learn how `//FILES` work)

To illustrate the problem I use the example below (ps: this PR makes this example work, but it's not the best solution. It was just a hack)

The entry point is this url: https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
So it'll be called:
```shell
jbang https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc
```

Two interesting things there:
  1. `//SOURCES` references a different url
  2. It uses Album, but this class is not in `//SOURCES`

```java
//SOURCES https://gist.github.com/tivrfoa/bb5deb269de39eb8fca9636dd3c9f123#file-gsonhelper-java
```

The first to thing to make this work is to make some methods like `collectDependencies` and `fetchIfNeeded` more generic accepting parameters, because today they handle just one file.

So for example, the referenced url (https://gist.github.com/tivrfoa/bb5deb269de39eb8fca9636dd3c9f123#file-gsonhelper-java) has: `//SOURCES Albums.java`
... but JBang can't find it, because `ScriptResourece.fetchIfNeeded` tries to find it relatively to the first url (`originalResource`).

The other thing we should do is like Max mentioned here:
@maxandersen in https://github.com/jbangdev/jbang/pull/435#discussion_r513233155_

Today `Script` has:
```java
private List<Path> resolvedSources;
```
I think we need to change that to:
```java
private List<Source (or ScriptResource)> resolvedSources;
```
And this class would have: `List<Path>`, dependencies, ...
Doing this refactoring now will make the code more maintainable and easier to extend.

So what are your thoughts on this?
What should be the relation between `Script` and `ScriptResource` now that JBang support multiple files?

PR #473, opened 2020-11-01
fix: handles when main file is inside a package

tivrfoa:
PR https://github.com/jbangdev/jbang/pull/469 introduced ability to search for `//SOURCES` and `//DEPS` in multiple files.
But there is one case (at least) where it doesn't work:
  1. main file is a local file that has `package` declared;
  2. It references other file(s) in `//SOURCES`  that is not in the same or child package.

Then it is not able to find the classes in the other packages because it tries to resolve their paths using the base path of the main file.

This PR does two things:
  1. it takes into account the package name and sets --source-path accordingly. eg:
      jbang /home/user/scripts/com/company/Main.java
      then --source-path will be /home/user/scripts
    This change should help cases where `//SOURCES` is not used, so javac will be able to resolve the paths. *ps: by not using `//SOURCES` you would not be able to use some of the JBang features.*
  2. it correctly resolves base path for `//SOURCES` when main file is in a package.

~~This PR does not handle `//SOURCES` with absolute path to local files. And relative paths only work if it's the same level or a sublevel of the main file.~~
The code below already works in JBang!! That is really cool!
```java
//SOURCES ../../../b.java
//SOURCES /home/leandro/dev/java/jbang/c.java

public class oi {
    
    public static void main(String args[]) {
        System.out.println("oi");
		b.say();
		c.say();
    }
}
```
```shell
leandro:distributions$ pwd
/home/leandro/dev/githubprojects/tmp-jbang/jbang/build/distributions
leandro:distributions$ rm -fR ~/.jbang/cache/jars/oi.java.*
leandro:distributions$ ./jbang/bin/jbang oi.java 
[jbang] Building jar...
oi
I am b
I am c
```

### DIFF
--- a/src/main/java/dev/jbang/Script.java
+++ b/src/main/java/dev/jbang/Script.java
@@ -74,6 +74,7 @@ public class Script {
 	private List<KeyValue> agentOptions;
 	private String preMainClass;
 	private String agentMainClass;
+
 	/**
 	 * if this script is used as an agent, agentOption is the option needed to pass
 	 * in
--- a/src/main/java/dev/jbang/ScriptResource.java
+++ b/src/main/java/dev/jbang/ScriptResource.java
@@ -1,10 +1,12 @@
 package dev.jbang;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.util.Scanner;
 
 public class ScriptResource {
 	// original requested resource
@@ -13,11 +15,19 @@ public class ScriptResource {
 	private File sourceCacheDir;
 	// physical file it is mapped to
 	private File file;
+	// This handles when main script file is in a package.
+	// eg: /home/user/scripts/org/acme/Main.java, then
+	// originalSourceFileBaseDir = /home/user/scripts
+	private final File originalSourceFileBaseDir;
 
 	public ScriptResource(String scriptURL, File urlCache, File file) {
 		this.originalResource = scriptURL;
 		this.sourceCacheDir = urlCache;
 		this.file = file;
+		if (Util.isJavaFile(scriptURL))
+			this.originalSourceFileBaseDir = setOriginalSourceFileBaseDir(scriptURL);
+		else
+			this.originalSourceFileBaseDir = null;
 	}
 
 	public static ScriptResource forFile(File file) {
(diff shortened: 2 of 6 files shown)


maxandersen:
Why do you think this fixes 464 with pattern matching ? The //sources is needed by edit and of classes not reachable by main class.

tivrfoa:
> Why do you think this fixes 464 with pattern matching ? The //sources is needed by edit and of classes not reachable by main class.

oh. I thought he just wanted to run. I updated to should *help* =)

So, on my machine (Ubuntu) all tests passed, but failed CI Ubuntu =(

maxandersen:
failure looks like github action infrastructure fail. restarted the jobs.

tivrfoa:
> failure looks like github action infrastructure fail. restarted the jobs.

This PR is a little big because I added the test that you didn't merge from PR #469 
I still don't know why CI is failing here, but works on my machine (Ubuntu)

tivrfoa:
> I still don't know why CI is failing here, but works on my machine (Ubuntu)

It didn't work because `--source-path` was introduced in Java 9.
So now I check the Java version, and if <= 8, then `-sourcepath` is used.

quintesse:
> So now I check the Java version, and if <= 8, then -sourcepath is used.

Why check? Why not always use `-sourcepath`?

tivrfoa:
> Why check? Why not always use `-sourcepath`?

Thank you Tako!
I thought the old way was removed, but from Java 9 both -sourcepath and --source-path are supported.

src/main/java/dev/jbang/Script.java
```diff
@@ -74,6 +74,7 @@
 	private List<KeyValue> agentOptions;
 	private String preMainClass;
 	private String agentMainClass;
+
 	/**
```
maxandersen, line 78: this looks like an unnecessary change?

src/main/java/dev/jbang/ScriptResource.java
```diff
@@ -13,11 +15,19 @@
...
+	// eg: /home/user/scripts/org/acme/Main.java, then
+	// originalSourceFileBaseDir = /home/user/scripts
+	private final File originalSourceFileBaseDir;
 
 	public ScriptResource(String scriptURL, File urlCache, File file) {
 		this.originalResource = scriptURL;
 		this.sourceCacheDir = urlCache;
 		this.file = file;
+		if (Util.isJavaFile(scriptURL))
+			this.originalSourceFileBaseDir = setOriginalSourceFileBaseDir(scriptURL);
```
maxandersen, line 28: this reads wrong - why call setOriginalSourceFileBaseDir(scriptUrl) to get the return value ...why is this not just calculated (and cached) in `getOriginalSourceFileBaseDir()` ?

src/main/java/dev/jbang/cli/BaseScriptCommand.java
```diff
@@ -155,49 +155,50 @@ public static Script prepareScript(String scriptResource, List<String> arguments
...
-
-			s.setResolvedSources(resolvedSourcePaths);
+			s.setResolvedSources(resolveSOURCESRecursively(s));
 		} catch (FileNotFoundException e) {
 			throw new ExitException(1, e);
 		}
 		return s;
 	}
 
+	public static List<Path> resolveSOURCESRecursively(Script script) throws IOException {
```
maxandersen, line 165: this does not seem related to this fix ?

src/main/java/dev/jbang/ScriptResource.java
```diff
@@ -43,19 +53,49 @@ public Path fetchIfNeeded(String resource) {
...
+				if (pkgLine.length == 1)
+					continue;
+				String packageName = pkgLine[1];
+				packageName = packageName.split(";")[0].trim(); // remove ';'
+				String pkg = packageName.replace(".", File.separator); // a.b.c -> linux a/b/c, windows a\b\c
+				String pkgFile = pkg + File.separator + Util.getFileName(scriptURL);
+				return new File(scriptURL.substring(0, scriptURL.lastIndexOf(pkgFile)));
+			}
+		} catch (FileNotFoundException e) {
+			throw new RuntimeException(e);
```
maxandersen, line 92: I would argue this methods must never fail. It can warn about some mismatched dir patterns but it should not fail hard.

src/main/java/dev/jbang/cli/BaseScriptCommand.java
tivrfoa: This one you didn't merge from previous PR. You also didn't merge the test.
I extracted to a function to make it easier to test. But it end up making the code more readable in my opinion.

maxandersen:
> This PR is a little big because I added the test that you didn't merge from PR #469

if I missed sometihng there lets put that in separate PR if possible. lets not make a habit of mixing two concerns in one PR :)

tivrfoa:
I'm closing this PR, as it seems everything just works if JBang is executed from the correct folder.

PR #482, opened 2020-11-06
feat: jbang edit: make source with 'package' work

tivrfoa:
This PR creates the directories for `jbang edit` if the Java file is in a package.

Example:
```java
package a.b.c;

class A {}
```
then the classes will be placed in the correct folder when the IDE is opened:
```
src/
    a/
       b/
          c/
            A.java
```

Today all classes are placed under `src`.

### DIFF
--- a/src/main/java/dev/jbang/Util.java
+++ b/src/main/java/dev/jbang/Util.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Scanner;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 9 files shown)


tivrfoa:
I still need to add a test for this.

src/main/java/dev/jbang/Util.java
```diff
@@ -820,4 +821,27 @@ public static boolean isURL(String str) {
...
 				|| str.startsWith("file:///");
 	}
+
+	/**
+	 *
+	 * @param content
+	 * @return the package as declared in the source file, eg: a.b.c
+	 */
+	public static Optional<String> getSourcePackage(String content) {
+		try (Scanner sc = new Scanner(content)) {
```
quintesse, line 831: I was wondering if we couldn't make this part of the code that already scans for `//SOURCES` and `//DEPS`. It's a place where the entire file is already scanned, doing it twice seems very inefficient.

maxandersen: Yes, we really should move to have a list of "scanners" that want to participate per line and then feed that to them.

Adding it in this case only affect jbang edit thus not a big deal and could be argued package/dir structure is only needed for edit thus its acutally fine to only pay for it there.

tivrfoa: Even a Scanner over a String is inefficient?

What are you suggesting is for each line to check if it's //SOURCES, //DEPS and package?
Then the return type would need to change to return all that, or maybe the methods would the receive a line instead of `List<line>`.
I'll think it'll be good for performance as you said.
Can this refactoring be done a next PR?
I wanted to just get this `jbang edit` working for package to not make PR complex. Today //SOURCES and //DEPS are already scanned in different places.

At least the String is in memory, so better than reading a File. :]

quintesse: it's not so much performance, I'm sure it'll be almost unnoticeable to humans :-) And perhaps inefficient is the wrong word, I just think that if you're analyzing a file for special lines to create a Script object then it's strange that we need to do yet another scan over the same file for some other special line. To me that seems that they belongs together, one scan that delivers all the required information.( Although I can imagine that it might cause problems with the current design. Perhaps Script is too "high level" for your needs.)

PR #488, opened 2020-11-07
fix: original source must be what's in ScriptResource.originalSource

tivrfoa:
This PR fixes the error below:

```java
./jbang/bin/jbang --verbose https://gist.github.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6#file-one-java
[jbang] Gist url api: https://api.github.com/gists/8e6ea001f168fd4ef14763ceca3e5ab6
[jbang] Downloaded file https://gist.githubusercontent.com/tivrfoa/8e6ea001f168fd4ef14763ceca3e5ab6/raw/f378d987f0f863d0b3bcecce1c91fa4290ec4816/one.java
[jbang] Gist url api: https://gist.github.com/tivrfoa/two.java
[jbang] Error when extracting file from gist url.
[jbang] [ERROR] java.io.FileNotFoundException: https://gist.github.com/tivrfoa/two.java
```

ScriptResource.originalSource has the value after swizzleURL, and that's important to find the other sources.

### DIFF
--- a/src/test/java/dev/jbang/TestScript.java
+++ b/src/test/java/dev/jbang/TestScript.java
@@ -15,6 +15,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


src/test/java/dev/jbang/TestScript.java
```diff
@@ -209,6 +209,36 @@ public static void writeContentToFile(Path path, String content) throws IOExcept
 		}
 	}
 
+	private void addUrlToTrustedSources(String url) {
```
maxandersen, line 212: this just looks wrong - trustedsources shouldn't be public and mutated from outside.

tivrfoa: yes. It feels wrong, but I'm changing just the test
the variable was already public =D
At least it allows a very good test

maxandersen: there is both an add and remove method with storage as option so just need  to try/finally that - no need to expose internal logic. I've pushed change that avoids this.

tivrfoa: Great!

PR #489, opened 2020-11-07
fix: it must allow same //SOURCES in different paths

tivrfoa:
This PR fixes the logic that checks if a source was already visited.
Today it considers exactly how it's declared in the //SOURCES

eg: `//SOURCES model/C`

So in the scenario below:
```
.
├── A.java
├── model
│   └── C.java
└── person
    ├── B.java
    └── model
        └── C.java
```

This error happens:
```java
/home/leandro/dev/java/jbang/scripts/test-multiple-child-folders-package/person/B.java:5: error: package person.model does not exist
import person.model.C;
```

So the solution is to check if the full resolved path was already visited.

And why is the visited logic needed? What can go wrong if it's removed?

Two reasons:
  1. It avoids doing redundant work;
  2. It avoids a possible infinite loop if there's a mutual dependency

eg:
Main.java
```java
//SOURCES A.java

public class Main {
    
    public static void main(String args[]) {
        new A();
    }
}
```

A.java
```java
//SOURCES B.java

public class A {
    public A() { new B(); }
}
```

B.java
```java
//SOURCES A.java

public class B {
	public B() {
		System.out.println("B constructor.");
	}
}
```

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseScriptCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseScriptCommand.java
@@ -180,10 +180,10 @@ public abstract class BaseScriptCommand extends BaseCommand {
 			String[] tmp = sources.remove(0);
 			String originalSource = tmp[0];
 			String destinationSource = tmp[1];
-			if (!visited.add(destinationSource))
-				continue;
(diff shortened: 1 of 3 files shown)


maxandersen:
makes sense.

PR #491, opened 2020-11-08
refactor: Avoid loading files to look for //DEPS

tivrfoa:
Today it loads files once to look for //SOURCES and package,
and then later in another place it loads all files to search for //DEPS.
This change makes it read only once, at the same place where it
searches for //SOURCES and package.

I'm also caching the call to `collectDependencies`

### DIFF
--- a/src/main/java/dev/jbang/Script.java
+++ b/src/main/java/dev/jbang/Script.java
@@ -6,7 +6,6 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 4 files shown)


maxandersen:
im not fully grokking what you are tryin to do in this - //SOURCES and //DEPS is just a few of many things we extract...shouldn't that be the goal to clean up or are you seeing some specific advantage to jsut do it for //SOURCES and //DEPS ?

tivrfoa:
> im not fully grokking what you are tryin to do in this - //SOURCES and //DEPS is just a few of many things we extract...shouldn't that be the goal to clean up or are you seeing some specific advantage to jsut do it for //SOURCES and //DEPS ?

The title of this PR is probably not good.
With this PR we don't need this code:
```java
				if (getResolvedSources() != null) {
					for (Source source : getResolvedSources()) {
						try {
							dependencies.addAll(collectDependencies(Files.readAllLines(source.getResolvedPath())));
						} catch (IOException e) {
							throw new RuntimeException(e);
						}
					}
				}
```

Because now '//DEPS' is searched in `resolvesourceRecursively` where the source's content is already in memory. So we just loop sources once.

tivrfoa:
I'm closing because it needs more thought.

PR #494, opened 2020-11-09
fix: don't look for sources if not needed

tivrfoa:
If you run the command below twice, it shouldn't download the urls the second time.
And they are not used, because jbang uses the existing jar:
`jbang --verbose https://gist.github.com/tivrfoa/475c4add1b3df035d13820cb756dacfc`

### DIFF
--- a/src/main/java/dev/jbang/Script.java
+++ b/src/main/java/dev/jbang/Script.java
@@ -7,15 +7,18 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


src/main/java/dev/jbang/Script.java
```diff
@@ -150,6 +153,54 @@ public ScriptResource getScriptResource() {
 		return dependencies;
 	}
 
+	public List<Source> collectSourcesRecursively() throws IOException {
```
maxandersen, line 156: why the need to move it to script in this PR?

tivrfoa:
When an existing `jar` is run, jbang still needs to search for `//DEPS` because it doesn't remember.
So this PR breaks cases where `//DEPS` is/are declared in classes different from the main class.

PR #495, opened 2020-11-11
feat: force update. Add -f --fresh flag

tivrfoa:
Today jbang only will create a new jar if the main class was changed.

This flag enables force the update.

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -83,6 +83,11 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
 	@CommandLine.Option(names = { "--cp", "--class-path" }, description = "Add class path entries.")
 	List<String> classpaths;
 
+	@CommandLine.Option(names = {
+			"-f",
+			"--fresh" }, description = "Make it a fresh run - i.e. a new build with fresh (i.e. non-cached) resources.", defaultValue = "false")
+	boolean fresh;
+
 	PrintStream out = new PrintStream(new FileOutputStream(FileDescriptor.out));
 
 	// build with javac and then jar... todo: split up in more testable chunks
@@ -90,6 +95,8 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
 		File baseDir = Settings.getCacheDir(Settings.CacheClass.jars).toFile();
 		File tmpJarDir = new File(baseDir, script.getBackingFile().getName() +
 				"." + Util.getStableID(script.getBackingFile()));
+		if (tmpJarDir.exists() && fresh)
+			Util.deleteFolder(tmpJarDir.toPath(), true);
 		tmpJarDir.mkdirs();
 
 		File outjar = new File(tmpJarDir.getParentFile(), tmpJarDir.getName() + ".jar");
@@ -117,7 +124,7 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
 		// always build the jar for native mode
 		// it allows integrations the options to produce the native image
 		if (!outjar.exists() || JavaUtil.javaVersion(requestedJavaVersion) < script.getBuildJdk()
-				|| nativeBuildRequired) {
+				|| nativeBuildRequired || fresh) {
 			List<String> optionList = new ArrayList<String>();
 			optionList.add(resolveInJavaHome("javac", requestedJavaVersion));
 			optionList.addAll(script.collectCompileOptions());
--- a/src/main/java/dev/jbang/ScriptResource.java
+++ b/src/main/java/dev/jbang/ScriptResource.java
@@ -37,10 +37,10 @@ public class ScriptResource {
 	}
 
 	public Path fetchIfNeeded(String resource) {
-		return fetchIfNeeded(resource, originalResource);
+		return fetchIfNeeded(resource, originalResource, false);
 	}
 
-	public Path fetchIfNeeded(String resource, String originalResource) {
+	public Path fetchIfNeeded(String resource, String originalResource, boolean fresh) {
 		if (Util.isURL(resource) || Util.isURL(originalResource)) {
 			try {
 				URI thingToFetch = null;
@@ -50,7 +50,7 @@ public class ScriptResource {
 					URI includeContext = new URI(originalResource);
 					thingToFetch = includeContext.resolve(resource);
 				}
-				return Util.downloadAndCacheFile(thingToFetch.toString(), true);
+				return Util.downloadAndCacheFile(thingToFetch.toString(), fresh);
 			} catch (URISyntaxException | IOException e) {
 				throw new IllegalStateException("Could not download " + resource + " relatively to " + originalResource,
 						e);
--- a/src/main/java/dev/jbang/cli/BaseScriptCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseScriptCommand.java
@@ -108,7 +108,12 @@ public abstract class BaseScriptCommand extends BaseCommand {
 	}
 
 	public static Script prepareScript(String scriptResource, List<String> arguments, Map<String, String> properties,
-			List<String> dependencies, List<String> classpaths)
+			List<String> dependencies, List<String> classpaths) throws IOException {
+		return prepareScript(scriptResource, arguments, properties, dependencies, classpaths, false);
+	}
+
+	public static Script prepareScript(String scriptResource, List<String> arguments, Map<String, String> properties,
+			List<String> dependencies, List<String> classpaths, boolean fresh)
 			throws IOException {
 		ScriptResource scriptFile = getScriptFile(scriptResource);
 
@@ -156,14 +161,14 @@ public abstract class BaseScriptCommand extends BaseCommand {
 			s.setOriginal(scriptResource);
 			s.setAdditionalDependencies(dependencies);
 			s.setAdditionalClasspaths(classpaths);
-			s.setResolvedSources(resolvesourceRecursively(s));
+			s.setResolvedSources(resolvesourceRecursively(s, fresh));
 		} catch (FileNotFoundException e) {
 			throw new ExitException(1, e);
 		}
 		return s;
 	}
 
-	public static List<Source> resolvesourceRecursively(Script script) throws IOException {
+	public static List<Source> resolvesourceRecursively(Script script, boolean fresh) throws IOException {
 		List<Source> resolvedSourcePaths = new ArrayList<>();
 		// It's important to know which sources we already visited,
 		// because many files can declare the same source.
@@ -181,7 +186,7 @@ public abstract class BaseScriptCommand extends BaseCommand {
 			String originalSource = tmp[0];
 			String destinationSource = tmp[1];
 			destinationSource = Util.swizzleURL(destinationSource); // base path for new sources
-			Path path = script.getScriptResource().fetchIfNeeded(destinationSource, originalSource);
+			Path path = script.getScriptResource().fetchIfNeeded(destinationSource, originalSource, fresh);
 			if (!visited.add(path.toString()))
 				continue;
 			String sourceContent = new String(Files.readAllBytes(path), Charset.defaultCharset());
--- a/src/main/java/dev/jbang/cli/Run.java
+++ b/src/main/java/dev/jbang/cli/Run.java
@@ -50,7 +50,7 @@ public class Run extends BaseBuildCommand {
 			enableInsecure();
 		}
 
-		script = prepareArtifacts(prepareScript(scriptOrFile, userParams, properties, dependencies, classpaths));
+		script = prepareArtifacts(prepareScript(scriptOrFile, userParams, properties, dependencies, classpaths, fresh));
 
 		String cmdline = generateCommandLine(script);
 		debug("run: " + cmdline);
--- a/src/test/java/dev/jbang/TestScript.java
+++ b/src/test/java/dev/jbang/TestScript.java
@@ -172,7 +172,7 @@ public class TestScript {
 		String scriptURL = mainPath.toString();
 		Script script = prepareScript(scriptURL);
 
-		List<Source> resolvesourceRecursively = BaseScriptCommand.resolvesourceRecursively(script);
+		List<Source> resolvesourceRecursively = BaseScriptCommand.resolvesourceRecursively(script, false);
 		assertTrue(resolvesourceRecursively.size() == 7);
 	}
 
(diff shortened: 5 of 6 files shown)


quintesse (changes requested):
Given the fact that you put the option in BaseBuildCommand I don't think the work should be done in Run, because Build now has the option as well. Instead I think it should be part of the `build()` method.

Second, although deleting the JAR file does work, up to a point, I think it's not the best solution. Instead I'd take a look at the if-statement (in build()):

```
if (outjar.exists()) {
```

And add the test `&& !forceUpdate` to it.

And finally I think just creating a new JAR file isn't enough. I also think that the call to `BaseScriptCommand::prepareScript()` should turn off caching too. Meaning that the call to `getScriptFile()` probably needs an extra argument `forceUpdate` too.

tivrfoa:
> Given the fact that you put the option in BaseBuildCommand I don't think the work should be done in Run, because Build now has the option as well. Instead I think it should be part of the `build()` method.
> 
> Second, although deleting the JAR file does work, up to a point, I think it's not the best solution. Instead I'd take a look at the if-statement (in build()):
> 
> ```
> 		if (!outjar.exists() || JavaUtil.javaVersion(requestedJavaVersion) < script.getBuildJdk()
> 				|| nativeBuildRequired) {
> ```
> 
> And add the test `|| forceUpdate` to it.
> 

Much better this way! Thank you. =)

> And finally I think just creating a new JAR file isn't enough. I also think that the call to `BaseScriptCommand::prepareScript()` should turn off caching too. Meaning that the call to `getScriptFile()` probably needs an extra argument `forceUpdate` .
> 

Here do you mean that the value in `forceUpdate` should reach `Util::downloadAndCacheFile` second parameter (`boolean updateCache`)?
Today `ScriptResource::fetchIfNeeded` always passes `true`, so it never uses the cache when a file or url is run with jbang:
```java
return Util.downloadAndCacheFile(thingToFetch.toString(), true);
```

quintesse:
> Today ScriptResource::fetchIfNeeded always passes true

Indeed. That's something that has to change. The `updateCache` is something that will have to be passed from the toplevel (using your `forceUpdate`) all the way down.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -83,8 +83,17 @@ void setJavaVersion(String javaVersion) {
 	@CommandLine.Option(names = { "--cp", "--class-path" }, description = "Add class path entries.")
 	List<String> classpaths;
 
+	@CommandLine.Option(names = {
+			"-f", "--fresh" }, description = "Force script upate. A new jar is created.", defaultValue = "false")
```
maxandersen, line 87:
```suggestion
			"-f", "--fresh" }, description = "Make it a fresh run - i.e. a new build with fresh (i.e. non-cached) resources.", defaultValue = "false")
```

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -83,8 +83,17 @@ void setJavaVersion(String javaVersion) {
...
+	boolean forceUpdate;
```
maxandersen, line 88:
```suggestion
	boolean fresh;
```

lets not mix terms if we can avoid it.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -83,8 +83,17 @@ void setJavaVersion(String javaVersion) {
...
+
 	PrintStream out = new PrintStream(new FileOutputStream(FileDescriptor.out));
 
+	protected void deleteJarFile(String scriptOrFile) throws IOException {
+		if (Util.getJarFile(scriptOrFile).delete())
```
maxandersen, line 93: is this even necessary ? don't it just overwrite the existing one ? 

and if it needs deleting the jar, how about the .jar directory that has the actual built content ?

tivrfoa: It was needed, but applying @quintesse suggestion all this jar logic is not necessarily anymore. =)

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -117,7 +122,7 @@ void build(Script script) throws IOException {
 		// always build the jar for native mode
 		// it allows integrations the options to produce the native image
 		if (!outjar.exists() || JavaUtil.javaVersion(requestedJavaVersion) < script.getBuildJdk()
-				|| nativeBuildRequired) {
```
maxandersen, line 120: if you do:

`jbang build xyz.java` or `jbang xyz.java` and then do `jbang --fresh xyz.java` you will end up with the same cached jar folder; thus to be fresh that cached jar folder needs to be deleted before build happens.

tivrfoa: > if you do:
> 
> `jbang build xyz.java` or `jbang xyz.java` and then do `jbang --fresh xyz.java` you will end up with the same cached jar folder; thus to be fresh that cached jar folder needs to be deleted before build happens.

Did you test that? I tried and it worked.
It ends up in the same folder, but the files are replaced.
Here's the structure before and after `-f`
```
jars$ tree
.
├── xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4
│   ├── a.class
│   ├── META-INF
│   │   └── maven
│   │       └── g
│   │           └── a
│   │               └── v
│   │                   └── pom.xml
│   └── xyz.class
└── xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4.jar

6 directories, 4 files
leandro:jars$ tree
.
├── xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4
│   ├── a.class
│   ├── META-INF
│   │   └── maven
│   │       └── g
│   │           └── a
│   │               └── v
│   │                   └── pom.xml
│   └── xyz.class
└── xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4.jar

6 directories, 4 files
```

Output:
```
$ ./jbang/bin/jbang --verbose xyz.java
[jbang] Building jar...
[jbang] compile: /usr/lib/jvm/java-11-openjdk-amd64/bin/javac -d /home/leandro/.jbang/cache/jars/xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4 xyz.java /home/leandro/dev/githubprojects/tmp/jbang/build/distributions/a.java
[jbang] run: /usr/lib/jvm/java-11-openjdk-amd64/bin/java -classpath /home/leandro/.jbang/cache/jars/xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4.jar: xyz
xyz!
a!
leandro:distributions$ vim a.java 
leandro:distributions$ ./jbang/bin/jbang -f --verbose xyz.java
[jbang] Building jar...
[jbang] compile: /usr/lib/jvm/java-11-openjdk-amd64/bin/javac -d /home/leandro/.jbang/cache/jars/xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4 xyz.java /home/leandro/dev/githubprojects/tmp/jbang/build/distributions/a.java
[jbang] run: /usr/lib/jvm/java-11-openjdk-amd64/bin/java -classpath /home/leandro/.jbang/cache/jars/xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4.jar: xyz
xyz!
a!!!!
```

tivrfoa: > if you do:
> 
> `jbang build xyz.java` or `jbang xyz.java` and then do `jbang --fresh xyz.java` you will end up with the same cached jar folder; thus to be fresh that cached jar folder needs to be deleted before build happens.

opss. I see.
It the main file change it goes to a different folder.
It works, so delete is more to keep the cache clean.
Maybe it can be done in a different thread?

tivrfoa: > if you do:
> 
> `jbang build xyz.java` or `jbang xyz.java` and then do `jbang --fresh xyz.java` you will end up with the same cached jar folder; thus to be fresh that cached jar folder needs to be deleted before build happens.

So, if the main file changes it goes to a *different* folder, eg:
```
$ tree
.
├── xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4
│   ├── a.class
│   ├── META-INF
│   │   └── maven
│   │       └── g
│   │           └── a
│   │               └── v
│   │                   └── pom.xml
│   └── xyz.class
├── xyz.java.0a7a631f89dd5ab9e831a8c7ccb9fa4ab103d2462cf92a5c2ea4ef39bc4fa7d4.jar
├── xyz.java.fd496fd4a0345243df6a0f34bf2ba7d68aba36af1e767185bb58b291bc46f3db
│   ├── a.class
│   ├── META-INF
│   │   └── maven
│   │       └── g
│   │           └── a
│   │               └── v
│   │                   └── pom.xml
│   └── xyz.class
└── xyz.java.fd496fd4a0345243df6a0f34bf2ba7d68aba36af1e767185bb58b291bc46f3db.jar

12 directories, 8 files
```
We could delete the old jar file/folders, but I don't think we should do that.
As the hash changed, we would need to search all files/folders in the `jars` folder for the prefix (eg xyz.java*) and delete them.

quintesse: Would it perhaps be an option that the `build()` method first makes sure that the folder that's going to be used to do the compilation is empty? So instead of having different threads or clean-up phases we could just wipe the folder clean first and that way be sure that we always get a clean compilation. Or did I misunderstand and that wasn't the actual problem @maxandersen ?

tivrfoa: > Would it perhaps be an option that the `build()` method first makes sure that the folder that's going to be used to do the compilation is empty? So instead of having different threads or clean-up phases we could just wipe the folder clean first and that way be sure that we always get a clean compilation. Or did I misunderstand and that wasn't the actual problem @maxandersen ?

There are 2 problems I see if no clean up is done:
  1. Lots of unused/old jars in the cache folder: if the main class changes, then it's a different hash, hence a different jar/folder.
  2. Unused classes in the new jar: if the main class does **not** change, then it's the same jar/folder, and classes that *used to be* referenced in //SOURCES would remain. eg: a -> b -> c, then b does not use c anymore, c would remain.

We can solve these problems deleting both the jar/folder.

tivrfoa: > 1. Lots of unused/old jars in the cache folder: if the main class changes, then it's a different hash, hence a different jar/folder.

Although I think it's not possible to solve this one (when hash changes) the way Jbang works today.
For example this jar:
`Main.java.de29d2e6984d99c2750d0224c9b76f57e043c3012fc9a039d20e20d1e3c5a904.jar`

We can't simply delete by prefix (Main.java*), because the user can have lots of scripts called Main.java in different folders.

quintesse: Personally I don't care _that_ much about cleaning up the cache for things we no longer use. That's what `jbang cache clear` is for IMO. I just want to make sure that the _folder_ where we will do our compilation is clean. But that we could "fix" by simply running `Util.deleteFolder(tmpJarDir, true)` at the proper place (somewhere before `// add additional files`) in the `build()` method.

src/main/java/dev/jbang/Util.java
```diff
@@ -844,4 +844,12 @@ public static boolean isURL(String str) {
 		return Optional.empty();
 	}
 
+	public static void deleteDirectory(File directory) {
```
quintesse, line 847: This is not necessary, there's already a method for this: `Util.deleteFolder()`

quintesse (changes requested):

src/main/java/dev/jbang/Util.java
tivrfoa: Thank you! I changed.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
maxandersen: my issue was not about many folders being created - that is separate issue and something to handle separately (we basically need a garbage collector :) 

Here i'm saying that currently the cache key is calculated by the main file thus if the main file is the same but some of the included resources have changed/moved then you can end up including left over files.

Thus we should have a check and a possible clean that ensure there are 0 files in the target classes dir before building starts happening.

maxandersen (approved):
this looks good - enables --fresh for run and build.

we might consider later on to make `--fresh` a global flag similar to `--verbose` so it can be applied to any of our operations, i.e. `jbang --fresh catalog list`

quintesse:
> we basically need a garbage collector

One way to "ease the pain" is by leaving less garbage behind. For example there's no real reason to leave the folder with the source files around after compilation has finished and the JAR file has been created. There's no way we can re-use them anyway.

tivrfoa:
> > we basically need a garbage collector
> 
> One way to "ease the pain" is by leaving less garbage behind. For example there's no real reason to leave the folder with the source files around after compilation has finished and the JAR file has been created. There's no way we can re-use them anyway.

Great idea! =)
Probably this change would be enough:
```diff
diff --git a/src/main/java/dev/jbang/cli/BaseBuildCommand.java b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
index 6d87300..17e0256 100644
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -266,6 +266,7 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
                        }
                        script.setPersistentJvmArgs(integrationResult.javaArgs);
                        script.createJarFile(tmpJarDir, outjar);
+                       Util.deleteFolder(tmpJarDir.toPath(), true);
                }
 
                if (nativeBuildRequired) {
```

tivrfoa:
> > we basically need a garbage collector
> 
> One way to "ease the pain" is by leaving less garbage behind. For example there's no real reason to leave the folder with the source files around after compilation has finished and the JAR file has been created. There's no way we can re-use them anyway.

And if we do this, there would be no need to delete in the beginning of the method, because they would exist. =)

quintesse:
> there would be no need to delete in the beginning of the method

Indeed, although my motto in these cases is "better safe than sorry". They _should_ not exist, but if for some reason a previous run actually crashed the JVM they could still be there. Very rare occurrence, I know, but Murphy tells us it will happen and you'll get all kinds of hard to track down errors. I assume trying to delete a non-existing directory won't take up any noticeable time.

PR #499, opened 2020-11-13
refactor: `build()` now cleans up after itself

quintesse:
Just a minor refactor so it's easier to have `build()` clean up after itself.

It now deletes the directory with sources that was used to compile and create the JAR file.

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -95,9 +95,6 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
 		File baseDir = Settings.getCacheDir(Settings.CacheClass.jars).toFile();
 		File tmpJarDir = new File(baseDir, script.getBackingFile().getName() +
 				"." + Util.getStableID(script.getBackingFile()));
-		if (tmpJarDir.exists() && fresh)
-			Util.deleteFolder(tmpJarDir.toPath(), true);
-		tmpJarDir.mkdirs();
 
 		File outjar = new File(tmpJarDir.getParentFile(), tmpJarDir.getName() + ".jar");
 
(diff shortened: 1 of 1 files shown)


src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -125,187 +122,209 @@ void build(Script script) throws IOException {
...
-			// add source files to compile
-			optionList.addAll(Arrays.asList(script.getBackingFile().getPath()));
-			if (script.getResolvedSources() != null) {
-				optionList.addAll(
-						script	.getResolvedSources()
-								.stream()
-								.map(x -> x.getResolvedPath().toFile().getAbsolutePath())
-								.collect(Collectors.toList()));
+			// set up temporary folder for compilation
+			Util.deleteFolder(tmpJarDir.toPath(), true);
```
tivrfoa, line 126: I think if it will be deleted afterwards (in `finally`), then this is probably not necessary anymore.

quintesse: Indeed, forgot to delete it after introducing that try-finally. Good catch.

maxandersen: lets not assume it was successfully deleted...lots of things can go wrong - if it exists it should go. no matter what.

maxandersen:
...is there a reason why we would want it to nuke this folder always after build ? 
there are usecases where having the classes avaialble are useful (for instrumentation etc.)

tivrfoa:
> ...is there a reason why we would want it to nuke this folder always after build ?
> there are usecases where having the classes avaialble are useful (for instrumentation etc.)

Can't they use the `.jar` file itself for this case?
The content is the same.

quintesse:
What @tivrfoa says seems sensible to me as well.

maxandersen:
we can always revert :) merging.

quintesse:
Another option would be to have some debug-system property `-Djbang.no.cleanup=true` that skips the cleanup.

PR #505, opened 2020-11-15
docs: Explain wrapper's .jbang/ and suggest to .gitignore it

vorburger:
Just a small first contrib with a minor doc clarification about something I've run into when using jbang in a project.

PS: Hey @maxandersen I'd known about jbang since early days, but only today found a reason to actually play with it - it's so cool!

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -264,10 +264,12 @@ Unzip the https://github.com/jbangdev/jbang/releases/latest[latest binary releas
 
 === Wrapper install
 
-If you would like to have jbang available in a local directory and committed into a source code repository you akin to maven and gradle wrappers you can use the `wrapper` command.
+If you would like to have `jbang` available in a local directory and committed into a source code repository (akin to Maven and Gradle wrappers) you can use the `jbang wrapper` command.
 
 If you have `jbang` already installed you call `jbang wrapper install` in a folder to install a local `jbang`
 that will run out of that directory using `./jbang`.
+
+The `./.jbang` directory which `jbang wrapper install` creates is just a cache which you typically would not commit to a source code repository, so you can e.g. `echo .jbang/ >>.gitignore`.
 
 === "Zero" Install
 


readme.adoc
```diff
@@ -264,11 +264,13 @@ Unzip the https://github.com/jbangdev/jbang/releases/latest[latest binary releas
 
 === Wrapper install
 
-If you would like to have jbang available in a local directory and committed into a source code repository you akin to maven and gradle wrappers you can use the `wrapper` command.
+If you would like to have `jbang` available in a local directory and committed into a source code repository (akin to Maven and Gradle wrappers) you can use the `jbang wrapper` command.
 
 If you have `jbang` already installed you call `jbang wrapper install` in a folder to install a local `jbang`
 that will run out of that directory using `./jbang`.
 
+The `./.jbang` directory which `jbang wrapper install` creates is just a cache which you typically would not commit to a source code repository, so you can e.g. `echo .jbang/ >>.gitignore`.
```
maxandersen, line 272: this debatable - the `.jbang/jbang.jar` file is currently the only way we know what version of jbang you installed. we don't have a `.jbang/jbang-wrapper.properties` similar to ` .mvn/wrapper/maven-wrapper.properties` which mvn wrapper uses to know which version of jbang you use.

As is today if you don't commit the .jbang folder you will get the latest downloaded ...which is probably just fine - but just saying that the folder isn't 100% a cache as it encodes that "

so we'll wanna find a different way to phrase it.

maxandersen (changes requested):
if you can come with an update that doesn't say its "just a cache" I'm good ;)

maxandersen:
> PS: Hey @maxandersen I'd known about jbang since early days, but only today found a reason to actually play with it - it's so cool!

watch out - it can be addicting! : )

maxandersen:
actually - as implemented now it is pure cache as if you remove it and jbang is otherwise in path or ~/.jbang/bin it will use that jbang ...okey committing this for now.

quintesse:
Thing is `./.jbang` can be used for other things as well, like the `jbang-catalog.json` file as well. So we can either update the docs to specifically add `.jbang/jbang.jar` to .gitignore, or put the JAR in a subdirectory, eg. `.jbang/cache` and add that to .gitignore.

maxandersen:
wanna do a new PR to clarify ? :)

PR #507, opened 2020-11-15
docs: Clarify how to install the native-image utility binary

vorburger:
This wasn't super obvious to me at first and took me a second to figure out - perhaps it's useful to explicitly document this here.

If there is an easier way, especially e.g. for Fedora users, let's add that here as well? (What was that GraalVM distribution from Red Hat again?)


maxandersen (approved):

maxandersen:
@all-contributors please add @vorburger for doc

maxandersen:
thank you!

Issue #508, opened 2020-11-15
Native Build -o option to --output built binary to specified file instead of ~/.jbang/cache


vorburger:
@maxandersen it took me a few minutes longer than perhaps you would like it to be for users to figure out just WHERE :smiley_cat:  my first shiny native binary was.

It seems like currently (`jbang version 0.53.2`) it ends up (only) e.g. in `~/.jbang/cache/jars/first.java.f611b03e9f0d0b9d21bf300aca3bc30ccdc1ea53e189d6243d5f2619b9f84be2.jar.bin`, is that right or am I just stupid?

For building distributions of tools, and even just easier local usage in development, it would probably be nice to have if that was either copied somewhere into the working directory, or perhaps GraalVM `native-image` can easily be made to produce it right where one would want it, instead of buried in `cache`? (I haven't looked into it.)

Perhaps `jbang build --native hello.java` could make the built executable binary show up as `hello` e.g. in the current working directory? (There ideally would even be an optional `-o` to specify the desired output name (or directory...), à la `gcc hello.c -o hello`.)

Would you potentially welcome a PR doing this? Just as described above, or do you perhaps have different ideas about this?

PS: I FYI originally used `strace` to find it (and later noticed that `--verbose` prints it..), but as a very very first quick action, an initial PR just adding a line about this to https://github.com/jbangdev/jbang#build-and-run-native-image-experimental may be handy; I can do that after I get an initial reaction to this issue.

maxandersen:
For now it's quite intentional it does not end up inside the same directory as main java file as it could be inside a source directory where it does not really make sense. Thus it is intentionally going into jbangs internal cache so it can be considered just temporary. 

Intent is twofold: 1) have a '-o/--output <dir>' 2) have a 'jbang install' (which would Also allow --native) to put the binary (wether native or via a script) that put it into your path. Think like 'go installment

Makes sense ?

quintesse:
Yeah, Jbang of course isn't meant as a compiler that gets you a result that you can use afterwards. A "normal use of Jbang" (if such a thing exists) is to always do `jbang source.java`. So just like `jbang source.java` doesn't give give you easy access to the JAR file `jbang --native source.java` doesn't do that for the executable either. In both cases the output format is more of an "implementation detail".

Doesn't mean that a feature to give the user the option can't be added (as Max mentions above) but it's at least the reason why it doesn't do so by default.

quintesse:
We now have `jbang export` so this can be closed, right?

[...]

PR #509, opened 2020-11-15
docs: Explain where native binary currently ends up (re. #508)

vorburger:
see #508

This would be updated to document `-o/--output` when that's implemented.

I'll see if I can have a go at that some time.


vorburger:
This PR has 2 commits because its based on / includes my earlier #507, just to avoid having to resolve a merge conflict when that's in.

readme.adoc
```diff
@@ -391,9 +391,14 @@ There is support for using `native-image` from GraalVM project to produce a bina
 
 Since not all java libraries can automatically be built with `native-image` - especially if using reflection feature are considered highly experimental.
 
-Just run `jbang --native helloworld.java` and `jbang` will use `native-image` from either `JAVA_HOME/bin` or `GRAALVM_HOME/bin` to
+Just run `jbang --native helloworld.java` and `jbang` will use `native-image` from either `$JAVA_HOME/bin` or `$GRAALVM_HOME/bin` or `$PATH` to
 produce a native image binary.
 
+The built native image binary will be in `~/.jbang/cache/jars/YOUR.java.HASH.jar.bin` (see [#508] for an `-o/--output`).
```
maxandersen, line 397:
```suggestion
The built native image binary will be placed in jbang's cache and used on next run.
```

I don't want to make the contents of ~/.jbang/cache "api" ...at leats for now it is to be considered a black box for users.

Definitely ok adding a `-o/--out` to make this possible to get..

maxandersen (changes requested):
lets see if we can find phrasing not making cache content documented.

maxandersen:
merging with a hint rather than documenting internal details.

Issue #18, opened 2020-01-11
enable gradle run


maxandersen:
by adding application plugin plus specifiying main class you can do gradle run.

```
plugins {
    id 'java'
    id 'application'
}

repositories {
    mavenLocal()
        jcenter()
}

dependencies {
        
}


application {
    mainClassName = 'piping'
}
```

maxandersen:
not sure how useful this actually is if we have the default launchers for the ides?

PR #518, opened 2020-11-18
fix(#18): enable gradle run

alban-auzeill:
Without the support of `gradle run`, first users could be a little bit disappointed if they are used to execute:
image
Or if in the terminal of their IDE they try `./gradlew run` instead of `./gradlew build`

The fix also removes redundant tabs in `build.gradle`.

**Before**
```
apply plugin: 'java'

repositories {
	mavenLocal()
		jcenter()
}

dependencies {
			compile "org.openjfx:javafx-graphics:11.0.2"
}

sourceSets.main.java.srcDirs 'src'
sourceSets.test.java.srcDirs 'src'
```

**After**
```
plugins {
	id 'java'
	id 'application'
}

repositories {
	mavenLocal()
	jcenter()
}

dependencies {
	compile "org.openjfx:javafx-graphics:11.0.2"
}

application {
	mainClass = 'edit'
}

sourceSets.main.java.srcDirs 'src'
sourceSets.test.java.srcDirs 'src'
```

### DIFF
--- a/src/main/resources/build.qute.gradle
+++ b/src/main/resources/build.qute.gradle
@@ -1,14 +1,21 @@
-apply plugin: 'java'
+plugins {
+	id 'java'
+	id 'application'
+}
 
 repositories {
 	mavenLocal()
-		jcenter()
+	jcenter()
 }
 
 dependencies {
-		{#for item in dependencies}
-			compile "{item}"
-		{/for}
+{#for item in dependencies}
+	compile "{item}"
+{/for}
+}
+
+application {
+	mainClass = '{baseName}'
 }
 
 sourceSets.main.java.srcDirs 'src'
--- a/src/test/java/dev/jbang/cli/TestEdit.java
+++ b/src/test/java/dev/jbang/cli/TestEdit.java
@@ -102,7 +102,10 @@ public class TestEdit {
 
 		File gradle = new File(project, "build.gradle");
 		assert (gradle.exists());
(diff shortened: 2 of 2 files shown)


quintesse:
Good idea. The only question I have is: why "edit" as main class?

maxandersen:
nice!

@quintesse I assume because the example was done on `jbang edit edit.java`

alban-auzeill:
> why "edit" as main class?

It's not always `edit`, only in the `testEditDeps` unit test.
My understanding is: `{baseName}` is the main class of the application. See [Edit.java#L163](https://github.com/jbangdev/jbang/blob/c0fb96d81924c60f5b317c6c3e8aa8558134759d/src/main/java/dev/jbang/cli/Edit.java#L163)

quintesse:
Understood! I got side-tracked by the fact that we have an Edit class too :-)

maxandersen:
so...I tried this but when run happens it does not seem to take the classpath into consideration so it does not work unless I use zero non-platform dependencies.

maxandersen:
never mind - that seems to be another issue :)

maxandersen:
merged in + fix so if file has package declaration it still works.

This is actually extra nice as it avoid the IntelliJ bug https://youtrack.jetbrains.com/issue/IDEA-233074 so debugging is now working more consistently - albeit slower as its through gradle.

maxandersen:
@all-contributors please add @alban-auzeill for code

PR #535, opened 2020-11-22
Antora site version of the readme.adoc

djencks:
Related to the conversation at https://discuss.asciidoctor.org/have-asciidoctorj-output-html-to-be-include-in-jekyll-generated-site-tp8353p8361.html, I spent a few minutes setting up an Antora project version of the readme.adoc.

To build it you need to have npm installed (although running mvn install in the docs dir might install it, I didn't check).
```
cd docs
npm run clean-build-all
```
The site will be at build/site/index.html.

I took out the TOC since the right sidebar works better.

The docs project is set up with the tabs-block extension from couchbase, and the UI is set up with the clipboard.js from couchbase. However...
* I don't know where you might want to use tabs or for what, so the tabs-block extension is untested.
* the copy to clipboard functionality seems to be active but it relies on some fontawesome icons and I can't figure out how to get them to appear.

The ui is set up with https://gitlab.com/djencks/antora-ui-builder which makes it very obvious what the changes from the default ui are. If you pursue using Antora you may wish to switch to modifying a clone of the default ui; I'm happy to provide advice on how to do that.


maxandersen:
trying to get this to build to see how to use it...seems like gulp needs installed too and there is this in package.json that breaks things:

`"@djencks/antora-ui-builder": "file:///Users/david/projects/antora/ui-builder/builder/djencks-antora-ui-builder-0.0.2.tgz",`

..whats the right way to refer to it without local system specifc files ?

maxandersen:
ah seems like `https://gitlab.com/djencks/antora-ui-builder.git` makes it go further

djencks (commented):
OOPS! sorry about that.  You might try (just updated)
`    "@djencks/antora-ui-builder": "https://experimental-repo.s3-us-west-1.amazonaws.com/djencks-antora-ui-builder-v0.0.3.tgz",
`
which should be more stable.

maxandersen:
replaced with #563

PR #540, opened 2020-11-24
Jbang setup

quintesse:
It works fine on Linux.

But it's the Windows implementation I'm worried about, it's more complex because of the fact that we can't overwrite any files that are currently in use. So I had to come up with a ping-pong mechanism of handing execution over from one script to another. I haven't been able to test any of that because I don't have access to a Windows machine right now. So I will need help validating what I implemented (especially the powershell version is most likely incorrect).

### DIFF
--- a/src/main/java/dev/jbang/Util.java
+++ b/src/main/java/dev/jbang/Util.java
@@ -510,8 +510,8 @@ public class Util {
 			Path saveTmpDir = urlCache.getParent().resolve(urlCache.getFileName() + ".tmp");
 			Path saveOldDir = urlCache.getParent().resolve(urlCache.getFileName() + ".old");
 			try {
-				Util.deleteFolder(saveTmpDir, true);
-				Util.deleteFolder(saveOldDir, true);
+				Util.deletePath(saveTmpDir, true);
+				Util.deletePath(saveOldDir, true);
 
 				Path saveFilePath = downloadFile(fileURL, saveTmpDir.toFile());
 
@@ -522,12 +522,12 @@ public class Util {
 				// rename the folder to its final name
 				Files.move(saveTmpDir, urlCache);
 				// remove any old content
-				Util.deleteFolder(saveOldDir, true);
+				Util.deletePath(saveOldDir, true);
 
 				return urlCache.resolve(saveFilePath.getFileName());
 			} catch (Throwable th) {
 				// remove the temp folder if anything went wrong
-				Util.deleteFolder(saveTmpDir, true);
+				Util.deletePath(saveTmpDir, true);
 				// and move the old content back if it exists
 				if (!Files.isDirectory(urlCache) && Files.isDirectory(saveOldDir)) {
 					try {
@@ -733,33 +733,31 @@ public class Util {
 		return null;
 	}
 
-	public static boolean deleteFolder(Path folder, boolean quiet) {
-		boolean result[] = new boolean[] { true };
-		if (Files.isDirectory(folder)) {
-			Util.verboseMsg("Deleting " + folder);
-			try {
-				Files	.walk(folder)
+	public static boolean deletePath(Path path, boolean quiet) {
+		Exception err[] = new Exception[] { null };
+		try {
+			if (Files.isDirectory(path)) {
+				Util.verboseMsg("Deleting folder " + path);
+				Files	.walk(path)
 						.sorted(Comparator.reverseOrder())
 						.forEach(f -> {
 							try {
 								Files.delete(f);
 							} catch (IOException e) {
-								if (quiet) {
-									result[0] = false;
-								} else {
-									throw new ExitException(-1, "Could not delete folder " + folder.toString(), e);
-								}
+								err[0] = e;
 							}
 						});
-			} catch (IOException e) {
-				if (quiet) {
-					result[0] = false;
-				} else {
-					throw new ExitException(-1, "Could not delete folder " + folder.toString(), e);
-				}
+			} else if (Files.exists(path)) {
+				Util.verboseMsg("Deleting file " + path);
+				Files.delete(path);
 			}
+		} catch (IOException e) {
+			err[0] = e;
 		}
-		return result[0];
+		if (!quiet && err[0] != null) {
+			throw new ExitException(-1, "Could not delete " + path.toString(), err[0]);
+		}
+		return err[0] == null;
 	}
 
 	public static boolean createLink(Path src, Path target) {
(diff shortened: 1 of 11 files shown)


src/main/java/dev/jbang/Util.java
```diff
@@ -733,33 +733,31 @@ public static String runCommand(String... cmd) {
...
 	}
 
-	public static boolean deleteFolder(Path folder, boolean quiet) {
-		boolean result[] = new boolean[] { true };
-		if (Files.isDirectory(folder)) {
-			Util.verboseMsg("Deleting " + folder);
-			try {
-				Files	.walk(folder)
+	public static boolean deletePath(Path path, boolean quiet) {
+		Exception err[] = new Exception[] { null };
```
tivrfoa, line 737: Does it need to be an array?

maxandersen:
I can try on windows - hopefully later tonight. 

Btw. Did you do this all from scratch or copied others install approach ?

src/main/java/dev/jbang/Util.java
quintesse: Yes, unfortunately this is the trick one has to use when updating a var from within a lambda block.

quintesse:
> Btw. Did you do this all from scratch or copied others install approach ?

From scratch, it just came out this way in one go, but it might be good to see if there are others doing a much better job, just preferred the coding over doing research :wink: 

Btw, I just updated it with a `--with-java` option which will also add Jbang's JDK to the user's environment (PATH and JAVA_HOME).

Also, running the setup multiple times won't update already added configuration on Linux/Mac, that's still a TODO.

src/main/java/dev/jbang/Util.java
tivrfoa: Interesting hack to know. ty =D

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -0,0 +1,171 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.*;
+import java.security.CodeSource;
+
+import dev.jbang.*;
+
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "setup", description = "Make Jbang available for the user, either in the current session or permanently")
```
maxandersen, line 13:
```suggestion
@CommandLine.Command(name = "setup", description = "Make jbang available for the user, either in the current session or permanently")
```

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -0,0 +1,171 @@
...
+public class Setup extends BaseCommand {
+	private static final String jbangUrl = "https://github.com/jbangdev/jbang/releases/latest/download/jbang.zip";
+
+	@CommandLine.Option(names = {
+			"--with-java" }, description = "Add Jbang's Java to the user's environment as well")
+	boolean withJava;
+
+	@CommandLine.Option(names = {
+			"--fresh" }, description = "Force re-download and re-install of Jbang")
```
maxandersen, line 22:
```suggestion
			"--fresh" }, description = "Force re-download and re-install of jbang")
```

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -0,0 +1,171 @@
...
+				Util.deletePath(toDir, true);
+				copyJbangFiles(fromDir, toDir);
+				createUpdateScript(toDir);
+			} else {
+				copyJbangFiles(fromDir, binDir);
+			}
+		}
+		String cmd = "";
+		// Permanently add Jbang's bin folder to the user's PATH
+		Util.infoMsg("Adding Jbang to PATH...");
```
maxandersen, line 64: log how/where it does it - at least in verbose mode.

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	private static void changeScript(Path binDir, Path javaHome, Path bashFile) throws IOException {
+		// Detect if Jbang has already been set up before
+		boolean jbangFound = Files	.lines(bashFile)
+									.anyMatch(ln -> ln.trim().startsWith("#") && ln.toLowerCase().contains("jbang"));
+		if (!jbangFound) {
+			// Add lines to add Jbang to PATH
+			String lines = "\n# Add Jbang to environment\n" +
+					"alias j!=jbang\n";
```
maxandersen, line 112: That was fast :)

maxandersen:
in powershell i get this (after nuking path for jbang refs with `rundll32 sysdm.cpl,EditEnvironmentVariables`) and run build/install/jbang setup:
```
PS Z:\code\personal\jbang\build\install> .\jbang\bin\jbang setup
[jbang] Adding Jbang to PATH...
& : The term '[Environment]::SetEnvironmentVariable' is not recognized as the name of a cmdlet, function, script file,
or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and
try again.
At line:1 char:3
+ & [Environment]::SetEnvironmentVariable('Path', 'C:\Users\max\.jbang\ ...
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: ([Environment]::SetEnvironmentVariable:String) [], CommandNotFoundExcept
   ion
    + FullyQualifiedErrorId : CommandNotFoundException
```

maxandersen:
`[System.Environment]::SetEnvironmentVariable(` seem to be the right thing that works on powershell.

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -0,0 +1,190 @@
...
+			}
+			String env = "[Environment]::SetEnvironmentVariable('Path', '" + newPath + "' + " +
+					"[Environment]::GetEnvironmentVariable('Path', [EnvironmentVariableTarget]::User), " +
+					"[EnvironmentVariableTarget]::User)";
+			if (withJava) {
+				env += " ; [Environment]::SetEnvironmentVariable('JAVA_HOME', '" + jdkHome + "', " +
+						"[EnvironmentVariableTarget]::User)";
+			}
+			if (Util.isUsingPowerShell()) {
+				cmd += env + " ; ";
```
maxandersen, line 79:
```suggestion
				cmd += env + " ; ";
				cmd = "{ " + cmd + " }";
```

maxandersen: fixes the error on powershell

PR #546, opened 2020-11-25
docs: Make sure zero-install on Windows always works

quintesse:
with the `-useb` (short for `-UseBasicParsing`) we can make sure that it will always run correctly


maxandersen:
can you make fix for jbang.dev too while you are at it ? :)

PR #553, opened 2020-11-26
fix: running tests won't affect user's environment anymore

quintesse:


maxandersen:
Nice!

quintesse:
Can we merge this? :-)

src/test/java/dev/jbang/BaseTest.java
```diff
@@ -0,0 +1,23 @@
...
+import java.io.IOException;
+
+import org.junit.Rule;
+import org.junit.contrib.java.lang.system.EnvironmentVariables;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.rules.TemporaryFolder;
+
+public abstract class BaseTest {
+
+	@BeforeEach
```
tivrfoa, line 12: This PR is nice @quintesse . It also avoided many duplications.

Would be better to use BeforeAll instead `BeforeEach`?

src/test/java/dev/jbang/BaseTest.java
```diff
@@ -0,0 +1,23 @@
...
+	void initEnv() throws IOException {
+		jbangTempDir.create();
+		environmentVariables.set("JBANG_DIR", jbangTempDir.getRoot().getPath());
```
tivrfoa, line 15: Isn't necessary to use `AfterAll` to set JBANG_DIR to its previous value?

src/test/java/dev/jbang/BaseTest.java
quintesse: Then it would only do it for each test _file_, I actually want to do it for _each test_ so one test can't affect the result of another.

src/test/java/dev/jbang/BaseTest.java
quintesse: No need. It won't affect the value in the user's environment. And within the test run I also don't care because no test should ever use the original value anyway. And if a test uses a value from a previous test then that test is wrong and should be fixed  :-)

PR #554, opened 2020-11-26
feat: added `install` command. Refactored `setup`

quintesse:
Added `jbang install`. Current implementation is extremely simple. It creates scripts in `~/.jbang/bin` that run `jbang run xxx`.
There is a special case for `install jbang` that installs Jbang itself. (Useful for zero-install)
`setup` was refactored to only set up the user's environment. Installing jbang is now handled by `install`.


src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -4,35 +4,53 @@
...
+		return setup(withJava, force);
+	}
+
+	public static boolean needsSetup() {
+		String envPath = System.getenv("PATH");
+		Path binDir = Settings.getConfigBinDir();
+		return !envPath.toLowerCase().contains(binDir.toString().toLowerCase());
+	}
+
+	public static boolean guessWithJava() {
```
maxandersen, line 43: what does this do ? guessing if it needs java installed or how ?

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -42,75 +60,74 @@ public static int setup(boolean withJava, boolean fresh) throws IOException {
...
+				changed = changeScript(binDir, jdkHome, zshRcFile) || changed;
+				if (changed) {
+					Util.infoMsg("Adding jbang/bin to PATH...");
+				}
+			} else {
+				Util.infoMsg("jbang/bin already available in PATH.");
 			}
 		}
+
 		if (Util.isWindows()) {
```
maxandersen, line 112: when already installed it prints:

```
 jbang setup
[jbang] jbang/bin already available in PATH.
Please start a new Shell for changes to take effect
```

if already available in PATH a restart shouldn't be required should it ?

src/main/java/dev/jbang/cli/Setup.java
quintesse: Indeed, if the user doesn't specify the option it will try to figure out if it should be installed or not. Basically if a JDK was installed by jbang and no JAVA_HOME is set it will assume it's okay to set up JAVA_HOME ourselves.

src/main/java/dev/jbang/cli/Setup.java
quintesse: Fixed

src/main/java/dev/jbang/cli/Setup.java
quintesse: Added javadoc

maxandersen: what if java is available in the PATH from your system or some other package manager ?

this feels a  bit too magical to me...

quintesse: Jbang is all about magic :-)

If you have Java installed from other package managers you most likely have JAVA_HOME set, right? But still, we can easily add a check if java(c) is on the user's PATH and use that as an additional hint. I just want `jbang setup` to do the most useful thing for the user without them having to know about any extra options.

quintesse:
@maxandersen `install` now does something useful, it can now actually install commands.

src/main/java/dev/jbang/cli/Setup.java
maxandersen: no, neither brew, sdkman or rpm/yum installs sets a JAVA_HOME afaik. I wish they do but I don't think they do - exactly to avoid having two sources of "truth" conflicting (PATH vs JAVA_HOME).

jbangs magic of giving you a java if the one you have does not match //JAVA or no java available at all is fine as its pretty explicit/non-ambigious; but I think in this case there too many examples of JAVA_HOME not being set.

quintesse: I just tested it, but SDKMAN does set JAVA_HOME :-)

maxandersen:
fails if install with relative path to script.

```
jbang --verbose app install examples/helloworld.java
[jbang] [ERROR] /Users/max/.jbang/bin/examples/helloworld.java
java.nio.file.NoSuchFileException: /Users/max/.jbang/bin/examples/helloworld.java
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:219)
	at java.base/java.nio.file.spi.FileSystemProvider.newOutputStream(FileSystemProvider.java:478)
	at java.base/java.nio.file.Files.newOutputStream(Files.java:219)
	at java.base/java.nio.file.Files.write(Files.java:3489)
	at java.base/java.nio.file.Files.write(Files.java:3540)
	at dev.jbang.cli.AppInstall.installShellScript(App.java:97)
	at dev.jbang.cli.AppInstall.installScripts(App.java:89)
	at dev.jbang.cli.AppInstall.install(App.java:78)
	at dev.jbang.cli.AppInstall.doCall(App.java:60)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:80)
	at picocli.CommandLine.access$1200(CommandLine.java:145)
[jbang] If you believe this a bug in jbang open issue at https://github.com/jbangdev/jbang/issues
```

maxandersen:
then I try:

```
cd examples
jbang --verbose app install helloworld.java
[jbang] Command installed: helloworld.java
```

now when I run `helloworld.java` I get:
```
helloworld.java
[jbang] [ERROR] Could not read script argument helloworld.java
```

maxandersen:
helloworld.java has:

```
eval "exec jbang run helloworld.java $*"
```

so it is not putting the absolute filename in which explains why it wont run outside `jbang/examples`

maxandersen:
this works: `jbang --verbose app install qs cli@quarkusio`

maxandersen:
so i would say that for files, the script would need to put resolved absolute paths into the script as otherwise it wont work work

quintesse:
Yeah, exactly the one thing I didn't test :-)

But as you can see there are no tests yet. Just wanted to let you play with it first to see if you liked it.

quintesse:
Another thing is that we probably want to limit the names, right? So we won't generate commands like `helloworld.java`. Perhaps even make the name required as well.

quintesse:
Fixed. Only need to add tests.

quintesse:
Added tests. If this passes CI it's in a mergeable state as far as I'm concerned.

maxandersen (changes requested):
seem to work well on osx at least ;) 

`jbang app install vertx vertx` isn't nice...lets make it `jbang app --name vx vertx` and just have basic magic name that takes name before first `.` as default name. i.e. `hello.java` -> `hello`

quintesse:
Do you want any more changes to be made @maxandersen ? Or is it mergeable as-is?

src/main/java/dev/jbang/cli/Setup.java
```diff
@@ -1,170 +0,0 @@
-package dev.jbang.cli;
```
maxandersen, line 1: remind me again why `setup` shouldn't stay as top level command ? since it also setup j! alias + PATH it seems more like related but independent from `app` ?

quintesse: Not sure what `j! alias + PATH` is but basically `app` does not make sense without `setup` and vice versa. The only thing `setup` does is make `app` work, it doesn't really do anything more than that. Yes, it seems it also sets up Jbang for zero-install but you could say that zero-install is like doing `jbang app install jbang`

maxandersen: in setup you add "alias j!=jbang", thats what is deocupled from app....

so zero install is now:  iex "& { $(iwr -useb https://ps.jbang.dev) } app setup" ?

quintesse: The alias seems unrelated yes, but you can't run setup just to set up an alias. It will always do the setup for the PATH. (And `j!` is nice but it might be difficult to get working for PowerShell and will most likely never work for CMD. So we can't really use it in documentation etc)

And yes, that would the command be right now.

maxandersen: powershell: `Set-Alias j! jbang`

cmd: `echo jbang > j!.cmd`

it actually works ;)

maxandersen:
this is good now! sorry it took a bit extra time to get in - worth it!

src/main/java/dev/jbang/cli/Setup.java
quintesse: PS1: and where do we put that command? :-)
Although I've heard there is some "startup folder" for PS scripts.
But we could do the same as for:

CMD: ok, so create a copy of the original script with a different name (or make it call the original script)? That could work I guess.

PR #555, opened 2020-11-26
fix: attempt of a more explicit edit message

### LINKED ISSUE #551: confusing message for first time user
You can now run it with 'jbang test.java' or edit it using 'jbang edit --open=code test.java`'

1. You don't explain what "code" is here.
2. It's actually not the best syntax to use for Unix.
3. There's a weird unmatched backtick.

JBang should know it's running on Unix, and suggest me the best syntax for me. Also, it should clarify what "code" means.

Note that this is only a problem the *very first* time you use jbang.

maxandersen:
Fixes #551

Now `jbang init test2.java` prints: 

`[jbang] File initialized. You can now run it with 'jbang test2.java' or edit it using 'jbang edit --open=[editor] test2.java' where [editor] is your editor or IDE, e.g. 'code'`

the editor is one of the known supported ones (code, idea, eclipse, etc.)
ok or too short ?

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -45,8 +45,9 @@ public class Init extends BaseScriptCommand {
 			}
 
 			info("File initialized. You can now run it with 'jbang " + scriptOrFile
-					+ "' or edit it using 'jbang edit --open=" + knowneditors[new Random().nextInt(knowneditors.length)]
-					+ " " + scriptOrFile + "`'");
+					+ "' or edit it using 'jbang edit --open=[editor] "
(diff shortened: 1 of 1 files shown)


quintesse (approved):
LGTM. Although if we have a limited list of supported IDEs (eg. you can't just type the name of an unknown one) then perhaps we should just show the full list?

gavinking:
LGTM

maxandersen:
@quintesse there isn't a fixed list just the most popular ones I added to be some kind of fair :)

gavinking:
Yeah but still, it's helpful to the user to know what that list is.

PR #562, opened 2020-11-30
fix: Don't attempt to minimse jboss-logging

stuartwdouglas:
It results in a broken jboss-logging that can't be
used with Quarkus.


maxandersen:
...was this always broken and we just avoided it or something change in qurkus dependency chain ?

stuartwdouglas:
I think you added qute to jbang, which pulls in jboss-logging.

maxandersen:
qute was added back in 704f391414a22c0745f468e492a9b1247b8ebc2f, 31st December 2019 :) i'll merge this but not sure its the long term right soluton - probably better to isolate in external process..

PR #563, opened 2020-11-30
Antora docs, 2nd try, with external tabset-block extension and working copy-to-clipboard.

djencks:
- Hopefully does not include any overlooked file-system access!
- I've made the tabset extension a project of it's own, https://gitlab.com/djencks/asciidoctor-tabset-block, and now it works pretty well. The tabset UI bits in the jbang UI are pulled from the tabset-block-ui subproject.
- I added a tabset-example.adoc page to show how to set it up.  The `+` and `--` (open block) are required.
- By eliminating any attempt to use icons I got the copy to clipboard functionality to work.  The positioning seems odd to me, and you need to include the language in any [source] block or you get something really bizarre.  It's a start.


maxandersen:
Thanks! I think i've come to terms with how to approach this:

1) moved readme.md into a antora comptible location in this repo (i.e. antora.yml, modules/...)
   Can do the same for the other "parts" like jbang-action, jbang-examples, etc.
2) update jbang.dev to have its /documentation be an antora site grabbing content from those other sites.
3) somehow get the styling of jbang.dev to show up in the antora layout...

1 and 2 I think I can figure out - #3 i'm a total css/html noob so lets see how bad that gets :)

maxandersen:
antora would be on jbang.dev repo thus closing it here for now.

PR #565, opened 2020-11-30
Jar output via jbang export

quintesse:
This has become a bit of a mix of related changes. The 2 main features are:

1. get a copy of the script JAR by specifying ~`--output` / `-O`~ `jbang export <script-ref>`

But that JAR won't work without setting up all the necessary deps, so:

2. add dependencies to the JAR manifest so the JAR can simply be executed with `java -jar` **on the same system**. (It won't work on other computers, but it also won't hurt)

3. There was already code for setting Boot-Class-Path but it was using the wrong format, so fixed that while adding 2. 

4. Refactored the Script object to make it stateless which was a change I had lying around for weeks already and given the fact that these changes used that exact same code I decided to add it in the PR.

### DIFF
--- a/src/main/java/dev/jbang/cli/Build.java
+++ b/src/main/java/dev/jbang/cli/Build.java
@@ -13,7 +13,7 @@ public class Build extends BaseBuildCommand {
 			enableInsecure();
 		}
 
-		script = prepareScript(scriptOrFile, userParams, properties, dependencies, classpaths);
+		script = prepareScript(scriptOrFile, null, properties, dependencies, classpaths);
 
 		if (script.needsJar()) {
 			build(script);
--- a/src/main/java/dev/jbang/cli/Build.java
+++ b/src/main/java/dev/jbang/cli/Build.java
@@ -1,18 +1,11 @@
 package dev.jbang.cli;
 
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
 
-import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
 @Command(name = "build", description = "Compiles and stores script in the cache.")
 public class Build extends BaseBuildCommand {
-
-	@CommandLine.Option(names = { "-O",
-			"--output" }, description = "Directory to write application binary to.")
-	Path outputDir;
 
 	@Override
 	public Integer doCall() throws IOException {
@@ -24,17 +17,6 @@ public class Build extends BaseBuildCommand {
 
 		if (script.needsJar()) {
 			build(script);
-		}
-
-		if (outputDir != null) {
-			// TODO make sure the output names are nicer
-			Path jar = script.getJar().toPath();
-			if (nativeImage) {
-				Path img = getImageName(jar.toFile()).toPath();
-				Files.copy(img, outputDir.resolve(img.getFileName()));
-			} else {
-				Files.copy(jar, outputDir.resolve(jar.getFileName()));
-			}
 		}
 
 		return EXIT_OK;
--- a/src/main/java/dev/jbang/cli/Build.java
+++ b/src/main/java/dev/jbang/cli/Build.java
@@ -1,11 +1,18 @@
 package dev.jbang.cli;
 
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
+import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
 @Command(name = "build", description = "Compiles and stores script in the cache.")
 public class Build extends BaseBuildCommand {
+
+	@CommandLine.Option(names = { "-O",
+			"--output" }, description = "Directory to write application binary to.")
+	Path outputDir;
 
 	@Override
 	public Integer doCall() throws IOException {
@@ -17,6 +24,17 @@ public class Build extends BaseBuildCommand {
 
 		if (script.needsJar()) {
 			build(script);
+		}
+
+		if (outputDir != null) {
+			// TODO make sure the output names are nicer
+			Path jar = script.getJar().toPath();
+			if (nativeImage) {
+				Path img = getImageName(jar.toFile()).toPath();
+				Files.copy(img, outputDir.resolve(img.getFileName()));
+			} else {
+				Files.copy(jar, outputDir.resolve(jar.getFileName()));
+			}
 		}
 
 		return EXIT_OK;
--- a/src/main/java/dev/jbang/cli/Export.java
+++ b/src/main/java/dev/jbang/cli/Export.java
@@ -24,7 +24,7 @@ public class Export extends BaseBuildCommand {
 			enableInsecure();
 		}
 
-		script = prepareScript(scriptOrFile, userParams, properties, dependencies, classpaths);
+		script = prepareScript(scriptOrFile, null, properties, dependencies, classpaths);
 
 		if (script.needsJar()) {
 			build(script);
--- a/src/main/java/dev/jbang/cli/Export.java
+++ b/src/main/java/dev/jbang/cli/Export.java
@@ -1,0 +1,61 @@
+package dev.jbang.cli;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import dev.jbang.Util;
+
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+@Command(name = "export", description = "Export the result of a build.")
+public class Export extends BaseBuildCommand {
+
+	@CommandLine.Option(names = { "-O",
+			"--output" }, description = "The name or path to use for the exported file. If not specified a name will be determined from the original source ref")
+	Path outputFile;
+
+	@Override
+	public Integer doCall() throws IOException {
+		if (insecure) {
+			enableInsecure();
+		}
+
+		script = prepareScript(scriptOrFile, userParams, properties, dependencies, classpaths);
+
+		if (script.needsJar()) {
+			build(script);
+		}
+
+		// Determine the output file location and name
+		Path cwd = Util.getCwd();
+		Path outputPath;
+		if (outputFile != null) {
+			outputPath = outputFile;
+		} else {
+			String outName = AppInstall.chooseCommandName(script);
+			if (nativeImage) {
+				outName = getImageName(new File(outName)).getName();
+			} else {
+				outName += ".jar";
+			}
+			outputPath = Paths.get(outName);
+		}
+		outputPath = cwd.resolve(outputPath);
+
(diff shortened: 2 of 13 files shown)


maxandersen:
> 3\. Refactored the Script object to make it stateless which was a change I had lying around for weeks already and given the fact that these changes used that exact same code I decided to add it in the PR.

wdym stateless ?

maxandersen:
> 1. add dependencies to the JAR manifest so the JAR can simply be executed with `java -jar` **on the same system**. (It won't work on other computers, but it also won't hurt)

this is nice; i thought about doing that as alternative to have the ~/.jbang/depcache file...as in could still have it to speed up compiling the same again and again but if jar has the right manifest can just run it...actually - no...if do that then you can add/use -classpath....

maxandersen:
> 1. get a copy of the script JAR by specifying `--output` / `-O`

so what i thought of was that -o would take a dir name not file name. so we can output .jar, .bin, etc. so this would work for --native too.

quintesse:
> wdym stateless ?

The previous version of Script is stateful, it remembers if `buildJar()` was ever called on it or not, which I didn't think was the proper place to store that information given a) that Script is basically the result of parsing a file, a data object so to speak. and b) the build process is 80% in BaseBuildCommand why is there another 20% in Script? It just didn't seem right.

quintesse:
> so what i thought of was that -o would take a dir name not file name. so we can output .jar, .bin, etc. so this would work for --native too.

Done

quintesse:
Btw, I'd wait with merging this one until we have merged #554 so I can re-use the filename "guessing" I implemented there. Right now it will use the actual JAR name which has the hash in it.

src/main/java/dev/jbang/cli/Build.java
```diff
@@ -1,12 +1,19 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
 
+import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
 @Command(name = "build", description = "Compiles and stores script in the cache.")
 public class Build extends BaseBuildCommand {
 
+	@CommandLine.Option(names = { "-O",
```
maxandersen, line 13: why capital O ?

maxandersen:
> * There was already code for setting Boot-Class-Path but it was using the wrong format, so fixed that while adding 2.
> *

whats that wrong format/change you are referring to ? boot-class-path was added to support javaagent add their external deps...so it worked at *some* point ;)

src/main/java/dev/jbang/cli/Build.java
quintesse: lowercase is already taken by `--offline` :shrug:

quintesse:
> > * There was already code for setting Boot-Class-Path but it was using the wrong format, so fixed that while adding 2.
> 
> whats that wrong format/change you are referring to ? boot-class-path was added to support javaagent add their external deps...so it worked at _some_ point ;)

Most likely because it was never tested with paths that needed quoting (eg. paths with spaces) :-)

src/main/java/dev/jbang/cli/Build.java
tivrfoa: Maybe this is a good case for a breaking change? =)
As `-o` is used everywhere for output, a capital O can be an annoyance.

maxandersen:
Ah, that could be :)

tivrfoa:
> > so what i thought of was that -o would take a dir name not file name. so we can output .jar, .bin, etc. so this would work for --native too.
> 
> Done

Is `output` still a good name as it's a dir and not a file? Maybe: outDir, toFolder, destination ?

maxandersen:
> Btw, I'd wait with merging this one until we have merged #554 so I can re-use the filename "guessing" I implemented there. Right now it will use the actual JAR name which has the hash in it.

you wanna push that change before merge?

maxandersen:
> Is `output` still a good name as it's a dir and not a file? Maybe: outDir, toFolder, destination ?

i think its fine - we could of course fully jump the shark and call it `target` or `buildDir` ;)

src/main/java/dev/jbang/cli/Build.java
maxandersen: > As `-o` is used everywhere for output, a capital O can be an annoyance.

well, if we do things right offline would be used more than output ..I don't think there is really a good choice here. for maven it is `mvn -o` and gradle its `gradle --offline` ...i'm +0 and just keep it as is.

quintesse:
> so what i thought of was that -o would take a dir name not file name. so we can output .jar, .bin, etc. so this would work for --native too.

So, I've now used it a couple of times and personally I don't really like that I can't decide what to call the resulting JAR or binary file.

So remind me, why is it more useful to specify a directory name instead of the actual file name?

I mean I can type `jbang build -o out.jar source.java` and I can also type `jbang build -o out.exe --native source.java`....

maxandersen:
The issue is that output aren't guaranteed to be a single file always. Even in some cases --native or even just build won't produce any output. 

Multi-file cases:
--native produces two files a .jar and a .bin

Using quarkus you can have extra buildresults in form of kubernetes yaml and more. 

No file case:
--native xyz.jsh prints a warning no binary artifact. 

One option would be to say that when explicit builddir is specified we drop hash generation and just use the base name as basis for the artifact name ?

We could also have two option is, one for build dir and another for naming pattern. 
But I kinda feel that is unnecessary complexity.

quintesse:
My point of view is that there are never 2 files, the JAR and the native file are two different artifacts. If you _really_ _really_ want both of them you can just run the build twice, once with `--native` and once without. The second time will be fast anyway.

To me the hash in the name is not the problem (I removed that anyway), it's that you can't always predict what the final name will be. In the case of URLs / Gists for example. Which makes scripts more complicated, if the URL is not known you'd have to a) try to detect what the final name is or b) look into the target directory to see what files were created, If you can specify the name you can simply always use that, eg: `jbang build -O app.exe <some_unknown_url>` and then simply run it with `app.exe`.

The fact that there's no artifact... well that's easily detected if you know what name should be generated. But still, if we fail to generate the artifact shouldn't we fail with an error + exit code?

quintesse:
Also: even if the step from script name to binary result name is simple: eg. `jbang build foo.java` -> `foo.jar`, `jbang build foo@catalog` -> `foo`, etc. it might be relatively difficult to implement in a script. You'd have to strip `.java` or `.jsh` or `@somename` from the reference. If it's a path or an URL you might have to use different stripping strategies etc. All things that make using Jbang more complicated IMO.

tivrfoa:
I think if the user is using `build` then he wants the jar where he is running the command or in a *target* or *build* folder (I would prefer to go to a folder).
So an additional `output` should only be necessary if he wants to override the default location/name

So if running `jbang build xyz.java`, the jar would go to *target/xyz.jar*

=)

quintesse:
@tivrfoa even if that would be what `build` would do (which it isn't really, it's more so you can prepare for the next `run` to be fast) you could just as easily argue that by default the value of `--output` is "./<scriptname>.jar" thereby making it possible for a user to say "I want my jar to be `target/myname.jar`".

IMO the option of specifying the folder name _takes away_ a very useful feature and makes things unnecessarily complex in certain cases.

maxandersen:
> My point of view is that there are never 2 files

There is one artifact run that is correct, but to get the native there is at least also 1 jar and for quarkus we need to hand a "build dir" over where extra output can go for stuff like geenrated docker file and kubernetes resources; aka a target / build dir.

Thus assuming there is only *one* artifact in a build output is too narrow.

quintesse:
Not sure I understand. What does the `build` command have to do with Quarkus and docker files and K8s resources?

To me the `build` command is the step before `run`. `run` just runs a single jar/binary without needing anything else.
Personally I don't think we should start overloading `build` to do other kinds of builds as well or its direct connection to `run` will get lost.

If we want to generate other kinds of output I think we should create new commands for them.

maxandersen:
quarkus can (optionally) during its build generate additional artifacts - those need a "home"/target.

That could be separate PR for a --build-dir option.

But still even if --output was a single name/file location how do you know which is going to be output ? a jar, native or nothign ?

quintesse:
> quarkus can (optionally) during its build generate additional artifacts - those need a "home"/target.

Ok, but those artifacts won't show up even if we use a target dir instead of a target file. The build isn't re-run when you specify the `-O` option, it just copies the artifact from the cache to the target you specify. Jbang doesn't know anything about any extra Quarkus artifacts that might have been generated so it won't copy them.

> how do you know which is going to be output ? a jar, native or nothign ?

You do. If you added the option `--native` it's native, if you didn't it's a JAR. And _nothing_ shouldn't happen IMO without an exit code specifying something went wrong. But at least testing for file existence tends to be easy.

maxandersen:
> And _nothing_ shouldn't happen IMO without an exit code

`jbang --native xyz.jsh` creates nothing

maxandersen:
> Ok, but those artifacts won't show up even if we use a target dir instead of a target file. The build isn't re-run in the target folder when you specify the `-O` option, it just copies the artifact from the cache to the target you specify.

okey, then this is something else than `build`, its `jbang export --output xyz.jar xyz.java` - wdyt ?

quintesse:
export: ok, that might actually be a better idea.

quintesse:
Ok, introduced an `export` command. Let me know what you think @maxandersen .
Btw it's based on the same base as `run` and `build` which unfortunately means that it has a `userParams` argument which doesn't make any sense for `export` but I didn't yet want to introduce yet another base class. But if you prefer that I'll add it.

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -0,0 +1,61 @@
...
+		// Copy the JAR or native binary
+		Path jar = script.getJar().toPath();
+		if (nativeImage) {
+			Path img = getImageName(jar.toFile()).toPath();
+			Files.copy(img, outPath);
+		} else {
+			Files.copy(jar, outPath);
+		}
+
+		return EXIT_OK;
```
maxandersen, line 59:
```suggestion
        Util.info("Exporting to " + outPath);
		return EXIT_OK;
```

maxandersen:
> Ok, introduced an `export` command. Let me know what you think @maxandersen .

This feels much more right! kudos!

> Btw it's based on the same base as `run` and `build` which unfortunately means that it has a `userParams` argument which doesn't make any sense for `export` but I didn't yet want to introduce yet another base class. But if you prefer that I'll add it.

The same applies to 'build' I would say, `userParams` is not of much use there - we should probably fix that; but as its a known/exsting bug I'm fine we fix that in separate PR once we feel we can't live with it anymore ;)

quintesse:
Ok, added message and also fixed that userParams issue, it was easier than I thought.

PR #570, opened 2020-12-02
feat: minimal jitpack and //REPOS support in edit/live mode

rpardini:
Minimal patch; DependencyUtils.resolveDependencies() needs a bigger refactor.

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -187,9 +187,9 @@ public class Edit extends BaseScriptCommand {
 
 		// Turn any URL dependencies into regular GAV coordinates
 		collectDependencies = collectDependencies
-				.stream()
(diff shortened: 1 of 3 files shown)


rpardini:
Hmm. I meant //REPOS support, not //SOURCES.

maxandersen:
Nice - can't believe We missed that! That explains a few weird errors in IntelliJ I had but worked fine in vscode/eclipse :)

Any chance you could expand tests in TestEdit to include one for this scenario ?

rpardini:
There you go. Dunno which jitpack dep to use for test, just used one of mine.
It shouldn't actually resolve during tests (except for string-based conversion) so shouldn't break in case it goes away.

maxandersen:
Perfect! Thank you @rpardini !

@all-contributors please add @rpardini for code

PR #573, opened 2020-12-02
Fix zero-install

quintesse:
A couple of small but necessary fixes for zero-install on Windows


quintesse:
Tested this personally so it should all work now :-)

PR #577, opened 2020-12-03
fix: watchService's event is relative to watched directory

rpardini:
Fix for running `edit --live` but with an absolute path to the script.
Without this fix, "changed" is a relative path, which then turns out to not exist.

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -2,6 +2,7 @@ package dev.jbang.cli;
 
 import static dev.jbang.Settings.CP_SEPARATOR;
 import static dev.jbang.Util.isWindows;
+import static dev.jbang.Util.verboseMsg;
 import static java.lang.System.out;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -98,8 +98,9 @@ public Integer doCall() throws IOException {
 					final WatchKey wk = watchService.take();
 					for (WatchEvent<?> event : wk.pollEvents()) {
 						// we only register "ENTRY_MODIFY" so the context is always a Path.
-						final Path changed = (Path) event.context();
-						// info(changed.toString());
+						// but relative to the watched directory
+						final Path changed = watched.resolve((Path) event.context());
+						info("Changed file: " + changed.toString());
```
maxandersen, line 103: this is debug logging - should remove it or put in a verboseMsg()

maxandersen:
that one nagged me a few times and i couldn't always reproduce it - great find!

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -98,8 +99,9 @@ public Integer doCall() throws IOException {
 					final WatchKey wk = watchService.take();
 					for (WatchEvent<?> event : wk.pollEvents()) {
 						// we only register "ENTRY_MODIFY" so the context is always a Path.
-						final Path changed = (Path) event.context();
-						// info(changed.toString());
+						// but relative to the watched directory
+						final Path changed = watched.resolve((Path) event.context());
+						verboseMsg("Changed file: " + changed.toString());
```
rpardini, line 104: Done, I was led to info by the commented-out info() living there

rpardini:
@maxandersen

maxandersen:
Thank you!

@all-contributors please add @rpardini  for code

PR #592, opened 2020-12-06
fix: Make `app install --force` work

### LINKED ISSUE #589: jbang install app for app that already exist should have a way to force/overwrite
at the moment you have to figure out to uninstall first.

maybe add a --force and at least write in response that there exist alreayd app what user needs to do.

quintesse:
Fixes #589


maxandersen (changes requested):

quintesse:
Don't forget to merge this :-)

PR #600, opened 2020-12-08
Don't add tar.gz files to classpath

### LINKED ISSUE #598: Should ignore non jar jars
JBang having issues with SVM dependencies:

```
$ jbang ./example.java
[jbang] [WARN] Detected missing dependencies in cache.
[jbang]     Resolving org.graalvm.nativeimage:svm:20.3.0...Done
[jbang] Dependencies resolved
jbang --native ./example.java
[jbang] Building jar...
[jbang] log: /tmp/jbang14277912176923607405native-image
Error: Invalid or corrupt jarfile /home/g/.m2/repository/org/graalvm/nativeimage/svm-hosted-native-linux-amd64/20.3.0/svm-hosted-native-linux-amd64-20.3.0.tar.gz
Caused by: java.util.zip.ZipException: zip END header not found
[jbang] [ERROR] Error during native-image

make: *** [Makefile:7: native] Error 1
```

The dependency is `//DEPS org.graalvm.nativeimage:svm:20.3.0`. JBang should be ignoring that `tag.gz` file.

galderz:
Closes #598

I'm not sure if this is the best way. Maybe the filtering could happen earlier? The svm pom has this:

```
    <dependency>
      <groupId>org.graalvm.nativeimage</groupId>
      <artifactId>svm-hosted-native-linux-amd64</artifactId>
      <version>20.3.0</version>
      <type>tar.gz</type>
    </dependency>
```

Maybe Aether could be configured/instructed to skip `tar.gz` type dependencies?

### DIFF
--- a/src/main/java/dev/jbang/ModularClassPath.java
+++ b/src/main/java/dev/jbang/ModularClassPath.java
@@ -25,6 +25,7 @@ public class ModularClassPath {
 	public String getClassPath() {
 		if (classPath == null) {
 			classPath = artifacts	.stream()
+									.filter(it -> it.asFile().getPath().endsWith(".jar"))
 									.map(it -> it.asFile().getAbsolutePath())
 									.map(it -> it.contains(" ") ? '"' + it + '"' : it)
 									.distinct()


src/main/java/dev/jbang/ModularClassPath.java
```diff
@@ -24,6 +24,7 @@ public ModularClassPath(List<ArtifactInfo> artifacts) {
 	public String getClassPath() {
 		if (classPath == null) {
 			classPath = artifacts	.stream()
+									.filter(it -> !it.asFile().getPath().endsWith(".tar.gz"))
```
quintesse, line 27: I was wondering, wouldn't a positive filter testing for .jar be better? There might be all kinds of files as part of artifacts but only JARs can be added to the classpath, right? Or do other options exist?

galderz:
@quintesse PR updated.

maxandersen:
Not sure if you saw the comment on the issue. It is part of Java classpath spec non jars are ignored by compiler/runner. 

This looks like a bug in native image to me. 

Can you Give the usecase where this ever happens? If we fix it it should be specific for native image.

Issue #602, opened 2020-12-13
jbang should be able honor common settings.xml maven repository location


maxandersen:
not sure if it should honor settings.xml location all the time but as noted in #431 jbang currently default the local maven repository to be JBANG_REPO if set and otherwise defaults to `~/.m2/repository`

It does *NOT* currently honor any repo setting specified in `~/.m2/settings.xml` about local repository.

PR #603, opened 2020-12-13
feat: Add `mavenLocal` to allow resolving dependencies from local repository

### LINKED ISSUE #431: Allow mavenlocal repository to be configured
Sometimes you want to test out things locally and would like mavenlocal to resolve first.

maxandersen:
Fixes #431

Note, this is currently implemented by calling shrinkwrap method to uselegacylocal repo option.

This means the order of when `mavenLocal` shows up does not matter.

I wonder if we should instead add the local repo location as file url so it gets treated like any other repo 
and can be ordered ?

Interested in feedback before merging this. You can test this behavior against using file url by just manually specifying `//REPOS mylocal=file:/...path to m2`

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -551,8 +551,11 @@ For ease of use there are also a few shorthands to use popular commonly availabl
 
 |===
 |Short name | Description
-|`mavencentral`
+|`mavenCentral`
 |Maven Central
+
+|`mavenLocal`
+|Local maven repository (only use if you know you need it)
 
 |`jcenter`
 |`https://jcenter.bintray.com/`
--- a/src/main/java/dev/jbang/DependencyUtil.java
+++ b/src/main/java/dev/jbang/DependencyUtil.java
@@ -31,6 +31,8 @@ public class DependencyUtil {
 	public static final String ALIAS_JCENTER = "jcenter";
 	public static final String ALIAS_GOOGLE = "google";
 	public static final String ALIAS_MAVEN_CENTRAL = "mavenCentral";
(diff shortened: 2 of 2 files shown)


maxandersen:
@aalmiray @cplerch wdyt ?

cplerch:
No knowing if this is at all meaningful to you in any way, I want to share some related findings, nevertheless.

1) It looks like the dependency resolution machinery that works under the hood uses localMaven repo for resolution of declared public and transitive dependencies as a cache anyway. When you define some public dependency (with possible transitive dependencies) all those get downloaded to localMaven correctly, whether redirection via `~/.m2settings.xml` is in place or not. The resulting classpath also correctly points to localMaven, wherever it may be located.

2) A surprisingly simple workaround for resolving the current problem with local-only dependencies in a redirected localMaven repo is to create a symbolic link `~/.m2/repository` that points to the correct localMaven path. At least in my case everything then works as expected, without any changes in the software, and I believe that this little fix also proves my assertion 1) above.

Now, since localMaven is already been utilized correctly, except for primary lookup of dependencies, I think this is the only thing that need to be changed: 
**Before trying to lookup any dependency in public repos, first check localMaven if its already there. Only if not found, try to find it in any of the declared public repos.** 
That's all.

Is this reasonable? wdyt?

quintesse:
I think @cplerch 's issue will be solved by #602, it's @aalmiray that mentioned he was still seeing the problem even though he doesn't have a different repo cache location defined.

I'm still not sure what it is this is trying to fix, even after having re-read the issues :sweat_smile:

maxandersen:
no usecases so closing.

if we find one we can revisit.

PR #604, opened 2020-12-16
Jbang config

quintesse:
This is a first setup of an implementation for handling config files (in the same way Jbang does for catalog files).

An assumption I made was to use a JSON formatted file because we already use JSON everywhere and going full git-like would mean having to implement an entire parser.

Also, there's an example config value for setting the default value for the editor used for `jbang edit --open=`, but I just went for the most basic, unstructured, option: "editor". Before we merge any of this we should come up with a plan on how we want to structure option names.

(eg. is "editor" enough? do we put it inside an "edit" block? Or do we flatten options inside a single list, eg "edit.editor"? etc etc)


maxandersen:
I tried this:

```
❯ less ~/.jbang/config.json
{
    "editor": "codexxx"
}
```

but `jbang edit --open=. test.java` didn't seem to take effect ? 

about naming I would go after vscode style of comments + property:

```
{
	// Controls whether the editor shows CodeLens.
	"diffEditor.codeLens": false,

	// When enabled, the diff editor ignores changes in leading or trailing whitespace.
	"diffEditor.ignoreTrimWhitespace": true,

	// Timeout in milliseconds after which diff computation is cancelled. Use 0 for no timeout.
	"diffEditor.maxComputationTime": 5000,
...
```

quintesse:
Comments? I'm pretty sure JSON doesn't support comments... :thinking:

quintesse:
Seems VSCode might be using JSONC "JSON With Comments". Not sure if that's supported by GSON but I couldn't find any mention of it so bets are off.

maxandersen:
Same format as trusted sources JSON uses.

quintesse:
Not sure I understand. Trusted sources just has a single fixed header. Is that how you want to do it?
Or do you always want to write the same fixed file of settings (including comments)?
(The latter would retains the comments but feels really restrictive, you could never add a comment of your own for example)

maxandersen:
I'll rather limit config edits now and introduce later on to preserve comments than open it up and instantly loose ability to always have updated and documented settings.

maxandersen:
@quintesse got a status on this one ? would be cool to have in ?

maxandersen:
test fails because I missed a indrect reference to examples. should be fixed by a rebase.

quintesse:
Just added a `config` command with `get`, `set`, `unset` and `list` subcommands.

There's just a single important issue left to handle: what to do when the user wants to set values that are not supported (the user might at that moment be using an older version of Jbang than the one that created the config file). This also happens when the file currently already contains unknown values and we want to rewrite it (because the user used set/unset).

With set/unset we could warn the user that the option name is unknown and prevent any changes. But that leaves the big problem that even if the user tries to set/unset a known value any unknown values would disappear when writing the file.

Or we could keep a list of unknown values and just put them at the end of the file. They couldn't have any comments of course which also means that if we rewrite a file created by a newer Jbang we would lose the comments of any options that are unknown if the user somehow happened to run an older Jbang.

Not sure right now what the best solution would be.

A possible solution would be to find a JSON parser that knows how to handle comments, but then we'd probably have to change all our JSON handling. Not really a fan of that idea.

Another option might be to do some clever post-processing of the result of a JSON serialization. We could perhaps serialize to a String, assume everything is pretty-printed and go looking for lines that match known `"some.known.key": "some value"` and insert their comments in the line above. Sounds a bit hacky but at least it's simple.

quintesse:
> Sounds a bit hacky but at least it's simple.

And we would still lose any comments made by newer Jbang versions of course. (That's why it's simple :-) )

maxandersen:
json sucks thats what it is :)

quintesse:
Well, we're not at 1.0 yet, we can still switch to YAML ;-)

quintesse:
(and YAML is backward compatible to JSON so it can still read the current files :-) )

gavinking:
Do we ever expect the user to edit this thing by hand? No, right?

quintesse:
Well.... I definitely edit Git config files by hand, I never use the CLI for that. But that file also doesn't have any comments either. So perhaps it's the comments that should go... You want to edit them by hand? You gotta know what ya're doing!

gavinking:
I mean, I can't see myself ever editing it by hand. But I never edited a git confit file by hand either 🤷‍♂️

maxandersen:
> Well, we're not at 1.0 yet, we can still switch to YAML ;-)

is there any comment preserving yaml parsers in java ?

maxandersen:
i'm in the mixed boat - I tend to use cli's ability to update their config if present but I definitely also edit by hand as often that is faster when trying to mimick/reproduce an existing setup.

hence why I liked vscode config style where I just rarely ever have to go look in docs to figure out what thing to change/modify.

Might as well go xml route if not for humans - there we can at least preserve comments - semi-just-kidding ;)

src/main/java/dev/jbang/ConfigUtil.java
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Returns a Config containing all the settings from local config files merged
+	 * into one. This follows the system where settings that are "nearest" have
+	 * priority. The Config starts out with all the values from `defaults`, any
+	 * values read from files have priority.
+	 *
+	 * @param cwd The current working directory
+	 * @return a Config object
+	 */
+	public static Config getMergedConfig(Path cwd) {
```
maxandersen, line 47: the docs says files as in possibly more than 2 ? 

isn't the normal behaviour that you get max 2, as in nearest local + global or are intending on capturing and merging all potential parents recursively ?

maxandersen:
tried using this but not sure how ;) 

```
❯ jbang --verbose config set mmm yyy
[jbang] Reading configuration from /Users/max/.jbang/jbang-config.json
[jbang] Option 'mmm' set to 'yyy' in /Users/max/.jbang/jbang-config.json
❯ jbang config list
❯ less /Users/max/.jbang/jbang-config.json
{
}
```

I would have expected either `mmm=yyy` to end up in the jbang-config.json or some error when trying to set it...?

maxandersen:
and just in case I missed it - atm there isn't actually a way to have `jbang edit` use the configured editor, right ?

src/main/java/dev/jbang/ConfigUtil.java
quintesse: It has the same behavior as aliases, so yes it recurses up the file system tree.

quintesse:
> tried using this but not sure how ;)
> <snip>
> I would have expected either `mmm=yyy` to end up in the jbang-config.json or some error when trying to set it...?

This has to do what I said before:

> There's just a single important issue left to handle: what to do when the user wants to set values that are not supported

The current implementation doesn't allow you to set values that this version of Jbang doesn't know about, so they are lost.
(This has to do with the fact that we're using Qute, it contains only the known settings, somehow adding the unknown values to the output of that template is something I don't know how to handle right now. Which is partly the reason why I wanted to move away from Qute)

quintesse:
> is there any comment preserving yaml parsers in java ?

Dammit, there isn't :-)

> and just in case I missed it - atm there isn't actually a way to have jbang edit use the configured editor, right ?

hmmm, supposedly it should work, but I didn't actually test it :-)

See https://github.com/jbangdev/jbang/pull/604/files#diff-37de0e1a64e7a866c486914a9e9f0d3254dcbc6c2c9acea3ed03cd7aa084b391R51-R62

quintesse:
This has been bit-rotting for quite some time now, basically because we can't really make up our mind what to do with the comments.
 - Yes, JSONC (JSON with Comments) is a thing, but no Java parsers seem to exist.
 - Trying to use Qute files just results in a whole bunch of weird corner cases

So to move things forward I propose we decide between two different possibilities:
 1. We stick with JSON and get rid of comments (we can possibly make this palatable by making the `jbang config` command smarter and have it provide help for attributes it knows about)
 2. We switch to a simpler format that _does_ support comments, like Java properties files for example.

The pro of the JSON format is that all our files are JSON so it makes some sense to make the config file JSON as well. On the other hand our configuration is going to be very basic, I don't see much use for nested options and complex data types, so a simpler format is very likely good enough.

Wdyt?

quintesse:
So I've decided to go with JSON for now and finish the PR. It won't be that much work to change it to some other format if we want to.

So right now this implements:

- Configuration settings stored in files named `jbang-config.json`
- Their lookup follows the exact same rules as for catalog files (`jbang-catalog.json`)
- The format of the file is a simple JSON object with key-values
- The keys follow the format `xxx.yyy.zzz`
- All command options can be set using keys of the form `<command_name>.<longest_option_name>`, eg `"run.debug": "5000"` is the same as `jbang run --debug=5000` on the command line
- For subcommands the key would be `<command_name>.<subcommand_name>.<longest_option_name>`, and so forth
- For options on the `jbang` command itself the command name is actually `all`, not `jbang`
- They can also be set using system properties using the same keys, only prefixed with `jbang.`, eg. `-Djbang.run.debug=5000`
- There is a new `jbang config` command to get, set and list configuration settings

quintesse:
We discussed this and it really would be nice to be able to have comments inside the config file.
Problem is that there's no parser for Java that support JSONC (both for reading and writing) and creating one is outside the scope of this project.
So we decided that we should try plain .properties files for our configuration. There is an Apache project that supports reading/writing .properties files with comments here: https://commons.apache.org/proper/commons-configuration/userguide/howto_properties.html

maxandersen:
@quintesse what was it you were missing for this one ?
publish of bits to maven central?

src/main/java/dev/jbang/cli/Init.java
```diff
@@ -31,15 +31,15 @@
...
 	@CommandLine.Option(names = { "--template",
-			"-t" }, description = "Init script with a java class useful for scripting", defaultValue = "hello")
-	String initTemplate;
+			"-t" }, description = "Init script with a java class useful for scripting")
+	public String initTemplate;
 
 	@CommandLine.Option(names = {
 			"--force" }, description = "Force overwrite of existing files")
-	boolean force;
+	public boolean force;
```
maxandersen, line 39: why do these need to be public now ?

src/main/java/dev/jbang/cli/Wrapper.java
```diff
@@ -26,9 +26,9 @@
 	@CommandLine.Command(name = "install", description = "Install/Setup jbang as a `wrapper` script in a folder")
 	public Integer install(
 			@CommandLine.Option(names = { "-d",
-					"--dir" }, description = "The folder to install the wrapper into.", defaultValue = ".") Path dest,
+					"--dir" }, description = "The folder to install the wrapper into.") Path dest,
```
maxandersen, line 29: why remove default values? they show in the docs.

src/main/java/dev/jbang/cli/Init.java
quintesse: For testing. I'm testing in the configurations tests (which are not in the same package as the CLI commands) that they correctly pick up any configuration settings. Because they are in a different package I can only do that if the CLI commands have public fields.

Is there a issue with that? Having those fields public is not going to cause any problems AFAIK, right?
If yes, then we'd need to come up with a different way to run those tests.

src/main/java/dev/jbang/cli/Wrapper.java
quintesse: Because default values are now handled by the configuration, if a user doesn't enter a value Picocli will fall back on the default value, for which it provides an interface that you can implement to supply those default values in case they are not fixed/static.

> they show in the docs.

But do they? I checked when I started implementing this and now I checked again and I don't see the default values mentioned anywhere when I execute `jbang wrapper --help` (or any other help).

I also see examples in the Picocli docs like this:

```
@Option(names = "-c", description = "The count (default: ${DEFAULT-VALUE})")
```

which seem to suggest you have to write it explicitly into the docs. This is also the recommended way to do things according to: https://picocli.info/#_show_default_values

Now there used to be a `showDefaultValues` option before Picocli version 3.0 but it's deprecated and we're not using it anywhere in our code, so perhaps it was once like this but it was lost some time ago? I don't remember seeing any default values at least.

maxandersen:
love it but a few kinks.

`jbang config list` gives me this when no ~/.jbang/jbang.properties:

```
jbang config list
init.template = hello
run.debug = 4004
run.jfr = filename={baseName}.jfr
wrapper.dir = .
```

default values.

Now do `jbang config set editor.open idea` and its:
```
edit.open = idea
```

where did the other values go ?

then if I try do:

`jbang config unset edit.open` nothing seem to happen. value stil in list and in thefile.

if I empty the file `jbang config list` is empty.

if I remove the file default values are back again.

So unset doesn't work and config list feels like its not quite right.

quintesse:
Things should work now. The `list` was indeed a simple problem where it wasn't looking at the "merged" config but only at the "nearest" one. (Which explains why a set followed by an unset behaved the way it did)

src/main/java/dev/jbang/cli/Config.java
```diff
@@ -0,0 +1,194 @@
...
+			Util.infoMsg("No configuration option found with that name: " + key);
+			return EXIT_INVALID_INPUT;
+		}
+	}
+}
+
+@CommandLine.Command(name = "list", description = "List active configuration values")
+class ConfigList extends BaseConfigCommand {
+	@CommandLine.Option(names = {
+			"--show-origin" }, description = "Show the origin of the catalog")
```
maxandersen, line 135:
```suggestion
			"--show-origin" }, description = "Show the origin of the configuration")
```

maxandersen:
seems like we need a guard against removing default values:

```
 jbang --verbose config unset run.debug
[jbang] jbang version 0.84.0.17
[jbang] [ERROR] public class java.lang.NullPointerException
	at java.base/java.nio.file.Files.provider(Files.java:100)
	at java.base/java.nio.file.Files.isRegularFile(Files.java:2271)
	at dev.jbang.Configuration.read(Configuration.java:267)
	at dev.jbang.util.ConfigUtil.unsetConfigValue(ConfigUtil.java:29)
	at dev.jbang.util.ConfigUtil.unsetNearestConfigValue(ConfigUtil.java:24)
	at dev.jbang.cli.ConfigUnset.doCall(Config.java:121)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:97)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:15)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
	at dev.jbang.cli.JBang$3.handle(JBang.java:141)
	at dev.jbang.cli.JBang$3.handle(JBang.java:136)
[jbang] If you believe this a bug in jbang, open an issue at https://github.com/jbangdev/jbang/issues
```

quintesse:
Fixed.

maxandersen:
`jbang config list --show-origin` shows default list with origins. good.
jbang config list --show-origin --show-available` shows a different list, no values shown nor which origin ...not good or at least not consistent.

can we do better?

src/main/java/dev/jbang/cli/Init.java
maxandersen: I don't see any test referring to them ?

maxandersen: I would rather put the tests in the right package or have an accessor for the field if we absolutely need them.

src/main/java/dev/jbang/cli/Wrapper.java
maxandersen: it works using fallback-value which is nice!

quintesse:
> can we do better?

Yeah, those two options aren't combinable, we should put them in an exclusive group.

PR #621, opened 2020-12-27
fix: stop jbang app install --force from appending

### LINKED ISSUE #620: jbang app install --force appends instead of overwrites
after using --force I get this:

```
#!/bin/sh
eval "exec jbang run https://github.com/ikwattro/kubectl-plugin-java-jbang/blob/master/lp.java $*"
#!/bin/sh
eval "exec jbang run https://github.com/maxandersen/kubectl-plugin-java-jbang/blob/master/lp.java $*"
```

maxandersen:
Fixes #620

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -1,7 +1,11 @@
 package dev.jbang.cli;
 
 import java.io.IOException;
-import java.nio.file.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


quintesse:
Hadn't seem your PR, I made one too: https://github.com/jbangdev/jbang/pull/623 :-)

I think mine is perhaps a little more thorough because it checks for the exact expected contents instead of just this specific error case.

maxandersen:
Ack. Merged yours.

Issue #579, opened 2020-12-04
add notion of default app file


maxandersen:
its all great with single sourced script files but you often tend to end up with auxillary files and in some cases (like quarkus or gum) 
it would be nice not having to specify which java file to use but there just be a default. 

i.e.

`jbang .` -> `jbang ./app.java` 
`jbang https://github.com/jbangdev/myapp/` -> `jbang https://github.com/jbangdev/myapp/blob/master/app.java`

quintesse:
Yeah, that could be useful. In the buildpack I went with `main.java`.

quintesse:
Btw, I just realized that we might have to think carefully on how to implement this. It might not always be easy to detect if that the default app file name should be added to a Path or URL. I mean there's nothing that says that I can't have a Path or URL to a Java file that doesn't end in `.java`. So we can't simply go "ok, there's no .java/.jsh at the end, let's just add `app.java`"

For file paths it would be easy enough I guess. We can detect that the path points to a folder and that's it. We can even actively detect if a file with the name `app.java` (maybe even `app.jsh` as well?) exists.

But for URLs? I think the only possibility is to try the original URL and if it fails try to add `app.java` (`app.jsh`?) and retry. But with what kind of error do we consider it failure? I don't think we can easily detect if it's a failure of the file not existing (many web servers will of course return a result if you query a folder name, resulting in errors of javac trying to compile an HTML index file or something like that) or a failure in the source code.

Something like that might lead to trying to add `app.java` to an URL like `https://server/path/sourcefile` just because sourcefile had a compile error.

Is of course all doable and perhaps there's another clean and nice solution. But just wanted to set down the possible problems that popped into my mind.

maxandersen:
couldn't we just say that for urls you would need to explicitly designate it is the folder you want - i.e. `https://server/path/sourcefilebutitsadir/` makes it clear with the `/` at the end ?

maxandersen:
> Yeah, that could be useful. In the buildpack I went with `main.java`.

ah yes, buildpack is another case where it is nice not having to specify an exact name.

maxandersen:
I guess `main.java` works too its just a bit repetitive: "create a main.java, with a class named main with a method named main" vs 'create app.java with app class and a main method"

also not everything needs a main() method (i.e. quarkus, jfx, etc.)

thats why I went with "app"

quintesse:
Sure, I definitely don't mind either option (or another). It's just the one I chose at that time.

[...]

PR #630, opened 2020-12-28
feat: add notion of default application

maxandersen:
should now treat `jbang .` as `jbang ./app.java` and any urls with `/` in end gets app.java added.

i.e. `jbang https://jbang.dev/` works :) 

note, `jbang https://github.com/jbangdev/jbang.dev/` does *not* currently work - not sure if should make that work...thoughts?


maxandersen:
fixes #579

quintesse (approved):
LGTM. Except perhaps for changing it to use `main.java` :-D

maxandersen:
updated to use `main.java` assuming twitter poll won't have a landslide to the more sensible `app.java` ;)

PR #637, opened 2021-01-01
WIP: Add export --style slim

### LINKED ISSUE #619: jbang export --slimjar / --standalone
currently jbang assumes the app built runs on same machine as its built on.

Could be nice to have export support a "standalone" option (i.e. `jbang export --standalone xyz.jar` where it would not only output a jar, but add a `libs/*` folder with all its dependencies and a matching `xyz`/`xyz.cmd` with the proper argument setup/generated.

besides normally "portability" usecase, it has a usecase in lego mindstorms (ev3) where the host is too slow to build but good enough for running: https://twitter.com/juanantoniobm/status/1341895888034263040

maxandersen:
Example: `jbang export --style slim myrobot.java`

A fairly naive export mechanism to have an option on export
that says instead of just copying the built jar into a directory; make it a
slim jar with a `libs` directory next to it.

the slim jar has all its classpath dependencies updated to point to libs/* instead
with relative paths letting you move it anywhere to just run.

Fixes #619

Thoughts:
To be real useful there should be a matching script generated that includes
the specified/necessary java arguments user specified. Possibly also the full classpath declaration
as not all jars would be runnable with -jar.

I made it part of export and not build as it did not feel right we add another run option at the default
level - but i'm not 100% convinced that is the right thing. i.e. can imagine quarkus integration would
be able to participate in generation of more optimized jar.

other option is also to make this implemented externally as a jbang script - but even if that is done
it at least still make sense to add enough of the metadata about javac/java options to the manifest
which jbang or external tools could use to repackage as all info would be in the jar.


src/main/java/dev/jbang/cli/Export.java
```diff
@@ -22,6 +31,105 @@
...
+						tempManifest = new File(tempDirectory.toFile(), "MANIFEST.MF");
+						jf.getManifest().getMainAttributes().put(Attributes.Name.CLASS_PATH, newPath.toString());
+						// System.out.println("writing manifest " + jf.getManifest() + " to " +
+						// tempManifest);
+						jf.getManifest().write(new FileOutputStream(tempManifest));
+					}
+
+					List<String> optionList = new ArrayList<String>();
+					optionList.add("jar"); // TODO locate it on path ?
+					optionList.add("ufm");
```
quintesse, line 109: Why are you using the `jar` command?

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -22,6 +31,105 @@
...
+
+				Files.copy(source, outputPath);
+				export.info("Exported to " + outputPath);
+				return EXIT_OK;
+			}
+		},
+		slim {
+			@Override
+			public int apply(Export export, Script script, Path outputPath) throws IOException {
+				// Copy the JAR or native binary
```
quintesse, line 65: I think this part until line 80 is the same as the `local` style, right? We should probably factor that out.

src/main/java/dev/jbang/cli/Export.java
quintesse: If we really need to use it I'd at least use the method that resolves commands against the active Jdk : `BaseBuildCommand.resolveInJavaHome()`

quintesse (approved):
LGTM

src/main/java/dev/jbang/cli/Export.java
maxandersen: > Why are you using the `jar` command?

fastest way I could find to modify a jar "in place".

and yes, should make it use similar lookup in javahome

quintesse: But we're already generating JAR files with MANIFESTs without using the `jar` command, right? Can't we somehow re-use that?

maxandersen: we could, but then it would need to be done at build time, not export as the code in build is about creating a full new jar.

quintesse: Ok, but using similar code to update an existing JAR should be possible too, right?

maxandersen: yes, but it would be quite alot more verbose.

maxandersen:
while writing readme.adoc I realized it might be better to name this "style" for `portable` as that what it is - its about making a portable (non machine dependent jar) + should generate a app command/shell that contains the necessary flags.

maxandersen:
@quintesse thoughts about `--portable` vs `--style portable` ?

quintesse:
Hmmm, I like `--portable`, also because `--style` feels strange for export. `--format` might be slightly better but is still not a great fit.

maxandersen:
`--format` could work, but not sure what other formats I can think of atm ;)

maxandersen:
like,  I kinda feel jbang just supporting its own "local" format and this `--portable` would be enough and anything else can be derived by a external script on top of one of those two.

quintesse:
Well `--fatjar` is still a very interesting option too IMO. And that would be "portable" too, so not sure that "portable" is a name that really covers the actual situation.

And who knows, perhaps we want to export to a container image in the future?

maxandersen:
Those are what I would say can be done as plugins not core :)

'--portable=fatty' could be option if we find core need it.

quintesse:
ok :+1:

PR #638, opened 2021-01-01
feat: Download and setup vscodium if no editor available.

### LINKED ISSUE #632: have jbang edit offer to download default usable java enabled editor (like https://vscodium.com)
idea is that `jbang edit` would in cases that no default editor is specified offer (like with java runtime) to download vscodium (free and opensource version of vscode) into `~/.jbang/vscodium` (or similar) with java tooling included and enable [portable install](https://code.visualstudio.com/docs/editor/portable) mode.

it might sound a bit crazy but I feel like it is in line with goal of making java more accessible for new comers.

With that on a bare bones machine with zero java or even IDE configured you can do this:

```
$ curl -Ls https://sh.jbang.dev | bash -s - setup 
Downlading Java 11...
$ jbang init -t cli myapp.java
myapp.java created - use jbang edit myapp.java to edit it
$ jbang edit myapp.java
No default editor configured - Downloading Editor in 3 seconds...Press Ctrl+C if you do not want that
Downloading vscodium
 Running `sh -c ~/.jbang/bin/jbang-edit /Users/max/.jbang/cache/projects/echo.java_jbang_c3307ece9f3e7b927787ad496512c8dccae4e1368b4e15bf5cf50e90109e48f2/echo`
```

and you would be editing full blown java on any (desktop) machine.

for those with pure terminals they can use emacs/vi as is :)

maxandersen:
WIP: Just POC for now. should support linux, windows and osx atm

Depends on #639 for notion of default editor.

Now when no editor is specified it:

1) downloads vscodium
2) create portable dir data
3) setup redhat.java and java.debug extensions

uses the just installed vscodium as editor.

Fixes #632

### DIFF
--- a/src/main/java/dev/jbang/EditorManager.java
+++ b/src/main/java/dev/jbang/EditorManager.java
@@ -1,0 +1,125 @@
+package dev.jbang;
+
+import static dev.jbang.cli.BaseCommand.EXIT_UNEXPECTED_STATE;
+
+import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class EditorManager {
+	private static final String CODIUM_DOWNLOAD_URL = "https://github.com/VSCodium/vscodium/releases/download/%s/VSCodium-%s-%s-%s.%s";
+
+	public static Path getVSCodiumPath() {
+		if (Util.isMac()) {
+			return Settings.getConfigEditorDir().resolve("vscodium.app");
+		} else if (Util.isWindows()) {
+			return Settings.getConfigEditorDir().resolve("vscodium");
+		} else {
+			return Settings.getConfigEditorDir().resolve("vscodium");
+		}
+	}
+
+	public static Path downloadAndInstallEditor() {
+		// https://github.com/VSCodium/vscodium/releases/download/1.52.1/VSCodium-darwin-x64-1.52.1.zip
+
+		String version = getLatestVSCodiumVersion();
+
+		Util.infoMsg("Downloading VSCodium " + version
+				+ ". Be patient, this can take several minutes...(Ctrl+C if you want to cancel)");
+		String url = getVSCodiumDownloadURL(version);
+		Util.verboseMsg("Downloading " + url);
+		Path editorDir = getVSCodiumPath();
+		Path editorTmpDir = editorDir.getParent().resolve(editorDir.getFileName().toString() + ".tmp");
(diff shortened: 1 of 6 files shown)


src/main/java/dev/jbang/EditorManager.java
```diff
@@ -0,0 +1,61 @@
...
+		// https://github.com/VSCodium/vscodium/releases/download/1.52.1/VSCodium-darwin-x64-1.52.1.zip
+		String version = "1.52.1";
+
+		Util.infoMsg("Downloading VSCodium " + version
+				+ ". Be patient, this can take several minutes...(Ctrl+C if you want to cancel)");
+		String os = Util.getOS().name().equals("mac") ? "darwin" : Util.getOS().name();
+		String url = String.format(CODIUM_DOWNLOAD_URL, version, os, Util.getArch().name(), version);
+		Util.verboseMsg("Downloading " + url);
+		Path editorDir = getEditorPath();
+		Path editorTmpDir = editorDir.getParent().resolve(editorDir.getFileName().toString() + ".tmp");
```
quintesse, line 26: I think it becomes time to create a standard "create and switch" system. We must have implemented this old/tmp/move-dance at least three times now :-)

maxandersen: https://github.com/jbangdev/jbang/issues/640

quintesse:
Btw, only after you merged this I realized: shouldn't we do this with the `app install` command, just like we do for `jbang` itself? That way instead of doing it automatically we could tell people "You don't have an editor installed, run 'jbang app install vscodium' to install one".

It just seems that downloading and installing something like that without even asking only because you didn't/forgot to specify an editor seems a bit rash?

maxandersen:
App install is a misfit for this - I don't want there to be another step needed. Should be as easy as possible to get started. The jdk is bigger :)


How about adding a 3-5 second delay for the download similar as done for trusts to autoexit ?

quintesse:
Just tried simply running `jbang edit --open xxx` on my system an dthe first thing it does is download vscodium. To me that's just unacceptable default behavior :-( I've got a gazillion IDEs and editors installed on my system, having it download another by default, even with a delay to me feels wrong. You make is easier perhaps for utter noobs but you make it more difficult to the (in my opinion much larger group) of experienced devs.

quintesse:
If you really really want this I'd suggest something like:

```
$ jbang edit --open xxx
You didn't specify an editor and no default editor was set.
The following editors were found on your system, make a choice:
 1. IntelliJ IDEA (/opt/idea/bin/idea)
 2. Eclipse (/usr/bin/eclipse)
 3. Visual Studio Code (~/bin/code)
 4. Download and install Visual Studio Codium
Type selection: 2
How do you want to use Eclipse?
 1. Only this time?
 2. Use and set as default for the future
Type selection: 2
Setting Eclipse as default editor.
... 
```

PR #639, opened 2021-01-02
feat: default editor --open without editor opens default editor feat: default key value, i.e. -Dkey interpreted as -Dkey=true

maxandersen:
default editor and default key value for system properties made possible
with picocli 4.6.0 release.

Means `jbang edit --open my.java` will now actually honor JBANG_EDITOR value
as documented. Removed the EDITOR and VISUAL fallbacks as that would conflict with want
of supporting vscodium as a fallback.

And means that `jbang -DskipTests properties@jbangdev` will print out
`skipTests = true` which are useful for easy toggles.
"breaking" change here as before `-DskipTests` would have resulted in error
being printed; but now it actually does something useful instead.

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -396,6 +396,11 @@ The script will have the output:
 
 Please note that `.jsh` files are source only, they are not compiled thus they are https://github.com/jbangdev/jbang/issues/506[not cached] nor can they be https://github.com/jbangdev/jbang/issues/510[built as native images].
 
+[TIP]
+====
+If you use `-Dkey` where no value is specified `jbang` will interpret this as `-Dkey=true` allowing you to easily have
+flags passed into, i.e. `jbang -DskipTests mytestrunner.java`. Now within your script `Boolean.getBoolean('skipTests') will return true.
+====
 
 ==== Running script from standard input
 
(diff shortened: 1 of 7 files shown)


readme.adoc
```diff
@@ -883,7 +888,7 @@ Above does require using a shell that allows for variable evaluation, if you are
 jbang edit --open=[editor] helloworld.java
 ----
 
-The editor used will be what is specified as the argument to `--open` or default to `$JBANG_EDITOR`, `$VISUAL` or `$EDITOR` in that order.
```
quintesse, line 886: Btw, I have always doubted was would happen exactly here: does setting JBANG_EDITOR mean that automatically `--open` is assumed and an editor will be opened or do you still need to provide `--open`? (The latter makes more sense to me but the wording seems to suggest the former)

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -42,11 +43,12 @@
 	boolean live;
 
 	@CommandLine.Option(names = {
-			"--open" }, description = "Opens editor/IDE on the temporary project.")
-	String editor;
+			"--open" }, description = "Opens editor/IDE on the temporary project.", fallbackValue = "${JBANG_EDITOR:-}", preprocessor = StrictParameterPreprocessor.class)
+	Optional<String> editor;
```
quintesse, line 47: Just curious: is the Optional necessary here or did you just like it better this way?

quintesse (approved):
LGTM!

readme.adoc
maxandersen: --open needed.

src/main/java/dev/jbang/cli/Edit.java
maxandersen: New feature of picocli that makes it possible to use optional to signal if option was present or not on the command line - no matter if value is null or default. 

Technically could still use null but I liked it better to be explicit andntypesafe

quintesse: Sure. Then we should just gradually convert the rest of the optional options as well. For consistency sake :-)

PR #649, opened 2021-01-04
feat: ask first before downloading default editor

### LINKED ISSUE #641: Jbang edit is too fast at installing default editor
Should give user choice before downloading.

maxandersen:
Now asks user before downloading vscodium with a 10s delay to answer.

Fixes #641

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -19,8 +19,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import dev.jbang.ConsoleInput;
 import dev.jbang.DependencyUtil;
 import dev.jbang.EditorManager;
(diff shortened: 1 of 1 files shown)


maxandersen:
does this by default:

```
 jbang edit --open itests/quote.java
[jbang] You requested to open default editor but no default editor configured.
jbang can download and configure a visual studio code with Java support to use
See https://vscodium.com for details

Do you want jbang to download VSCodium for you into /Users/max/.jbang/editor/vscodium.app ?

0) Yes, please.

Any other response will result in exit.

[jbang] Type in your choice (0) and hit enter. Times out after 10 seconds.
[jbang] [ERROR] Could not parse answer as a number. Aborting
[jbang] [ERROR] No default editor configured and automatic download not accepted.
 Please try again accepting the download or use an explicit editor, i.e. `jbang edit --open=eclipse xyz.java`
```

did not do the whole scan for eclipse, vscode, etc. in path as we don't (yet) have  a way to store the config. could actually be we should say `~/.jbang/bin/jbang-edit` should be the way to configure default editor ? but more complex logic for that would be later anyways.

maxandersen:
hearing no objections - merging ;)

PR #652, opened 2021-01-06
fix: Make sure scripts are actually built before printing info

quintesse:
If we don't do this the result might be incomplete. Running `info classpath`
before building doesn't include the actual script JAR while running it after
building does. That's obviously not consistent.

### DIFF
--- a/src/main/java/dev/jbang/cli/Info.java
+++ b/src/main/java/dev/jbang/cli/Info.java
@@ -19,7 +19,7 @@ import picocli.CommandLine;
 public class Info {
 }
 
-abstract class BaseInfoCommand extends BaseScriptCommand {
+abstract class BaseInfoCommand extends BaseBuildCommand {
 
 	class ScriptInfo {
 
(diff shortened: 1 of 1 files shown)


maxandersen:
This does not sound right. 

This would make printing of this info fail if build errors. 

This needs to work even without  build working.

quintesse:
Why does it need to work even when build errors? What use is this information if it's not to have something that works?

maxandersen:
the main usecase for info is/was for IDE's to get the classpath.

the resulting jar is secondary/optional for that usecase.

maxandersen:
..and not being able to get the classpath for the IDE's just because of a typo in the code would be bad

quintesse:
Ok, but it makes it impossible to get correct info from that call with a single jbang call. Which is a pity because it precludes doing some smart one-liners. But ok, closing.

maxandersen:
Add a --build option or make a built-info command that assumes build can fail ?

quintesse:
Dunno. That seems we would get into the murky twilight zone of `jbang info --build` vs `jbang build --info` ^^

maxandersen:
there is a diff between `jbang info --build` vs `jbang build --info`.

`jbang info --build` is saying do what `jbang info` does but please, ensure build is done before it so you got all the info you I need and okey to fail then.

`jbang build --info` would be saying do what `jbang build` does but change the output to include info making it look like i'm actually calling `jbang info`

...of those two I much prefer `jbang info` as it is not changing expected output it just become a short cut for what `jbang build; jbang info` would do...

quintesse:
I agree that between those two one is preferable. But personally both options feel a bit "off". So I'd rather not rush into adding that feature. I already switched to away from trying to use a one-liner in this case.

PR #656, opened 2021-01-09
fix: do not fail on unrecognized OS like arm64 for new macs. should at least try run located java. Fixes #650

### LINKED ISSUE #650: Add support for macOS arm64
image

maxandersen:

### DIFF
--- a/src/main/java/dev/jbang/Util.java
+++ b/src/main/java/dev/jbang/Util.java
@@ -1,7 +1,5 @@
 package dev.jbang;
 
-import static dev.jbang.cli.BaseCommand.EXIT_UNEXPECTED_STATE;
-
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -181,11 +179,11 @@ public class Util {
 	}
 
 	public enum OS {
-		linux, mac, windows
+		linux, mac, windows, unknown
 	}
 
 	public enum Arch {
-		x32, x64, aarch64
+		x32, x64, aarch64, unknown
 	}
 
 	public enum Vendor {
@@ -289,7 +287,8 @@ public class Util {
 		} else if (os.startsWith("win")) {
 			return OS.windows;
 		} else {
-			throw new ExitException(EXIT_UNEXPECTED_STATE, "Unsupported Operating System: " + os);
+			Util.verboseMsg("Unknown OS: " + os);
+			return OS.unknown;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


maxandersen:
idea for this is that jbang wont fail early on unknown OS but at least go use java available in path.

@quintesse something I might have missed ?

quintesse:
We will have to fix the startup scripts as well, eg: https://github.com/jbangdev/jbang/blob/master/src/main/scripts/jbang#L51

quintesse:
I think the best thing to do there is to move the `exit` to the part of the code where we try to download the JDK. If in that part `$os == ""` we can just exit. Perhaps with an extra message like "No Java found and OS is not supported for automatic download".

maxandersen:
Good catch. Is that a change you can add to this branch/pr or want me to recall my bad scripting skills? :)

quintesse:
Added my own branch becuase I didn't have the rights to update yours: https://github.com/jbangdev/jbang/pull/657

maxandersen:
ack - thought you already had push access; my bad - fixed now so don't push to origin by accident - now it won't fail ;)

maxandersen:
closing.

quintesse:
> fixed now so don't push to origin by accident

Have you turned off pushing to "master"? That at least makes things safer. Accidentally overwriting a branch is most likely less of a problem than overwriting master :-)

maxandersen:
> Have you turned off pushing to "master"? That at least makes things safer. Accidentally overwriting a branch is most likely less of a problem than overwriting master :-)

that would require some bot to do merges to master? (but no, haven't turned off master pushes)

quintesse:
The idea is that all changes are done using PRs, no direct pushes to master ever.
(As admin you can always turn it off temporarily if you really need to push, for example to do a force-push to remove sensitive information)

maxandersen:
So when/what does the pushes ?
I'm all for enabling it - just haven't tried it yet so not sure what to enable on GitHub to make it happe.

quintesse:
There are no pushes to master, only PRs that get merged. You can only do pushes to branches.

quintesse:
You can set that protection here: https://github.com/jbangdev/jbang//settings/branch_protection_rules/new (that should be the correct URL, can't test it cause I don't have the rights). You set the name to "master", enable any other options you like (by default it only protects against force pushes from non-admins) and hit Create.

maxandersen:
Okey. Enabled :)

PR #657, opened 2021-01-09
fix: do not fail on unrecognized OS like arm64 for new macs. should at least try run located java.

quintesse:
Overrides https://github.com/jbangdev/jbang/pull/656


maxandersen:
looks great! thanks!

PR #659, opened 2021-01-10
fix: have app install use better derived name

### LINKED ISSUE #653: auto naming of jbang install not honoring extension less naming like kubectl-example
`jbang app install https://github.com/jbangdev/k8s-cli-java/blob/jbang/kubectl-example` results in `KubectlExample` being installed.

it should be `kubectl-example` to be consistent with  https://github.com/jbangdev/jbang#extension-lessnon-java-files-for-cli-plugins

maxandersen:
i.e. `kubectl-example` stays `kubectl-example` not kebabified `KubectlExample`

Fixes #653

### DIFF
--- a/itests/kubectl-example
+++ b/itests/kubectl-example
@@ -1,0 +1,297 @@
+//#!/usr/bin/env jbang
+// when editing comment out the line above to avoid errors in IDE's.
+// Necessity to have it as kubectl don't recognize it as a script otherwise.
+// See https://github.com/kubernetes/kubectl/issues/822
+
+// client-java for k8s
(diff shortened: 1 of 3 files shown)


maxandersen:
@quintesse changed this to use original ref as it probably have the "proper" name rather than any unkebabified file names.

maxandersen:
some windows specific failures - investigating

maxandersen:
> some windows specific failures - investigating

fixed - now it handles urls, linux and windows paths references.

quintesse (approved):
LGTM

PR #660, opened 2021-01-11
feat: add //DESCRIPTION tag in scripts

### LINKED ISSUE #190: We should add a //DESC or //DESCRIPTION tag for scripts
This way we can extract that and use it for the "description" field in the `jbang-catalog.json` files.

Multiple tags would be joined together as a block of text.

jbardel:
A proposition that fixes #190.
Support of multiple tag DESCRIPTION that creates a merged description.
The description in argument has more precedence than the DESCRIPTION tag in the script.

### DIFF
--- a/src/main/java/dev/jbang/AliasUtil.java
+++ b/src/main/java/dev/jbang/AliasUtil.java
@@ -201,7 +201,7 @@ public class AliasUtil {
 			Map<String, String> props = a1.properties != null && !a1.properties.isEmpty() ? a1.properties
 					: a2.properties;
 			Catalog catalog = a2.catalog != null ? a2.catalog : a1.catalog;
-			return new Alias(a2.scriptRef, null, args, props, catalog);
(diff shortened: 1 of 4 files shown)


quintesse (dismissed):
LGTM.

maxandersen:
Nice patch, I noticed a few things while testing:

- if I have no `//DESCRIPTION` I get `"description": ""` added to the .json - it shouldn't be there in that case.

- If we have multiple `//DESCRIPTION` should we not treat it as having a newline instead of just one large string ?

maxandersen (approved):
+1 on LGTM. I added the handling of null and multilines so can get a release out ;)

jbardel:
Indeed, I forgot to test case with no //DESCRIPTION.

For the case with multiple descriptions, I thought about adding new line each time we get the tag, but wasn't sure it was the behavior expected.

Anyway, ty for the fix and the merge ! :D

maxandersen:
@all-contributors please add @jbardel for code

thanks for the contribution!

PR #672, opened 2021-01-16
Major refactor, Scripts can now be nested (for //SOURCE)

quintesse:
This is still a WIP, not all tests pass yet, so do NOT merge!

This PR is a major refactor of `Script`, basically to make it possible that any `//` directives in included source files are picked up.
For that it was necessary that a method like `collectSources()` would actually return instances of `Script`, not just references. And then it becomes easy to say for example: give me all the dependencies defined in this Script or in any of the `//SOURCE` scripts, etc so forth recursively.

But Script contained a lot of stuff that have nothing to do with Scripts. First of all it keeps information about JARs but it also contains a lot of "secondary" information that we basically just put there because necessary info and the class was available.

So I started by splitting off most of that extraneous information into a subclass: `ExtendedScript`, so `Script` would mostly only contain info that could be read from the source files and nothing more.

But it still contained info about JARs, so I split that off into two separate files: `Script` and `Jar`, both implementing `RunUnit.`

I then turned `ExtendedScript` into a container class for either a Script or a Jar, so I changed its name to `ExtendedRunUnit`.

Now, I'm not happy with the names `RunUnit` (but okay, that's only an interface, not very visible) and `ExtendedRunUnit` (I especially hate this name). So perhaps we can come up with something better before merging.

This PR still needs tests as well.

But given the fact that this PR touches a lot of code I also don't want to wait too long and let it bit-rot. So it would be nice to get some early feedback :-)


quintesse:
Added description...

maxandersen:
Just as I'm exploring refactoring dependency resolution :) good timing. 

Will look later tonight.

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -79,13 +74,15 @@ public Integer doCall() {
...
 		try {
-			Script script = BaseScriptCommand.prepareScript(scriptOrFile);
+			ExtendedRunUnit script = RunUnit.forResource(scriptOrFile);
 			if (name == null) {
 				name = AppInstall.chooseCommandName(script);
 			}
 
-			if (description == null && script.getDescription() != null && script.getDescription().length() > 0) {
-				description = script.getDescription();
+			if (description == null && script.runUnit instanceof Script) {
```
maxandersen, line 82: why need instanceof to do things right ? that feels wrong...

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -79,13 +74,15 @@ public Integer doCall() {
...
+				if (script.script().getDescription() != null && script.script().getDescription().length() > 0) {
```
maxandersen, line 83: maybe rename `script` to `unit` ?

maxandersen: so its unit.script()

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -95,7 +92,7 @@ public Integer doCall() {
 				catFile = AliasUtil.addNearestAlias(null, name, scriptOrFile, description, userParams, properties);
 			}
 			info(String.format("Alias '%s' added to '%s'", name, catFile));
-		} catch (IOException e) {
+		} catch (Exception e) {
```
maxandersen, line 95: why need need for catching this broad?

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -148,36 +135,33 @@ void build(Script script) throws IOException {
...
-							.stream()
-							.map(x -> x.getResolvedPath().toFile().getAbsolutePath())
-							.collect(Collectors.toList()));
-		}
+		optionList.add(script.getBackingFile().getPath());
+		optionList.addAll(
+				script	.script()
+						.collectAllSources()
+						.stream()
+						.map(x -> x.getBackingFile().getPath())
```
maxandersen, line 158: this doesn't change the semantics as using different files/paths?

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -220,17 +209,19 @@ File createProjectForEdit(Script script, boolean reload) throws IOException {
...
 
-		for (Source source : script.getResolvedSources()) {
-			File sfile = null;
-			if (source.getJavaPackage().isPresent()) {
-				File packageDir = new File(srcDir, source.getJavaPackage().get().replace(".", File.separator));
-				packageDir.mkdir();
-				sfile = new File(packageDir, source.getResolvedPath().toFile().getName());
-			} else {
-				sfile = new File(srcDir, source.getResolvedPath().toFile().getName());
+		if (script.runUnit instanceof Script) {
```
maxandersen, line 212: when is script never a Script ?

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -249,52 +240,52 @@ File createProjectForEdit(Script script, boolean reload) throws IOException {
 
 		// both collectDependencies and repositories are manipulated by
 		// resolveDependencies
-		List<MavenRepo> repositories = script.getRepositories();
+		List<MavenRepo> repositories = script.script().collectAllRepositories();
```
maxandersen, line 243: why this rename ? before collect* was for non-cached and get* was caching collected?

maxandersen:
> that any `//` directives in included source files are picked up.

we don't necessarily want *all* directives picked up. are there any in particular you are missing ?

maxandersen:
I agree on the names not being great but i'm not still fully grokking what the various subclasses of RunUnit represents...could you add some basic javadoc to header of new classes - might help my brain to get it faster.

src/main/java/dev/jbang/cli/Alias.java
quintesse: Yeah, but didn't feel like renaming all those yet :-)
Most of that is manual work, no good way to do a search & replace.

src/main/java/dev/jbang/cli/Alias.java
quintesse: For some reason I thought multi-catch wasn't part yet of Java 8 and I hate repeating the same code in multiple catch-blocks :-) will change

src/main/java/dev/jbang/cli/Alias.java
quintesse: Why? It's functionality that is only supported by Scripts. To me it seems more wrong to add a `getDescription()` to Jar, where it doesn't make any sense, just to avoid an `instanceof`?

src/main/java/dev/jbang/cli/Edit.java
quintesse: When it's a JAR. In fact, in the current version of Jbang you can do `jbang edit /path/to/a/jar.jar` and it will accept it and then fail with an ugly error. (This doesn't fix that, but it shows you can get to that code and not have an instance of Script)

src/main/java/dev/jbang/cli/Edit.java
quintesse: Ok, that distintion must have been lost at some point, because we definitely have `collectXxxx()` calls that do caching right now. Not seeing and not knowing the difference I just decided to unify the names. But I can definitely change them all to `getXxxx()`.

quintesse:
> we don't necessarily want all directives picked up. are there any in particular you are missing ?

What do you mean? I'm not missing any and not _all_ directives are recursively handled. For example the `getDescription()` is not recursive because that wouldn't make sense. But most of the others all make sense to retrieve recursively. (eg. repos, dependencies, options, etc).

quintesse:
> could you add some basic javadoc to header of new classes - might help my brain to get it faster.

Sure will do, but I thought they were pretty self-explanatory:

`Script` - Scripts, sources, .java/.jsh files
`Jar` - .jar files

src/main/java/dev/jbang/cli/Alias.java
quintesse: In fact that code doesn't throw an exception anymore...

src/main/java/dev/jbang/cli/Alias.java
quintesse: I've added the `getDescription()` to `RunUnit`, who knows JAR files might have a description at some point.

src/main/java/dev/jbang/cli/Edit.java
quintesse: Done

src/main/java/dev/jbang/cli/Alias.java
quintesse: I renamed all of them to `xrunit`. Not a beautiful name but it's one that's very easy to search & replace for a future rename :-)

src/main/java/dev/jbang/ExtendedRunUnit.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.stream.Stream;
+
+/**
+ * This class wraps a reference to an runnable/executable resource in the form
+ * of a RunUnit. It also holds all additional information necessary to be able
+ * to actually run/execute that resource. This is information that can not be
+ * induced or extracted from the resource itself but is information that is
+ * provided by the user or by the environment at runtime.
+ *
+ * This class also implements RunUnit, passing all calls directly to the wrapped
```
maxandersen, line 18: I kinda feel that these are different things and should use delegation rather than inheritance ...

src/main/java/dev/jbang/FileRef.java
```diff
@@ -10,40 +10,34 @@
...
-	final Script source;
-
-	String ref;
-
-	String destination;
-
-	public FileRef(Script source, String destination, String ref) {
-		assert (destination != null);
-		this.source = source;
+	public FileRef(String base, String ref, String destination) {
```
maxandersen, line 17: reason I had Script passed in here was that in case of error you should be able to report specificly where it was included from at least. is base enough for that?

src/main/java/dev/jbang/Jar.java
```diff
@@ -0,0 +1,72 @@
...
+ * It's a reference to an already existing JAR file, either as a solitary file
+ * on the user's file system or accessible via a URL. But it can also be a Maven
+ * GAV (group:artifact:version) reference resolving to a JAR file in the Maven
+ * cache (~/.m2/repository).
+ *
+ * NB: The Jar contains/returns no other information than that which can be
+ * extracted from the JAR file itself. So all Jars that refer to the same JAR
+ * file will contain/return the exact same information.
+ */
+public class Jar implements RunUnit {
```
maxandersen, line 17: JarUnit would be better here

src/main/java/dev/jbang/ExtendedRunUnit.java
quintesse: Well it uses delegation. That's why RunUnit is an interface. So the ExtendedRunUnit delegates to the wrapped RunUnit (Being Script or Jar)

src/main/java/dev/jbang/FileRef.java
quintesse: The code wasn't using anything else from the Script than it's base, that's why I removed it, to simplify it as much as possible. First I changed it to (what is now) ResourceRef but then I realized it wasn't even using all of that.

src/main/java/dev/jbang/Jar.java
quintesse: JarUnit and ScriptUnit?

quintesse: An other option instead of "RunUnit" would be something like "Executable" (or "Runnable" but then you get all the conflicts with the existing Java Runnable)

quintesse: Ok, I see you wrote stuff in the chat too. ScriptUnit, RunnableUnit, etc. Let's discuss that some more there first...

maxandersen:
okey - been looking at this some more (or rather thought about it more ;) 

Here is some names my brain seem to deal with better - let me know if they match:

`RunUnit` -> `RunnableUnit` (because `RunUnit` just rubbed me the wrong way ;)
`Jar` -> `RunnableJar` (since it is not just a Jar)
`Script` -> `RunnableScript` (to match RunnableJar and just because I would keep thinking Script is what is now `BuildableUnit`
`ExtendedRunUnit` -> `BuildableUnit` (as it is not really extended; it is the "thing" that is buildable and what is actually passed into `build`)

src/main/java/dev/jbang/cli/Run.java
```diff
@@ -185,8 +184,8 @@ String generateCommandLine(Script script) throws IOException {
 					optionalArgs.add(classpath);
 				}
 
-				if (optionActive(cds(), script.enableCDS())) {
-					String cdsJsa = script.getJar().getAbsolutePath() + ".jsa";
+				if (xrunit.runUnit instanceof Script && optionActive(cds(), xrunit.script().enableCDS())) {
```
maxandersen, line 187: I don't grok why this should only apply for Scripts and not Jars ?

src/main/java/dev/jbang/cli/Run.java
```diff
@@ -210,42 +209,44 @@ String generateCommandLine(Script script) throws IOException {
...
+									"No jar found for agent " + agent.getResourceRef().getOriginalResource());
 						}
 						fullArgs.add("-javaagent:" + jar
 								+ (agent.getJavaAgentOption() != null ? "=" + agent.getJavaAgentOption() : ""));
 
 					});
 
-			fullArgs.addAll(script.collectRuntimeOptions());
-			fullArgs.addAll(script.getAutoDetectedModuleArguments(requestedJavaVersion, offline));
+			if (xrunit.runUnit instanceof Script) {
```
maxandersen, line 224: this is not only scripts - jars could have runtime options....but yeah right now it won't.

maxandersen:
I pushed a commit with the suggested names - if completely wrong we can revert it.

maxandersen (changes requested):

src/main/java/dev/jbang/cli/Run.java
quintesse: Because the original code checks for the existence of `//CDS` which can only happen in scripts? Or did I understand this wrong? (But then I don't understand how it works)

src/main/java/dev/jbang/cli/Run.java
quintesse: Sure, but `collectRuntimeOptions()` is only for `//JAVA_OPTS` which only exists on scripts.

quintesse:
> ExtendedRunUnit -> BuildableUnit (as it is not really extended; it is the "thing" that is buildable and what is actually passed into build)

I'm perfectly fine with the other suggestions but not with this one. A Jar isn't "buildable" so wrapping a JarRunnable inside a `BuildableUnit` doesn't make sense.

You could also easily build a ScriptRunnable so IMO it doesn't really explain correctly what it does.

If you look at the fields of ExtendedRunUnit, you'll see that most of them are things that are either not fixed or that the user passes to change the default behavior:

	- arguments: the arguments are of course not known beforehand and are therefore not part of Script or Jar which are fixed representations of code
	- properties: the properties that the user passes. Will be merged at runtime with any options that are found in the Script or in the JAR's manifest
	- additionalDeps: added by the user, will be merged with the Script dependencies at runtime
	- additionalClasspaths: same
	- forcejsh: can't be determined from the Script itself, must be supplied by the user
	- mainClass: the user wants to use a different main class
```

So they are mostly things that are supplied by the user, things that together with the wrapped Script or Jar objects determine the _actual_ thing to run and _how_ to run it.

File: src/main/java/dev/jbang/ExtendedRunUnit.java [hunk 12]
```diff
@@ -0,0 +1,264 @@
...
+ * RunUnit object. This makes it easier to use this class in places where it's
+ * not really important to know with what type or RunUnit we're dealing.
+ */
+public class ExtendedRunUnit implements RunUnit {
```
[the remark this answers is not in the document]
quintesse: Just one example but there are dozens:

Info.java:48

```
			javaVersion = xrunit.javaVersion();
```

Meaning it will return the Java version requested by the Scxript or Jar without having to know if what you have is a Script or a Jar.

Otherwise you either have to do (stupidly):

```
if (xunit.runUnit instanceof Script) {
   javaVersion = xrunit.script().javaVersion();
} else {
   javaVersion = xrunit.jar().javaVersion():
}
```

or add a new helper method just for this, eg:

```
javaVersion = xrunit.unit().javaVersion();
```

Personally I prefer my solution over this last one.

A completely different solution, which work too, is to just split `ExtendedRunUnit` off from the wrapped object inside it. So instead of passing a single `Extended RunUnit` which wraps a `Script` or `Jar` object we could pass two object everytwhere: one `RunUnit` (script or Jar) and one ... `ExtraInfoSuppliedByUser` object. But I'm not positive this would be any better.

src/main/java/dev/jbang/cli/Run.java
maxandersen: Technically you can specify via env vars too.

src/main/java/dev/jbang/cli/Run.java
maxandersen: I think the old code was wrong :)

CDs is applicable to any launch Afaics. (Except native)

maxandersen:
DecoratedRunnable...it's an unnamable thing :)

quintesse:
> DecoratedRunnable...it's an unnamable thing :)

Hahaha that's ok with me (if you were serious hehe)

src/main/java/dev/jbang/cli/Run.java
quintesse: Ok, I can certainly add that.

Btw, how do you force CDS *off* if you don't want to use it?

src/main/java/dev/jbang/cli/Run.java
```diff
@@ -175,18 +178,25 @@ String generateCommandLine(Script script) throws IOException {
 					// TODO: find way to generate ~/.jbang/script.jfc to configure flightrecorder to
 					// have 0 ms thresholds
 					String jfropt = "-XX:StartFlightRecording=" + flightRecorderString.replace("{baseName}",
-							Util.getBaseName(script.getBackingFile().toString()));
+							Util.getBaseName(xrunit.getResourceRef().getFile().toString()));
 					optionalArgs.add(jfropt);
 					Util.verboseMsg("Flight recording enabled with:" + jfropt);
 				}
 
+				if (xrunit.getJar() != null) {
```
maxandersen, line 186: this looks new ? what does it do ?

maxandersen (approved):
lgtm

PR #673, opened 2021-01-17
feat: support for BOM POM's i.e. maven dependencyManagement

### LINKED ISSUE #63: support for adding a BOM pom as dependency
would be cool to be able to use a BOM pom to define the versions for related dependencies so you don't need to provide specific versions

//DEPS com.azure:azure-sdk-bom:1.0.0-beta.1 com.azure:azure-identity com.azure:azure-security-keyvault-secrets

maybe prefix it with something pom: or it should be possible to determine it from the type

maxandersen:
jbang will now if and only if the very first dependency has a `@pom`
classifier, download it and load it as pom file in shrinkwrap.

Example:

```java
//DEPS io.quarkus:quarkus-bom:1.11.0.Final@pom
//DEPS io.quarkus:quarkus-smallrye-openapi
//DEPS io.quarkus:quarkus-swagger-ui
```

Then that pom file dependency management (and dependencies) gets listened to.

Fixes #63

Caveats:
 - I would prefer if could load arbitrary amount of bom/pom's but I haven't found
   a way to do that with shrinkwrap; we could potentially if more than one found generate
   a temporary one that has them all listed (TBD)
- This also means that we can't support the sequence being arbitrary; basically @pom's has to come first
  Should add check that for now is only *one* @pom and reject otherwise we risk breaking users in future when we have it.
- I had to create a fork of shrinkwrap resolver, to circumvent it so it would not filter out pom files. 
  Could probably find a way to call maven directly to get it but for now this is what we got :)

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -530,6 +530,28 @@ $ ./classpath_example.java
 1 [main] INFO classpath_example  - Hello from Java!
 ----
 
+=== Managed dependencies ("BOM POM"'s) [Experimental]
+
+When using libraries and frameworks it can get tedious to mange and update multiple versions.
+For that jbang started since 0.62 to support so called "BOM POM"'s which are commonly used for managing versions.
+
+You use it by having as the very first dependency a `@pom` reference. This first reference will be used
+to define your managed dependences. Below is an example how that could look like when using Quarkus:
+
+```java
+//DEPS io.quarkus:quarkus-bom:1.11.0.Final@pom
+//DEPS io.quarkus:quarkus-resteasy
+//DEPS io.quarkus:quarkus-smallrye-openapi
+//DEPS io.quarkus:quarkus-swagger-ui
+```
+
+Notice the `@pom` at first line and then following dependencies are not required to use explicit versions.
+
+[NOTE]
+At the moment jbang support only one bom pom; in future it should be expanded to multiple.
+For now you can workaround this by reusing a published pom that includes all the dependency management
+sections you need.
+
 === Using links to Git sources
 
 Instead of gradle-style locators you can also use URLs to projects on GitHub, GitLab or BitBucket.
--- a/src/main/java/dev/jbang/DependencyUtil.java
+++ b/src/main/java/dev/jbang/DependencyUtil.java
@@ -21,9 +21,12 @@ import org.jboss.shrinkwrap.resolver.api.maven.Maven;
 import org.jboss.shrinkwrap.resolver.api.maven.MavenFormatStage;
 import org.jboss.shrinkwrap.resolver.api.maven.MavenResolvedArtifact;
 import org.jboss.shrinkwrap.resolver.api.maven.MavenStrategyStage;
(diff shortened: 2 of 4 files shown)


readme.adoc
```diff
@@ -530,6 +530,28 @@ $ ./classpath_example.java
 1 [main] INFO classpath_example  - Hello from Java!
 ----
 
+=== Managed dependencies ("BOM POM"'s) [Experimental]
+
+When using libraries and frameworks it can get tedious to mange and update multiple versions.
```
tivrfoa, line 535: "tedious to **manage**"?

readme.adoc
```diff
@@ -530,6 +530,28 @@ $ ./classpath_example.java
...
+//DEPS io.quarkus:quarkus-bom:1.11.0.Final@pom
+//DEPS io.quarkus:quarkus-resteasy
+//DEPS io.quarkus:quarkus-smallrye-openapi
+//DEPS io.quarkus:quarkus-swagger-ui
+```
+
+Notice the `@pom` at first line and then following dependencies are not required to use explicit versions.
+
+[NOTE]
+At the moment jbang support only one bom pom; in future it should be expanded to multiple.
```
tivrfoa, line 551: support**s**?

src/main/java/dev/jbang/DependencyUtil.java
```diff
@@ -177,7 +219,7 @@ public String decodeEnv(String value) {
 	}
 
 	public static boolean looksLikeAGav(String candidate) {
```
tivrfoa, line 221: maybe change the name to looksLikeAFullGav ?

Don't you need to look for gav (not full) anymore?

src/main/java/dev/jbang/DependencyUtil.java
```diff
@@ -131,26 +137,62 @@ public ModularClassPath resolveDependencies(List<String> deps, List<MavenRepo> r
 			mavenRepo.apply(resolver);
 		});
 
+		MavenWorkingSession xyz = ((MavenWorkingSessionContainer) resolver).getMavenWorkingSession();
```
tivrfoa, line 140: some side effect here? Because it seems xyz is not being used.

readme.adoc
maxandersen: thx fixed!

readme.adoc
maxandersen: fixed

src/main/java/dev/jbang/DependencyUtil.java
maxandersen: this method is to recognize GAV's on command line so I think its named appropriately. better if we remove the need to have a full GAV pattern.

src/main/java/dev/jbang/DependencyUtil.java
maxandersen: whoop nope - leftover from me exploring to use raw maven. turned out not needed for now.

PR #692, opened 2021-01-31
feat: Allow adding repositories from cli using `--repos`

### LINKED ISSUE #611: add `--repos` support
would be nice not having to have intermediate `.java` file and instead just do:

``` jbang --repo https://repo.spongepowered.org/maven org.spongepowered.spongevanilla:spongevanilla:1.12.2-7.3.1-RC395```

### PR DESCRIPTION
**Pull request by maxandersen**
Add `--repo` to cli including a refactoring of how dependency context is
setup.

Now DependencyContext is created and pass through rather than having
individual deps, repos and classpath entries.

Fixes #611

WIP as turns out we have no tests for --deps nor --classpath and thus 
adding test for repos needs more work. Submitting PR just to check 
haven't broken something thus far.

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -30,6 +30,7 @@ import org.jboss.jandex.Indexer;
 import org.jboss.jandex.Type;
 
 import dev.jbang.DecoratedSource;
+import dev.jbang.DependencyContext;
 import dev.jbang.ExitException;
 import dev.jbang.FileRef;
 import dev.jbang.IntegrationManager;
@@ -81,6 +82,9 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
 	@CommandLine.Option(names = {
 			"-n", "--native" }, description = "Build using native-image", defaultValue = "false")
 	boolean nativeImage;
+
+	@CommandLine.Option(names = { "--repos" }, description = "Add additional repositories.")
+	List<String> repositories;
 
 	@CommandLine.Option(names = { "--deps" }, description = "Add additional dependencies.")
 	List<String> dependencies;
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: src/main/java/dev/jbang/cli/BaseBuildCommand.java [hunk 1]
```diff
@@ -82,6 +83,9 @@ void setJavaVersion(String javaVersion) {
 			"-n", "--native" }, description = "Build using native-image", defaultValue = "false")
 	boolean nativeImage;
 
+	@CommandLine.Option(names = { "--repos" }, description = "Add additional repositories.")
```
aimiebell, line 86: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

src/test/java/dev/jbang/cli/TestExternalDeps.java
```diff
@@ -0,0 +1,86 @@
...
+
+import dev.jbang.BaseTest;
+import dev.jbang.DecoratedSource;
+import dev.jbang.Util;
+
+import picocli.CommandLine;
+
+class TestExternalDeps extends BaseTest {
+
+	String checkdeps = "public class test {"
```
aimiebell, line 20: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

maxandersen: Jbang encourages writing scripts that are used from cli. Mixed casing does not work well here. Thus we use lower case for main script entry points. 

I also realize you might just be a bot and don't see this but I do hope you'll take that into your consideration when adding comments elsewhere. 

Peace.

maxandersen:
replaced with #1081

PR #695, opened 2021-02-02
feat: `--deps` and `--cp` now available on `info` and `edit`

quintesse:
It's a minor change which adds the flags `--deps` and `--cp` to the `info` and `edit` commands.
It's especially useful for `info` where we can add extra dependencies using `--deps`. The `--cp` is perhaps less useful but it's just to be complete.
The same for `edit`, not sure if these options are very useful for it, but it uses the same kind of code that `info` does, so just to be complete I've added the options there too.

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseScriptDepsCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseScriptDepsCommand.java
@@ -1,0 +1,15 @@
+package dev.jbang.cli;
+
+import java.util.List;
+
+import picocli.CommandLine;
+
+public abstract class BaseScriptDepsCommand extends BaseScriptCommand {
+
+	@CommandLine.Option(names = { "--deps" }, description = "Add additional dependencies.")
+	List<String> dependencies;
+
+	@CommandLine.Option(names = { "--cp", "--class-path" }, description = "Add class path entries.")
+	List<String> classpaths;
+
+}
--- a/src/main/java/dev/jbang/DecoratedSource.java
+++ b/src/main/java/dev/jbang/DecoratedSource.java
@@ -275,6 +275,17 @@ public class DecoratedSource implements Source {
 		return classpath.getAutoDectectedModuleArguments(requestedVersion);
 	}
 
+	public void importJarMetadata() {
+		File outjar = getJar();
+		if (outjar.exists()) {
+			JarSource jar = JarSource.prepareJar(
+					ResourceRef.forNamedFile(getResourceRef().getOriginalResource(), outjar));
+			setMainClass(jar.getMainClass());
+			setPersistentJvmArgs(jar.getRuntimeOptions());
+			setBuildJdk(jar.getBuildJdk());
+		}
+	}
+
 	public static DecoratedSource forResource(String resource) {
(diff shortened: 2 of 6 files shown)


quintesse:
Btw if you're wondering why, it's because I want to be able to do:

```
$ java -cp $(jbang info classpath --deps some.group.artifact:version app.java) MainClass
```

And the reason I want to be able to use `--deps` and not add `//DEPS` to `app.java` is twofold:

 1. Not sure they'll allow me to add `//DEPS` to that file, as it's not mine
 2. Still thinking about ServiceLoaders https://github.com/jbangdev/jbang/discussions/266

So basically if I'm allowed to do `jbang run --deps xxx --cp yyy ....` then I should also be able to do `java -cp $(jbang info --deps xxx --cp yyy ....)` and be able to get the same result.

quintesse:
@maxandersen if possible I'd like to have this merged and released ASAP, it's something I could really use for a project :-)

maxandersen:
> @maxandersen if possible I'd like to have this merged and released ASAP, it's something I could really use for a project :-)

agree on intent but would feel so much better merging it quickly if there was just a basic test for it - can you add that ?

src/main/java/dev/jbang/cli/BaseScriptDepsCommand.java
```diff
@@ -0,0 +1,15 @@
+package dev.jbang.cli;
+
+import java.util.List;
+
+import picocli.CommandLine;
+
+public abstract class BaseScriptDepsCommand extends BaseScriptCommand {
+
+	@CommandLine.Option(names = { "--deps" }, description = "Add additional dependencies.")
```
maxandersen, line 9: fine for now doing it by inheritance but I think picocli has notion of compositional reuse which might be better in long run...but ok for this pr to do it this way

quintesse: Yeah, it does, but it doesn´t make the code nicer, because in the end Java doesn't have a notion of "mixins", so suddenly you introduce an indirection everywhere you need to use those fields. (Now, indirections aren't bad or anything, but it just feels weird to me that among all the options we have suddenly there are 2 that need indirection).

But there might indeed come a point where just more levels of inheritance won't work, we'll think about what to do when we get there :-)

quintesse:
> A can you add that ?

Sure, will do!

quintesse:
Done. There were _no_ tests for `info` actually. And the tests did expose a couple of minor issues :-)

src/main/java/dev/jbang/cli/BaseScriptDepsCommand.java
maxandersen: yes, for this one its not a big win - but i think we are on the cusp where having mixins for everything feeding into "run" would be nice to get applied uniformly across all the commands.

maxandersen (approved):

src/main/java/dev/jbang/cli/BaseScriptDepsCommand.java
quintesse: Yeah, you're probably right

PR #697, opened 2021-02-02
Jar classpath + cleanup

quintesse:
The first commit is simple:

The script's classpath was already stored in the JAR but we extracted it only in the specific place where it was needed. The `JarSource` now takes care of extracting that information.
This also cleans up the usage of MANIFEST attribute names a bit.

The second commit is a bit trickier:

The script's JAR is now used exclusively when it exists. Before Jbang would still parse the script file for information about dependencies and options, but those are also stored in the JAR itself so we should skip that and just use the JAR. Which is what this commit does. It also allows us the remove that "specific place" mentioned above.

### DIFF
--- a/src/main/java/dev/jbang/JarSource.java
+++ b/src/main/java/dev/jbang/JarSource.java
@@ -8,6 +8,8 @@ import java.util.List;
 import java.util.Properties;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * A Jar represents a Source (something runnable) in the form of a JAR file.
@@ -23,6 +25,7 @@ import java.util.jar.JarFile;
 public class JarSource implements Source {
 	private final ResourceRef resourceRef;
 
+	private String classPath;
 	private String mainClass;
 	private List<String> javaRuntimeOptions;
 	private int buildJdk;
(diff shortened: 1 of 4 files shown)


quintesse:
The reason why the second commit is "trickier" is because while it works just fine there is a corner-case where using the JAR exclusively would fail while the current code would work. And that's if between when the JAR is built and when it's run the user moves or deletes Maven's `.m2` cache. (This is because the JAR's classpath is hard-coded to the resolved dependencies).

So I'm actually not convinced we should use the second commit but I wanted to run the option by you anyway @maxandersen .

maxandersen:
Cant we like we do today check file classpath existence before launch ? 

Yes it will cost a bit but only be penalized when users nuked deps ?

quintesse:
> Cant we like we do today check file classpath existence before launch ?

Do we do that today? Can you point me to that?

quintesse:
Ok, I really need to give that second commit a bit more thought. I removed it for now. So let's just merge this as a minor code addition + some cleanup.

maxandersen:
> Do we do that today? Can you point me to that?

https://github.com/jbangdev/jbang/blob/master/src/main/java/dev/jbang/Settings.java#L287

basically if we find that the deps asked for already are in the resolved cache we check the list of files - if all good we run if fail we force resolution.

maxandersen (approved):

PR #698, opened 2021-02-03
refactor: Script refactor 2

quintesse:
Ok, major refactor again of the same Script code :-)

This is Phase 2 where I get rid of that weird DecoratedSource.

So what I basically did was split it into it's 2 component parts:

 - **Source** - we already know and have that one, not much changed here
 - **RunContext** - this is the new one. It's basically the rest of `DecoratedSource` wihen you take out the Source (so the "Decorated" part)

It made it possible to get rid of that whole bunch of static factory methods with all those overloads. `Source` now has just a couple of very straight-forward ones and `RunContext` the same.

It does mean that instead of passing 1 parameter there's now 2 parameters being passed in a bunch of places. But on the positive side a couple of those places are now more strictly typed (you can't pass a `JarSource` to `build()` for example, it doesn't make sense).

These big refactors probably aren't you making happy, but I promise I'm done now with this :-)

Edit: Oh, and no weird inheritance anymore. We just have Source->(ScriptSource|JarSource). The RunContext is a separate class. No shared interfaces or anything.


src/main/java/dev/jbang/RunContext.java
```diff
@@ -0,0 +1,259 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class RunContext {
```
maxandersen, line 11: lets add some basic javadoc ;)

src/main/java/dev/jbang/RunContext.java
```diff
@@ -0,0 +1,281 @@
...
+
+	public String getPreMainClass() {
+		return preMainClass;
+	}
+
+	public void setPreMainClass(String name) {
+		preMainClass = name;
+	}
+
+	public class SourceContext {
```
maxandersen, line 207: what is this thing? ..and should it not be static?

maxandersen: iread it as a AgentContext where source is representing an agent and then the runcontext for that agent ?  is that about right?

maxandersen (changes requested):
overall looks straight forward - looks like just a sensible reshuffling. only tihng I cant grok is the "sourcecontext" class.

Issue #478, opened 2020-11-04
Dependency cache can be wrong


dmlloyd:
Sometimes, particularly when handling snapshots, the dependency cache can be aggressively out of date.  It should compare timestamps or something to ensure that the cached dependency set still makes sense.

In conjunction with #475 this can cause some awkwardness.

The workaround is, if classpaths are coming up wrong, `rm ~/.jbang/cache/dependency_cache.json`.  Once #475 is fixed, `jbang cache clear` should do the job.

quintesse:
I also think a `--rebuild` option or something similar is still useful. The cache clear is bit of a scattergun approach.

maxandersen:
'jbang --fresh run' was my thinking to indicate to not do cache lookups but still cache the result. 

Instead of 'jbang --no-cache ...'

quintesse:
Fixed by https://github.com/jbangdev/jbang/pull/699

PR #699, opened 2021-02-04
feat: improved dependency handling

quintesse:
We now check much better if an app's dependencies are missing or out-of-date.


quintesse:
This now works really nicely @maxandersen :-)

maxandersen:
this fixed #478, correct?

maxandersen (approved):

quintesse:
> this fixed #478, correct?

Indeed!

PR #708, opened 2021-02-05
fix: jbang doesn't store additional deps in JAR Class-Path anymore

### LINKED ISSUE #707: jbang shouldn't store additional deps in JAR Class-Path
It doesn't for the additional classpath (as it should) but it does for the additional deps (it shouldn't).

quintesse:
Fixes #707

Dumb mistake. I should have double-checked.

If possible I'd do a quick-fix release @maxandersen . It's not _that_ bad , but I don't want people depending on this "feature" and then later complain it got broken :-)

To explain that better: right now you could do `jbang run --dep fu:bar:1.0 myfile.java`  and also `java -jar .../path/to/myfile.java.jar` and that would work.... but it shouldn't. The `--dep` is for adding dependencies at runtime, it shouldn't be "baked in".

But of course people might think it's very useful, you could have provided a whole bunch of deps at compile time but at runtime you don't have to anymore, yay! No! Not yay! :-)

It goes back to the idea of being able to use service loaders etc, it must be possible to specify deps at runtime for that.

Sorry for the long explanation for such a small PR, but I just wanted to make clear the "severity" of the issue.

### DIFF
--- a/src/main/java/dev/jbang/dependencies/DependencyUtil.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyUtil.java
@@ -15,6 +15,7 @@ import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.jboss.shrinkwrap.resolver.api.maven.ConfigurableMavenResolverSystem;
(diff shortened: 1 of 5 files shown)


maxandersen:
Test is failing because piccolo shows up twice.

quintesse:
> Test is failing because piccolo shows up twice.

Yes, this is fixed now.

PR #712, opened 2021-02-06
feat: Jbang now checks if a new version of itself is available

quintesse:
And tells the user and provides instruction for the update.

Nice little present for you @maxandersen :-)

### DIFF
--- a/itests/karate-config.js
+++ b/itests/karate-config.js
@@ -22,7 +22,8 @@ function fn() {
 
         !('JBANG_REPO' in env) && (env.JBANG_REPO = sc + sep + "karate-m2")
         !('JBANG_DIR' in env) && (env.JBANG_DIR = sc + sep + "karate-jbang")
-        
+        !('JBANG_NO_VERSION_CHECK' in env) && (env.JBANG_NO_VERSION_CHECK = "true")
+
         var proc = karate.fork({ redirectErrorStream: false, useShell: true, line: line, env: env});
         proc.waitSync();
         karate.set('out', proc.sysOut);
--- a/src/main/java/dev/jbang/Main.java
+++ b/src/main/java/dev/jbang/Main.java
@@ -12,7 +12,6 @@ public class Main {
 		CommandLine cli = Jbang.getCommandLine();
 		args = handleDefaultRun(cli.getCommandSpec(), args);
 		int exitcode = cli.execute(args);
-		((Jbang) cli.getCommandSpec().userObject()).checkLatestVersion();
 		System.exit(exitcode);
 	}
 
--- a/src/main/java/dev/jbang/Main.java
+++ b/src/main/java/dev/jbang/Main.java
@@ -12,6 +12,7 @@ public class Main {
 		CommandLine cli = Jbang.getCommandLine();
 		args = handleDefaultRun(cli.getCommandSpec(), args);
 		int exitcode = cli.execute(args);
+		((Jbang) cli.getCommandSpec().userObject()).checkLatestVersion();
 		System.exit(exitcode);
 	}
 
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -45,7 +45,7 @@ public class App {
 
 @CommandLine.Command(name = "install", description = "Install a script as a command.")
 class AppInstall extends BaseCommand {
-	private static final String jbangUrl = "www.jbang.dev/releases/latest/download/jbang.zip";
+	private static final String jbangUrl = "https://www.jbang.dev/releases/latest/download/jbang.zip";
 
 	@CommandLine.Option(names = {
 			"--native" }, description = "Enable native build/run")
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -45,7 +45,7 @@ public class App {
 
 @CommandLine.Command(name = "install", description = "Install a script as a command.")
 class AppInstall extends BaseCommand {
-	private static final String jbangUrl = "https://github.com/jbangdev/jbang/releases/latest/download/jbang.zip";
+	private static final String jbangUrl = "www.jbang.dev/releases/latest/download/jbang.zip";
 
 	@CommandLine.Option(names = {
 			"--native" }, description = "Enable native build/run")
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -224,7 +224,7 @@ class AppInstall extends BaseCommand {
 		if (force || !managedJbang) {
 			// Download Jbang and unzip to ~/.jbang/bin/
 			Util.infoMsg("Downloading and installing jbang...");
-			Path zipFile = Util.downloadAndCacheFile(jbangUrl, force);
+			Path zipFile = Util.downloadAndCacheFile(jbangUrl);
 			Path urlsDir = Settings.getCacheDir(Cache.CacheClass.urls);
 			Util.deletePath(urlsDir.resolve("jbang"), true);
 			UnpackUtil.unpack(zipFile, urlsDir);
--- a/src/main/java/dev/jbang/cli/Version.java
+++ b/src/main/java/dev/jbang/cli/Version.java
@@ -1,18 +1,38 @@
 package dev.jbang.cli;
 
+import java.io.IOException;
 import java.io.PrintWriter;
 
 import dev.jbang.Settings;
 import dev.jbang.util.Util;
+import dev.jbang.util.VersionChecker;
 
 import picocli.CommandLine;
 
 @CommandLine.Command(name = "version", description = "Display version info.")
 public class Version extends BaseCommand {
 
+	@CommandLine.Option(names = { "--check" }, description = "Check if a new version of jbang is available")
+	boolean checkForUpdate;
+
+	@CommandLine.Option(names = { "--update" }, description = "Update jbang to the latest version")
+	boolean update;
+
 	@Override
 	public Integer doCall() {
 		System.out.println(Util.getJbangVersion());
+
+		if (update) {
+			if (VersionChecker.updateOrInform(checkForUpdate)) {
+				try {
+					AppInstall.installJbang(true);
+				} catch (IOException e) {
+					throw new ExitException(EXIT_INTERNAL_ERROR, "Could not install command", e);
+				}
+			}
+		} else if (checkForUpdate) {
+			VersionChecker.checkNowAndInform();
+		}
 
 		if (isVerbose()) {
 			PrintWriter out = spec.commandLine().getOut();
@@ -20,6 +40,7 @@ public class Version extends BaseCommand {
 			out.println("Config: " + Settings.getConfigDir());
 			out.println("Repository:" + Settings.getLocalMavenRepo());
 		}
+
 		return EXIT_OK;
 	}
 }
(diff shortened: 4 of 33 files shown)


src/main/java/dev/jbang/util/VersionChecker.java
```diff
@@ -0,0 +1,124 @@
...
+import java.nio.file.Path;
+import java.nio.file.attribute.FileTime;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+
+import dev.jbang.Settings;
+
+public class VersionChecker {
+	private static final String jbangVersionUrl = "https://github.com/jbangdev/jbang/releases/latest/download/version.txt";
```
maxandersen, line 16: Please use `https://www.jbang.dev/releases/latest/download/version.txt` with follow redirect enabled.

then we get a poor-mans stat on checks.

src/main/java/dev/jbang/util/VersionChecker.java
```diff
@@ -0,0 +1,124 @@
...
+			} else {
+				Util.verboseMsg("jbang is up-to-date.");
+			}
+		}
+	}
+
+	// Determines and returns the latest Jbang version from GitHub
+	private static String retrieveLatestVersion() {
+		try {
+			Path versionFile = Util.downloadFile(jbangVersionUrl, Settings.getCacheDir().toFile());
```
maxandersen, line 74: what happens if no network available or url slow to respond ?

maxandersen (changes requested):
nice one - but fails in integration tests and lets change the url to something we can get stats on.

src/main/java/dev/jbang/util/VersionChecker.java
quintesse: ok, I didn´t know about that one. Shouldn't we then do that for downloads as well?

quintesse:
> but fails in integration test

Hmm that's the tests themselves failing, not because something is wrong. Is there an option to set an ENV var for all tests and have that block version checking? (So we'd check for that ENV var, let's say JBANG_NO_VERSION_CHECK, and never do any checks if it's set to "true").

src/main/java/dev/jbang/util/VersionChecker.java
quintesse: Not having a network is unnoticeable. A slow network would make Jbang seem to hang (but so would anything that tries to access the network).

Two options I can think of:

 - We could either do something in a separate thread, but I'm not sure I like that option because then the output could appear in unexpected places. (We could wait and always show the text at the end, but personally that feels wrong, if we show it, it should appear first)
 - Or we could perform the check in a place where we already know we're doing network access. For example in the `downloadAndCacheFile()` method. Then if it's slow then it's just slow for everything, one simple query isn't going to majorly change that. It does mean that if you never run anything that accesses the network then you'll never be notified of newer versions.

quintesse: Third option:

 - Keep it like it is now, but make sure the timeout is set to 2s or so.

Thought of this option because option 2 above is kind of intrusive. We have this nice clean code for downloading stuff and suddenly there's this version check in there. Just feels ugly.

maxandersen:
Should be doable in karate.Java.

src/main/java/dev/jbang/util/VersionChecker.java
maxandersen: I just created. I have 3 free rules. This is the last rule I have left :)

quintesse: Not sure I understand. You have 3 free rules. On is for version.txt. BUt what are the other 2 then?

maxandersen: Sh1.Jbang and ps1.jBang

quintesse: Ah yes. Hmmm. Does GitHub provide download number at least for the releases?

maxandersen: Yes. There are download stats per artifact.

maxandersen: just realized I could use `*` and `$1` to have it work for multiple files - so now `www.jbang.dev/releases/latest/download/jbang.zip` works too.

maxandersen: the benefit for these are that at least in theory we should be able to get the agent string on requests and get a bit more interesting data.

quintesse:
Updated the PR. Checking is now done async.

Btw, when running scripts there might still be a pause (max 2 secs) because of the way jbang passes off control to the java command that does the actual execution. But I don't think there's anything we can do to improve that. It will be a max 2s pause every 3 days. I think we can live with that? (People can still set the ENV var to disable checking altogether).

The code does take into account the offline state now. The flag for which has been made toplevel.
I also made the --fresh a toplevel flag, simplifying the cache handling a bit.

itests/karate-config.js
```diff
@@ -22,7 +22,8 @@ function fn() {
 
         !('JBANG_REPO' in env) && (env.JBANG_REPO = sc + sep + "karate-m2")
         !('JBANG_DIR' in env) && (env.JBANG_DIR = sc + sep + "karate-jbang")
-        
+        !('JBANG_NO_VERSION_CHECK' in env) && (env.JBANG_NO_VERSION_CHECK = "true")
```
maxandersen, line 25: since it is named JBANG_NO_VERSION_CHECK I think we should just check for presence so doing `export JBANG_NO_VERSION_CHECK` would be enough ?

src/main/java/dev/jbang/Main.java
```diff
@@ -12,6 +12,7 @@ public static void main(String... args) {
 		CommandLine cli = Jbang.getCommandLine();
 		args = handleDefaultRun(cli.getCommandSpec(), args);
 		int exitcode = cli.execute(args);
+		((Jbang) cli.getCommandSpec().userObject()).checkLatestVersion();
```
maxandersen, line 15: shouldn't we do this check before final result is done ? or is the intention to always show it - even for runs that reuse the cache?

src/main/java/dev/jbang/cli/Version.java
```diff
@@ -1,24 +1,46 @@
...
 
 import picocli.CommandLine;
 
 @CommandLine.Command(name = "version", description = "Display version info.")
 public class Version extends BaseCommand {
 
+	@CommandLine.Option(names = { "--check" }, description = "Check if a new version of jbang is available")
+	boolean checkForUpdate;
+
+	@CommandLine.Option(names = { "--update" }, description = "Update jbang to the latest version")
```
maxandersen, line 18: Should we have a dedicated 'update' command for symmetry with install ?

itests/karate-config.js
quintesse: Perhaps, I never trust my intuition on what `export JBANG_NO_VERSION_CHECK` will do :-)
The check itself in Jbang is pretty broad, anything that is not "false" (ignoring case) is considered to be "true".

src/main/java/dev/jbang/util/Util.java
```diff
@@ -427,22 +444,44 @@ public static Path downloadFileSwizzled(String fileURL, File saveDir) throws IOE
 	static private String getAgentString() {
 		if (agent == null) {
 			String version = System.getProperty("java.version");
-			agent = "JBang/" + BuildConfig.VERSION + " (" + System.getProperty("os.name") + "/"
+			agent = "JBang/" + getJbangVersion() + " (" + System.getProperty("os.name") + "/"
 					+ System.getProperty("os.version") + "/" + System.getProperty("os.arch") + ") " + "Java/" + version;
```
maxandersen, line 431: Should add Java.vendor too

itests/karate-config.js
quintesse: (which actually means that `export JBANG_NO_VERSION_CHECK` should just work fine)

maxandersen (changes requested):
The change to catalog would have been nicer in separate pr - but hey - it's awesome :)

Should change interval to once max 24 hrs and don't wait for result if not easy at end.

src/main/java/dev/jbang/Main.java
quintesse: It's actually misnamed (historical reasons), renamed it to `informLatestVersion()`. The (async) check was already started in the beginning, that line is just retrieving the result.

src/main/java/dev/jbang/util/Util.java
quintesse: WAnt me to put that in this PR?

src/main/java/dev/jbang/cli/Version.java
quintesse: Perhaps, but that means we probably need to have a way to detect that newer versions of installed scripts are available (HTTP HEAD requests?).

maxandersen: Actually, we don't have versions in scripts anyway. -_force can be used there to force update. 

Thus fine to have --update part of version for jbang

src/main/java/dev/jbang/util/Util.java
maxandersen: Feel free. I can also add it before releasing.

quintesse: I'll leave it to you then because you have a better idea of what it is you want exactly :-)

maxandersen:
running `jbang version --check` I get double output:

```
❯ jbang version --check
[jbang] There is a new version of jbang available!
[jbang] You have version 0.65.0.9 and 0.65.1 is the latest.
[jbang] Use your package manager to update jbang to the latest version,
[jbang] or visit https://jbang.dev to download and install it yourself.
[jbang] There is a new version of jbang available!
[jbang] You have version 0.65.0.9 and 0.65.1 is the latest.
[jbang] Use your package manager to update jbang to the latest version,
[jbang] or visit https://jbang.dev to download and install it yourself.
0.65.0.9
```

src/main/java/dev/jbang/cli/App.java
```diff
@@ -45,7 +45,7 @@ public static void deleteCommandFiles(String name) {
 
 @CommandLine.Command(name = "install", description = "Install a script as a command.")
 class AppInstall extends BaseCommand {
-	private static final String jbangUrl = "https://github.com/jbangdev/jbang/releases/latest/download/jbang.zip";
+	private static final String jbangUrl = "www.jbang.dev/releases/latest/download/jbang.zip";
```
maxandersen, line 48: no https:// ?

quintesse: DUnno, shouldn't it be https always nowadays?

src/main/scripts/container/Dockerfile
```diff
@@ -7,6 +7,7 @@ ADD ./entrypoint /bin/entrypoint
 
 ENV SCRIPTS_HOME /scripts
 ENV JBANG_VERSION @projectVersion@
+ENV JBANG_NO_VERSION_CHECK true
```
maxandersen, line 10: why setting it here ? that will make jbang github action invisible.

src/main/java/dev/jbang/cli/App.java
maxandersen: yes, thats what i'm asking ;) sholdn't it be "https://www.jbang../"

src/main/scripts/container/Dockerfile
quintesse: I don't understand what you're trying to say?

maxandersen (approved):

src/main/scripts/container/Dockerfile
maxandersen: this dockerfile is what jbang github action uses. not for testin

maxandersen:
merged to master

PR #714, opened 2021-02-07
fix: move to use mavencentral instead of jcenter

maxandersen:
Please delete this text, and add description of the topic solved by this PR.

To make releases as automated as possible we use conventional commits formats.
Thus commits and pull-requests titles should before merge follow the Conventional Commits spec.

Details in https://github.com/zeke/semantic-pull-requests

If in doubt then open the pull-request and we'll help you - Thank you!
-->

### DIFF
--- a/src/main/java/dev/jbang/dependencies/DependencyUtil.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyUtil.java
@@ -39,6 +39,7 @@ public class DependencyUtil {
 
 	static {
 		aliasToRepos = new HashMap<>();
+		aliasToRepos.put("mavencentral", "https://repo1.maven.org/maven2/");
 		aliasToRepos.put("jbossorg", "https://repository.jboss.org/nexus/content/groups/public/");
 		aliasToRepos.put("redhat", "https://maven.repository.redhat.com/ga/");
 		aliasToRepos.put("jcenter", "https://jcenter.bintray.com/");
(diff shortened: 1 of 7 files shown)


maxandersen:
the extra commit is due to missing merge from tag - will be fixed when merging this

src/main/java/dev/jbang/dependencies/DependencyUtil.java
```diff
@@ -39,6 +39,7 @@
 
 	static {
 		aliasToRepos = new HashMap<>();
+		aliasToRepos.put("mavencentral", "https://repo1.maven.org/maven2/");
 		aliasToRepos.put("jbossorg", "https://repository.jboss.org/nexus/content/groups/public/");
 		aliasToRepos.put("redhat", "https://maven.repository.redhat.com/ga/");
 		aliasToRepos.put("jcenter", "https://jcenter.bintray.com/");
```
quintesse, line 45: this one you're keeping for now?

maxandersen: Yeah. Don't break those who used it explicitly.

quintesse: Yeah. Should we perhaps show a warning to people using it?

maxandersen: not sure where to best put that - we can do that in separate PR if we really feel for it.

quintesse: Perhaps right where we parse //REPOS?

PR #728, opened 2021-02-15
fix: Second runs where special flags are ignored

quintesse:
Flags like //JAVA were ignored while //JAVA_OPTIONS were not
even stored in the JAR like they should. Both problems have
been fixed.


maxandersen:
cool you found it ! did you manage to have test for it so we don't get caught again ?

quintesse:
@maxandersen yes, I made a change to an existing test that will trigger at least the lack of options. Not for the //JAVA though because I'm not sure how to easily test that.

maxandersen:
merged on command line.

PR #738, opened 2021-02-18
fix: Using correct backingResoure for JarSources

### LINKED ISSUE #734: jfr saves to cache not local file
`jbang --jfr myapp.java` no longer saves to `myapp.jfr` but not to some file inside the cache.

quintesse:
A JarSource that is created from a ScriptSource should have the
same backingResource as the original ScriptSource.

Fixes #734


maxandersen:
that makes sense!

PR #747, opened 2021-02-21
fix: jbang-native setup on Windows

### LINKED ISSUE #746: Instructions for using jbang to install jbang don't work on Windows
From Git Bash on Windows, following these instructions:  https://github.com/jbangdev/jbang#using-jbang 

> curl -Ls https://sh.jbang.dev | bash -s - app setup

it fails with:
> bash: line 70: <....home...>/jbang.cmd: No such file or directory

which seems to make sense... I haven't installed jbang yet.      

There's a bunch of other ways to install that work, but since this was called out as the simplest, probably worth fixing.

quintesse:
Was originally tested in an unclean test environment which hid
the problem of performing a zero-install run on Windows when
using Cygwin/GitBash/etc

Fixes #746

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -39,7 +39,7 @@ download() {
   fi
 }
 
-abs_jbang_path=$(resolve_symlink $(absolute_path $0))
+abs_jbang_dir=$(dirname $(resolve_symlink $(absolute_path $0)))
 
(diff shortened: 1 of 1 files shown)


quintesse:
Tested on Windows (in a clean environment this time) and it seems to work fine.

maxandersen (approved):

PR #755, opened 2021-02-24
Add support for running jbang on AIX (also removes GNU tar dependency)

sxa:
This PR (Only about two months after I told @maxandersen I'd do it!) adds support for IBM's AIX:
- Detection of `os` and `arch` variables
- Removal of dependency on GNU tar for `--strip-components` parameter (Similar logic to macos is now used everywhere

Alternatives considered:
- Detect whether GNU tar is present and use the `--strip-components` parameter if it is, otherwise use the alternate mechanism
- Using a weird regex to see if `uname -m` is a hex string and if so assume AIX
- Use uname -p on AIX to set `arch`

Also worth noting that this does still rely on `readline` which isn't standard on AIX but can be obtained from e.g. the `coreutils` package on the AIX toolbox

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -48,6 +48,8 @@ case "$(uname -s)" in
     os=mac;;
   CYGWIN*|MINGW*|MSYS*)
     os=windows;;
+  AIX)
+    os=aix;;
   *)
     os=
 esac
@@ -62,7 +64,12 @@ case "$(uname -m)" in
   armv7l)
     arch=arm;;
   *)
-    arch=
+    ## AIX gives a machine ID for `uname -m` but it only supports ppc64
+    if [ "$os" = "aix" ]; then
+      arch=ppc64
+    else
+      arch=
+    fi;;
 esac
 
 ## when mingw/git bash or cygwin fall out to just running the bat file.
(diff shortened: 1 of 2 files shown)


src/main/scripts/jbang
```diff
@@ -61,7 +63,12 @@ case "$(uname -m)" in
   armv7l)
     arch=arm;;
   *)
-    echo "Unsupported Architecture: $(uname -m)" 1>&2; exit 1;;
+    ## AIX gives a machine ID for `uname -m` but it only supports ppc64
```
quintesse, line 66: Are you saying that on AIX `uname -m` does not return "ppc64"?

quintesse: I think I've found the answer already online: no it doesn't. It seems to be a long list of possible version numbers.

quintesse: https://sites.ualberta.ca/dept/chemeng/AIX-43/share/man/info/C/a_doc_lib/cmds/aixcmds5/uname.htm

sxa: Correct - `uname -p` does though

sxa:
Moving back to draft until I can test after a rebase

maxandersen:
That's hilarious :) awesome!

src/main/scripts/jbang
quintesse: So should we be using `uname -p` or is there some other hardware where that won't work?

quintesse: NB: Git Bash on WIndows returns "unknown" for `-p`

maxandersen (approved):
Looks good to me but I'll let it be up to @quintesse as he did the multiplatform support the most in past.

quintesse (approved):
LGTM 👍

src/main/scripts/jbang
sxa: Yeah there isn't a sensible standard choice to use universally unfortuantely (dealt with this many times in the past - here's one example which had more options: https://github.com/sxa/qinstall/blob/26f3e7e4cc4993596da416f32ffbea1138b29820/qinstall#L34 - I did use `uname -p` on AIX there) which is why I figured just basing it on the OS name would be a reasonable solution for jbang (for now ...)

sxa:
What's the recommended way to test jbang's java vs shell JDK download implementations explicitly instead of letting it use what it wants?

src/main/scripts/jbang
quintesse: Indeed, I had actually considered a solution where we'd just use both the value of -m and -p and if one of those matches the pattern it would be okay. Concatenating the two values would work as well indeed.

quintesse:
> What's the recommended way ...

That depends a bit on what it is you want to do. But the easiest way would be to just set the JAVA_HOME to that JDK. It will then use that one as long as a) no other version is requested on the command line (using `--java`) and b) no other version is specified inside the .java file (using `//JAVA`).

src/main/scripts/jbang
sxa: Annoyingly `uname -p` returns `powerpc` on AIX so while I could change it to say `arch="$(uname -p)"` within the AIX section it would still have to be translated to get the correct name for the adoptopenjdk download - so we'd likely end up with `arch="$(uname -p | sed s/powerpc/ppc64/)"` in the AIX case. I can change to that if you'd prefer it over a fixed `arch=ppc64`

quintesse: No, don't worry :-)

maxandersen:
@all-contributors please add @sxa  for code

PR #756, opened 2021-02-24
Add support for Linux/ppc64le and Linux/s390x

sxa:
Support a couple more Linux architectures which can be pulled from adoptopenjdk :-)

Have verified on one ppc64le and one s390x system and both were able to download a JDK with `jbang jdk install 11`. Not sure what the best way is of testing both shell and java download methods though.

Signed-off-by: Stewart X Addison <sxa@redhat.com>


sxa:
Moving back into draft until I can test following a rebase

quintesse:
Nice. For full support we'll also need some changes to the `jbang` startup script. That way Jbang should also work when no JDK is currently available on the system.

sxa:
> Nice. For full support we'll also need some changes to the `jbang` startup script. That way Jbang should also work when no JDK is currently available on the system.

The PR already contains the changes to the `jbang` startup script - are they incomplete? I have teested in the non-java case.

quintesse:
I could swear that yesterday it showed me only the changes in the Java code, and I thought I double-checked. But you're absolutely right those changes are there too :-)

quintesse (approved):
LGTM

maxandersen:
@all-contributors please add @sxa for code

PR #758, opened 2021-02-25
fix: Now correctly handles out-of-date dependencies

### LINKED ISSUE #757: Still problems detecting out-of-date dependencies
When we try to run a Maven artifact that hasn't been released yet (it has a -SNAPSHOT version) and it gets a new dependency Jbang will not correctly handle that change.

It shows that it has detected that it needs to resolve the dependencies again, but in the end it doesn't do this correctly and the new dependency doesn't get added to the classpath, nor to the dependency cache.

quintesse:
Updating the dependency cache now correctly only writes out the
information that was actually retrieved and updated. Before the
code would always write out the latest timestamps, updating them
for *all* files in the cache, even the ones we weren't even
retrieving. Because resolving dependencies is a multi-step
process this basically ruined the information that was needed
for the following steps.

Fixes #757

### DIFF
--- a/src/main/java/dev/jbang/dependencies/ArtifactInfo.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactInfo.java
@@ -30,8 +30,12 @@ public class ArtifactInfo {
 		return coordinate;
 	}
 
-	public File asFile() {
+	public File getFile() {
 		return file;
+	}
+
+	public long getTimestamp() {
+		return timestamp;
 	}
 
 	public boolean isUpToDate() {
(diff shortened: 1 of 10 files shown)


quintesse:
Btw I didn't add any tests because it's a bit hard to do. We'd need to have two identically named Maven artifacts with a different number of dependencies and with different file timestamps.

quintesse:
Btw I also renamed `asFile()` to `getFile()` because having to `getXxx()` methods and one `asXxx()` method, while all were basically just getters felt a bit weird.

quintesse:
Added a second commit that assures that `--fresh` also updates the dependency cache.

maxandersen (approved):

PR #760, opened 2021-03-01
Fix quoting in jbang startup scripts

quintesse:
Today while using Jbang on Windows I found out that the hand-off from the bash script (running in Git Bash) to the .cmd file wasn't taking into account quoted arguments.

So something like `jbang hello@jbangdev one "two three"` would display:

```
Hello one
Hello two
Hello three
```

After fixing that I went on to check if other hand-offs had similar problems and indeed, running `curl -Ls https://sh.jbang.dev | bash -s - hello@jbangdev one "two three"` gave the same problem (both on Linux as on Windows).

So the second commit fixes that.

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -101,7 +101,8 @@ else
     rm -f "$JBDIR/bin/jbang" "$JBDIR/bin"/jbang.*
     cp -f "$TDIR/urls/jbang/bin"/* "$JBDIR/bin"
   fi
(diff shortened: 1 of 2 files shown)


quintesse:
This is probably simple enough for me to go ahead and merge, right @maxandersen ? Here goes ...

quintesse:
Ha, I can't actually.

maxandersen:
Doh. I'll fix that when back from lunch

maxandersen (approved):

PR #767, opened 2021-03-06
fix: make git urls work with jitpack again

maxandersen:
HEAD-SNAPSHOT does not seem to work with jitpack thus moving to use -SNAPSHOT instead

Fixes #766


quintesse (approved):
LGTM

PR #769, opened 2021-03-07
feat: add mavenrepo export option

maxandersen:
this adds a `--mavenpublish` which will generate .jar / .pom
with group, artifact and version.

draft, as for now im not happy with the defaults of group and version and that i must use --fresh as otherwise 
pom.xml generation fails every second time - somehow classpath are not accessible.

### DIFF
--- a/src/main/java/dev/jbang/source/ScriptSource.java
+++ b/src/main/java/dev/jbang/source/ScriptSource.java
@@ -506,6 +506,11 @@ public class ScriptSource implements Source {
 		return Stream.concat(func.apply(this).stream(), subs).collect(Collectors.toList());
 	}
 
+	@Override
+	public boolean isCreatedJar() {
+return getJarFile().exists();
+	}
+
 	public static ScriptSource prepareScript(String resource) {
 		ResourceRef resourceRef = ResourceRef.forResource(resource);
 		return prepareScript(resourceRef);
--- a/src/main/java/dev/jbang/source/ScriptSource.java
+++ b/src/main/java/dev/jbang/source/ScriptSource.java
@@ -508,7 +508,7 @@ public class ScriptSource implements Source {
 
 	@Override
 	public boolean isCreatedJar() {
-return getJarFile().exists();
+		return getJarFile().exists();
 	}
 
 	public static ScriptSource prepareScript(String resource) {
--- a/itests/export.feature
+++ b/itests/export.feature
@@ -2,13 +2,13 @@ Feature: export
 
 Scenario: basic export no classpath
   When command('rm helloworld.jar')
-  When command('jbang export helloworld.java')
+  When command('jbang export local helloworld.java')
   Then match err contains "Exported to"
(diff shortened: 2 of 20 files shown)


maxandersen:
jitpack using this as jbang wrapper is here https://jitpack.io/p/maxandersen/jbang-to-maven/

maxandersen:
Failure on windows is bug where I use / instead of path separator. Easy fix

quintesse:
Is "mavenpublish" a good name? It doesn't actually do any publishing, right? Perhaps "mavenpackage"?

maxandersen:
Technically it is a publish to a local maven repo. 

But yeah I'm not committed to the name. 

Mainly coming from "publishToMavenLocal" referenced in https://jitpack.io/docs/BUILDING/

But without requiring it to be in maven local.

maxandersen:
Just realized in jitpack.yml
We can just copy the output of this export to ie. M2/repository.

Should we make the default just export the artifacts to the current dir (still in right group/artifact versioned dir) as the other export styles or should we keep it where it generated by default into whatever local maven repo is configured to use ?

quintesse:
In that case the "mavenpackage" might be an even better name, it would just generate the correct structure locally

maxandersen:
refactor/redid this so it is now subcommands.

still need polishing and to see if somehow can avoid breaking existing export calls.

feel free to look and comment ..but still probably not good ..need sleep first :)

maxandersen:
cleaned up - this will break anyone who used jbang export before but couldn't find a way to support both old and new syntax without adding a lot of ambiguity

but otherwise just a pr that moves from -- flags to commands: local, portable, mavenrepo

maxandersen:
note, could probably be cleaned up further - main refactor just made build methods static so can be called from "anywhhere" and started to use mixins to share properties rather than locked into inheritance.
would have used subcommand methods  but our testing infratsructure assumes commands has a doCall method so kept the more verbose class usage.

src/main/java/dev/jbang/source/ScriptSource.java
```diff
@@ -506,6 +506,11 @@ private void collectAllSources(Set<ResourceRef> refs, List<ScriptSource> scripts
 		return Stream.concat(func.apply(this).stream(), subs).collect(Collectors.toList());
 	}
 
+	@Override
+	public boolean isCreatedJar() {
```
quintesse, line 510: Not sure if "created" is completely correct. It could already exist from a previous build. It just exists and from that you _could_ infer that it has been built because otherwise we'd have taken the JarSource code path.
Also, the fact that JarSource returns false feels counter-intuitive.

maxandersen: Yes the name sucks. Not sure what to use for "created just now" :)

If there is a existing jar it becomes a jarsource hence why it returns false. 

The use is for CDs where the jdk stupidly enough can't figure out to handle by itself if the jar is preexisting or not.

PR #780, opened 2021-03-16
chore: mention -ti with docker command for those who might need console input

brunoborges:


maxandersen:
it is not technically needed to have `-ti` but I can see how it can catch you by surprise command input is not possible - so lets merge it :)

PR #781, opened 2021-03-16
chore: remove gradle 7 deprecations to allow testing using gradle 7

maxandersen:


maxandersen:
this was handled elsewhere.

PR #787, opened 2021-03-18
fix: Not duplicating or misquoting java options anymore

### LINKED ISSUE #786: Improper JAVA_OPTIONS handling
Issue #782 shows that there is some problem with the handling of JAVA_OPTIONS:

```
[jbang] run: /opt/data/sfw/tools/sdkman/candidates/java/current/bin/java '-Djava.util.logging.manager=org.jboss.logmanager.LogManager -Djava.util.logging.manager=org.jboss.logmanager.LogManager' -classpath ... '-Djava.util.logging.manager=org.jboss.logmanager.LogManager -Djava.util.logging.manager=org.jboss.logmanager.LogManager' io.quarkus.runner.GeneratedMain
```

The options appear quadrupled!

The main reason we're seeing this is because the script activates the Quarkus integration which causes a) the duplication of the options and b) making the options appear both at the start of the command as well as the end.

But it also shows a second problem where multiple options are quoted as if they were a single argument.

quintesse:
Fixes #786

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -283,7 +283,7 @@ public abstract class BaseBuildCommand extends BaseScriptDepsCommand {
 				throw new ExitException(1, e);
 			}
 		}
-		ctx.setPersistentJvmArgs(integrationResult.javaArgs);
+		ctx.setRuntimeOptions(integrationResult.javaArgs);
 		createJarFile(src, ctx, tmpJarDir, outjar);
 		createdJar = true;
 		return integrationResult;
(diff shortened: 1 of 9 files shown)


src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -329,8 +329,10 @@ static void createJarFile(ScriptSource src, RunContext ctx, File path, File outp
 			}
 		}
 
-		String runtimeOpts = Stream	.concat(src.getRuntimeOptions().stream(), ctx.getPersistentJvmArgs().stream())
-									.collect(Collectors.joining(" "));
+		// When persistent JVM args are set they override any runtime options set on the
```
maxandersen, line 332: are they no longer persistent ?

quintesse: yes, but all runtime options are always persistent, so it wasn't really necessary to give them different names

Issue #806, opened 2021-03-24
v0.69.1 breaks compatibility with JDK 8.


michael-simons:
Given `demo.java`

```java
public class demo {
    public static void main(String...a) {
        System.out.println("Hallo");
    }
}
```

Running

```
docker run -v `pwd`:/ws --workdir=/ws jbangdev/jbang-action:0.69.1 --java 8 demo.java
```

fails with

```
jbang --java 8 demo.java
[jbang] Building jar...
[jbang] Downloading JDK 8. Be patient, this can take several minutes...
[jbang] Installing JDK 8...
[jbang] Default JDK set to 8
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: demo has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:601)
```

Where as `docker run -v `pwd`:/ws --workdir=/ws jbangdev/jbang-action:0.68.0 --java 8 demo.java` say

```
[jbang] Downloading JDK 8. Be patient, this can take several minutes...
[jbang] Installing JDK 8...
[jbang] Default JDK set to 8
[jbang] Building jar...
[jbang] There is a new version of jbang available!
[jbang] You have version 0.68.0 and 0.69.1 is the latest.
[jbang] Use your package manager to update jbang to the latest version,
[...]

PR #807, opened 2021-03-24
fix: build and runtime not following same java version. fixes 806

maxandersen:


maxandersen:
fixes #806

maxandersen:
not enough - //JAVA 16 inside a java file still gets ignored.

maxandersen:
ok, i know what the error is - need sleep before i can fix it.

for now if using --java or //java feature best to stay on previous versions.

will get upodate out shortly

PR #817, opened 2021-03-28
feat: //FILES included in ide project Fixes #749

### LINKED ISSUE #749: //FILES should be mapped in as files under src in ide project

maxandersen:
needed this for qute templates and realized they were not editable....seem to easy but I think this is
all it takes :)

in draft as still need tests ...


maxandersen:
tests added and fixed the issue it found :)

quintesse (approved):
LGTM

PR #821, opened 2021-03-29
fix: Invalid comment in jbang.cmd

quintesse:
Small but very wrong typo


quintesse:
Still can't merge stuff by myself @maxandersen , will wait for your review :-)

maxandersen (approved):
lgtm - surprised how that could pass integration tests?

maxandersen (commented):
lgtm

quintesse:
> how that could pass integration tests?

Because errors in .cmd files are not fatal, so it's just this extra message that appears in the output

PR #822, opened 2021-03-29
docs: fix usages of app subcommand

aalmiray:
`jbang list` and `jbang uninstall` require the use of the `app` command.


quintesse:
@aalmiray if you could also fix the typo of "uinstall" at the same time that would be great :-)

quintesse (dismissed):
LGTM

quintesse:
Also if you could prefix the commit itself with "docs: " that would be even greater.

(Btw, I'm asking because I can okay and merge your work, but I can't okay my own work 🙄 . If not @maxandersen would have to do it)

quintesse (approved):
LGTM

quintesse:
Thanks @aalmiray ! 👍

PR #839, opened 2021-04-21
feat: Support Alpine Linux

### LINKED ISSUE #838: Support Alpine Linux
Alpine support is now available at AdoptOpenJDK so we should now be able to make this work.

There are several issues though:
 - [x] We have to set the minimum Java version to 16 for Alpine
 - [x] AdoptOpenJDK uses a special "OS name" to distinguish Alpine from any other Linux
 - [ ] Alpine has no `bash` by default
 - [ ] Alpine has no `curl` by default and the `wget` is has is brain dead

We can easily do the first two items but I think we should just expect people to install bash and curl, at least for now, we can alwys try to revisit in the future.

quintesse:
Fixes #838

### DIFF
--- a/src/main/java/dev/jbang/Settings.java
+++ b/src/main/java/dev/jbang/Settings.java
@@ -5,6 +5,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 
 import dev.jbang.catalog.Catalog;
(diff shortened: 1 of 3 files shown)


maxandersen:
Ha. That's a nice "workaround" :)
Great!

PR #852, opened 2021-05-01
Add quarkus config property for swagger-ui in template

ledinscak:
Without setting configuration property "quarkus.swagger-ui.always-include=true" feature swagger-ui is not installed and path /q/swagger-ui is not build.
This is taken from quarkus tutorial https://quarkus.io/guides/scripting


ledinscak:
making new with semantic commit

PR #853, opened 2021-05-01
Add quarkus config property for swagger-ui in template

ledinscak:
Without setting configuration property "quarkus.swagger-ui.always-include=true" feature swagger-ui is not installed and path /q/swagger-ui is not build.
This is taken from quarkus tutorial https://quarkus.io/guides/scripting


maxandersen:
What are you after here ?

To get swagger ui run with -Dquarkus.dev to get same behavior as mvn quarkus:dev ?

ledinscak:
Yes, you are right, thanks for pointing that out :-) Im closing PR

PR #855, opened 2021-05-05
fix: Don't use `eval` in app scripts

quintesse:
It messes with arguments containing spaces

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -133,7 +133,7 @@ class AppInstall extends BaseCommand {
 	private static void installShellScript(Path file, String scriptRef, boolean benative) throws IOException {
 		List<String> lines = Arrays.asList(
 				"#!/bin/sh",
-				"eval \"exec jbang run" + (benative ? " --native " : " ") + scriptRef + " $*\"");
+				"exec jbang run" + (benative ? " --native " : " ") + scriptRef + " \"$@\"");
 		Files.write(file, lines, StandardOpenOption.TRUNCATE_EXISTING, StandardOpenOption.CREATE);
 		setExecutable(file);
 	}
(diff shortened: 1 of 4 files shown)


quintesse:
The tests are failing because of an unrelated issue. This URL isn't accessible (anymore?): https://bintray.com/cardillo/maven/download_file?file_path=joinery%2Fjoinery-dataframe%2F1.9%2Fjoinery-dataframe-1.9-jar-with-dependencies.jar

Not exactly sure what to do about that (besides adding `@Ignore` :-) )

src/main/java/dev/jbang/cli/App.java
```diff
@@ -133,7 +133,7 @@ private static void installScripts(String name, String scriptRef, boolean benati
 	private static void installShellScript(Path file, String scriptRef, boolean benative) throws IOException {
 		List<String> lines = Arrays.asList(
 				"#!/bin/sh",
-				"eval \"exec jbang run" + (benative ? " --native " : " ") + scriptRef + " $*\"");
+				"exec jbang run" + (benative ? " --native " : " ") + scriptRef + " \"$@\"");
```
maxandersen, line 136: so it looks okey but exec creates a process where eval is in-process - does that change things if we have a long running java app...will ctrl-c and similar still kill the process or can we end up with detached processes?

quintesse: Isn't `exec` used to replace the current process? So shouldn't that handle ctrl+Cs just fine? 🤔

quintesse: I've done some simple testing and ctrl+c seems to work without any problems.

maxandersen:
> Not exactly sure what to do about that (besides adding `@Ignore` :-) )

jcenter/bintray death caused it. you can fix it by just taking any other jar hosted by https.

even just pick one from https://search.maven.org :)

maxandersen (changes requested):
windows test issue should be fixed and readme.adoc refer to the old form so would be good to fix that too.

quintesse:
CI passed

PR #856, opened 2021-05-06
chore: Alias list again shows @ catalog where appropriate

quintesse:
The `alias list` command wasn't showing the `@catalogname` in the appropriate circumstances giving the impression that you could run the alias simply by using it's name (which in case of the implicit catalogs isn't the case).

### DIFF
--- a/src/main/java/dev/jbang/source/ResourceRef.java
+++ b/src/main/java/dev/jbang/source/ResourceRef.java
@@ -313,9 +313,9 @@ public class ResourceRef implements Comparable<ResourceRef> {
 
 	private static ResourceRef fetchFromURL(String scriptURL) {
 		try {
-			scriptURL = swizzleURL(scriptURL);
-			Path path = Util.downloadAndCacheFile(scriptURL);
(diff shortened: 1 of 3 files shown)


src/main/java/dev/jbang/source/ResourceRef.java
```diff
@@ -313,8 +313,8 @@ private static ResourceRef fetchScriptFromUntrustedURL(String scriptURL) {
 
 	private static ResourceRef fetchFromURL(String scriptURL) {
 		try {
-			scriptURL = swizzleURL(scriptURL);
-			Path path = Util.downloadAndCacheFile(scriptURL);
+			String url = swizzleURL(scriptURL);
```
maxandersen, line 316: this breaks tests.

maxandersen: testFindDependenciesWithURLInsource to be precise.

maxandersen: fixing - seems like a bad rebase caused it.

PR #858, opened 2021-05-06
Implemented all items from issue #854

### LINKED ISSUE #854: Make `catalog` commands more like `alias`
The two commands have "grown apart" somewhat:

- [x] Right now the `catalog add` command still requires 2 arguments, a name and a url or path, but `alias add` was changed to use an optional `--name` flag. We should do the same for `catalog add`.
- [x] `catalog list` doesn't have a `--show-origin` flag like `alias list` has. It would perhaps be nice if it did.
- [x] `catalog list` also doesn't have colored output
- [x] docs

quintesse:
Fixes #854

### DIFF
--- a/itests/catalog.feature
+++ b/itests/catalog.feature
@@ -1,29 +1,29 @@
 Feature: catalog testing
 
 Scenario: java catalog list
-  When command('jbang catalog add --global tc jbang-catalog.json')
+  When command('jbang catalog add --global --name tc jbang-catalog.json')
   And command('jbang catalog list')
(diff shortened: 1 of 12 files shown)


maxandersen:
fails on windows so not merging yet.

maxandersen (changes requested):
windows test issue should be fixed and readme.adoc refer to the old form so would be good to fix that too.

maxandersen:
damn - wrong issue i commented on ;)

quintesse:
CI passed

PR #876, opened 2021-05-22
feat: Kotlin support

evanchooly:
making a draft PR so it easier to see and talk about


build.gradle
```diff
@@ -44,6 +44,8 @@ dependencies {
 	//implementation 'org.apache.maven:maven-aether-provider:3.0.5'
 	//implementation 'com.google.guava:guava:28.2-jre'
 
+	implementation 'org.jetbrains.kotlin:kotlin-compiler:1.5.0'
```
maxandersen, line 47: Do we need these when we are downloading the kotlin tools ?

Im very eery on growing the size of jbang itself if we can avoid it. 

This change moves it from 5.9mb to 7.7mb ... can it be avoided ?

maxandersen (changes requested):

build.gradle
maxandersen: looks like the *only* thing these are used for is @NotNull and a default string version for kotlin. That isn't justifiable; but also super easy to fix by just removing them.

src/main/java/dev/jbang/util/Util.java
```diff
@@ -67,6 +69,7 @@
 			"^.*(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*",
 			Pattern.MULTILINE);
 
+	private static final List<String> EXTENSIONS = asList(".java", ".jsh", ".kt");
```
maxandersen, line 72: ".kts" is what kscript defaults to so we so should consider adding that too so you can do:

" jbang https://github.com/holgerbrandl/kscript/blob/master/examples/classpath_example.kts"

maxandersen: ah - seem there is a little more to support .kts ...at least renaming from .kts to .kt isn't enough :)

src/main/java/dev/jbang/source/ScriptSource.java
```diff
@@ -16,22 +39,14 @@
...
-import dev.jbang.cli.BaseCommand;
-import dev.jbang.cli.ExitException;
-import dev.jbang.dependencies.DependencyUtil;
-import dev.jbang.dependencies.Detector;
-import dev.jbang.dependencies.MavenRepo;
-import dev.jbang.dependencies.ModularClassPath;
-import dev.jbang.util.JavaUtil;
-import dev.jbang.util.PropertiesValueResolver;
-import dev.jbang.util.Util;
+import static dev.jbang.cli.BaseBuildCommand.createJarFile;
```
maxandersen, line 48: we shouldn't have references from base back to cli package. Should move those out.

src/main/java/dev/jbang/source/ScriptSource.java
```diff
@@ -16,22 +39,14 @@
...
+import static dev.jbang.cli.BaseBuildCommand.resolveInJavaHome;
```
maxandersen, line 49: same as above.

maxandersen:
I haven't looked into all in full detail but this does feel like the right direction so let's see if we can get it cleaned up so things are not (too) cyclic and add more tests - Happy to release kotlin as experimental early so we can always iterate from there.

maxandersen:
@quintesse you should give this a look :)

maxandersen:
whats the simplest way to test this? I get this for now:

```
 jbang init -t kotlin kotlin.kt
[jbang] File initialized. You can now run it with 'jbang kotlin.kt' or edit it using 'jbang edit --open=[editor] kotlin.kt' where [editor] is your editor or IDE, e.g. 'vi'
❯ jbang kotlin.kt
[jbang] Building jar...
/Users/max/code/personal/jbangdev/jbang/build/install/jbang/bin/jbang: line 186: exec: originalResource: not found
❯ jbang --verbose kotlin.kt
[jbang] jbang version 0.70.0.14
[jbang] run: /Users/max/.sdkman/candidates/java/current/bin/java -classpath /Users/max/.jbang/cache/jars/kotlin.kt.364999e9df120025a4ca69dc4abfe52c631fa38c3669e6653532179224eb1a1e.jar KotlinKt
/Users/max/code/personal/jbangdev/jbang/build/install/jbang/bin/jbang: line 186: exec: originalResource: not found
```

src/main/java/dev/jbang/Cache.java
```diff
@@ -11,7 +11,7 @@
 public class Cache {
 
 	public enum CacheClass {
-		urls, jars, jdks, projects, scripts, stdins, deps
```
maxandersen, line 14: are `kotlins` any different than `jars` ?

src/main/java/dev/jbang/source/ScriptSource.java
```diff
@@ -521,6 +707,12 @@ public static ScriptSource prepareScript(String resource) {
 	}
 
 	public static ScriptSource prepareScript(ResourceRef resourceRef) {
-		return new ScriptSource(resourceRef);
+		String originalResource = resourceRef.getOriginalResource();
+		System.out.println("originalResource = " + originalResource);
```
maxandersen, line 711: this system out breaks jbang with error as its executed.

src/main/java/dev/jbang/Cache.java
evanchooly: they get expanded out like jdks do for `kotlinc` access, etc.

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: i tend to agree but i was aiming for "working" and "general approval" before massive disruptions to code org.  :)

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: stupid debug.  one moment.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -137,161 +126,7 @@ static Source build(ScriptSource src, RunContext ctx) throws IOException {
 		return result;
 	}
 
-	// build with javac and then jar... todo: split up in more testable chunks
-	static private IntegrationResult buildJar(ScriptSource src, RunContext ctx, File tmpJarDir, File outjar,
```
quintesse, line 141: Why is this refactor necessary? (see also below)

src/main/java/dev/jbang/util/Util.java
```diff
@@ -226,7 +229,11 @@ public static String unkebabify(String name) {
 		if (name.endsWith(".sh")) {
 			name = name.substring(0, name.length() - 3);
 		}
-		if (!(name.endsWith(".java") || name.endsWith(".jsh"))) {
+		boolean valid = false;
+		for (String extension : EXTENSIONS) {
+			valid |= name.endsWith(extension);
+		}
+		if (!valid) {
```
evanchooly, line 236: honestly this whole method feels sketchy to me but I didn't want to change it as it's technically (mostly) unrelated to the kotlin support.

src/main/java/dev/jbang/source/ScriptSource.java
```diff
@@ -86,6 +101,181 @@ private ScriptSource(ResourceRef resourceRef, String content) {
...
+	protected Predicate<ClassInfo> getMainFinder() {
+		return pubClass -> pubClass.method("main", BaseBuildCommand.STRINGARRAYTYPE) != null;
+	}
+
+	protected String getExtension() {
+		return ".java";
+	}
+
+	// build with javac and then jar... todo: split up in more testable chunks
+	public IntegrationResult buildJar(RunContext ctx, File tmpJarDir, File outjar,
```
quintesse, line 121: Why is this refactor necessary? The `ScriptSource` is meant as a data class to hold all the information that could be extracted from a source file, it shouldn't contain code that "does stuff".

src/main/java/dev/jbang/source/KotlinScriptSource.java
```diff
@@ -0,0 +1,45 @@
+package dev.jbang.source;
```
quintesse, line 1: I'm not per se against introducing subclasses of `ScriptSource`, but we should then take it to its logical conclusion and introduce `JavaScriptSource` as well.

quintesse (changes requested):
The refactor of `buildJar()` is the part I'd like to see undone, it goes against the spirit of a major refactor we did recently to split the code that analyses scripts from the code that does the building and the running. (It's still not perfect but this PR would move us farther away from that goal instead of bringing us closer to it :-) )

build.gradle
evanchooly: yeah.   old code and gone now.

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: moved some stuff around for you to check out

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: moving `resolveInJavaHome` to `JdkManager`

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: this is gone now

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: because this process is different for kotlin projects (or any other lang, e.g., scala) and static methods are fiendishly difficult/impossible specialize whereas with inheritance you just need a few simple overloads, like i did in this diff, to customize things while reusing the bulk of the code.  Static methods, in general, make for lousy extensibility.

maxandersen:
@quintesse i missed that. What would be the alternative - we introduce a separate "buildtools" mechanism that gets passed scriptsources or what were your thinking ?

quintesse:
@maxandersen basically exactly what I was thinking, yes :-)

That way we don't even need a `KotlinScriptSource` so the current design stays the same, we'd just add a new "builder" with its own implementations. Using a factory method that gets passed a `ScriptSource` it can return the appropriate builder.

src/main/java/dev/jbang/Cache.java
maxandersen: Ah. Sure.

maxandersen: Is it called kotlin development kits ? So kdks?

src/main/java/dev/jbang/source/ScriptSource.java
maxandersen: yes - we should just have a separate set of classes/hiearchy for the tools. those can do what is needed.

evanchooly:
Should I do that as part of this PR or should we separate the two?  I'm happy to do the work but don't want to overstep.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
maxandersen: i pushed commit that "undoes" most of it.

src/main/java/dev/jbang/source/KotlinScriptSource.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class KotlinScriptSource extends ScriptSource {
+
+	protected KotlinScriptSource(ResourceRef script) {
+		super(script);
+	}
+
+	@Override
+	public List<String> getCompileOptions() {
+		return Collections.emptyList();
```
maxandersen, line 22: should we add "KOTLINC_OPTIONS" here ? 

or maybe just "reuse" "JAVAC_OPTIONS" - is there ever a time where you need to specify/use both ?

src/main/java/dev/jbang/util/Util.java
maxandersen: should be fine :)

src/main/resources/jbang-catalog.json
```diff
@@ -12,6 +12,10 @@
       "file-refs": {"{filename}":  "init-hello.java.qute"},
       "description": "Basic Hello World template"
     },
+    "kotlin": {
+      "file-refs": {"{filename}":  "init-kotlin.kt.qute"},
```
maxandersen, line 16:
```suggestion
      "file-refs": {"{filename}":  "init-hello.kt.qute"},
```

src/main/resources/jbang-catalog.json
```diff
@@ -12,6 +12,10 @@
       "file-refs": {"{filename}":  "init-hello.java.qute"},
       "description": "Basic Hello World template"
     },
+    "kotlin": {
```
maxandersen, line 15:
```suggestion
    "hello.kt": {
```

src/main/java/dev/jbang/net/KotlinManager.java
```diff
@@ -0,0 +1,83 @@
...
+		} catch (Exception e) {
+			Util.deletePath(kotlinTmpDir, true);
+			if (!Files.isDirectory(kotlinDir) && Files.isDirectory(kotlinOldDir)) {
+				try {
+					Files.move(kotlinOldDir, kotlinDir);
+				} catch (IOException ex) {
+					// Ignore
+				}
+			}
+			Util.errorMsg("Required Java version not possible to download or install. You can run with '--java "
```
maxandersen, line 68:
```suggestion
			Util.errorMsg("Required Kotlin version not possible to download or install. You can run with '--kotlin "
```

src/main/java/dev/jbang/net/KotlinManager.java
```diff
@@ -0,0 +1,83 @@
...
+					+ JavaUtil.determineJavaVersion() + "' to force using the default installed Java.");
```
maxandersen, line 69: this would need some other detection for kotlin

src/main/java/dev/jbang/source/KotlinScriptSource.java
maxandersen: it doesn't work that great though as java is also .jsh....so this should probably be handled differently.

src/main/java/dev/jbang/source/ScriptSource.java
```diff
@@ -86,6 +92,22 @@ private ScriptSource(ResourceRef resourceRef, String content) {
...
+
+	public String getCompilerBinary(String requestedJavaVersion) {
+		return resolveInJavaHome("javac", requestedJavaVersion);
+	}
+
+	public Predicate<ClassInfo> getMainFinder() {
+		return pubClass -> pubClass.method("main", BaseBuildCommand.STRINGARRAYTYPE) != null;
+	}
+
+	public String getExtension() {
```
maxandersen, line 107: this wouldn't not always be true - i.e. `.jsh` for java shell...so maybe change to be a "deriveMainClassName" or similar ?

src/main/java/dev/jbang/Cache.java
maxandersen: `kotlincs' ? as its the kotlin compiler or is it more ?

maxandersen: pushed update to use `kotlincs` and added the support in `cache clear` command.

src/main/java/dev/jbang/util/Util.java
maxandersen: .kts would be using kotlinc to run ...so slightly different. lets make that a future idea.

src/main/java/dev/jbang/source/KotlinScriptSource.java
evanchooly: I can't imaging that most of the JAVAC_OPTIONS would even work with kotlinc. some would but many would not, i would think.  i agree with your latest commit to just leave them out completely until we have a sense of how someone wants/needs them.

src/main/java/dev/jbang/net/KotlinManager.java
evanchooly: I used to have a `determinKotlinVersion()` but it looks like it got refactored away.  I'm not sure this concept even makes sense for kotlin because it's not like we're running a KotlinVM here.  the version of kotlin is whatever is configured in the script or on the command line.  Other than that, there's nothing really left to discover, as such.

maxandersen:
I added first initial fixes I could find + some very basic doc and are quite keen try go out with this and we can cleanup / finish the two main things I noticed:

* remove need to have subclass KotlinSource
* ensure jbang edit works

wdyt @quintesse - does it look decent ?

src/main/java/dev/jbang/source/ScriptSource.java
evanchooly: vs `getMainFinder()`  ?  i'm not sure i'm following.  But if it's that, i'm not married to that name.  i just need to call it something so I could refactor and clean up.  :)

maxandersen: its just a sligthly mismatch to have it on the ScriptSource as it represents something that can have varied extensions. They make more sense on a separate "BuildTool" interface/class.

maxandersen (approved):
im ok putting this is in as experimental and we can start iterating based on feedback/issues we find.

src/main/java/dev/jbang/net/JdkManager.java
```diff
@@ -88,6 +88,17 @@ public static Path downloadAndInstallJdk(int version) {
 		}
 	}
 
+	public static String resolveInJavaHome(String cmd, String requestedVersion) {
```
quintesse, line 91: I think this is best moved to `JavaUtil`. The `JdkManager` really only handles install/uninstall of JDKs and telling you where they can be found.

quintesse (changes requested):

quintesse:
> im ok putting this is in as experimental

Agreed

src/main/java/dev/jbang/net/JdkManager.java
maxandersen: moved.

src/main/java/dev/jbang/net/KotlinManager.java
maxandersen: there isn't even a explicit control of --kotlin yet thus for now remove this to avoid confusion. we can add back if/when usecase arise for the external control.

src/main/java/dev/jbang/source/KotlinScriptSource.java
maxandersen: leaving this in for now - should be refactored to not need/want a subclass

maxandersen (approved):
this is good to go as experimental feature!

maxandersen:
@all-contributors please add @evanchooly  for code

PR #878, opened 2021-05-24
fix: Update should now work on Windows

### LINKED ISSUE #823: Jbang update fails on Windows
When you do `jbang version --update` or `jbang app install --force jbang` you'll get an:

```
[jbang] Downloading and installing jbang...
[jbang] [ERROR] Could not copy jbang.jar
```

Which is because you can't overwrite the `jbang.jar` while it's in use. (I thought I tested that and that it was possible to do so, but obviously I tested wrong somehow).

quintesse:
When a new update gets downloaded we won't try to overwrite
the jbang.jar on Windows because that would fail anyway.
Instead we copy the new jar next to the old jar as `jbang.jar.new`
and we overwrite the old version when the user runs jbang the
next time.

Fixes #823

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -194,7 +194,12 @@ class AppInstall extends BaseCommand {
 			files	.map(from::relativize)
 					.forEach(f -> {
 						try {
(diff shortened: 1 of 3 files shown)


maxandersen (approved):
lgtm

maxandersen:
awesome! thx!

PR #883, opened 2021-05-25
fix: Make startup scripts a little more foolproof

quintesse:
The scripts will now detect if an install is missing
a required file and try to re-download.


quintesse:
NB: Didn't do this for the .cmd script because IF-statements in .cmd files don't support AND/OR making the code a little unwieldy. Given the fact that a user shouldn't be missing files anyway I just decided to skip this change for jbang.cmd.
(I had missing files because I was testing the jbang update on Windows)

PR #929, opened 2021-06-09
Kotlin init command not working

mikrethor:
Command jbang init -t hello.kt hello not working jbang init -t hello.kt hello.kt


maxandersen:
works for me - what error are you getting ?

mikrethor:
I don't have any error on the generate command but when I execute I get an error on both MacOS laptop
Capture d’écran, le 2021-06-09 à 20 49 10
And it works with the following
Capture d’écran, le 2021-06-09 à 20 52 21

quintesse:
Indeed, if you have a source file without an extension it assumes it's a Java source file.
We could try to be more intelligent about it and analyse the source code to see what language it's written in.
And we should probably add an argument like `--jsh` to force a Kotlin interpretation.
(Not sure if we should add `--kotlin` or think of the future and introduce something like `--source=kotlin`)

maxandersen:
@quintesse shouldn't we be able to figure that out already ? if I use `hello.java` it complains that the template expects `.kt`

maxandersen:
@mikrethor last night I redid the docs so its now split up in multiple files and rendered on jbang.dev/documentation.

do you wanna update your PR to that format ?

mikrethor:
@maxandersen sure

quintesse:
@maxandersen it's not the `init` command that's confused, it's the `build` command that doesn't know what to do with the file because it has no extension. It's the same reason we have a `--jsh`  option because just by looking at the code we can't (easily) determine if it's proper Java or if it's meant for Jshell.

mikrethor:
so @quintesse  do I update the doc to add the extension?

maxandersen:
> so @quintesse do I update the doc to add the extension?

for now update the doc as I hadn't considered the implication @quintesse talks about and its not a quick fix to handle :)

mikrethor:
I redid the change in usage.adoc but I don't understand why other files are here. My branch seems patch-1 seems sync with main.

maxandersen:
looks like bad rebase? 

anyhow, i cherry picked your change and comitted it.

@all-contributors please add @mikrethor  for docs

PR #941, opened 2021-06-18
fix: Docs don't explain several JBANG_ env vars #785

### LINKED ISSUE #785: Docs don't explain several JBANG_ env vars
Missing are:

 - [ ] `JBANG_REPO`
 - [ ] `JBANG_DIR`
 - [ ] `JBANG_CACHE_DIR`

karstentorp:
Fixes #785
Add short documentation of the 3 undocumented environment variables JBANG_DIR, JBANG_REPO and JBANG_CACHE_DIR listed in #785

### DIFF
--- a/docs/modules/ROOT/pages/caching.adoc
+++ b/docs/modules/ROOT/pages/caching.adoc
@@ -16,4 +16,4 @@ In previous versions of `jbang`, Java 10+ direct launch of `.java` was used, but
 
 The caching goes to `~/.jbang/cache` by default, you can run `jbang cache clear` to remove all cache data from this folder.
 
-The default cache location can be overwritten by the environment variable `JBANG_CACHE_DIR`. If `$JBANG_DIR` environment variable is set, the `cache` folder will be placed there.
+The default cache location can be overwritten by the environment variable `JBANG_CACHE_DIR`. If `JBANG_DIR` environment variable is set, the `cache` folder will be placed there.
--- a/docs/modules/ROOT/pages/caching.adoc
+++ b/docs/modules/ROOT/pages/caching.adoc
@@ -15,3 +15,5 @@ endif::[]
 In previous versions of `jbang`, Java 10+ direct launch of `.java` was used, but since v0.6 `jbang` works with Java 8 and thus it needs to do a separate compile step. Besides now working with Java 8 this also allows to cache the compiled script and thus launch faster on consecutive runs.
 
 The caching goes to `~/.jbang/cache` by default, you can run `jbang cache clear` to remove all cache data from this folder.
+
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/caching.adoc
```diff
@@ -15,3 +15,5 @@ endif::[]
 In previous versions of `jbang`, Java 10+ direct launch of `.java` was used, but since v0.6 `jbang` works with Java 8 and thus it needs to do a separate compile step. Besides now working with Java 8 this also allows to cache the compiled script and thus launch faster on consecutive runs.
 
 The caching goes to `~/.jbang/cache` by default, you can run `jbang cache clear` to remove all cache data from this folder.
+
+The default cache location can be overwritten by the environment variable `JBANG_CACHE_DIR`. If `$JBANG_DIR` environment variable is set, the `cache` folder will be placed there.
```
quintesse, line 19: I'd remove the `$` from `$JBANG_DIR` (on Windows it would be `%JBANG_DIR%`, so let's just use the name)

docs/modules/ROOT/pages/installation.adoc
```diff
@@ -22,6 +22,11 @@ Once you hav installed from one of the below approaches it is recommended you ru
  your `PATH` to include jbang app scripts + it will on operating systems that supports
 it setup a `j!` alias you can use instead of `jbang`.
 
+[NOTE]
```
quintesse, line 25: I think this is a bit too "prominent". We should probably find a better place for this, wdyt @maxandersen ?

quintesse (changes requested):
In general okay but I think we need a couple of changes.

See the comments I left, but in addition to that I'd also try to make all the texts more similar in wording.  Personally I prefer this one:

```
By default `jbang` uses <some text>. This can be changed using <some more text>
```

maxandersen:
Looks good. Agree on make wording similar.

docs/modules/ROOT/pages/installation.adoc
maxandersen: We have a cache section - would put it there.

karstentorp: Good idea. I'll move it to the caching.adoc and write up a common text for JBANG_CACHE_DIR and JBANG_DIR . They are related after all.
But the the part hinting that JBANG_DIR is also used for some configurations is still valid for installation.adoc I think, or possible some other more relevant page?

docs/modules/ROOT/pages/caching.adoc
karstentorp: Was an error, fixed.

maxandersen:
@all-contributors please add @karstentorp  for docs

maxandersen:
Thanks @karstentorp - merging this now to avoid conflicts as i'm rearranging a few other sections anyway! :)

PR #961, opened 2021-07-04
feat: properties for init templates

### LINKED ISSUE #960: Is there a way to supply parameters to the init command
### Discussed in https://github.com/jbangdev/jbang/discussions/959

<div type='discussions-op-text'>

<sup>Originally posted by **grumpyf0x48** July  4, 2021</sup>
I've seen the init templates uses {baseName} to refer to the script name that is generated.

But is there a way to provide other parameters to the init command ?

I've seen the init command does not accept -Dkey=value, so i guess the answer is no :-(</div>

maxandersen:
adds initial support for properties in `init`

makes -Dkey=value available as `{properties.key}` and `{key}` to avoid overshadow but also to prevent overwriting jbang defined values such as `{baseName}`.

not fully sold both are needed ....wdyt ?

fixes #960

### DIFF
--- a/docs/modules/ROOT/pages/templates.adoc
+++ b/docs/modules/ROOT/pages/templates.adoc
@@ -72,3 +72,11 @@ $ jbang template add showlogo.java img.jpg some.properties
 [jbang] No explicit target pattern was set, using first file: \{basename}.java=showlogo.java
 [jbang] Template 'showlogo' added to '.../jbang-catalog.json'
 ----
+
+== Properties
+
+Templates can refer to properties passed in using `-Dkey=value` on the `jbang init` command.
+
+Any key specified via `-Dkey=value` becomes available as `{key}` inside the templates.
+
+It is recommended you write templates where there is some kind of valid default or fallback on behavior as users might not be aware there is need for a key.
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -149,7 +149,7 @@ public class Init extends BaseScriptCommand {
 			if (properties == null) {
 				properties = new HashMap<>();
 			}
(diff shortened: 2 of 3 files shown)


maxandersen:
Kinda feel should add ability for templates to declare which properties are required / supported so jbang init Can help User if anything missing.

quintesse:
LGTM. I do agree that a system to define props would be nice. Although perhaps another option would be to make it possible to specify default values. That's probably easier to implement than a system where you can specify which props are required and being able to show which props are needed etc (and once you do that you probably also need docs/descriptions. It's easy to fall into a rabbit hole :-) )

maxandersen:
I agree it can be a rabbit hole but I also hate maven archetypes which has zero selfdocumentation :) 

And yes I was thinking at least enable templates to declare something like (using yaml in example as i hate writing json :):

```
properties:
      - classname:
           -  description: "Class to be tested"
           -  default: "MyTest.java" 
 ```

but since default could also be calculated by the template probably still need `required: true|false` option

WDYT properties. vs no prefix - just have both ?

quintesse:
> properties. vs no prefix - just have both ?

Personally I'd perhaps just not worry about any clashes. It's not as we're expecting to add a whole bunch of pre-defined ones.

maxandersen:
if we only put {key} which wins if I do `jbang init -t cli -DbaseName=blah MyName.java` ?

maxandersen:
last update adds a mention to the docs and for now doing most restrictive one and just do {key} and any jbang vars will override/win if conflicts.

quintesse:
> if we only put {key} which wins

I'd say the one that's explicitly mentioned wins. This ensures that old templates that use values that suddenly become built-ins keep working (assuming no default values), while new templates should never use builtin values for themselves, that's just wrong :-)

PR #968, opened 2021-07-26
feat: be able to refer to scriptref in init templates

maxandersen:
This allows you to have a scriptref in your init template
    so you can do things like:

```
before_install:
  -  curl -Ls https://sh.jbang.dev | bash -s - app setup
install:
  - ~/.jbang/bin/jbang export mavenrepo --force -O target -Dgroup=$GROUP -Dartifact=$ARTIFACT -Dversion=$VERSION {scriptref or 'main.java'}
  - mkdir -p ~/.m2/repository
  - cp -rv target/* ~/.m2/repository/
```

 and then `jbang init -t jitpack@jbangdev whatever.file`

  and then the templates know its `whatever.file` it should use as its target.

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -37,7 +37,7 @@ public class Init extends BaseScriptCommand {
 	boolean force;
 
 	@CommandLine.Option(names = { "-D" }, description = "set a system property", mapFallbackValue = "true")
-	Map<String, Object> properties;
+	Map<String, Object> properties = new HashMap<>();
 
 	@Override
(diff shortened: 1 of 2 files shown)


quintesse:
I don't really understand the explanation 😅 

The title seems simple enough but then the example with the before install and install parts confuses me. Perhaps an example of what the template would look like could help?

maxandersen:
seem copy/paste error when writing - fixed it - the code is the actual template :P

quintesse:
Ooooh! Do we also have defaults now? (ie `{scriptref or 'main.java'}`) I didn't know that, cool!

maxandersen:
That's default qute syntax.

PR #972, opened 2021-08-01
enable jsh by detecting main on stdin input.

maxandersen:
if you really want .jsh you can use `--jsh`

### DIFF
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -193,6 +193,9 @@ i.e. then you can "pipe" the script to jbang:
 
 `echo 'System.out.println("Hello World!");' | jbang -`
 
+jbang will try and guess if the piped content is a java class and if not assume it it is jshell code.
+To force jbang to run it as a jshell script you can use `--jsh` flag.
+
 [TIP]
 ====
 If you use `--interactive` `jbang` will let `jshell` enter into interactive/REPL mode. You can write `/exit` to leave this mode.
--- a/itests/run.feature
+++ b/itests/run.feature
@@ -10,6 +10,11 @@ Scenario: parameter passing
 * match err == "[jbang] Building jar...\n"
 * match out == "Hello jbangtest\n"
 
+Scenario: std in
+* command('cat helloworld.java | jbang - jbangtest')
+* match err == "[jbang] Building jar...\n"
+* match out == "Hello jbangtest\n"
+
 Scenario: java launch helloworld with jfr
   When command('jbang --jfr helloworld.java')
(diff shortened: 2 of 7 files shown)


docs/modules/ROOT/pages/usage.adoc
```diff
@@ -193,6 +193,9 @@ i.e. then you can "pipe" the script to jbang:
 
 `echo 'System.out.println("Hello World!");' | jbang -`
 
+jbang will try and guess if the piped content is a java class and if not assume it it is jshell code.
+To force jbang to run it as a jshell script you can use `--jsh` flag.
```
quintesse, line 197: The thing of course is that for this to work really well the default should be to assume .java and then to be able to override that with --jsh. Because right now if jbang misdetects code as being Jshell there is no way to override that. But of course we probably can't suddenly invert how jbang worked up to now :-/

quintesse (approved):
LGTM

docs/modules/ROOT/pages/usage.adoc
maxandersen: I don’t think Default to assume java is right for stdin. Jshell is more apropriate. 

What we are doing here is that if I’m that rare occasion you try and pipe code in and it has a main method we’ll attempt to run it. 

Ultimately you should run it with jbang your.java as java syntax and semantics is tied to the file name. Something stdin doesn’t allow for.

quintesse: You're probably right when you think of things like `echo 'System.out.println("Hello World");' | jbang -`, although personally I'm fan of introducing an option for that, eg: `jbang run -s 'System.out.println("Hello World");'`

maxandersen: that would be an additional option. 

jshells would normally be mulitline things anyways so `-s` not really a good option for that anyway

quintesse: Sure, but if they are multi-line you're probably piping them in from a file or a curl command anyway so the statement that it's most likely to be JShell code doesn't hold as much truth IMO. If it's something somebody would type on the CLI, yes, but once you start piping in files it's much more likely to be .java code (again IMO).

And of course we'd still use the heuristics to decide what it is, but the default if we can't figure it out could be .java which then still leaves the `--jsh` option to override.

I also say this because the amount of Java code out there is vastly larger than the Jshell code. Even in our examples we hardly ever use Jshell.

maxandersen: What i mean Is that JShell code is evaluated Line by Line. Java is not and must have certain Structure to work. So it's Lead likely to work via a pipe. With This pr we detect the specific case so I think it's good.

PR #976, opened 2021-08-08
feat: Add runtime options for aliases

### LINKED ISSUE #975: Add runtime options for aliases
It would be a very useful feature to be able to define runtime options for aliases, eg. options that set memory limits.

This would mean adding a new section to the jbang-catalog.json file, eg:

```
{
  "aliases": {
    "hello": {
      "script-ref": "hello@quintesse",
      "java-options": [
        "-showversion"
      ]   
    }   
  }
}
```

Edit: renamed `runtime-options` to `java-options`

quintesse:
Fixes #975


src/main/java/dev/jbang/cli/Run.java
```diff
@@ -59,6 +59,9 @@ boolean debug() {
 	@CommandLine.Option(names = { "--interactive" }, description = "activate interactive mode")
 	boolean interactive;
 
+	@CommandLine.Option(names = { "--catalog" }, description = "path to catalog file")
```
maxandersen, line 62: what is this used for in context of run ?

maxandersen (changes requested):
looks good but not following why this change means run now has a --catalog option ?

quintesse:
Ah sorry yes, I added that because the `jbang alias` and `jbang catalog` commands have a `-f`/`--file` option to explicitly specify the catalog file to read/write, but `run` didn't have that so it wasn't possible to tell it to "use this particular catalog file when resolving aliases/catalogs". Which might not seem that useful for everyday use but was essential for testing :-)
I can remove it but then testing possibly becomes somewhat harder, especially because the normal way the `run` command does lookups is pretty complex. So you have to make sure you override all the proper paths so it will read the test file and not some other file on the system.

maxandersen:
Ack. Makes sense. Merged. Thanks!!

PR #979, opened 2021-08-16
feat: added `java` and `main` options to aliases

quintesse:
Added the `java` (alt `java-version`) option we need for the MSFT support and also added `main` (alt `main-class`) because that was the only other option that might prevent code from running if not available (even if we most likely don't need it in this particular case).

### DIFF
--- a/src/main/java/dev/jbang/catalog/Alias.java
+++ b/src/main/java/dev/jbang/catalog/Alias.java
@@ -21,12 +21,18 @@ public class Alias extends CatalogItem {
 	@SerializedName(value = "java-options")
 	public final List<String> javaOptions;
 	public final Map<String, String> properties;
+	@SerializedName(value = "java")
+	public final String javaVersion;
+	@SerializedName(value = "main")
+	public final String mainClass;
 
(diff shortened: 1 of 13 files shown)


src/main/java/dev/jbang/catalog/Alias.java
```diff
@@ -21,19 +21,27 @@
 	@SerializedName(value = "java-options")
 	public final List<String> javaOptions;
 	public final Map<String, String> properties;
+	@SerializedName(value = "java", alternate = { "java-version" })
```
maxandersen, line 24: why the alternate names? shouldn't we not just use the same as long form command line flag ?

quintesse: no special reason, if you don't like them I'll remove them :-)

maxandersen: Less ways to configure. Less bugs :) so yes please. Let's keep it clean.

quintesse: Done

maxandersen (approved):

Issue #418, opened 2020-10-10
Allow specifying JVM flags


alblue:
It would be good if you could also specify JVM flags when launching the script from JBang, such as `jbang --XX:+UseShenandoahGC` that would be passed to the JVM itself. You have this somewhat with the `-D` and `--jfr` flags, but they're specific for those two use cases. For some common options, like shenandoah, you could even map those flags directly to a --gc shenandoah flag.

maxandersen:
We have support for that In the Java file using javac_options and Java_options - would make sense to to expose on cli but for now I've been holding back as haven't found a way to nicely separate javac vs java target flags.

maxandersen:
Maybe just a raw --javac-options and --java-options option to mimick what you can do in the file would work. They would need splitting to be appended to the command line arguments.

quintesse:
As just noticed while talking here https://github.com/jbangdev/jbang/discussions/988 it's strange that we can now set runtime options at least using aliases but we can still not do so with `jbang run`. Perhaps we should just use the same `--runopts` flag? And a `--buildopts` as well for good measure?

quintesse:
Btw, it also shows that we should perhaps take another look at that `--runopt` name on `alias add`. It's abbreviated and not even in plural.

maxandersen:
hmm - I honestly missed that :/ it should not be `--runopt` and most definitely should be availabe using jbang directly not just alias.

it should be the same setting/property/key we use across.

So `--java-options` and `--javac-options` would be the right names - wanna make a PR to fix that for 0.80th release?

quintesse:
Sure, no problem!

brunoborges:
I'd rather prefer `--jvm-options` to remove ambiguity with `--javac-options`, and the risk of failing due to a missing `c`.

quintesse:
> I'd rather prefer `--jvm-options` to remove ambiguity with `--javac-options`, and the risk of failing due to a missing `c`.

Hmmm true. Wdyt @maxandersen ? It would mean that we're not using the same name as used in the code: `//JAVA_OPTIONS`. But perhaps this is something we could still change before 1.0? We could leave `//JAVA_OPTIONS` as an alias for `//JVM_OPTIONS` for a while before deprecating it.

quintesse:
(Also is we'd ever want to consider supporting other languages perhaps we shouldn't use "JAVA" at all but more generic `RUN_OPTIONS` and `BUILD_OPTIONS`)

[...]

PR #989, opened 2021-09-10
feat: add `--java-options`

quintesse:
Both for #418 and to fix the flag used by `alias add`


src/main/java/dev/jbang/cli/Run.java
```diff
@@ -80,7 +83,7 @@ public Integer doCall() throws IOException {
 	}
 
 	RunContext getRunContext() {
-		RunContext ctx = RunContext.create(userParams, null,
```
maxandersen, line 83: observation: this line is going to grow as we  add all of the options here :)

maxandersen (approved):

src/main/java/dev/jbang/cli/Run.java
quintesse: Yeah, we'll see how long it gets before we get sick of it and refactor the shit out of that :-)

PR #994, opened 2021-09-12
fix!: make edit open by default and add --no-open

### LINKED ISSUE #993: jbang edit should open editor by default
This is a breaking and possibly controversial change but I constantly find two things about the edit command:

1) users don't realize what it does as it just prints out folder name by default
2) I have to constantly type out `jbang edit --open test.java` when it could just be `jbang edit test.java`

I've been thinking for a long time we should instead make `jbang edit xyz.java` simply just open editor (if not defined will ask to install vscodium) and if you don't want to open and work as before you do jbang edit --no-open xyz.java`

`jbang edit --open=idea xyz.java` will still work and so will `JBANG_EDITOR=idea jbang edit xyz.java`

This is "less" linux like but I feel it is a better modern default behavior.

maxandersen:
fixes #993 by making open ide the default and add --no-open option to use it in a terminal context.

### DIFF
--- a/docs/modules/ROOT/pages/editing.adoc
+++ b/docs/modules/ROOT/pages/editing.adoc
@@ -14,45 +14,39 @@ endif::[]
 
 toc::[]
 
-You can edit your script in an IDE/editor by using `jbang edit helloworld.java`. This will generate a project in a temporary location with symbolic links to your script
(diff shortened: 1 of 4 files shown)


quintesse (approved):
LGTM

PR #995, opened 2021-09-13
fix: Fixed version updates

### LINKED ISSUE #990: `jbang version --update` often doesn't work
This happens when a previous version is in the cache (the downloads have the same name).

A workaround is to add `--fresh`.

Possible solutions are to just implicitly add `--fresh` or knowing the version to upgrade to we could specifically download that version.

quintesse:
We're not trying to resolve new version downloads from cache anymore

Fixes #990


quintesse:
There's no test for this as I have no idea how to easily create one. We'd have to introduce an entire mocking layer for that and that seems like a lot of work right now :-)

PR #996, opened 2021-09-13
[fix] Make sure `version --update` works on Windows

### LINKED ISSUE #991: `jbang version --update` doesn't work on Windows
Jbang can't overwrite the .jar file that's currently in use.

There are two possible solutions:

- Write the new jar file as `jbang.jar.new` and have the `jbang.cmd`/`jbang.ps1` scripts rename it next time they get invoked.
- Have `jbang version` return a CLI command to rename the jar file

quintesse:
Fixes #991


quintesse:
Here too creating a test is kinda hard, so we'll just have to test after a new release has been done. But I've stepped through the code and simulated as much as possible the update from one version to another and it should work now.

PR #999, opened 2021-09-15
fix: `baseName` is no longer ambiguous

### LINKED ISSUE #987: `jbang init` has `{baseName}` ambiguities in templates
See discussion here https://github.com/jbangdev/jbang-catalog/pull/8#discussion_r702797278

The TL;DR is that using any complex file names using `{baseName}` cause problems.

The longer explanation is that when file names and/or class names are generated that are not plain `{baseName}` but have some string prepended or appended, eg `{baseName}Test` are problematic because `{baseName}` can have different values depending on context, it's just that in the simple cases we've used so far the `{baseName}` just happened to resolve to the same value.

For example when you write `jbang init -t foo Foo.java` the value of `{baseName}` used in the `jbang-catalog.json` file resolves to "Foo". So if that catalog file has a section like this:

```
"file-refs": {
    "{basename}Test.java": "templates/junit4b.java.qute"
},
```

it means that jbang will create a file named "FooTest.java" with it's contents based on `templates/junit4b.java.qute`, but... and this is the issue, once that template gets processed the `{baseName}` has the value "FooTest" because that's what its name is, so that's correct too.

quintesse:
Fixes #987

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -52,6 +52,7 @@ public class Init extends BaseScriptCommand {
 		String outName = outFile.getFileName().toString();
 
 		properties.put("scriptref", scriptOrFile);
+		properties.put("baseName", Util.getBaseName(Paths.get(scriptOrFile).getFileName().toString()));
 
 		List<RefTarget> refTargets = tpl.fileRefs	.entrySet()
 													.stream()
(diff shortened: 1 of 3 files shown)


maxandersen:
And this doesn't break existing templates referring to base name ?

quintesse:
It shouldn't break ours at least, the "wrong" values should always have been the same as the "right" values for the simple cases. And for anybody who tried to use the current values in a more complex case I'm guessing they would have run into the same problem that the PR is trying to fix: that the values weren't the ones they were expecting.

quintesse:
Btw, as you can see I hardly had to fix any template or init tests, just one place where the expected baseName value was hard-coded to be a certain value.

maxandersen:
looks good!

Issue #487, opened 2020-11-07
jbang "plugins"


maxandersen:
until now i've been hesitant to make jbang pluggable but I think I found a usecase where it could be the best option: extend jbang to do things like `test`, `codecov` etc.

these are things I don't really want to put into the core of jbang but I would like to be more accessible to jbang users.

i.e. today you can do the following to instrument your code for codecoverage:

`jbang --javaagent=agent@jacoco Snake3dApp.java`

that is easy enough; but when it comes to the report:

`jbang cli@jacoco report --classfiles ~/.jbang/cache/jars/Snake3dApp.java.18fdbbb06380ec9b425414e14b60f3aab129ea87e6941197850153239801f846 --html html --sourcefiles . jacoco.exec`

now..we can simplify some of this using:

```
jbang cli@jacoco report --classfiles `jbang info classpath Snake3dApp.java` --html html --sourcefiles . jacoco.exec
```

but what if you could do:

```
jbang codecov report Snake3dApp.java
```

and it would be able to transform that into the right command.

We kinda could make it work with aliases today we just need to find a way to have the alias "talk back" to jbang...i.e. if alias marked 'plugin' we pass a json doc to it via env var or similar for it to extract get info needed, possibly make a small client lib.

quintesse:
What would that needed info passed in the json be exactly?

maxandersen:
basically `jbang info tools`

maxandersen:
btw. thinkinga bout this we probably be better of just making a jbang api aliases can use/do to get access to classpath, caache info etc.

quintesse:
What would that API look like? An actual API? Another option would be to write that info (to yet another temp file? hehe) and pass in a path to it using a system property for example. (We could then create a simple Maven module to obtain and parse the file to make it even easier)

maxandersen:
The Challenge is we don't know what the parameters after the plugin means so we can't really know which file to parse. 

So we would kinda need to give access to some boiled down API to get access to something similar like Script.

But right now that is a mess :)

quintesse:
But isn't that something we could define? That plugins always get passed the path to that file as their first parameter?

maxandersen:
Wouldn't that be too limiting ? For example then doing other things like, 'alias' wouldn't work as it has name then script.

quintesse:
[...]

PR #1000, opened 2021-09-15
feat: Added support for "plugins"

quintesse:
Any aliases or commands on the user's PATH that start with "jbang-"
will be considered "plugins" and can be run as if they are part of
the built-in jbang commands.

See #487

@maxandersen this is some code I had still lying around. I've cleaned it up and created this PR so you can take a look and decide if you like the idea or not :-)


maxandersen:
Congrats on 1000th pull-request!

I like the idea and been pondering how to have a way to pass on info or some kind of call back.

I was thinking jbang would have a http end point but once the jbang-* scripts runs the jbang process would be gone. 

Best thing I can think of is to dump output of "info tools" in a temporary folder and have JBANG_TOOLS_INFO env var set to be passed on ... but that would require some magic in the shell scripts to work.

quintesse:
I was actually thinking that, as long as the info is limited enough, it could just be directly passed as an ENV var. That way there's nothing to create and nothing to clean up.

You also have to think about _what_ useful information could be passed, there doesn't seem to be that much. You have any examples of data that you'd like to pass that wouldn't be easy enough for the plugin to discover by itself?

quintesse:
> Congrats on 1000th pull-request!

Yay! Do I get a T-Shirt? 😁

maxandersen:
> You also have to think about _what_ useful information could be passed, there doesn't seem to be that much. You have any examples of data that you'd like to pass that wouldn't be easy enough for the plugin to discover by itself?

the challenge is that we can't really pass much in as jbang don't even know what script/file it is targeting - or are you assuming that any non-flag argumentsis the script ref to look for ?

if not, then the plugin would need to have some jbang api dependency it could depend on to include to then do the jbang work.

quintesse:
Well right now these "plugins" are more like Git plugins: they look like they're part of the main command but they're completely separate and therefore have basically nothing to do with it.

For commands that want to deal with script refs and such we could make sure that JBang has a proper public API that people can use. Then a path to the `jbang.jar` would be the only information we'd have to pass to the command.

quintesse:
Rebased old PR that still seems useful. It has an extremely simple implementation that follows how Git does things. Like I've explained before I think the simplicity is an advantage. We can already do useful things with just this code.

In the future we could extend this with for example:

- a mechanism to pass information from the main executable to these plugins
- a command to manage plugins
- ways to have plugins predefined but installed on first use (imagine for example that the "export" command was a plugin, JBang could come with it predefined without it being part of the main code)
- etc?

maxandersen:
```
jbang app install --name=jbang-env env@jbangdev
jbang env
```

does not result in jbang-env being run.

that should work should it not?

quintesse:
> that should work should it not?

It should indeed. Will have to see if something broke during the rebase

quintesse:
Ok, tests pass now.

maxandersen:
merged as it does what it says - simple and no major catches :)

* a mechanism to pass information from the main executable to these plugins

I gave this a think and I dont think we should/can do much here besides tell subcommand with some env var where jbang is installed/run from to let them call `jbang info ...` commands or similar. we can't really know what comes "after" the command - which part is alias etc.

* a command to manage plugins

yes!

* ways to have plugins predefined but installed on first use (imagine for example that the "export" command was a plugin, JBang could come with it predefined without it being part of the main code)

yes - we should have that - could even have it go look for them if not present....though need to be careful not breaking any ci scripts.

maxandersen (dismissed):

PR #1001, opened 2021-09-16
chore: it is JBang not Jbang thus search/replace to fix it once and for all

maxandersen:
in writing an article I noticed we still have places in docs and output we print "Jbang" instead of "Jbang".

So did a search and replace to get it all removed.


maxandersen:
@quintesse just to give you a headsup as this touches a lot of files but is just cleanup. so prefer to merge it when you dont have too many conflicting branches in parallel :)

quintesse:
I think there won't be too much of a problem, so go ahead :-)

PR #1002, opened 2021-09-19
fix: migrate to adoptium

maxandersen:
first attempt on migrating to use adoptium as requested at https://blog.adoptopenjdk.net/2021/08/goodbye-adoptopenjdk-hello-adoptium/ but it is not working.

opened https://github.com/adoptium/api.adoptium.net/issues/165 to understand better why.


maxandersen:
closing as foojay can handle all of the needed versions we need/want access to.

PR #1003, opened 2021-09-19
fix: use disco/foojay instead of adoptopenjdk to get api that works with old and new jdk versions

maxandersen:
initial attempt on using foojay.io to fetch jdks so we can get both old and new java versions via 
the same API.

Opens up for in future doing nice things like `//JAVA 18-ea` to get earlyaccess versions or even `//JAVA 11+(zulu)` to get 11 or higher using zulu.

For now made default for jbang to use termurin for 8, 11 and 17+, rest is adoptopenjdk.
For bash and windows download of default jdk this still uses adoptopenjdk api as it gives 
a url that returns the right binary.

### DIFF
--- a/src/main/scripts/jbang.cmd
+++ b/src/main/scripts/jbang.cmd
@@ -9,7 +9,7 @@ set os=windows
 set arch=x64
 
 set jburl="https://github.com/jbangdev/jbang/releases/latest/download/jbang.zip"
-set jdkurl="https://api.adoptopenjdk.net/v3/binary/latest/%javaVersion%/ga/%os%/%arch%/jdk/hotspot/normal/adoptopenjdk"
+set jdkurl="https://api.foojay.io/disco/v2.0/directuris?distro=aoj&libc_type=libc&archive_type=zip&operating_system=%os%&package_type=jdk&version=%javaVersion%&release_status=ga&architecture=%arch%&latest=available"
 
 if "%JBANG_DIR%"=="" (set JBDIR=%userprofile%\.jbang) else (set JBDIR=%JBANG_DIR%)
 if "%JBANG_CACHE_DIR%"=="" (set TDIR=%JBDIR%\cache) else (set TDIR=%JBANG_CACHE_DIR%)
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -8,6 +8,12 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      debug_enabled:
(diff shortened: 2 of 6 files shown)


maxandersen:
thinking about this I was thinking we could make a GAV "like" syntax for the java version/reference.

| java ref  | meaning |
| ------------- | ------------- |
| 11+ | get 11 or higher java Latest stable version using aoj/termurin  |
| zulu:11+ | get 11 or higher java latest stable version using zulu  |
| oracle_openjdk:18-ea | get 18 early access version using oracle openjdk builds |

quintesse:
I wonder if `11+` shouldn't mean "any JDK of version 11 or above". And only refer to aoj/termurin when having to install a JDK if one isn't already present. Or was that already what you meant to do?

maxandersen:
> I wonder if `11+` shouldn't mean "any JDK of version 11 or above". And only refer to aoj/termurin when having to install a JDK if one isn't already present. Or was that already what you meant to do?

Yes. That is what I meant.

maxandersen:
updated to use new endpoint at foojay that does redirect directly.

Would be great to have others test that things like:

```
 curl https://raw.githubusercontent.com/maxandersen/jbang/foojay/src/main/scripts/jbang | bash -s - app setup
```

works on mac/linux

and 

```
iex "& { $(iwr https://raw.githubusercontent.com/maxandersen/jbang/foojay/src/main/scripts/jbang.ps1) } app setup"
```

works on windows 10

maxandersen:
something is preventing jdk download  working on github actions/azure setup for *one* test. added some logging about what it is trying to  download.

quintesse:
> works on windows 10

Can't test this until next week, otherwise I'd do it. Hopefully someone else can :-)

src/main/scripts/jbang.cmd
```diff
@@ -9,7 +9,7 @@ set os=windows
 set arch=x64
 
 set jburl="https://github.com/jbangdev/jbang/releases/latest/download/jbang.zip"
-set jdkurl="https://api.adoptopenjdk.net/v3/binary/latest/%javaVersion%/ga/%os%/%arch%/jdk/hotspot/normal/adoptopenjdk"
+set jdkurl="https://api.foojay.io/disco/v2.0/directuris?distro=aoj&libc_type=libc&archive_type=zip&operating_system=%os%&package_type=jdk&version=%javaVersion%&release_status=ga&architecture=%arch%&latest=available"
```
quintesse, line 12: You didn't change the .ps1 and bash shell versions yet?

maxandersen:
made tests pass. @quintesse spot anything else missing?

maxandersen:
tested this in fedora containers, osx and windows....lets let it loose into the world and iterate fast if find issues.

PR #1011, opened 2021-09-22
chore: eliminate warnings from code base

fbricon:
- removed unused variables/fields
- used static access of static methods when necessary

0 warnings in VS Code \o/

Screenshot 2021-09-22 at 11 06 52

Signed-off-by: Fred Bricon <fbricon@gmail.com>


quintesse:
Good to see that almost all of these are in tests ;-)

maxandersen:
@all-contributors please add @fbricon  for code

PR #1015, opened 2021-09-23
fix: kotlin edit now link to source rather than cache location

### LINKED ISSUE #1004: kotlin jbang edit has broken symlink, easy to lose changes
Hello

zsh 19222 %-./jbang version 0.79.0

Editing a kotlin script sets a wrong symlink, its target is the `.jbang/cache/scripts` instead of the original file. So when you save, it saves into the cached workspace, not the original workspace ; so you may lose changes if you clear the cache without first making a backup copy of the cache, or use a --fresh parameter.

Editing a .java script works fine, the symlink target is the original file. See below

```
./jbang init -t hello.kt hello.kt (incorrect, symlink symlinks the cache)
ll /home/lperez/.jbang/cache/projects/hello.kt_jbang_4213db2048b7d3e23f509e98ca472a24b9c9a7d1429d73690dfa350b05e70092/hello/src/hello.kt
lrwxrwxrwx 1 lperez lperez 107 sept. 21 09:43 /home/lperez/.jbang/cache/projects/hello.kt_jbang_4213db2048b7d3e23f509e98ca472a24b9c9a7d1429d73690dfa350b05e70092/hello/src/hello.kt -> /home/lperez/.jbang/cache/scripts/364999e9df120025a4ca69dc4abfe52c631fa38c3669e6653532179224eb1a1e/hello.kt
```

```
./jbang init helloworld.java (correct, symlink is ok)
ll /home/lperez/.jbang/cache/projects/helloworld.java_jbang_a98bb232e47d9a44a00ceb11e94e330bed194064d9bfc7185cd63451f73d32b5/helloworld/src/helloworld.java
lrwxrwxrwx 1 lperez lperez 45 sept. 21 09:35 /home/lperez/.jbang/cache/projects/helloworld.java_jbang_a98bb232e47d9a44a00ceb11e94e330bed194064d9bfc7185cd63451f73d32b5/helloworld/src/helloworld.java -> /home/lperez/Bureau/work/bang/helloworld.java
```

I've looked into the code but I can't pinpoint the symlink file :/

Thanks, great project BTW

maxandersen:
Fixes #1004

### DIFF
--- a/src/main/java/dev/jbang/source/ResourceRef.java
+++ b/src/main/java/dev/jbang/source/ResourceRef.java
@@ -171,6 +171,7 @@ public class ResourceRef implements Comparable<ResourceRef> {
 		try {
 			if (probe != null && probe.canRead()) {
 				if (!probe.getName().endsWith(".jar") && !probe.getName().endsWith(".java")
+						&& !probe.getName().endsWith(".kt")
 						&& !probe.getName().endsWith(".jsh")) {
 					if (probe.isDirectory()) {
 						File defaultApp = new File(probe, "main.java");


src/main/java/dev/jbang/source/ResourceRef.java
```diff
@@ -171,6 +171,7 @@ private static ResourceRef forScript(String scriptResource) {
 		try {
 			if (probe != null && probe.canRead()) {
 				if (!probe.getName().endsWith(".jar") && !probe.getName().endsWith(".java")
+						&& !probe.getName().endsWith(".kt")
 						&& !probe.getName().endsWith(".jsh")) {
 					if (probe.isDirectory()) {
 						File defaultApp = new File(probe, "main.java");
```
quintesse, line 177: Seeing this line I just remembered that we had a discussion about this. Perhaps we should open an issue to track wether we should remove the support for `main.java` or not...

maxandersen: why would we want to remove support for main.java?

quintesse: See, this shows we should turn our talks into issues ASAP, otherwise we keep rehashing things :-)

It was because we wanted to reclaim URLs that point to folders instead of files. Right now I don't remember the exact reason anymore. I'll update when I remember the details :-)

maxandersen: What I recall was that we would need to treat urls  with a path ending with / as folder. Main thing was to be able to do jbang https://acme.org/folder/

quintesse: Perhaps. But to have completely different behaviours between http://acme.com and http://acme.com/ is a bit ... awful IMO. Besides the fact that there might be situations where trailing slashes get stripped automatically by the client your're using (officially there should be no semantic difference between `/folder` and `/folder/` after all).

PR #1017, opened 2021-09-25
fix: --deps now gets special javafx treatments as //DEPS does

maxandersen:
This now works:

```bash
jbang --java 17 --deps org.openjfx:javafx-graphics:17:mac --deps eu.hansolo.fx:charts:RELEASE https://github.com/HanSolo/charts/blob/master/src/test/java/eu/hansolo/fx/charts/SankeyPlotTest.java
```

Before `--deps` was added to the classpath but was not considered
part of the modular classpath thus the magic of detecting javafx dependencies
never took affect.

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -13,7 +13,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
@@ -90,9 +89,11 @@ public abstract class BaseBuildCommand extends BaseScriptDepsCommand {
 	static Source build(ScriptSource src, RunContext ctx) throws IOException {
 		Source result = src;
 
-		for (Map.Entry<String, String> entry : ctx.getProperties().entrySet()) {
-			System.setProperty(entry.getKey(), entry.getValue());
-		}
+		// todo: pass properties through instead of changing system properties
+		/*
+		 * for (Map.Entry<String, String> entry : ctx.getProperties().entrySet()) {
+		 * System.setProperty(entry.getKey(), entry.getValue()); }
+		 */
 
 		File outjar = src.getJarFile();
 		boolean nativeBuildRequired = ctx.isNativeImage() && !getImageName(outjar).exists();
--- a/src/main/java/dev/jbang/source/RunContext.java
+++ b/src/main/java/dev/jbang/source/RunContext.java
@@ -8,13 +8,15 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.stream.Collectors;
 
-import dev.jbang.Settings;
 import dev.jbang.catalog.Alias;
 import dev.jbang.cli.BaseCommand;
 import dev.jbang.cli.ExitException;
+import dev.jbang.dependencies.Detector;
 import dev.jbang.dependencies.ModularClassPath;
 import dev.jbang.util.JavaUtil;
+import dev.jbang.util.PropertiesValueResolver;
 
 /**
  * This class contains all the extra information needed to actually run a
@@ -51,6 +53,8 @@ public class RunContext {
 	private ModularClassPath additionalMcp;
 	private boolean nativeImage;
 	private String javaVersion;
+	private ModularClassPath combinedMcp;
+	private Properties contextProperties;
 
 	public static RunContext empty() {
 		return new RunContext();
@@ -245,6 +249,17 @@ public class RunContext {
 		this.catalogFile = catalogFile;
 	}
 
+	public Properties getContextProperties() {
+		if (contextProperties == null) {
+			contextProperties = new Properties(System.getProperties());
+			// early/eager init to property resolution will work.
+			new Detector().detect(contextProperties, Collections.emptyList());
+
+			contextProperties.putAll(getProperties());
+		}
+		return contextProperties;
+	}
+
 	public static class AgentSourceContext {
 		final public Source source;
 		final public RunContext context;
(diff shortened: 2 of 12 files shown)


src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -90,6 +90,7 @@ static Source buildIfNeeded(Source src, RunContext ctx) throws IOException {
 	static Source build(ScriptSource src, RunContext ctx) throws IOException {
 		Source result = src;
 
+		// todo: pass properties through instead of changing system properties
```
maxandersen, line 93: i noticed this and similar in alias where there is a reliance on system properties having been set upfront. That doesn't work well if want to have this as a library as it would then be part of global state.

Thus it is not relevant to this specific issue but just took note of it.

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -1437,4 +1435,70 @@ void testJVMOpts() throws IOException {
...
+	void testJavaFXMagicPropertyViaCommandline() throws IOException {
+
+		Assertions.assertThrows(ExitException.class, () -> {
+			JBang jbang = new JBang();
+
+			Path fileref = examplesTestFolder.resolve("SankeyPlotTest.java").toAbsolutePath();
+
+			// todo fix so --deps can use system properties
+			CommandLine.ParseResult pr = new CommandLine(jbang).parseArgs("run",
+					"--deps", "org.openjfx:javafx-graphics:17:${os.detected.jfxname}",
```
maxandersen, line 1490: this is currently not working because properties are not processed/accesssible when these --deps from command line are accessed. needs cleaning up in separate pr.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
maxandersen: i went ahead and fixed property passing so now runcontext have a getContextProperties() which is backed by system properties rather than require changing system properties.

quintesse (approved):
LGTM

maxandersen:
weird - dont grok why it fails on linux - works fine on my osx...and its some weird errors i don't see how they could be different per OS.

quintesse:
The errors are all of this form:

```
java.lang.AssertionError: 
Expected: a string containing " --module-path "
     but: was "java -classpath /tmp/junit3867832231155117106/jbang/cache/jars/SankeyPlotTest.java.13c9c8b57a4c7c72342a1673a3004aeb866e86ae12456532782119c2368c8e1a.jar:/home/runner/.m2/repository/org/openjfx/javafx-graphics/17/javafx-graphics-17-linux.jar:/home/runner/.m2/repository/org/openjfx/javafx-base/17/javafx-base-17.jar:/home/runner/.m2/repository/org/openjfx/javafx-base/17/javafx-base-17-linux.jar:/home/runner/.m2/repository/org/openjfx/javafx-controls/17/javafx-controls-17-linux.jar:/home/runner/.m2/repository/org/openjfx/javafx-graphics/17/javafx-graphics-17.jar:/home/runner/.m2/repository/eu/hansolo/fx/charts/17.0.0/charts-17.0.0.jar:/home/runner/.m2/repository/org/openjfx/javafx-base/17.0.0.1/javafx-base-17.0.0.1.jar:/home/runner/.m2/repository/org/openjfx/javafx-base/17.0.0.1/javafx-base-17.0.0.1-linux.jar:/home/runner/.m2/repository/org/openjfx/javafx-graphics/17.0.0.1/javafx-graphics-17.0.0.1.jar:/home/runner/.m2/repository/org/openjfx/javafx-graphics/17.0.0.1/javafx-graphics-17.0.0.1-linux.jar:/home/runner/.m2/repository/org/openjfx/javafx-controls/17.0.0.1/javafx-controls-17.0.0.1.jar:/home/runner/.m2/repository/org/openjfx/javafx-controls/17.0.0.1/javafx-controls-17.0.0.1-linux.jar:/home/runner/.m2/repository/org/openjfx/javafx-swing/17.0.0.1/javafx-swing-17.0.0.1.jar:/home/runner/.m2/repository/org/openjfx/javafx-swing/17.0.0.1/javafx-swing-17.0.0.1-linux.jar fakemain"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at dev.jbang.cli.TestRun.testJavaFXMagicPropertyViaCommandline(TestRun.java:1500)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

src/main/java/dev/jbang/source/RunContext.java
```diff
@@ -245,6 +247,12 @@ public void setCatalog(File catalogFile) {
 		this.catalogFile = catalogFile;
 	}
 
+	public Properties getContextProperties() {
+		Properties p = new Properties(System.getProperties());
```
quintesse, line 251: Is this what you meant to do here? 
I had no idea how `Properties` worked but this code does something very unexpected for somebody (like me) who doesn't.
This actually results in an *empty* Properties object!

I actually find it completely mystifying how this passes the tests. If I run them all there's no error. But when I run the (javaFX) tests one by one, the one using properties fail.

quintesse (changes requested):
Tests with --deps and system properties don't actually work.

quintesse:
I think this line might need to be updated: https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/source/ScriptSource.java#L145

src/main/java/dev/jbang/source/RunContext.java
maxandersen: Aaaah. That makes sense! Need the detector wired in to the properties rather than system properties. Totally agree. 

And yes it's intended the properties are empty - the system properties just used as fall back.

maxandersen:
not really grokking why this keeps failing in ci but not locally for me after i did the last changes.

exploring...

quintesse:
Perhaps the CI environment does a better job of cleaning up the environment after each test? Or it just runs them in a different order. The tests fail for me too, locally, as I described above. There's some `System.setProperty()` going on that shouldn't be happening.

maxandersen:
figured it out  - the tests runs on java 8 - which does not support modules hence the test is actually correct.

maxandersen:
looking good - fixed test to use path based check and things are looking good :)

quintesse (approved):

PR #1020, opened 2021-09-26
fix:  jbang edit suggests incorrect command when template renames out…

grumpyf0x48:
Use the first rendered file using `{baseName}` as the name to use in suggested edit command.

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -9,6 +9,7 @@ import java.util.AbstractMap;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Random;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


grumpyf0x48:
Didn't add tests because i guess you don't write tests to expect JBang output.

maxandersen:
We do. See itest folder for .feature.

grumpyf0x48:
> We do. See itest folder for .feature.

Great ! Not seen cucumber was used there so far.

To test the change in `init` output, we need to have a template using something like `"{baseName}Test.java": "templates/..."`.

So two possibilities here:

- Such a template exists somewhere and can be used in tests: for example in jbangdev catalog (`catalog.feature` already uses this catalog)
- The template needs to be created in tests by a JBang command line

The first option makes me think that we could use JUnit templates once they are merged.

For the second option, how to create a template looking like:

```
    "junit4b": {
      "file-refs": {
        "{basename}Test.java": "templates/junit4b.java.qute"
      },
      "description": "Basic template for JUnit4 tests"
    },
```

with a JBang command line ?

quintesse:
> how to create a template 

IIRC you can do: `jbang template add --name junit4b "{basename}Test.java=templates/junit4v.java.qute"`

maxandersen:
@all-contributors please add @grumpyf0x48  for code

maxandersen:
@all-contributors please add @grumpyf0x48 for code

PR #1028, opened 2021-09-30
fix: Prevent jbang edit --live from entering endless loop 

### LINKED ISSUE #1027: Changes in live editing mode lead to infinite "Regenerating project."
jbang Version 0.80.2

Using Windows 10 I have a problem while editing jbang files in live editing mode. I tested with Eclipse and VS Code.

Whenever I save a edited script jbang logs `Regenerating project.` infinitely.

quintesse:
Also using links now for files with improper names.

Fixes #1027

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -124,7 +124,6 @@ public class Edit extends BaseScriptDepsCommand {
 								// TODO only regenerate when dependencies changes.
 								info("Regenerating project.");
 								ctx = RunContext.empty();
-								src = Source.forResource(scriptOrFile, ctx);
 								createProjectForEdit((ScriptSource) src, ctx, true);
 							} catch (RuntimeException ee) {
 								warn("Error when re-generating project. Ignoring it, but state might be undefined: "
--- a/src/main/java/dev/jbang/source/ResourceRef.java
+++ b/src/main/java/dev/jbang/source/ResourceRef.java
@@ -196,7 +196,14 @@ public class ResourceRef implements Comparable<ResourceRef> {
 											.resolve(Util.unkebabify(probe.getName()))
 											.toFile();
 					tempFile.getParentFile().mkdirs();
-					Util.writeString(tempFile.toPath().toAbsolutePath(), original);
+					if (!Util.createLink(tempFile.toPath().toAbsolutePath(), probe.toPath())) {
+						// Creating a link didn't work so we just create a copy
+						// TODO: Handle case of not being able to create link better.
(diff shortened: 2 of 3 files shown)


quintesse:
Darn, the link change doesn't work because now the watch doesn't detect any changes :-)

So what to do now @maxandersen ? Set the watch on the original folder perhaps?

maxandersen:
Wait - why does this change affect watch? Watch if java files work so why would that change for differently named files ?

quintesse:
Because right now they get copied to the jbang cache with a "proper" name and then _that_ copy is being used by `edit`. So the edits never end up back in the original file. This must have been like this for a long time already because that mechanism is pretty old.

So that's why I thought to change it to using links instead of making a copy, but then the watch doesn't work anymore (because links don't change when you edit the file they point to, obviously). So then the next step would be to not watch the temp project folder but the actual individual files. Because that way you can watch the original file (called `abc-def`) but give the link (named `AbcDef.java`) to the edit command.

It sounds all a bit complicated perhaps, but it should work.

maxandersen:
I'll need to look as I've been using edit for ages and the edit happens on original file not cache.

quintesse:
> I'll need to look as I've been using edit for ages and the edit happens on original file not cache.

Sure, but you perhaps never did on files with invalid names?

maxandersen:
im not questioning that invalid names have an issue just not grokking why file watching wouldn't work just because the filenames changes... myapp.java is also symoblically linked...not copied.

quintesse:
> myapp.java is also symoblically linked...not copied.

They're symbolically linked into the `cache/project` folder , yes. But _before_ that happens files with illegal names are _first_ copied to a file with a "proper" name under `cache/scripts`... and _then_ they are linked to `cache/project`. So it's a double indirection.

maxandersen:
okey I grok the issue now - yeah; there aren't a good real quick fix - we'll need to skip the caching when editing and then do rename via link.

quintesse:
> we'll need to skip the caching when editing

Hmm that's probably going to be ugly in some way because that's pretty deep down into the `ResourceRef` code. And when you start getting into those weird little opt-outs for special cases it often means you have the wrong abstraction.

So why not watch the original file (before rename) as I suggested? Wouldn't that be an option?

maxandersen:
Yes that could work.

maxandersen:
@quintesse you Wanna finish this one ?

maxandersen:
@quintesse still an issue you wanna fix? if yes - lets keep it open but if not lets close it to not take up mental space :)

quintesse:
I'm okay with closing it 👍

Issue #1019, opened 2021-09-26
jbang edit suggests incorrect command when template renames output file


grumpyf0x48:
With the following template info:

`
    "junit4b": {
      "file-refs": {
        "{basename}Test.java": "templates/junit4b.java.qute"
      },
      "description": "Basic template for JUnit4 tests"
    }
`

And the init command:

` jbang init --fresh --template=junit4b@jbangdev_local Script.java`

The output is:

[jbang] File initialized. You can now run it with 'jbang Script.java' or edit it using 'jbang edit --open=[editor] Script.java' where [editor] is your editor or IDE, e.g. 'netbeans'

which is puzzling because the created file is ScriptTest.java.

grumpyf0x48:
We could take the first rendered file into account, which works here, but what if the template renders several files ?

quintesse:
The best solution would be to use the one that has `{basename}` in its target name, because that is the one that will have the name the user provided. (If no such file should exist should perhaps just not show the edit message)

grumpyf0x48:
Yes but you can have several files using `{basename}`. Most of the time it won't happen, but you can do it.

maxandersen:
Pick the first one in that case.

maxandersen:
fixed in #1020

PR #1033, opened 2021-10-03
fix: Add itest for issue #1019

grumpyf0x48:
Add itest for issue #1019
Fixes: 1032, 1034

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -182,9 +182,9 @@ public class Init extends BaseScriptCommand {
 			if (optionalRefTarget.isPresent()) {
 				Path path = optionalRefTarget.get().to(outDir);
 				if (absolute) {
-					return path.toFile().getPath();
+					return path.toString();
 				} else {
-					return Util.getCwd().relativize(path).toFile().getPath();
+					return Util.getCwd().relativize(path).toString();
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/main/java/dev/jbang/cli/Init.java
```diff
@@ -177,7 +180,12 @@ String getRenderedScriptOrFile(Map<String, String> fileRefs, List<RefTarget> ref
 																															.getValue()))
 																.findFirst();
 			if (optionalRefTarget.isPresent()) {
-				return Util.getCwd().relativize(optionalRefTarget.get().to(outDir)).toFile().getPath();
+				Path path = optionalRefTarget.get().to(outDir);
+				if (absolute) {
+					return path.toFile().getPath();
```
quintesse, line 185: This looks like a no-op (code that does nothing).

quintesse, line 185: This can be simplified somewhat to `path.toString()`

src/main/java/dev/jbang/cli/Init.java
```diff
@@ -177,7 +180,12 @@ String getRenderedScriptOrFile(Map<String, String> fileRefs, List<RefTarget> ref
...
+				} else {
+					return Util.getCwd().relativize(path).toFile().getPath();
```
quintesse, line 187: Same here `.toString()` is enough

src/main/java/dev/jbang/cli/Init.java
grumpyf0x48: done

src/main/java/dev/jbang/cli/Init.java
grumpyf0x48: done

maxandersen (approved):

PR #1037, opened 2021-10-07
feat: Startup scripts now use User-Agent header for downloads

quintesse:
Has to be applied after https://github.com/jbangdev/jbang/pull/1036

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,11 +67,11 @@ tasks.withType(AbstractArchiveTask) {
 
 
 sonarqube {
-properties {
-	property "sonar.projectKey", "jbangdev_jbang"
-	property "sonar.organization", "jbangdev"
-	property "sonar.host.url", "https://sonarcloud.io"
-}
+	properties {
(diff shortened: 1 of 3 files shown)


maxandersen:
Wont These be problematic when using sh.jbang.dev and ps1.jbang.dev for direct download using main github content ? The version wont be extracted…

maxandersen:
this fix result in all scripts being missing from the archive so rolling it back to make working release as it breaks curl.

quintesse:
That's weird, it seemed to work locally. Have any idea why it falls?

maxandersen:
the dependsOn is not working as we think it does.

See this twitter thread https://twitter.com/CedricChampeau/status/1454788194789101568 that cover some of it.

quintesse:
I've added this as a Maybe to the 1.0 roadmap. Not essential but I think it would be good to have a proper User-Agent from 1.0 onwards.

quintesse:
Oops, this was the PR, removing again :-)

Issue #984, opened 2021-08-26
`jbang init` failing if script name has not a valid Java class name


sebastienblanc:
with v 0.70
`jbang init --template=qcli kubectl-august ` is failing : 
`[jbang] [ERROR] 'kubectl-august' is not a valid class name in java. Remove the special characters`

If I remove the --template parameter it's working. 

it was working with 0.62.

quintesse:
@maxandersen AFAICS you implemented this on purpose like this, the `TestInit` has specific tests for these situations:

- `jbang init invalid-name.java` this is okay
- `jbang init valid-name`  this isn't

The logic in the code and the tests seem to be that when you give an extension the name will be taken as explicit, while an extension-less name can be processed (ie kebab names will be converted to camel-case).

We can change this of course but I'm guessing you had a specific reason for implementing it like this?

quintesse:
Wait, I see that the original issue doesn't have a `.java` extension either. In that case it just seems to work fine in 0.79. Perhaps you can update and confirm @sebastienblanc ?

PR #1039, opened 2021-10-07
fix: Do not force .java suffix for Quarkus based templates

rsvoboda:
Do not force `.java` suffix for Quarkus based templates

Fixes https://github.com/jbangdev/jbang/issues/984 as in case `jbang init --template=qcli kubectl-august` file `kubectl-august.java` was created, but with default template `kubectl-august`

Checked all the 3 templates (qcli, qmetrics, qrest) and are working fine when `jbang kubectl-august` is called

Example for `qrest` based file:
```java
$ cat kubectl-august

// Update the Quarkus version to what you want here or run jbang with
// `-Dquarkus.version=<version>` to override it.
//DEPS io.quarkus:quarkus-bom:${quarkus.version:1.11.0.Final}@pom
// //DEPS io.quarkus:quarkus-smallrye-openapi
// //DEPS io.quarkus:quarkus-swagger-ui
//JAVAC_OPTIONS -parameters

import io.quarkus.runtime.Quarkus;
import javax.enterprise.context.ApplicationScoped;
import javax.ws.rs.GET;
import javax.ws.rs.Path;

@Path("/hello")
@ApplicationScoped
public class KubectlAugust {

    @GET
    public String sayHello() {
        return "Hello from Quarkus with jbang.dev";
    }

}
```

### DIFF
--- a/src/main/resources/jbang-catalog.json
+++ b/src/main/resources/jbang-catalog.json
@@ -17,15 +17,15 @@
       "description": "Basic kotlin Hello World template"
     },
     "qcli": {
-      "file-refs": {"{basename}.java":  "init-qcli.java.qute"},
+      "file-refs": {"{filename}":  "init-qcli.java.qute"},
       "description": "Quarkus CLI template"
     },
     "qmetrics": {
-      "file-refs": {"{basename}.java":  "init-qmetrics.java.qute"},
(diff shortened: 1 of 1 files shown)


maxandersen:
Thank you @rsvoboda - makes sense we loosen that up.

@all-contributors please add @rsvoboda  for code

PR #1042, opened 2021-10-13
feat: jbang direct edit

### LINKED ISSUE #462: Add option to generate IDE project and build files in current directory, instead of copying sources to temp dir
In the situation where you're working in a multi-file project, jbang's default handling of IDE workspaces uses a temporary directory to "serve"/regenerate IDE configurations.

JBang also copies the source script to this "virtual" workspace.

This is fine when you only have one source file, but jbang really needs to copy all source files, and ideally, it also needs to be able to sync files that are created by the user in the IDE once the project is open.

I think a solution would be to simply generate IDE files in the current directory. Let the user choose to omit them (via `.gitignore` or etc,) rather than hiding them away by default.

This could be enabled/disabled by a `--flag` when starting jbang.

```edit --open=code --live hello-world.java --workspace .```

### PR DESCRIPTION
**Pull request by maxandersen**
Fixes #462

work in progress but last time I lost it by accidental delete thus
opening this now :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -371,5 +371,5 @@ tasks.named("spotlessXml").configure { dependsOn("chocolatey") }
 tasks.named("spotlessXml").configure { dependsOn("copyITests") }
 
 group = "dev.jbang"
-sourceCompatibility = '8'
-targetCompatibility = '8'
+sourceCompatibility = "1.9"
+targetCompatibility = "1.9"
--- a/build.gradle
+++ b/build.gradle
@@ -371,5 +371,5 @@ tasks.named("spotlessXml").configure { dependsOn("chocolatey") }
 tasks.named("spotlessXml").configure { dependsOn("copyITests") }
 
 group = "dev.jbang"
-sourceCompatibility = "1.9"
-targetCompatibility = "1.9"
+sourceCompatibility = '8'
+targetCompatibility = '8'
--- a/docs/modules/ROOT/pages/editing.adoc
+++ b/docs/modules/ROOT/pages/editing.adoc
@@ -14,11 +14,18 @@ endif::[]
 
 toc::[]
 
-You can edit your script in an IDE/editor by using `jbang edit helloworld.java`. This will generate a project in a temporary location with symbolic links to your script and open a editor.
+You can edit your script in an IDE/editor by using `jbang edit helloworld.java`. This will generate a project in a temporary location with symbolic links to your script and open a editor. If your IDE supports JBang directly (see https://jbang.dev/ide) then you can open the folder instead and avoid the symbolic links and temporary location. That will generally be faster and more reliable.
 
 [source, bash]
 ----
 jbang edit helloworld.java
+----
+
+or 
+
+[source, bash]
+----
+jbang edit . helloworld.java
 ----
 
 By default `jbang` will offer to install https://vscodium.com[VSCodium] (free/libre version of Visual Studio code) with default java extensions enabled in so called https://code.visualstudio.com/docs/editor/portable["portable mode"]. Portable mode means all the installed binaries and configuration does not affect rest of your system; everything is stored in `~/.jbang/editor`.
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by maxandersen** [comment]
resetted to the simpler approach of jbang edit file will work as in past, but edit dir will assume the IDE understands jbang structured projects.

you can do `jbang edit . my.java` to have it open the folder and then open the specific file(s).

**Comment by maxandersen** [comment]
we should print info about jbang.dev/ide if ide is not vscodium that we installed.

then --single

**Comment by maxandersen** [comment]
maybe jbang edit --sandbox

**Comment by maxandersen** [comment]
updated to have edit always assume IDE is jbang enabled, and then `-b'/'--sandbox' as toggles to edit in "script" mode.

**Comment by maxandersen** [comment]
marking as draft as docs still needs updating but lets play with this and see if we like the feel of it.

**Comment by maxandersen** [comment]
A thing i'm torn on is wether the directory we open should just be parent file as is now or be smarter.

i.e.

```
jbang edit src/hello.java`
```

will open `src`

if you really meant to open where you are at you do:

```
jbang edit . src/hello.java`
```

then you force the location.

We could do some detection based on src name, package names etc. but they all kinda collapse if you do this:

```
cd ../..
jbang edit code/myproject/src/hello.java
```

should we use the current dir (.), parent dir (code/myprojet/src) or figure out the "project" is more likely `code/myproject` ?

currently it takes `code/myproject/src` ....which is also "wrong" but less hard and more deterministic ..but I could probably be convinced to use current dir.

**Comment by maxandersen** [comment]
new suggestion:

`jbang ../mymodule/src/a/b/my.java`

locate source root (using package) to find starting parent if it matches -> src
search through src and its parents for:

- jbang.build
- .jbang
- pom.xml
- build.gradle

if no hit on the above do second search for "project" ides:

- jbang.properties
- jbang-catalog
- .jbang (but not in home)
- .vscode
- .eclipse-project 
- intellij

stop and ignore home dir...

if that fails use parent of file.

**Comment by maxandersen** [comment]
add `--edit` to jbang init ?

**Comment by maxandersen** [comment]
do the above without any arguments?

if nothing found error

File: docs/modules/ROOT/pages/editing.adoc [hunk 1]
```diff
@@ -14,13 +14,20 @@ endif::[]
 
 toc::[]
 
-You can edit your script in an IDE/editor by using `jbang edit helloworld.java`. This will generate a project in a temporary location with symbolic links to your script and open a editor.
+You can edit your script in an IDE/editor by using `jbang edit helloworld.java`. This will generate a project in a temporary location with symbolic links to your script and open a editor. If your IDE supports JBang directly (see https://jbang.dev/ide) then you can open the folder instead and avoid the symbolic links and temporary location. That will generally be faster and more reliable.
```
quintesse, line 17: Is this explanation still relevant? Because by default the sandbox mode will never be used, right? Shouldn't it explain a bit that difference and suggest to use the sandbox option if you don't have an IDE with support for JBang?

docs/modules/ROOT/pages/editing.adoc
```diff
@@ -48,9 +55,26 @@ for your user or run your shell/terminal as administrator to have this feature w
 You can also use `jbang edit --live` and `jbang` will launch your editor while watching
 for file changes and regenerate the temporary project to pick up changes in dependencies.
 
-== IDE and Editor support
+== IDE Support
+
+Some IDE's have dedicated JBang support and is recommended to use those if you can; by using `jbang edit .` (the dot is important) to edit a directory directly. 
+Wit this approach you avoid the symbolic links and temporary location and simply have the IDE understand the implicit project structure as understood by JBang.
```
quintesse, line 61: [With]

docs/modules/ROOT/pages/organizing.adoc
```diff
@@ -76,7 +76,7 @@ Here `resource.properties` will be copied as is and `META-INF/resources/index.ht
 
 All locations are relative to the script location.
 
-WARNING: Currently `jbang edit` and http(s) based script do not work with `//FILES`. Will be added later.
+WARNING: Currently `` and http(s) based script do not work with `//FILES`. Will be added later.
```
quintesse, line 79: There's just empty quotes here?

docs/modules/ROOT/pages/editing.adoc
```diff
@@ -48,9 +55,26 @@ for your user or run your shell/terminal as administrator to have this feature w
...
+|image:{sicons}/visualstudiocode.svg[50,50] https://code.visualstudio.com[Visual Studio Code]
+|image:{sicons}/eclipseide.svg[50,50] https://www.eclipse.org/downloads/download.php[Eclipse]
+|image:{sicons}/intellijidea.svg[50,50] https://www.jetbrains.com/idea/download[IntelliJ Idea]
+|image:{sicons}/gitpod.svg[50,50] https://www.gitpod.io[GitPod]
+|===
+
+== Generic edit support
 
-The `edit` feature been tested with the following IDE's:
+The generic `edit` feature using temporary directory and symbolic links been tested with the following IDE's:
```
quintesse, line 77: Same here: shouldn't we mention that's called sandbox mode and that you have to enable it explicitly using an option?

build.gradle
```diff
@@ -373,5 +373,5 @@ tasks.named("spotlessXml").configure { dependsOn("chocolatey") }
 tasks.named("spotlessXml").configure { dependsOn("copyITests") }
 
 group = "dev.jbang"
-sourceCompatibility = '8'
-targetCompatibility = '8'
+sourceCompatibility = "1.9"
```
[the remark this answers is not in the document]
maxandersen: Hmm. That wasnt supposed to happen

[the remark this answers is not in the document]
maxandersen: reverted.

docs/modules/ROOT/pages/editing.adoc
maxandersen: updated with bit different text

docs/modules/ROOT/pages/editing.adoc
maxandersen: reworded

docs/modules/ROOT/pages/editing.adoc
maxandersen: reworded

docs/modules/ROOT/pages/editing.adoc
quintesse: Ok, and if I want to open that file directly (because it's a folder with a whole bunch of other scripts) with an IDE that supports JBang I'd have to use `jbang edit . helloworld.java`, right? It might be useful giving that example as well. Also, will `init` still say "use jbang edit newfile.java", etc? Because it would be nice if that actually told the user "use jbang edit . newfile.java" instead.

Issue #1056, opened 2021-10-31
Regression regarding issue 1019


grumpyf0x48:
Although the integration tests are still passing, it seems that changes in template code has consequences on the issue 1019 fix:

```
$ jbang template add --name test "{basename}Test.java=templates/test.java.qute" "{basename}SecondTest.java=templates/test.java.qute"
[jbang] Template 'test' added to '/home/pyfourmond/.jbang/jbang-catalog.json'

$ jbang init -t test /tmp/Script.java
[jbang] File initialized. You can now run it with 'jbang /tmp/Script.java' or edit it using 'jbang edit --open=[editor] /tmp/Script.java' where [editor] is your editor or IDE, e.g. 'idea'
```

The comparison on file paths no longer works (the use of `endsWith` method is hazardous and should be changed) and the initialized file displayed is erroneous (should be /tmp/ScriptTest.java).

I will submit a PR to fix that.

grumpyf0x48:
This issue has been fixed and can be closed.

PR #1057, opened 2021-10-31
fix: Regression regarding issue 1019

grumpyf0x48:
fix: Regression regarding issue 1019 (Issue #1056)

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -179,10 +179,11 @@ public class Init extends BaseScriptCommand {
 			Optional<RefTarget> optionalRefTarget = refTargets	.stream()
 																.filter(refTarget -> refTarget	.getSource()
 																								.getFile()
-																								.getPath()
-																								.endsWith(
-																										optionalFileRefEntry.get()
-																															.getValue()))
+																								.getName()
+																								.equals(
+																										new File(
+																												optionalFileRefEntry.get()
+																																	.getValue()).getName()))
 																.findFirst();
 			if (optionalRefTarget.isPresent()) {
 				Path path = optionalRefTarget.get().to(outDir);
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -178,12 +178,10 @@ public class Init extends BaseScriptCommand {
 		if (optionalFileRefEntry.isPresent()) {
 			Optional<RefTarget> optionalRefTarget = refTargets	.stream()
 																.filter(refTarget -> refTarget	.getSource()
-																								.getFile()
-																								.getName()
-																								.equals(
-																										new File(
-																												optionalFileRefEntry.get()
-																																	.getValue()).getName()))
+																								.getOriginalResource()
+																								.endsWith(
+																										optionalFileRefEntry.get()
+																															.getValue()))
 																.findFirst();
 			if (optionalRefTarget.isPresent()) {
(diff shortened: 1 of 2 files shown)


maxandersen:
what case is it that we broke ? can we add test for it so it does not happen again?

grumpyf0x48:
I discovered it was not working anymore when testing JUnit templates once merged in jbang-catalog:

```
$ jbang init -t junit@jbangdev Script.java
[jbang] File initialized. You can now run it with 'jbang Script.java' or edit it using 'jbang edit --open=[editor] Script.java' where [editor] is your editor or IDE, e.g. 'netbeans'
```

When i fixed issue 1019, i had tested that scenario with a similar template and it was ok.

So i don't know what broke the fix but the test validating that the output is correct already exists.

maxandersen:
My guess is the test are using absolute names not relative paths.

maxandersen:
hmm...no that didn't seem to make a difference. weird.

grumpyf0x48:
The file path for source RefTarget is different in itests and in manual run.

When running itests, i have:

```
refTarget.getSource().getFile().getPath()=/home/pyfourmond/Documents/Sources/mine/forks/jbang/build/itests/templates/test.java.qute
optionalFileRefEntry.get().getValue()=templates/test.java.qute
```

and the first value `endsWith` the second one.

If i use JBang junit template, i get:

```
refTarget.getSource().getFile().getPath()=/home/pyfourmond/.jbang/cache/urls/e4e45d1a0263c4b26597581c7a04f1be24675886fc1c6cac9fa8f93a7f5401db/junit.java.qute
optionalFileRefEntry.get().getValue()=templates/junit.java.qute
```

and the `endsWith`  fails.

That's why i now compare only the name (**test.java.qute** in the first example) to find the first rendered file.

maxandersen:
There is something wrong elsewhere here. 

I also noticed this morning that jbang edit filewithoutsuffix is again broken. 

Something is not behaving consistently when dealing with file names ...

@quintesse any ideas ?

quintesse:
> I also noticed this morning that jbang edit filewithoutsuffix is again broken.

Is it? It seems to work here. Do you have a way to reproduce that?

src/main/java/dev/jbang/cli/Init.java
```diff
@@ -179,10 +179,11 @@ String getRenderedScriptOrFile(Map<String, String> fileRefs, List<RefTarget> ref
 			Optional<RefTarget> optionalRefTarget = refTargets	.stream()
 																.filter(refTarget -> refTarget	.getSource()
 																								.getFile()
```
quintesse, line 181: Don't use `getFile()` here because that can be a cached/rewritten file that doesn't follow the original path and file names. That why the `endWith()` didn't work. Replace with 

```
.getSource()																						.getOriginalResource()																						.endsWith(...)
```

and things should work fine.

quintesse:
> @quintesse any ideas ?

@maxandersen yes, there isn't any big problem really, it's the implementation of `getRenderedScriptOrFile()` that was recently added that needs a bit of polishing.

Although perhaps I should take another look at the implementation of that command because right now retrieving that single bit of information needs an awful lot of code. It's probably one of those places where you need a bit of side-effects which means that the Stream API might not be the best solution. But with the suggestion I left it should work.

src/main/java/dev/jbang/cli/Init.java
grumpyf0x48: done. Works as expected. JUnit templates use now get correct output.

maxandersen:
thank you both!

PR #1067, opened 2021-11-07
Simpler symbolic link policy update instructions

quintesse:
It's actually pretty hard to extract the exact instructions from the MS documentation, so this patch just tells you what command to run (as an Administrator)


quintesse:
But it isn't enough *sigh*

quintesse:
Seems there isn't really an easy way. But the links given in the docs are just lousy. They don't really explain at all how to enable links.

This is perhaps because there all kinds of scary warnings that links should NOT be enabled for normal users!

Perhaps we should be looking into how to make JBang work without ever needing links....

quintesse:
I literally gave up after spending way too much time trying to get it to work on my wife's computer and downloaded the [Link Shell Extension](https://schinagl.priv.at/nt/hardlinkshellext/linkshellextension.html) that sets up everything correctly for you.

PR #1079, opened 2021-11-14
chore: Update release settings for JReleaser 0.9

aalmiray:
Update the JReleaser configuration ad GH workflows with latest changes required by v0.9

### DIFF
--- a/jreleaser.yml
+++ b/jreleaser.yml
@@ -11,6 +11,7 @@ project:
   authors:
     - Max Rydahl Andersen
   license: MIT
+  licenseUrl: https://github.com/jbangdev/jbang/blob/main/LICENSE
   tags:
     - jbang
     - bash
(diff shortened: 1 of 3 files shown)


aalmiray:
Converting to draft as JReleaser v0.9 is unreleased and JBang prefers using released versions.

jreleaser.yml
```diff
@@ -66,6 +67,7 @@ files:
 
 distributions:
   jbang:
+    executableExtension: cmd
```
maxandersen, line 70: what does this mean?

aalmiray: By default JReleaser uses .bat as the file extension for executables/launchers on Windows. JBang uses .cmd. The leading . is not necessary in the config

aalmiray:
@maxandersen feel free to merge any time now that JReleaser v0.9.0 is out.

maxandersen:
merged after bumping it to actually use 0.9 - thanks!

PR #1081, opened 2021-11-15
feat: CLI and aliases now allow all dependency options

quintesse:


quintesse:
The only thing I'm not sure about is whether the Alias JSON element for class paths should either be named `classpaths` or `class-paths` or `class-path`.

The CLI uses `--class-path` (`java` itself supports both `--class-path` and `-classpath`) while the code uses `classpaths` everywhere (suggesting it's a single word).

CLI naming follows the names in the sources, eg: `--deps` vs `//DEPS` but the JSON uses full naming: `dependencies`.

So I really have no idea what the best thing to do is, right now I've decided on `classpaths`, but I'll let you decide @maxandersen  :-)

maxandersen:
looks good Please add the tests from #692 which will test the lower level non-alias parts.

quintesse:
You mean these: https://github.com/jbangdev/jbang/pull/692/files#diff-1b82b18c947340fd87c42f19df84647fd965aec7bfd97da99ee36d9288fe85ea ?

maxandersen:
@quintesse yup

quintesse:
@maxandersen done

maxandersen (approved):
lgtm!

PR #1095, opened 2021-11-29
fix: JBang now correctly detects "fake" JDK on MacOS

### LINKED ISSUE #642: Java on new Mac is broken and jbang keeps trying to use it
/user/bin/java is broken on OS X. 

Jbang keeps picking it up but fails so doesn't even try install other option.

quintesse:
Fixes #642

Can someone with a Mac test this?

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -39,6 +39,10 @@ download() {
   fi
 }
 
+javacInPath() {
+  [[ -x "$(command -v javac)" && ( $os != "mac" || $(/usr/libexec/java_home &> /dev/null) ) ]]
+}
+
 abs_jbang_dir=$(dirname $(resolve_symlink $(absolute_path $0)))
 
 # todo might vary by os so can be overwritten below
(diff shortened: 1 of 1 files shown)


galderz:
Testing testing...

galderz:
```
❯ /opt/jbang/bin/jbang cache clear --all
Downloading JDK 11. Be patient, this can take several minutes...
Installing JDK 11...
mv: rename /Users/g/.jbang/cache/jdks/11.tmp/Contents/Home/*/* to /Users/g/.jbang/cache/jdks/11.tmp/*: No such file or directory
Error installing JDK
```

galderz:
I removed `~/.jbang` and still see the same error.

quintesse:
Ha, I'm wondering if this ever worked or that everybody who ever tested it on Mac already had Java pre-installed :-)

Anyway, could you see what the contents is of that `/Users/g/.jbang/cache/jdks/11.tmp/` folder? (Perhaps by copy&pasting the output of `tree /Users/g/.jbang/cache/jdks/11.tmp/` here).

quintesse:
Could you test the new version @galderz ? I think I found what the problem was.

galderz:
@quintesse That works now:

```
$ jbang version
Downloading JDK 11. Be patient, this can take several minutes...
Installing JDK 11...
[jbang] Default JDK set to 11
0.83.1.12
```

quintesse:
Well @maxandersen , it seems we can merge this

maxandersen:
I'll try on my wife laptop too first :D

One thing though - won't the max specific test return true if installing an oracle jdk manually and have javac in it ? Will we not then also download even though not needed ?

quintesse:
Well if it returns `true`, that's be good because it would mean there's a `javac` on the PATH *and* it's not fake.
But if manually downloading an Oracle JDK makes `/usr/libexec/java_home` fail (return != 0) then indeed it would be a problem.
I have no idea how all that stuff works on Mac, though.
The `java_home` only works for Apple-sanctioned installations?
If so we might have to try with `javac -version` instead (of `java_home`)...

maxandersen:
So I Just ran this on my own machine where `/usr/libexec/java_home` prints `/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home`

which is a proper java and there I get:

```
./jbang
JAVA_HOME is set but does not seem to point to a valid Java JDK
Downloading JDK 11. Be patient, this can take several minutes..
```

that ain't right.

p.s. javac -version wont work on the fake javac

can't we just check that java_home actually returns a directory ?

maxandersen:
btw. 

```
 /usr/libexec/java_home
/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
❯ echo $?
0
```

quintesse:
> that ain't right.

To get that message means that JBang found a `$HOME` var but it didn't point to a proper installation (meaning it couldn't find `$HOME/bin/javac`).

But that's basically a warning telling you your environment is not set up correctly. After that it will _still_ check for `javac` in the user's PATH.

Things is I just made a mistake in the test, that's what happens when I can't test it myself :-)

quintesse:
Try again please.

maxandersen:
@quintesse can you fix the conflict ?

quintesse:
Done.

PR #1102, opened 2021-12-01
chore: Typo in error message

ge0ffrey:


src/main/java/dev/jbang/cli/JBang.java
```diff
@@ -104,7 +104,7 @@ public int handleExecutionException(Exception ex, CommandLine commandLine, Comma
 			Util.errorMsg(null, ex);
 			if (Util.isVerbose()) {
 				Util.infoMsg(
-						"If you believe this a bug in jbang open issue at https://github.com/jbangdev/jbang/issues");
+						"If you believe this a bug in jbang, open an issue at https://github.com/jbangdev/jbang/issues");
```
quintesse, line 107: Perhaps even "If you believe this IS a bug"? The current version isn't wrong of course but perhaps this reads a bit easier, especially for non-native readers?

quintesse (approved):

src/main/java/dev/jbang/cli/JBang.java
ge0ffrey: Agreed, my fingers didn't type what my head was thinking. :)

PR #1104, opened 2021-12-01
fix: Improved `main()` method lookup

### LINKED ISSUE #1099: "Could not locate unique class. Found 0 candidates." when the first class does have main method.
To reproduce, take this gist:
  https://gist.github.com/ge0ffrey/9f4c558347f128790e774223f4a63fb3
And run it on:
  https://www.jdoodle.com/try-jbang

quintesse:
Before it would just assume that the `main()` method was located
in a class with the same name as the Java source file, but that's
not always the case. For example, when reading from stdin it's
hard to detect the right class and even more so when dealing with
non-public classes (non-public classes can still have `main()`
methods!)

Fixes #1099

### DIFF
--- a/src/main/java/dev/jbang/cli/BaseBuildCommand.java
+++ b/src/main/java/dev/jbang/cli/BaseBuildCommand.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.function.Predicate;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.regex.Pattern;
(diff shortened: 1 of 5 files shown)


src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -1250,6 +1250,30 @@ void multiSourcesHttp() throws IOException {
 
 	}
 
+	String script = "" +
+			"class One {\n" +
+			"}" +
+			"class Two {\n" +
+			"public static void main(String... args) { };\n" +
```
maxandersen, line 1257: when I remove this line the test still passes so its not actually testing the issue.

maxandersen: i.e. building is not the issue running is.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -411,56 +412,47 @@ protected static void searchForMain(ScriptSource src, RunContext ctx, File tmpJa
...
-
-					if (src.isAgent()) {
+				List<ClassInfo> mains = classes	.stream()
+												.filter(src.getMainFinder())
+												.collect(Collectors.toList());
+				String mainName = src.getSuggestedMain();
+				if (mains.size() > 1 && mainName != null) {
+					mains = mains.stream().filter(ci -> ci.simpleName().equals(mainName)).collect(Collectors.toList());
+				}
+				if (mains.size() > 1) {
```
maxandersen, line 433: can also be 0 so should be != as in previous version.

src/test/java/dev/jbang/cli/TestRun.java
maxandersen: actually it is build that is the issue but code and test doesn't catch it . i'll update the pr

src/main/java/dev/jbang/cli/BaseBuildCommand.java
maxandersen: actually if class is an agent it might be ok...

maxandersen:
not following how the build can fail ....works perfectly when I check it out.

maxandersen:
@quintesse can you confirm wether it fails for you too or github action being dumb?

quintesse:
@maxandersen no problem here, tried with Java 8 and 11

maxandersen:
GitHub actions, go home. you're drunk.

quintesse:
Hmmm... I just rebased against `main` locally and then I get the same errors (because the API has changed and your new tests haven't been updated yet). I just don't understand how that is happening though, the build should be against the PR code

quintesse:
Warning @maxandersen : I'm going to do a force-push of the rebase code !!

quintesse:
Well that seems to have "fixed" things 🤷‍♂️

quintesse:
DId you cancel @maxandersen or did something go wrong during the CI?

maxandersen:
Im making danish xmas food atm so it wasn't me.

quintesse:
Hmmm, is it already xmas overthere? :-)

maxandersen:
> Hmmm, is it already xmas overthere? :-)

It's December. It's Xmas 24-7

PR #1106, opened 2021-12-02
fix: `jdk` now works properly on bash in Windows

### LINKED ISSUE #1066: JBang running `jdk` in ming/cygwin/git bash returns unusable info
Running `jdk home` or `jdk java-env` return results that are unusable in the current shell:

```
$ jbang jdk home
C:\Users\foo\.jbang\currentjdk

$ jbang jdk java-env
set PATH=C:\Users\foo\.jbang\currentjdk\bin;%PATH%
set JAVA_HOME=C:\Users\foo\.jbang\currentjdk
rem Copy & paste the above commands in your CMD window or add
rem them to your Environment Variables in the System Settings.
```

We should detect that we're running in a bash-like environment and adjust our output. Eg. `C:\Users\foo\.jbang\currentjdk` should become `/c/Users/foo/.jbang/currentjdk`. And `java-env` should return the instructions we normally show for Linux.

quintesse:
Fixes #1066

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -125,6 +125,7 @@ class AppInstall extends BaseCommand {
 		if (Util.isWindows()) {
 			installCmdScript(binDir.resolve(name + ".cmd"), scriptRef, benative);
 			installPSScript(binDir.resolve(name + ".ps1"), scriptRef, benative);
+			installShellScript(binDir.resolve(name), scriptRef, benative);
 		} else {
 			installShellScript(binDir.resolve(name), scriptRef, benative);
 		}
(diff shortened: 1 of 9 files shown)


quintesse:
Still have to actually test this on Windows...

quintesse:
@maxandersen do you know why you ever added these lines: https://github.com/jbangdev/jbang/blob/main/src/main/scripts/jbang#L94-L98

Was there a specific reason to do that? Because right now these lines mess with the ability to detect in which Shell we're running :-)

quintesse:
Ok, already realized why: `jbang` doesn't know how to deal with Window paths.
Now, we could try to fix that, but that feels like a lot of work,
So perhaps the detection mechanism needs to change...

quintesse:
Hmm actually the problem is not in the jbang script...

quintesse:
Done. Can be merged now.

maxandersen:
What was the culprit ? Just adding more Shell specific code ?

quintesse:
It was this line: https://github.com/jbangdev/jbang/blob/dd7cf077dc23cf2ee40fd4cc4229f6968a792803/src/main/java/dev/jbang/util/Util.java#L1118

I had used of `getRoot().endsWith()` thinking I was dealing with a String (and having a `endsWith()` seemed to confirm that) but it's actually a `Path` and it doesn't like it when you pass `"C:/"` to it.

PR #1111, opened 2021-12-07
Better deps handling

quintesse:
Includes these two important commits:

> **refactor: Improved dependency handling …**
>
> `Source` dependency handling has been improved so that regardless of the sources or command line options like `--deps` we'll only do a single call to the dependency resolver.

and

> **refactor: Aether now resolves all deps in a single go …**
>
> This fixes any problems with duplicate and conflicting dependencies because we used to handle each dependency separately. The downside, for now, is that we can't log the dependency resolution to the output for the user to see what's happening.

### DIFF
--- a/itests/dependencies.feature
+++ b/itests/dependencies.feature
@@ -3,6 +3,6 @@ Feature: dependency fetching
 Scenario: fetch dependencies
 * command('jbang --verbose version')
 When command('jbang classpath_log.java', { JBANG_REPO: scratch + "/newrepo"})
-Then match err == '[jbang] Resolving dependencies...\n[jbang]     Resolving log4j:log4j:1.2.17...Done\n[jbang] Dependencies resolved\n[jbang] Building jar...\n'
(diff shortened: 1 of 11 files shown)


maxandersen:
This looks good and very much how I been thinking we would do it.

I've been investigating if we could get more fine grained output but we can't without adding more tweaks to jbangdev/jbang-resolver.

Users can though do this:

create log.properties file with:

```
# Specify the handlers to create in the root logger
# (all loggers are children of the root logger)
# The following creates two handlers
handlers= java.util.logging.ConsoleHandler

# Set the default logging level for new ConsoleHandler instances
java.util.logging.ConsoleHandler.level= FINEST

# Set global verbose level
#.level= INFO

# Set log verbose level for ShrinkWrap Resolvers
org.jboss.shrinkwrap.resolver.impl.maven.logging.LogTransferListener.level= FINE
#org.jboss.shrinkwrap.resolver.impl.maven.logging.LogRepositoryListener.level= FINEST
#org.jboss.shrinkwrap.resolver.impl.maven.logging.LogModelProblemCollector.level= FINEST
```

then run with 

```
JAVA_TOOL_OPTIONS=-Djava.util.logging.config.file=logging.properties jbang yourapp.java
```

and you get very noisy detailed output.

We can look into making that easier but for now more important we get more correct resolution.

PR #1112, opened 2021-12-08
fix: check for `javac` should now work on Mac

### LINKED ISSUE #642: Java on new Mac is broken and jbang keeps trying to use it

quintesse:
One character fix for inverted if-condition.

Fixes #642

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -40,7 +40,7 @@ download() {
 }
 
 javacInPath() {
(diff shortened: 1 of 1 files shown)


maxandersen:
weird windows fail?

quintesse:
Just GitHUb being weird it seems, reran the action and now everything is green 🤷

PR #1118, opened 2021-12-10
Better deps continued

quintesse:
This is a continuation of the work on #1111 

The `--deps` and `--repos` now also work when dealing with GAVs.

There's also a pretty decent refactor included that splits out the resolving of `ResourceRef`s out into a separate `ResourceResolver` that uses a plugin system where each type of resolving (remote, from classpath, from stdin, etc) is done by its own class. There's a bit more code now but it's much clearer who's responsible for what.

### DIFF
--- a/src/main/java/dev/jbang/source/ResourceResolver.java
+++ b/src/main/java/dev/jbang/source/ResourceResolver.java
@@ -1,0 +1,60 @@
+package dev.jbang.source;
+
+import java.util.function.Function;
+
+import dev.jbang.dependencies.ModularClassPath;
+import dev.jbang.source.resolvers.*;
+
+/**
+ * An interface used for analysing a resource string, resolving what exact file
+ * it refers to and returning a <code>ResourceRef</code> for it.
+ */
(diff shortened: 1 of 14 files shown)


src/main/java/dev/jbang/source/ResourceResolver.java
```diff
@@ -0,0 +1,60 @@
...
+	 * @param depResolver A function which, given a GAV string, returns a
+	 *                    <code>ModularClassPath</code>
+	 * @return A <code>ResourceRef</code> or <code>null</code>
+	 */
+	static ResourceResolver forScripts(Function<String, ModularClassPath> depResolver) {
+		return new CombinedResourceResolver(
+				new RenamingScriptResourceResolver(),
+				new StdinScriptResourceResolver(),
+				new RemoteResourceResolver(RemoteResourceResolver::fetchScriptFromUntrustedURL),
+				new ClasspathResourceResolver(),
```
maxandersen, line 43: what is this one actually used for ? have no recollection of loading classpath based resources ?

PR #1121, opened 2021-12-11
fix: Some regressions for Windows

quintesse:
Running inside a Bash shell on Windows we got a couple of regressions:

- the bash script needs to be able to perform the Windows update procedure. (The issue is that when using Cygwin/Git Bash on Windows and trying to update Jbang will download the new version but the switch to the new version won't actually be performed.)
-  `jbang edit` using wrong paths on Windows (All paths that get printed out to the output need to be passed through `Util.pathToString()`)


quintesse:
@maxandersen this should probably go in ASAP.

maxandersen:
thought i already merged it - now in place.

PR #1125, opened 2021-12-14
feat: Allow jbang to install itself

quintesse:
Normally when you run `jbang app install jbang` a fresh copy of the
latest version of Jbang will be downloaded, unpacked and installed.
But this patch allows the currently running jbang to be used as the
source for the files to install. This is done by supplying the
option `--offline` to prevent Jbang from downloading the latest version.


maxandersen:
What's the usecase for this One?
To test install using manually downloaded jbang?

quintesse:
> To test install using manually downloaded jbang?

Indeed, mostly made especially for you :-)

PR #1139, opened 2021-12-21
fix: support running markdown. Fixes #904

### LINKED ISSUE #904: Support literate programming via Mark down and asciidoc
Inspired by https://github.com/google/zx/blob/2.1.0/examples/markdown.md

Have super trivial extraction of code blocks in md and adoc files and run as jshell input.

Edit: The linked document was removed from later versions. I've updated the link to point to a much older version of the repo. -quintesse

maxandersen:
This will pick up code blocks with `` ```java`` and `` ```js`` markers. Anything else is ignored/commented out.

Would like to get asciidoc added but that's a bit trickier so marking that for later.


src/main/java/dev/jbang/source/resolvers/MarkupResourceResolver.java
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		if (probe != null && probe.canRead()) {
+			String ext = "." + Util.extension(probe.getName());
+			if (".md".equals(ext)) {
+				String scriptText = null;
+				try {
+					scriptText = Util.readString(probe.toPath());
+				} catch (IOException e) {
+					throw new IllegalStateException("Cuold not readed markup file " + probe.toPath());
```
quintesse, line 38: *Could

quintesse: *Readed

quintesse (approved):

maxandersen:
Yes - basically extract code and treat it as if it was from stdin. 

Need fix the windows test issue and verify it works on remote access Content too.

linux-china:
Could you add  ` ```jshelllanguage ` support for jsh?   IntelliJ IDEA use `jshelllanguage` for JShell file type, and it's very convenient to write jshell code with code completion. 

Screen Shot 2021-12-22 at 5 38 05 PM

maxandersen:
@linux-china makes sense - albeit "jshelllanguage" must be the worst  shortcut ever for a source syntax :)

maxandersen:
> IIUC what you basically do is collect all `'''java` and `'''jsh` code blocks, joining all the lines together as one big script and then treat that if is was a normal file without an extension, right? Which means it goes through the same heuristics for deciding if it's JShell or plain Java code. Seems reasonable.

so all works locally but when fetching remotely the resolvers does not seem to get engaged.

maxandersen:
mkay so this implementation only works when doing it on local file - wont work on remote URLs as we don't actually have a great way of dealing with that. same issue occurs for kotlin and groovy support; lots of small places you need to add some "if this prefix do x otherwise y".

I'll see if can find a way to make it feasible for this change ....

maxandersen:
Had to flip it around and not used the ResourceResolvers but instead treat .md as a separate language like kotlin to then let it transform the source, cache it and then run from there.

Definitely breaks jbang edit for remote files but that isn't new.

PR #1159, opened 2021-12-27
fix: rename ".kts" ot ".kt"

linux-china:
This bug  will stop Kotlin source to compile, and it was introduce by me to test kts file during Groovy PR.  Sorry about this fault.


maxandersen:
doh - I thought I fixed that already when merged. must been a  bad rebase.
Thanks for fixing it!

PR #1163, opened 2021-12-27
feat: add kotlin-stdlib-jdk8 as default dependency for Kotlin source 

linux-china:
`kotlin-stdlib-jdk8` is almost a must for Kotlin source code.   Introduce `org.jetbrains.kotlin:kotlin-bom:version@pom` and Kotlin other dependencies are not required to use explicit versions to avoid version confliction.

```kotlin

//DEPS org.jetbrains.kotlin:kotlin-reflect

fun main() {
    "First\nSecond".lines().forEach {
        println(it)
    }
}
```

### DIFF
--- a/itests/cache.feature
+++ b/itests/cache.feature
@@ -3,7 +3,7 @@ Feature: cache
 Scenario: clear cache default
 When command('jbang cache clear')
 * match exit == 0
-* match err == "[jbang] Clearing cache for urls\n[jbang] Clearing cache for jars\n[jbang] Clearing cache for scripts\n[jbang] Clearing cache for stdins\n[jbang] Clearing cache for deps\n"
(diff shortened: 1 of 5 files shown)


maxandersen:
No, this is not okey. Means users can't add any other boms nor remove/alter these.

We can add it to the kotlin template but as long as jbang only support one pom at a time this is not feasible to add as a forced set of dependencies.

maxandersen:
See in https://kotlinlang.org/docs/gradle.html#dependency-on-the-standard-library how gradle handles stdlib. This is a much more reasonable approach. Basically only added if user don't add it and option to turn off (albeit I would say we might want to do that with exclusions feature). 

We might not have the machinery to do this yet but that's what we should be doing over this pr which hard limits what you can use kotlin for with jbang.

linux-china:
Got.  It's better to add `kotlin-stdlib-jdk8` in script.

PR #1173, opened 2021-12-31
feat: support multiple bom poms

### LINKED ISSUE #877: Support multiple boms at the starting DEPS
Sometimes we might want to use multiple boms in a single script...

```java
//DEPS io.vertx:vertx-stack-depchain:4.0.3@pom
//DEPS org.apache.camel:camel-bom:3.9.0@pom
//DEPS io.vertx:vertx-core
//DEPS org.apache.camel:camel-core
//DEPS org.apache.camel:camel-stream
//DEPS org.apache.camel:camel-vertx
//DEPS org.slf4j:slf4j-simple:1.7.30
...
```

```console
$ jbang camel-vertx.java 
[jbang] Loading MavenCoordinate [io.vertx:vertx-stack-depchain:pom:4.0.3]
[jbang] [ERROR] POM imports as found in org.apache.camel:camel-bom:3.9.0@pom is only supported as the first import.
```

maxandersen:
ShrinkWrap does not have direct support for multiple pom files to use
for resolving versions for dependencies without version, i.e. called
`<dependencyManagement>` in Maven, platforms in Gradle.

To fix it we now generate a temporary pom.xml that has nothing but a
<dependencyManagement> section containg the list of managed pom dependencies.

This has the following consequences:

- we need to write a temporary file as ShrinkWrap does not allow passing a string directly.
- we avoid loading a bom pom as a pom which will activate parent poms features like adding repositories that generated warnings. Now this just don't happen.
- means that @pom references now can come in any order (before we required it to be the first)
- consequence of that is there in effect will be to `//DEPS` lists; one is pom's other is everything else.
- Also *could* resolve in change in behavior but as users only could load one pom.xml for this in past this should not be a problem;except maybe we don't generate so many warnings anymore if parent pom's were incomplete.

...and what this enables:

- having multiple bom poms which in particular is useful with larger frameworks like quarkus, springboot, groovy, kotlin which all rely on multiple boms.

Fixes #877

### DIFF
--- a/src/main/java/dev/jbang/dependencies/DependencyUtil.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyUtil.java
@@ -6,6 +6,9 @@ import static dev.jbang.util.Util.infoHeader;
 import static dev.jbang.util.Util.infoMsg;
 import static dev.jbang.util.Util.infoMsgFmt;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -32,6 +35,9 @@ import org.jboss.shrinkwrap.resolver.api.maven.coordinate.MavenCoordinates;
 
 import dev.jbang.Settings;
 import dev.jbang.cli.ExitException;
+import dev.jbang.util.Util;
+
+import picocli.CommandLine;
 
 public class DependencyUtil {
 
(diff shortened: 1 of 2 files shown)


maxandersen:
note, this really need some more manual/automatic testing before we merge it.

things to test:
- using multiple boms with quarkus  2+
- does jbang edit generate proper gradle file when using multiple boms?
- example from camel?

Issue #1175, opened 2021-12-31
AWS Lambda with JBang (and Quarkus) is not working properly


nandorholozsnyak:
### Motivation

I have recently started using AWS Lambda and experimenting with the possibilities to run Java code with them. 
I was attending projects with a few functions written in Python, I liked it, but with JBang I think me and other Java developers could do good things with Lambda as well.  
The goal would be to not create a lot of Maven/Gradle projects or overpopulating one Maven/Gradle project with tons of functions, with Quarkus I think it could be done with one project but with JBang I think it would be on another level. In our scenarios a lambda function is a "folder" with only one Python file, I would really like the idea if it could happen with our Java "script" files as well.
Of course, I would not like to focus only on AWS Lambda, but I'm having problems with that one right now.

#### Problems
Use this file as an example:
```java

///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS io.quarkus:quarkus-bom:2.6.1.Final@pom
//DEPS io.quarkus:quarkus-amazon-lambda
//JAVAC_OPTIONS -parameters
//JAVA_OPTIONS -Djava.util.logging.manager=org.jboss.logmanager.LogManager

import org.jboss.logging.Logger;

public class AwsLambda implements RequestHandler<Object, Object> {

    public static final Logger LOG = Logger.getLogger("AwsLambdaLogging");

    @Override
    public Object handleRequest(Object o, Context context) {
        LOG.info("Incoming object:" + o);
        return o;
    }
}

```
After calling the `jbang export portable AwsLambda.java` command JBang is going to create the `.jar` file and a folder called `libs`.
When this generated jar file is going to be uploaded as a Java function to AWS Lambda (with the handler class: `io.quarkus.amazon.lambda.runtime.QuarkusStreamHandler`) on execution the following error is going to happen:
```json
{
  "errorMessage": "Class not found: io.quarkus.amazon.lambda.runtime.QuarkusStreamHandler",
  "errorType": "java.lang.ClassNotFoundException"
}
```

The same would happen if the example would be using **_Funq_**, but then it would not be able to find this class: `io.quarkus.funqy.lambda.FunqyStreamHandler`

The problem is that the final `.jar` files does not contain the `libs` folder in itself, and if it would then it would not work as well because AWS Lambda requires this folder to be named `lib`.

Check out `Check for class file issues` section on: https://aws.amazon.com/premiumsupport/knowledge-center/lambda-troubleshoot-java-classnotfound/

[...]

PR #1176, opened 2021-12-31
fix: Renaming libs to lib when calling export portable

nandorholozsnyak:
- Option to put the lib folder into the final jar

### DIFF
--- a/src/main/java/dev/jbang/cli/Export.java
+++ b/src/main/java/dev/jbang/cli/Export.java
@@ -102,6 +102,7 @@ class ExportLocal extends BaseCommand implements Exporter {
 @Command(name = "portable", description = "Exports jar together with dependencies in way that makes it portable")
 class ExportPortable extends BaseCommand implements Exporter {
 
+	public static final String LIB = "lib";
 	@CommandLine.Mixin
 	ExportMixin exportMixin;
 
@@ -131,7 +132,7 @@ class ExportPortable extends BaseCommand implements Exporter {
 			try (JarFile jf = new JarFile(outputPath.toFile())) {
 				String cp = jf.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
 				String[] jars = cp == null ? new String[0] : cp.split(" ");
-				File libsdir = new File(outputPath.toFile().getParentFile(), "libs");
+				File libsdir = new File(outputPath.toFile().getParentFile(), LIB);
 				if (jars.length > 0) {
 					if (!libsdir.exists()) {
 						libsdir.mkdirs();
@@ -140,7 +141,7 @@ class ExportPortable extends BaseCommand implements Exporter {
 				StringBuilder newPath = new StringBuilder();
 				for (String jar : jars) {
 					Path file = downloadFile(new File(jar).toURI().toString(), libsdir);
-					newPath.append(" libs/" + file.toFile().getName());
+					newPath.append(" " + LIB + "/" + file.toFile().getName());
 				}
 
 				Path tempDirectory = Files.createTempDirectory("jbang-export");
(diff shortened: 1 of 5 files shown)


quintesse:
You're trying to create an "uber jar" that contains everything, right? But AFAIK doing it like this isn't correct. Did you try the resulting jar file on a clean system? I'm guessing it's really still obtaining them from your local file system. Also see: https://imagej.net/develop/uber-jars

quintesse:
Aha I just read https://github.com/jbangdev/jbang/issues/1175 so I understand what you're trying to do, but I don't think we should add an option for something that only works for AWS. So either we add a proper `export uberjar` option or I'd suggest you create a simple script that would turn the result from `export portable` into something that can be used with AWS.

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -162,6 +163,10 @@ public int apply(ExportMixin exportMixin, Source src, RunContext ctx) throws IOE
 			optionList.add("ufm");
 			optionList.add(outputPath.toString());
 			optionList.add(tempManifest.toString());
+            if (exportMixin.packageLibFolder) {
```
maxandersen, line 166: are you sure this is doing what you think it is ? just adding folder to the jar doesn't make it visible to the standard java class loader.

maxandersen (commented):
the change of `libs` to `lib` is fine but lets not mix that up with trying to make an uber-jar . The approach of adding lib/ folder to the jar I doubt works.

maxandersen (changes requested):
see comments

src/main/java/dev/jbang/cli/Export.java
nandorholozsnyak: That is true, you need a special class loader for it like what Spring Boot has (or at least I read about it). In this case I think it would be better to provide a new command which explicitely tells that it is going to create a .zip file that could be used on AWS Lambda. 
I think that would be cleaner, this first attempt was just a PoC. 

What do you think about a new command or new flags that will create a .zip with the lib folder and not the jar with the lib folder?

nandorholozsnyak:
@maxandersen I just deleted the packaging of the lib folder and made some changes to the documentation as well.
I can imagine it is going to cause some sort of backward compatibility issues if developers are expecting the existence the `lib` folder.

On the other topic, what would be the best way to implement this "zipping" mechanism with the `lib` folder? New command, or export command with --aws-function option ?

quintesse:
> On the other topic, what would be the best way to implement this "zipping" mechanism with the lib folder? New command, or export command with --aws-function option ?

My personal preference would be a simple `--zip` option to the `export portable` option, because that's basically what it does, the entire rest of the procedure is the same, except for the part where at the end a zip is created.

Personally I would stay away from AWS-specific naming unless there really is no way the export could be used by anything else but AWS. (And if that were the case I'd doubt the usefulness of adding such specific exports to the core Jbang code)

nandorholozsnyak:
> > On the other topic, what would be the best way to implement this "zipping" mechanism with the lib folder? New command, or export command with --aws-function option ?
> 
> My personal preference would be a simple `--zip` option to the `export portable` option, because that's basically what it does, the entire rest of the procedure is the same, except for the part where at the end a zip is created.
> 
> Personally I would stay away from AWS-specific naming unless there really is no way the export could be used by anything else but AWS. (And if that were the case I'd doubt the usefulness of adding such specific exports to the core Jbang code)

Okay, I can totally understand it, what you wrote earlier about creating a simple post processing on the end user end is still a good workaround. If the `--zip` gets green light then I will implement it on another branch.

maxandersen (approved):

maxandersen:
@all-contributors please add @nandorholozsnyak  for code

PR #1183, opened 2022-01-06
Use `tagName` instead of `projectVersion`

michael-simons:
GH-Action requires the `v`. Fixes the readme for the action repository.


maxandersen:
Thanks!

PR #1188, opened 2022-01-09
fix: config set/remove should tell what file it edits

### LINKED ISSUE #1184: jbang config set/unset etc. should tell what configuration it updated
example:

```
$ jbang config set -f . run.classpath src/main/resources
Updated ./jbang.properties
$ jbang config set run.deps g:a:v
Updated /Users/max/.jbang/jbang.properties
```

maxandersen:
fixes #1184


quintesse (approved):
LGTM

PR #1196, opened 2022-01-16
fix: failing `app setup` on Windows

### LINKED ISSUE #1195: Install fails downloading JDK on Windows with no existing java install, using Bash script; disco JDK endpoint URL incorrect?
**Describe the bug**

In a Windows 10, Git Bash shell (or cygwin I assume), I would like to install via:
>  curl -Ls https://sh.jbang.dev | bash -s - app setup

rather than using the PS scripts (since I'm not a PS user).  

Instead I get the following error:
 
> $ curl -Ls https://sh.jbang.dev | bash -s - app setup
> Downloading JBang...
> Installing JBang...
> Downloading JDK 11. Be patient, this can take several minutes...
> Error downloading JDK

**To Reproduce**

By trying and failing to recreate the problem on other Windows/Linux VMs, and by enabling bash trace, I see that the problem occurs specifically because:
1. I don't have JAVA_HOME or JAVA_EXEC set
2. I don't have any `javac` on my PATH, i.e. `command -v javac` in `javacInPath()` returns empty. 

So basically:
1. unset JAVA_HOME
2. unset JAVA_EXEC
3. make sure `command -v javac` returns empty (by adjusting PATH or however)
4. do the install: `curl -Ls https://sh.jbang.dev | bash -s - app setup`

### BAD URL - foojay disco 'directuris'
So, now that we see what's unique about this use case, we can see this takes us down the path to:
```bash
      jdkurl="https://api.foojay.io/disco/v2.0/directuris?distro=aoj&libc_type=$libc_type&archive_type=tar.gz&operating_system=$os&package_type=jdk&version=$javaVersion&release_status=ga&architecture=$arch&latest=available"
      download $jdkurl "$TDIR/bootstrap-jdk.tgz"
```

But if you do a ` curl -Lf` on this URL you'll get a 404.

If you hit:  **https://api.foojay.io/disco/v2.0/**
or **https://api.foojay.io/swagger-ui/** 
you'll see several other endpoints, but none aligned with:    `https://api.foojay.io/disco/v2.0/directuris`

**Expected behavior**
Just trying to install

**JBang version**
N/A

**Additional context**
See above

quintesse:
Fixes #1195

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -398,7 +398,7 @@ class AppSetup extends BaseCommand {
 		}
 		if (Util.getShell() == Util.Shell.bash) {
 			if (changed) {
-				System.out.println("Please start a new Shell for changes to take effect");
+				System.err.println("Please start a new Shell for changes to take effect");
 			}
 		} else {
 			if (changed) {


quintesse:
One of those things that are hard to test but it seems to work now :-)

src/main/java/dev/jbang/cli/App.java
```diff
@@ -398,7 +398,7 @@ public static int setup(boolean withJava, boolean force, boolean chatty) {
 		}
 		if (Util.getShell() == Util.Shell.bash) {
 			if (changed) {
-				System.out.println("Please start a new Shell for changes to take effect");
+				System.err.println("Please start a new Shell for changes to take effect");
```
maxandersen, line 401: What does that fix exactly ?

quintesse: I mistakenly believed that when running with a shell there would never be a need to execute code so I printed the text to stdout, but when running on Windows you actually need to run PowerShell anyway to be able to make changes to the Environment. So the text needs to be printed to stderr to avoid that the startup script tries to execute it.

maxandersen: okey, but is the url generated actually right ?

maxandersen:
It does make it work for me so merging.

PR #1198, opened 2022-01-16
fix: app install builds before installing scripts

### LINKED ISSUE #1192: jbang app install should ensure the alias can be built/resolved
this is to reduce users getting asked to trust a URL when they run the script rather than at install time.

maxandersen:
Trying to fix #1192 by just do a build of ref.


maxandersen:
this does fail for GAV's for some reason.

quintesse:
> this does fail for GAV's for some reason.

You need to use `buildIfNeeded()`

maxandersen:
> You need to use `buildIfNeeded()`

doh - of course.

PR #1200, opened 2022-01-17
fix: Fixed JDK download URLs for startup scripts

### LINKED ISSUE #1195: Install fails downloading JDK on Windows with no existing java install, using Bash script; disco JDK endpoint URL incorrect?

quintesse:
The startup scripts were not taking into account the `distro` argument
for the different java versions

Fixes #1195

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -39,6 +39,23 @@ download() {
   fi
 }
 
+unpack() {
+  if [[ "$1" == *.tar.gz ]]; then
+    gzip -cd "$1" | tar xf - -C "$2"
+    retval=$?
+    if [[ $os == mac && $retval -eq 0 ]]; then
+      mv "$TDIR/jdks/$javaVersion.tmp/"*/Contents/Home/* "$TDIR/jdks/$javaVersion.tmp/"
+      retval=$?
+    else
+      mv "$TDIR/jdks/$javaVersion.tmp/"*/* "$TDIR/jdks/$javaVersion.tmp/"
+    fi
+  else
+    unzip -qq -o $1 -d "$2"
+    retval=$?
+    mv "$TDIR/jdks/$javaVersion.tmp/"*/* "$TDIR/jdks/$javaVersion.tmp/"
+  fi
+}
+
 javacInPath() {
   [[ -x "$(command -v javac)" && ( $os != "mac" || $(/usr/libexec/java_home &> /dev/null) ) ]]
 }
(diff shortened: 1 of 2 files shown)


quintesse:
Pfff, you can't ask for *.tgz downloads on Windows. So the bash version of the startup script needs to be able to handle ZIP files.

maxandersen:
Makes sense. Now we are add it can we have it print out the attempted download url if things fails ? Helps debugging if it says more than just "error downloading jdk" :)

quintesse:
@maxandersen I know it might help debugging for us easier, but an error that spews out a very long and ugly URL is not going to do our users any good, I'd say...

quintesse:
This PR is now ready.

maxandersen (approved):
well that looks like a  nice and scary change :)

quintesse:
Yeah especially because we don't have any tests for this stuff 🙈 

We should have something that just does installations for all 3 main OSses and for a variety of versions.

Issue #1203, opened 2022-01-17
Extending templates with properties in catalog


nandorholozsnyak:
**Is your feature request related to a problem? Please describe.**
It is not a problem in my opinion, an idea, addition to the JBang toolset.

**Describe the solution you'd like**
In the current implementation of JBang the templates in the catalog are not describing, listing the available properties related to the templates and the description field is probably not enough if a JBang template is having a lot of property.
My proposal for this problem would be the following:  
In the `jbang-catalog.json` under the `templates` key each element could have the following new property: `properties`.

So for example, lets take a look on this template file:
```
{#if tf-providers}
terraform {
  required_version = ">= 0.14.0"
  required_providers {
    aws        = {
      source  = "hashicorp/aws"
      version = "{tf-provider.aws.version ?: '3.71.0'}"
    }
    archive    = {
      source  = "hashicorp/archive"
      version = "{tf-provider.archive.version ?: '2.2.0'}"
    }
    null = {
      source = "hashicorp/null"
      version = "{tf-provider.null.version ?: '3.1.0'}"
    }
  }
}
provider "aws" {
  region  = "{tf-provider.aws.region ?: 'eu-central-1'}"
}
{/if}

resource "aws_iam_role" "iam_for_lambda_{baseName}" {
  name = "iam_for_lambda_{baseName}"

  assume_role_policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "sts:AssumeRole",
      "Principal": {
        "Service": "lambda.amazonaws.com"
      },
      "Effect": "Allow",
      "Sid": ""
    }
  ]
}
EOF
}

resource "null_resource" "jar_file_{baseName}" {
  triggers = {
    file-hash = fileexists("{baseName}.java") ? filebase64sha256("{baseName}.java") : base64sha256("0")
  }
  provisioner "local-exec" {
    command = "sh generate-jar.sh {baseName}"
  }
}

resource "aws_lambda_function" "function_{baseName}" {
  filename         = "{baseName}.jar"
  source_code_hash = fileexists("{baseName}.jar") ? filebase64sha256("{baseName}.jar") : base64sha256("0")
  function_name    = "{baseName}"
  role             = aws_iam_role.iam_for_lambda_{baseName}.arn
  handler          = "io.quarkus.amazon.lambda.runtime.QuarkusStreamHandler::handleRequest"

  depends_on = [null_resource.jar_file_{baseName}]

  runtime = "java11"
[...]

PR #1205, opened 2022-01-18
feat: Catalog templates with property list

nandorholozsnyak:
- Property can have description and default value
- Adding multiple property at the same time with a specific format
- Listing templates returns the list of the template properties too

Associated issue: https://github.com/jbangdev/jbang/issues/1203

### DIFF
--- a/src/main/java/dev/jbang/catalog/Catalog.java
+++ b/src/main/java/dev/jbang/catalog/Catalog.java
@@ -59,7 +59,7 @@ public class Catalog {
 				new Alias(a.scriptRef, a.description, a.arguments, a.javaOptions, a.dependencies, a.repositories,
 						a.classpaths, a.properties, a.javaVersion, a.mainClass, this)));
 		templates.forEach((key, t) -> this.templates.put(key,
-				new Template(t.fileRefs, t.description, this)));
+				new Template(t.fileRefs, t.description, this, new HashMap<>())));
 	}
 
 	public static Catalog empty() {
--- a/src/main/java/dev/jbang/catalog/Catalog.java
+++ b/src/main/java/dev/jbang/catalog/Catalog.java
@@ -59,7 +59,7 @@ public class Catalog {
 				new Alias(a.scriptRef, a.description, a.arguments, a.javaOptions, a.dependencies, a.repositories,
 						a.classpaths, a.properties, a.javaVersion, a.mainClass, this)));
 		templates.forEach((key, t) -> this.templates.put(key,
-				new Template(t.fileRefs, t.description, this, new HashMap<>())));
+				new Template(t.fileRefs, t.description, t.properties, this)));
 	}
 
 	public static Catalog empty() {
--- a/src/main/java/dev/jbang/catalog/CatalogUtil.java
+++ b/src/main/java/dev/jbang/catalog/CatalogUtil.java
@@ -9,7 +9,10 @@ import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.AbstractMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
@@ -117,8 +120,9 @@ public class CatalogUtil {
 	/**
 	 * Adds a new template to the nearest catalog
 	 *
-	 * @param properties
 	 * @param name       The name of the new template
+	 * @param properties Properties that the template uses and can be used during
+	 *                   script initialization to customize the script.
 	 */
 	public static Path addNearestTemplate(String name, Map<String, String> fileRefs, String description,
 			Map<String, TemplateProperty> properties) {
@@ -130,9 +134,10 @@ public class CatalogUtil {
 	/**
 	 * Adds a new template to the given catalog
 	 *
-	 * @param properties
 	 * @param catalogFile Path to catalog file
 	 * @param name        The name of the new template
+	 * @param properties  Properties that the template uses and can be used during
+	 *                    script initialization to customize the script.
 	 */
 	public static Template addTemplate(Path catalogFile, String name, Map<String, String> fileRefs,
 			String description, Map<String, TemplateProperty> properties) {
@@ -145,7 +150,7 @@ public class CatalogUtil {
 															catalog.relativize(e.getValue())))
 													.collect(Collectors.toMap(AbstractMap.SimpleEntry::getKey,
 															AbstractMap.SimpleEntry::getValue));
-		Template template = new Template(relFileRefs, description, catalog, properties);
+		Template template = new Template(relFileRefs, description, properties, catalog);
 		catalog.templates.put(name, template);
 		try {
 			catalog.write();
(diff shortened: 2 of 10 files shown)


src/main/java/dev/jbang/catalog/Template.java
```diff
@@ -14,11 +14,14 @@
 	@SerializedName(value = "file-refs")
 	public final Map<String, String> fileRefs;
 	public final String description;
+	public final Map<String, TemplateProperty> properties;
 
-	public Template(Map<String, String> fileRefs, String description, Catalog catalog) {
+	public Template(Map<String, String> fileRefs, String description, Catalog catalog,
```
quintesse, line 19: This is not wrong of course, but could you change the order a bit? The `catalog` is meant to be the last argument in the list.

src/main/java/dev/jbang/catalog/Catalog.java
```diff
@@ -59,7 +59,7 @@ public Catalog(String baseRef, String description, ResourceRef catalogRef, Map<S
 				new Alias(a.scriptRef, a.description, a.arguments, a.javaOptions, a.dependencies, a.repositories,
 						a.classpaths, a.properties, a.javaVersion, a.mainClass, this)));
 		templates.forEach((key, t) -> this.templates.put(key,
-				new Template(t.fileRefs, t.description, this)));
+				new Template(t.fileRefs, t.description, this, new HashMap<>())));
```
quintesse, line 62: Use `Collections.emptyMap()` here.

src/main/java/dev/jbang/catalog/CatalogUtil.java
```diff
@@ -9,10 +9,7 @@
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.AbstractMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
quintesse, line 12: Make sure you run `./gradlew spotlessApply` before committing, no `*` imports are allowed.
I always run `./gradlew spotlessApply clean build installDist` to make sure all my code has been cleaned up.

src/main/java/dev/jbang/catalog/CatalogUtil.java
```diff
@@ -120,22 +117,25 @@ private static void removeAlias(Catalog catalog, String name) {
 	/**
 	 * Adds a new template to the nearest catalog
 	 *
-	 * @param name The name of the new template
+	 * @param properties
```
quintesse, line 120: Could you add some description here?

src/main/java/dev/jbang/catalog/CatalogUtil.java
```diff
@@ -120,22 +117,25 @@ private static void removeAlias(Catalog catalog, String name) {
...
 		Path catalogFile = Catalog.getCatalogFile(null);
-		addTemplate(catalogFile, name, fileRefs, description);
+		addTemplate(catalogFile, name, fileRefs, description, properties);
 		return catalogFile;
 	}
 
 	/**
 	 * Adds a new template to the given catalog
 	 *
+	 * @param properties
```
quintesse, line 133: Same here

src/main/java/dev/jbang/cli/Template.java
```diff
@@ -263,6 +313,22 @@ private static void printTemplate(PrintStream out, String catalogName, Catalog c
...
 			}
 		}
+		if (showProperties && template.properties != null) {
+			for (Map.Entry<String, TemplateProperty> entry : template.properties.entrySet()) {
+				out.print(Util.repeat(" ", indent));
+				StringBuilder propertyLineBuilder = new StringBuilder()
+																		.append(Util.repeat(" ", indent + 4))
+																		.append(ConsoleOutput.cyan(entry.getKey()))
+																		.append(" = ");
+				if (StringUtils.isNotBlank(entry.getValue().getDescription())) {
```
quintesse, line 323: Given the changes I suggested for the `TemplatePropertyConverter` I'd simply check for `null` here.

src/main/java/dev/jbang/catalog/TemplateProperty.java
```diff
@@ -0,0 +1,31 @@
+package dev.jbang.catalog;
+
+import com.google.gson.annotations.SerializedName;
+
+public class TemplateProperty {
+
+	private String description;
+	@SerializedName(value = "default-value")
```
quintesse, line 8: This is just my personal preference but I'd suggest simply using `default` instead of `default-value`.

quintesse (changes requested):
Looks good! Just needs a couple of minor changes.
It also needs tests and changes to the docs :-)

nandorholozsnyak:
Thank you for the feedback @quintesse . Will take a look on these. 
Just a quick feedback, in the Contribution file maybe some code standard and requirements should be visible, I understand I have to write the tests and make sure it is well formatted, but for the formatting issues I do remember I ran the spotless plugin before committing :D I'm just setting up the git hook for it to never happen again.
Is there any proper documentation about the code layering and structure?

quintesse (approved):
This would already be good enough to merge IMO.
But we need a couple of more tests, especially in `TestTemplate`, just to make sure that the information gets stored and retrieved correctly. A couple of edge cases perhaps (wrongly formatted property, things like that).

quintesse:
> Is there any proper documentation about the code layering and structure?

Not really, I'd just look at the surrounding code when you add/change things and try to follow that style as much as possible. And if you write something completely new then we tend to just let you write it as you want, as long as it is good code :-)

quintesse:
> This would already be good enough to merge IMO.

Once it builds and tests correctly of course 😉

maxandersen:
> > Is there any proper documentation about the code layering and structure?
> 
> 
> 
> Not really, I'd just look at the surrounding code when you add/change things and try to follow that style as much as possible. And if you write something completely new then we tend to just let you write it as you want, as long as it is good code :-)

...and we apply code formatting automatically so the way I code is just whatever I need with whatever formatting my ide uses - Then run the spotlessapply
And it will fix it. Avoids having to think about formatting :)

quintesse:
> And it will fix it

We should do something about the indenting, though. The fact that it tries to line up with the end of the first line (or whatever it is it tries to do) often creates some pretty horrible indents 😞

nandorholozsnyak:
I have never used Spotless before but .editorconfig. For me it was working well, of course if you are an Eclipse or Netbeans user you need a plugin for it in the IDE, but it has Gradle plugins as well. 
https://editorconfig.org/

quintesse:
@nandorholozsnyak Oh, Spotless works well enough, it's just the chosen configuration that I'm somewhat unhappy with :-)

nandorholozsnyak:
Hey @quintesse I added tests and small changes. I added the CLI docs as well, but I'm not sure if it would be enough or not.

nandorholozsnyak:
Btw, locally I use Java 17 (Temurin) and the build fails with that, so this is why I was not able to spot what is not working, Spotless is not working with that, or at least some things are not okay with that.
I ran the code you mentioned, now it should be working.

quintesse (approved):

quintesse:
Good job @nandorholozsnyak, thanks !

nandorholozsnyak:
Just a quick question @quintesse , did I just screw up the Codecov reports? I see that the master is failing because of the codecov report, and I just saw your new PR and checked the code as well and saw a lot of "misses" by the codecov. Should I open a PR to fix those?

quintesse:
Codecov is still in somewhat of an experimental phase for us right now, so don't worry too much about it. At some point we'll want to do something with that information though :-)

PR #1215, opened 2022-01-25
fix: headline levels

happyherp:
"Manual Install" and "Gradle Plugin" should not be under "Maven Plugin".


maxandersen:
Thank you @happyherp !

@all-contributors please add @happyherp  for docs

PR #1217, opened 2022-01-26
feat: Use dialog box for input

quintesse:
Detects when console input is not possible and will pop-up a dialog box instead.

This needs #1211 to be merged first.


maxandersen (changes requested):
lets make JBANG_GUI env that if false (and we for now assume its false by default) we work like before and if true we enable this behavior. Then we can release it and get some feedback before enabling it for every jbang install by switching to assume JBANG_GUI is true by default.

quintesse:
So how are we going to get feedback then if it's false by default?

quintesse:
You have to remember this will only kick in when somebody is using `curl | bash` _and_ they're trying to run something that isn't trusted. If something would go wrong they can still use the trust command to pre-trust and then the pop-up wouldn't show.

Or we should start creating beta versions for Jbang that people who like to live on the edge can install :-)

maxandersen:
> So how are we going to get feedback then if it's false by default?

By asking them enable it :)

maxandersen:
Btw. I didn't see it only happen using curl/pipe...are you sure ? I'll test again.

quintesse:
> Btw. I didn't see it only happen using curl/pipe...are you sure ? I'll test again.

It should definitely only happen when the stdin is not a shell/terminal.

quintesse:
> By asking them enable it :)

And how are you going to do that? Asking a bunch of people if they want to set `JBANG_GUI=true` in their environment? That's going to get as what? 3 people to test? :-)

Better to just ask them directly to download this version and test it for a bit. That way at least they won't forget to enable it.

But I think you're being a bit too careful :-)
If it works on my Windows and Linux machines and it works on your Mac we should be good.
If we notice that in some particular way it doesn't work we fix it and we release an update.
(Which is a good reason to also merge #1219 )

maxandersen:
btw. how are you testing this ?I can for the love of me not activate it :)

quintesse:
Hehe, one of the ways is simulating a curl by simply using cat:

```
cat build/install/jbang/bin/jbang | bash -s - edit itests/helloworld.java
```

quintesse:
Sorry, you actually need to be in the `bin` directory for that to work, so:

```
cd build/install/jbang/bin
cat jbang | bash -x -s - edit ../../../itests/helloworld.java
```

maxandersen:
okey so here is what I get 

:

notice the java duke icon..in the dialog AND in docker.

Can we get that to at leat be using jbang icon to  not look too javaish :)

maxandersen:
also needs rebase after #1211 got merged?

quintesse:
> notice the java duke icon..in the dialog AND in docker.

It's just a single method for opening a dialog with a single select.
It's not something that I designed, so I think it depends on your platform, because it doesn't look that in my systems.

image

quintesse:
Rebase done.

quintesse:
@maxandersen could you perhaps see if any of the other standard icon options give a better result?

https://github.com/jbangdev/jbang/blob/7bae2063651abbf542a0daf8ca46e0aa01c92301/src/main/java/dev/jbang/util/Util.java#L1365

It's easier if you test it manually instead of me making changes , committing and then asking you to test :-)

quintesse:
And it seems that the dock icon can only be changed with JDK 9+ : https://stackoverflow.com/a/56924202/2889926

maxandersen:
For Java 8+ we would need to set -Xdock:icon=/path/myIcon.png

maxandersen:
...I could fear that only work for some jdks though...

quintesse:
Yes, and it only works for MacOS and where would we get the icon from?

quintesse:
@maxandersen can you test if this new version works for you? There should now be Jbang icons both in the dialog and in the MacOS dock.

maxandersen:
works on java 11 at least

maxandersen (commented):
also tested on java 8 and here icon is great but task bar is duke but that I can live with :)

PR #1219, opened 2022-01-26
feat: "zero" install now lives up to its name

### LINKED ISSUE #1218: "zero install" should live up to its name
Right now when you do `curl sh.jbang.dev | bash -s -` you're basically performing a regular installation, the only difference being that no changes are being made to your environment (eg. PATH).

This is perhaps efficient because you don't have to download Jbang on each invocation, but it also means that we have a conflict when an older version Jbang is already installed. We'd either have to overwrite/update the version automatically, which the user might not want/expect, or we'd use the existing version, when the user perhaps expects access to the latest features.

I would suggest that we download to a temporary location that would be automatically cleaned up when Jbang exits. It's less efficient but it will make sure that we always run the latest version without any conflicts with installed Jbang versions.

quintesse:
We're not doing a "hidden install" anymore, but (inefficiently)
downloading Jbang to a temporary location and cleaning it up after
each invocation.

Fixes #1218

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -127,21 +127,15 @@ if [ -f "$abs_jbang_dir/jbang.jar" ]; then
 elif [ -f "$abs_jbang_dir/.jbang/jbang.jar" ]; then
   jarPath=$abs_jbang_dir/.jbang/jbang.jar
 else
-  if [[ ! -f "$JBDIR/bin/jbang.jar" || ! -f "$JBDIR/bin/jbang" ]]; then
-    echo "Downloading JBang..." 1>&2
-    mkdir -p "$TDIR/urls"
-    jburl="https://github.com/jbangdev/jbang/releases/latest/download/jbang.tar"
-    download $jburl "$TDIR/urls/jbang.tar"
-    if [ $retval -ne 0 ]; then echo "Error downloading JBang" 1>&2; exit $retval; fi
-    echo "Installing JBang..." 1>&2
-    rm -rf "$TDIR/urls/jbang"
-    tar xf "$TDIR/urls/jbang.tar" -C "$TDIR/urls"
-    if [ $retval -ne 0 ]; then echo "Error installing JBang" 1>&2; exit $retval; fi
-    mkdir -p "$JBDIR/bin"
-    rm -f "$JBDIR/bin/jbang" "$JBDIR/bin"/jbang.*
-    cp -f "$TDIR/urls/jbang/bin"/* "$JBDIR/bin"
-  fi
-  "$JBDIR"/bin/jbang "$@"
+  zidir=$(mktemp -d "$(basename $0).XXXXXXXXXX")
+  trap "rm -rf $zidir" EXIT
+  jburl="https://github.com/jbangdev/jbang/releases/latest/download/jbang.tar"
+  download $jburl "$zidir/jbang.tar"
+  if [ $retval -ne 0 ]; then echo "Error downloading JBang" 1>&2; exit $retval; fi
+  tar xf "$zidir/jbang.tar" -C "$zidir"
+  if [ $retval -ne 0 ]; then echo "Error unpacking JBang" 1>&2; exit $retval; fi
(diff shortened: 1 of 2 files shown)


quintesse:
WDYT about this @maxandersen . I feel this is a better way of handling the "zero install" issue.
I realized that the current way wasn't working as expected, for example with the GUI dialog popup PR I created.
For anyone running an older Jbang or for anyone who previously had run a `curl | bash` command the popup would not show unless they would first update their local Jbang to the latest version. That just felt really wrong.

So in this new version the startup scripts _only_ run if they find a local `jbang.jar` (either in the same folder of in a `.jbang` subfolder). In all other cases the zero-install kicks in and will download the latest Jbang to a temp folder, run jbang and then clean up afterwards. So the scripts will _not_ look anymore to see if you have Jbang installed and run that instead (which is what they do now).

maxandersen:
so just thinking it through - it makes sense; but side-effect is that doing curl-jbang twice now will "cost" more than before, right? not necessarily bad - but if the case it gets connected to how we currently recommend installing quarkus cli where we call curl twice - one for setup trust, other for a run...which is verbose but fast.

thus related to trust workflow.

quintesse:
It's indeed slightly slower to do two curls now, but Jbang isn't that big luckily.
Also, the moment you _must_ do two curl/jbang commands isn't it much better to just install jbang?
And thirdly, with the GUI popup we don't need to do two commands anymore (when not running remotely/in a container at least).

src/main/scripts/jbang
```diff
@@ -127,21 +127,15 @@ if [ -f "$abs_jbang_dir/jbang.jar" ]; then
...
-    echo "Installing JBang..." 1>&2
-    rm -rf "$TDIR/urls/jbang"
-    tar xf "$TDIR/urls/jbang.tar" -C "$TDIR/urls"
-    if [ $retval -ne 0 ]; then echo "Error installing JBang" 1>&2; exit $retval; fi
-    mkdir -p "$JBDIR/bin"
-    rm -f "$JBDIR/bin/jbang" "$JBDIR/bin"/jbang.*
-    cp -f "$TDIR/urls/jbang/bin"/* "$JBDIR/bin"
-  fi
-  "$JBDIR"/bin/jbang "$@"
+  zidir=$(mktemp -d "$(basename $0).XXXXXXXXXX")
```
maxandersen, line 130: whats the semantics for mktemp created directories? deleted at boot or ?

quintesse: No, they get deleted either by the `rm` on line 138 (if execution is all green) or by the `trap` statement on line 131 if anything goes wrong or the user cancels the job.

quintesse: And they get created in /tmp so if for some reason they still manage to evade any of that cleanup they will follow whatever cleanup is defined for your OS.

maxandersen: ack

maxandersen:
lets not merge this before we still have a working one-liner "app install" option that works, i.e.:

`curl -Ls https://sh.jbang.dev | bash -s - app install quarkus@quarkusio`

quintesse:
> lets not merge this before we still have a working one-liner "app install" option that works

Turns out that this worked all along, the `app install` already performs a setup if needed.

But while testing this I needed a clean environment so I started a container and realized that in those kind of situations it won't work anyway because a) we're redirecting stdin and b) there's no way to show a GUI popup.

And then I thought: what if we read directly from `/dev/tty`? It won't work on Windows (I tried using `con`, `con:` and `conIn$`, but none of these seemed to work) but that's fine, this is just so it will work in containers.

Therefore I created this PR: https://github.com/jbangdev/jbang/pull/1312 . We can merge this PR when that one is merged.

maxandersen (approved):

maxandersen:
merging this so can do a release with this in effect and test over the weekend :)

maxandersen:
had to revert this as "curl -Ls https://sh.jbang.dev | bash -s - app install --fresh --force quarkus@quarkusio" stopped working.

PR #1221, opened 2022-01-26
fix: `//SOURCES` now work in .jsh files too

### LINKED ISSUE #1220: `//SOURCE` doesn't do anything for .jsh scripts
It's seemingly ignored. Any mention of `//SOURCE` should probably simply be turned into extra file paths being passed to the `jshell` command.

quintesse:
Fixes #1220

### DIFF
--- a/src/main/java/dev/jbang/cli/Run.java
+++ b/src/main/java/dev/jbang/cli/Run.java
@@ -194,6 +194,8 @@ public class Run extends BaseBuildCommand {
 					}
 				}
 
+				optionalArgs.add("--execution=local");
+
 				if (!Util.isBlankString(classpath)) {
 					optionalArgs.add("--class-path=" + classpath);
 				}
--- a/src/main/java/dev/jbang/cli/Run.java
+++ b/src/main/java/dev/jbang/cli/Run.java
@@ -16,6 +16,7 @@ import org.apache.commons.text.StringEscapeUtils;
 
 import dev.jbang.Settings;
 import dev.jbang.source.RunContext;
+import dev.jbang.source.ScriptSource;
 import dev.jbang.source.Source;
 import dev.jbang.util.JavaUtil;
 import dev.jbang.util.Util;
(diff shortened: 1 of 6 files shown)


src/main/java/dev/jbang/cli/Run.java
```diff
@@ -194,6 +194,8 @@ String generateCommandLine(Source src, RunContext ctx) throws IOException {
 					}
 				}
 
+				optionalArgs.add("--execution=local");
```
maxandersen, line 197: what is that flag trying to achieve?

maxandersen: like, my jshell has no documentation for that flag afaics but it does not complain about it...

maxandersen:
should add a line or two about this behavior in the docs.

The semantics is that they are treated as includes or?

src/main/java/dev/jbang/cli/Run.java
quintesse: jshell, by default, starts up a VM in the background to which it connects. But this means you can't read stdin from that VM because it's not connected to the terminal. This option tells jshell to not do that and just use the local process for the VM, which then inherits the stdin from jshell. It even makes jshell a bit faster it seems.

quintesse:
> should add a line or two about this behavior in the docs.

What more do you want to say about it that the current docs aren't already explaining?

"The main script file defines all the dependencies and you add more source files into the application using `//SOURCES <filename>`. If included source has `//SOURCES` that will also get included recursively."

The current explanation is a bit vague perhaps but I think it's best not to go into too much technical detail. The sources just get included into the application, no need to know exactly how. Right?

src/main/java/dev/jbang/cli/Run.java
maxandersen: Hmmm...interesting...what's the downside ?

maxandersen:
Well; for java the ordering does not really matter as the classes just gets added to jar. 

In jshell ordering matters if they are treated as inclusion and executed in order. 

Basically if that is the semantics the description needs to be more specific than its vague now.

src/main/java/dev/jbang/cli/Run.java
quintesse: Not sure, the information on this is very limited. I'm assuming it's more efficient when you call jshell a lot.

quintesse:
Ok, added a bit of explanation

src/main/java/dev/jbang/cli/Run.java
maxandersen: i'm hesitant to do this specific change but lets merge in and we can revert if some usecase fails

PR #1225, opened 2022-01-30
Allow user to set JDK release

### LINKED ISSUE #1224: Allow user to set JDK release
Right now you can't install Java 18 yet, even though it's available, because it's an EA release instead of GA. So it would be nice to be able to override that with `JBANG_JDK_RELEASE`, just like there is a `JBANG_JDK_VENDOR`.

Secondly, the startup scripts should take these values into account as well (right now only the Java code knows about JBANG_JDK_VENDOR).

And finally is would perhaps be nice to accept some more vendor options.

quintesse:
Fixes #1224

### DIFF
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -61,11 +61,25 @@ it won't be as transparent. Here you will need to run using `jbang` directly, i.
 [WARNING]
 ====
 When no <<java-version,JDK version>> is available in the PATH, JDK 11 will be downloaded by default to bootstrap jbang.
-If your script requires a higher version and you don' want to download two JDK's, you can define an alternative default with `JBANG_DEFAULT_JAVA_VERSION` env variable.
+If your script requires a higher version and you don' want to download two JDK's, you can define an alternative default
+with `JBANG_DEFAULT_JAVA_VERSION` env variable.
 
  $ JBANG_DEFAULT_JAVA_VERSION=14 jbang my-script.java
 
 Note that if JDK is found in the PATH, `JBANG_DEFAULT_JAVA_VERSION` will be ignored.
+
+It's also possible to override the exact JDK distribution that will be used by setting the `JBANG_JDK_VENDOR` env variable.
+By default this will either be `temurin` or `aoj`, but it can also be set to `azul`, `debian`, `microsoft`, `openjdk`,
+`oracle`, `oracle_open_jdk` or `redhat`.
+
+By default JBang will only install final (GA) JDK releases, if you want to install an Early Access (EA) release you can do
+so by setting the `JBANG_JDK_RELEASE` env variable to `ea`.
+
+ $ JBANG_DEFAULT_JAVA_VERSION=18 JBANG_JDK_VENDOR=oracle JBANG_JDK_RELEASE=ea jbang my-script.java
+
+Note that if a JDK of matching version is already installed these env variables will have _no_ effect! They only take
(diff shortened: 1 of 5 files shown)


maxandersen:
How is This to be used ? Couldnt find a test ?

quintesse:
Well this is the same as `JBANG_JDK_VENDOR`, so there is no test, because it would mean testing the download of very specific JDKs which is very time dependent (so right now you can only download a Java 18 EA, but perhaps in a months time it is in GA and trying to get the EA version would fail).

But the way you use it is like this: `JBANG_JDK_RELEASE=ea jbang install jdk 18`.

Btw, the `JBANG_JDK_VENDOR` isn't actually documented anywhere, we should probably remedy that and at least explain what they do. They are _very_ specific options though that should only be used by people that really know what they're doing. (In this case I _really_ wanted to install Java 18 EA).

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -61,11 +61,25 @@ it won't be as transparent. Here you will need to run using `jbang` directly, i.
...
 
  $ JBANG_DEFAULT_JAVA_VERSION=14 jbang my-script.java
 
 Note that if JDK is found in the PATH, `JBANG_DEFAULT_JAVA_VERSION` will be ignored.
+
+It's also possible to override the exact JDK distribution that will be used by setting the `JBANG_JDK_VENDOR` env variable.
+By default this will either be `temurin` or `aoj`, but it can also be set to `azul`, `debian`, `microsoft`, `openjdk`,
+`oracle`, `oracle_open_jdk` or `redhat`.
+
+By default Jbang will only install final (GA) JDK releases, if you want to install an Early Access (EA) release you can do
```
maxandersen, line 75:
```suggestion
By default JBang will only install final (GA) JDK releases, if you want to install an Early Access (EA) release you can do
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -61,11 +61,25 @@ it won't be as transparent. Here you will need to run using `jbang` directly, i.
...
+so by setting the `JBANG_JDK_RELEASE` env variable to `ea`.
+
+ $ JBANG_DEFAULT_JAVA_VERSION=18 JBANG_JDK_VENDOR=oracle JBANG_JDK_RELEASE=ea jbang my-script.java
+
+Note that if a JDK of matching version is already installed these env variables will have _no_ effect! They only take
+effect when an actual JDK installation is performed. To force Jbang to install such a JDK one would first have to
```
maxandersen, line 81:
```suggestion
effect when an actual JDK installation is performed. To force JBang to install such a JDK one would first have to
```

maxandersen:
It would be nice //JAVA support something like `[vendor:][release]:17[.prefix]` in future instead of 3 different JBANG_* env vars...but I like this as a start. Will probably mark these env vars experimental.

maxandersen:
@quintesse can you rebase it ? conflicted with the cygwin fix.

maxandersen:
a shame this does not work but that's a disco API issue:

```
JBANG_JDK_VENDOR=oracle JBANG_JDK_RELEASE=ea-loom jbang --java 18
```

:)

quintesse:
> It would be nice //JAVA support something like [vendor:][release]:17[.prefix]

True, but that would probably introduce a bunch of issues that I definitely didn't want to deal with right now hehe

> a shame this does not work

Yeah, that would indeed be cool :-)

quintesse:
Rebase done

PR #1226, opened 2022-01-31
fix: Fixed Jbang failing to run correctly with Cygwin

### LINKED ISSUE #1223: windows cygwin shell fails for install
**Describe the bug**
found in https://stackoverflow.com/questions/70893995/jbang-on-cygwin-shell

```
curl -Ls https://sh.jbang.dev | bash -s - app install --fresh --force quarkus@quarkusio
Downloading JBang... 
Error downloading JBang
```

I  confirm this happens using cygwin shell. not sure why yet.

quintesse:
For Cygwin all kinds of path mangling is necessary, so reverting back
to calling `jbang.cmd` when we detect Cygwin. This does mean that
Jbang will fail to give proper usage instructions in certain cases
when running with Cygwin.

Fixes #1223

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -110,6 +110,12 @@ case "$(uname -m)" in
     ;;
 esac
 
+## when using cygwin fall out to just running the bat file.
+if [[ $os == windows && -f "$abs_jbang_dir/jbang.cmd" && "$(uname -s)" == CYGWIN* ]]; then
+  cmd /c "$(cygpath -m "$abs_jbang_dir"/jbang.cmd)" "$@"
+  exit $?
+fi
+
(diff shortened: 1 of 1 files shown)


maxandersen:
Proper usage instructions?

maxandersen:
Nm. Saw the issue.

maxandersen:
Why All the quote changes?

maxandersen:
quotes seem to be just cleanup. lets see how it goes :)

quintesse:
> quotes seem to be just cleanup. lets see how it goes

Yeah, it was just changes suggested by the IDE for dealing with spaces. Given that much of the data is paths they seemed like reasonable suggestions.

Issue #1231, opened 2022-02-07
Init command not using template properties's default value


nandorholozsnyak:
**Is your feature request related to a problem? Please describe.**
After this one implemented: https://github.com/jbangdev/jbang/issues/1203 a small feature, but a really important thing is still missing.
When an init command happens with a template flag and this template is having properties with default values they are not being applied to the final generated code.

**Describe the solution you'd like**
The `jbang init -t=<template_name>` command should take into account if the template is having properties with default values and apply to the final files. I would prefer a new option that would ignore the default values if need, maybe a `-i` and/or `--ignore-template-defaults`

**Additional context**

```terraform
{#if tf-providers}
terraform {
  required_version = ">= 0.14.0"
  required_providers {
    aws        = {
      source  = "hashicorp/aws"
      version = "{tf-provider.aws.version}"
    }
    archive    = {
      source  = "hashicorp/archive"
      version = "{archiveVersion}"
    }
    null = {
      source = "hashicorp/null"
      version = "{null-version}"
    }
  }
}
provider "aws" {
  region  = "{tf-provider.aws.region}"
}
{/if}
```
```json
...
"templates": {
        "q-aws-lambda-tf": {
            "file-refs": {
                "{filename}": "aws/aws-lambda.java.qute",
                "generate-jar.sh": "aws/generate-jar.sh",
                "generate-native-image.sh": "aws/generate-native-image.sh",
                "application.properties": "aws/application.properties",
                "lambda-aws.tf": "aws/lambda-aws.tf.qute"
            },
            "description": "Quarkus AWS Lambda template with Terraform template. Use the -Dnative-function flag to have native image based Terraform resources",
            "properties": {
                "tf-providers": {
                    "description": "If enabled extra Terraform related providers will be generated",
                    "default": false
                },
                "tf-provider.aws.version": {
                    "description": "Version of the AWS Terraform provider",
                    "default": "3.71.0"
                },
                "archiveVersion": {
                    "description": "Version of the Archive Terraform provider",
                    "default": "2.2.0"
                },
                "null-version": {
                    "description": "Version of the NULL Terraform provider",
                    "default": "3.1.0"
                },
                "tf-provider.aws.region": {
                    "description": "AWS Region",
                    "default": "eu-central-1"
                },
                "native-function": {
                    "description": "Generate native executable based lambda function or not",
                    "default": false
                }
            }
        }
    }
...
```

Result:
```terraform
terraform {
  required_version = ">= 0.14.0"
[...]

PR #1232, opened 2022-02-07
feat: Init command not using template properties's default value

nandorholozsnyak:
- Removing getter and setter pairs from TemplateProperty and using public modifier
- Applying template property defaults in init command by default, option to disable.
- Documentation for the new option

Solves this: https://github.com/jbangdev/jbang/issues/1231

### DIFF
--- a/docs/modules/cli/pages/jbang-init.adoc
+++ b/docs/modules/cli/pages/jbang-init.adoc
@@ -65,6 +65,10 @@ Initialize a script.
 *--quiet*::
   jbang will be quiet, only print when error occurs.
 
+*-i*, *--ignore-template-defaults*=_<true|false>_::
+  If *true* and a template is given with properties that have default values, these will not be applied in the render process.
+  By default *false*.
+
 *-t*, *--template*=_<initTemplate>_::
   Init script with a java class useful for scripting
 
--- a/docs/modules/cli/pages/jbang-init.adoc
+++ b/docs/modules/cli/pages/jbang-init.adoc
@@ -65,10 +65,6 @@ Initialize a script.
 *--quiet*::
   jbang will be quiet, only print when error occurs.
 
-*-i*, *--ignore-template-defaults*=_<true|false>_::
-  If *true* and a template is given with properties that have default values, these will not be applied in the render process.
-  By default *false*.
-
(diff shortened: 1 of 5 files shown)


quintesse:
Personally I don't see much use for an option that doesn't apply default values. Do you @maxandersen ?

maxandersen:
Yeah. Same. No need to be able to turn this off.

docs/modules/cli/pages/jbang-init.adoc
```diff
@@ -65,6 +65,10 @@ Initialize a script.
 *--quiet*::
   jbang will be quiet, only print when error occurs.
 
+*-i*, *--ignore-template-defaults*=_<true|false>_::
```
maxandersen, line 68: I don't see value in having this flag. why would you not want the defaults applied?

src/main/java/dev/jbang/catalog/TemplateProperty.java
```diff
@@ -9,31 +9,15 @@
  */
 public class TemplateProperty {
 
-	private String description;
+	public String description;
```
maxandersen, line 12: why this change?

src/main/java/dev/jbang/cli/Init.java
```diff
@@ -43,6 +44,10 @@
 			"--deps" }, converter = CommaSeparatedConverter.class, description = "Add additional dependencies (Use commas to separate them).")
 	List<String> dependencies;
 
+	@CommandLine.Option(names = { "-i",
```
maxandersen, line 47: just remove it?

maxandersen (dismissed):
I pushed update that removes the flag. if pass ci good to go.

src/main/java/dev/jbang/cli/Init.java
```diff
@@ -109,6 +112,16 @@ public Integer doCall() throws IOException {
 		return EXIT_OK;
 	}
 
+	private void useTemplatePropertiesIfAllowed(dev.jbang.catalog.Template tpl) {
```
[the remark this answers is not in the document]
maxandersen: fixed.

nandorholozsnyak:
Sorry for not responding for the change requests, I'm on "holiday", thank you for the changes and merge.

Issue #1248, opened 2022-02-18
add `jbang jdk install 17 <path>`


maxandersen:
to enable using `jbang jdk install 17  ~/.sdkman/candidates/java/17-open/` to reuse existing sdkman (or any other systems java)

maxandersen:
more specifically to do:

    jbang jdk install 17 `sdk home java 11.0.9.hs-adpt`

quintesse:
Should be doable 🤔

nandorholozsnyak:
I started to work on it, but I see a few problems with this approach, although I like it.

The first approach is easy `jbang jdk install 17 ~/.sdkman/candidates/java/17-open/` - the codes checks if the input parameter is a folder and validates the specific java version using the `release` file, after that a soft link will be created to that folder, I think it is a valid solution.

BTW, I think this code part is "broken", but I may not get something with that: `dev.jbang.util.JavaUtil#parseJavaOutput`
```
public static int parseJavaOutput(String output) {
		if (output != null) {
			Matcher m = javaVersionPattern.matcher(output);
			if (m.find() && m.groupCount() == 2) {
				return parseJavaVersion(m.group(1));
			}
		}
		return 0;
	}
```
The if should be `m.find() && m.groupCount() == 1`, for me it is constantly returning 0 for these inputs `JAVA_VERSION="11.0.14"`. After the change it will return 11. - And the `dev.jbang.util.JavaUtil#determineJavaVersion(java.nio.file.Path)` only works because it has a fallback with the System.getProperty. 

The second approach could work as well, I will try to implement it ASAP, but my only question is the expression will be executed before JBang starts or should JBang resolve it?

nandorholozsnyak:
It does resolve the expression so now it is easier :)

nandorholozsnyak:
Just opened a PR for it: https://github.com/jbangdev/jbang/pull/1252
Tests are missing, will be making them soon.

PR #1252, opened 2022-02-23
feat: add jbang jdk install <version> <path>

nandorholozsnyak:
- Ability to link pre-installed JDK directory to JBang
- Documentation update

### DIFF
--- a/docs/modules/cli/pages/jbang-jdk-install.adoc
+++ b/docs/modules/cli/pages/jbang-jdk-install.adoc
@@ -19,7 +19,7 @@ jbang-jdk-install - Installs a JDK.
 // tag::picocli-generated-man-section-synopsis[]
 == Synopsis
 
-*jbang jdk install* *-o* [*-f*] *--fresh* *--quiet* *--verbose* _version_
+*jbang jdk install* *-o* [*-f*] *--fresh* *--quiet* *--verbose* _version_ _[existingJdkPath]_
 
 // end::picocli-generated-man-section-synopsis[]
 
@@ -55,6 +55,9 @@ Installs a JDK.
 
 _version_::
   The version to install
+
+_[existingJdkPath]_::
+  Already existing JDK's path on the filesystem. If path is given, its version will be checked against the incoming one, if not the same, the process will not finish successfully.
 
 // end::picocli-generated-man-section-arguments[]
 
--- a/docs/modules/cli/pages/jbang-jdk-install.adoc
+++ b/docs/modules/cli/pages/jbang-jdk-install.adoc
@@ -57,7 +57,7 @@ _version_::
   The version to install
 
 _[existingJdkPath]_::
-  Already existing JDK's path on the filesystem. If path is given, its version will be checked against the incoming one, if not the same, the process will not finish successfully.
+  Already existing JDK's path on the filesystem. If path is given, its version will be checked against the incoming one, and if the same, the given JDK will be linked successfully.
 
 // end::picocli-generated-man-section-arguments[]
 
--- a/src/main/java/dev/jbang/cli/Jdk.java
+++ b/src/main/java/dev/jbang/cli/Jdk.java
@@ -7,6 +7,8 @@ import java.io.PrintStream;
 import java.nio.file.Path;
 import java.util.Set;
 
+import org.apache.commons.lang3.StringUtils;
+
 import dev.jbang.Settings;
 import dev.jbang.net.JdkManager;
 import dev.jbang.util.Util;
@@ -23,10 +25,15 @@ public class Jdk {
 	public Integer install(
 			@CommandLine.Option(names = { "--force",
 					"-f" }, description = "Force installation even when already installed") boolean force,
-			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version)
+			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version,
+			@CommandLine.Parameters(paramLabel = "existingJdkPath", index = "1", description = "Pre installed JDK path or command to evaluate", arity = "0..1") String path)
 			throws IOException {
 		if (force || !JdkManager.isInstalledJdk(version)) {
-			JdkManager.downloadAndInstallJdk(version);
+			if (StringUtils.isNotBlank(path)) {
+				JdkManager.linkToExistingDirectory(path, version);
+			} else {
+				JdkManager.downloadAndInstallJdk(version);
+			}
 		} else {
 			Util.infoMsg("JDK " + version + " is already installed");
 		}
--- a/src/main/java/dev/jbang/cli/Jdk.java
+++ b/src/main/java/dev/jbang/cli/Jdk.java
@@ -26,11 +26,11 @@ public class Jdk {
 			@CommandLine.Option(names = { "--force",
 					"-f" }, description = "Force installation even when already installed") boolean force,
 			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version,
-			@CommandLine.Parameters(paramLabel = "existingJdkPath", index = "1", description = "Pre installed JDK path or command to evaluate", arity = "0..1") String path)
+			@CommandLine.Parameters(paramLabel = "existingJdkPath", index = "1", description = "Pre installed JDK path", arity = "0..1") String path)
 			throws IOException {
 		if (force || !JdkManager.isInstalledJdk(version)) {
 			if (StringUtils.isNotBlank(path)) {
-				JdkManager.linkToExistingDirectory(path, version);
+				JdkManager.linkToExistingJdk(path, version);
 			} else {
 				JdkManager.downloadAndInstallJdk(version);
 			}
--- a/src/main/java/dev/jbang/cli/Jdk.java
+++ b/src/main/java/dev/jbang/cli/Jdk.java
@@ -7,8 +7,6 @@ import java.io.PrintStream;
 import java.nio.file.Path;
 import java.util.Set;
 
-import org.apache.commons.lang3.StringUtils;
-
 import dev.jbang.Settings;
 import dev.jbang.net.JdkManager;
 import dev.jbang.util.Util;
@@ -29,7 +27,7 @@ public class Jdk {
 			@CommandLine.Parameters(paramLabel = "existingJdkPath", index = "1", description = "Pre installed JDK path", arity = "0..1") String path)
 			throws IOException {
 		if (force || !JdkManager.isInstalledJdk(version)) {
-			if (StringUtils.isNotBlank(path)) {
+			if (!Util.isNullOrBlankString(path)) {
 				JdkManager.linkToExistingJdk(path, version);
 			} else {
 				JdkManager.downloadAndInstallJdk(version);
--- a/src/main/java/dev/jbang/net/JdkManager.java
+++ b/src/main/java/dev/jbang/net/JdkManager.java
@@ -19,8 +19,6 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.apache.commons.io.FileUtils;
-
 import dev.jbang.Cache;
 import dev.jbang.Settings;
 import dev.jbang.cli.ExitException;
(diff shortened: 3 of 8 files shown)


src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -81,6 +81,7 @@ private static int determineJavaVersion(Path javaCmd) {
 		String output = Util.runCommand(javaCmd.toString(), "-version");
 		int version = parseJavaOutput(output);
 		if (version == 0) {
+			Util.verboseMsg("Version is 0, reading it from java.version property");
```
quintesse, line 84: say something like "Version could not be determined from: '$javaCmd -version', trying 'java.version' property"

src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -104,7 +105,7 @@ public static Path getJdkHome() {
 	public static int parseJavaOutput(String output) {
 		if (output != null) {
 			Matcher m = javaVersionPattern.matcher(output);
-			if (m.find() && m.groupCount() == 2) {
+			if (m.find() && m.groupCount() == 1) {
```
quintesse, line 108: That does indeed seem to have been wrong all this time, good catch!

quintesse (changes requested):
Looks good in general! Good job!

quintesse:
Oh and add "Fixes #1248" to the end of the commit message, then GitHub will automatically close the issue when your PR gets merged (you probably know that and just forgot 😄)

docs/modules/cli/pages/jbang-jdk-install.adoc
```diff
@@ -56,6 +56,9 @@ Installs a JDK.
 _version_::
   The version to install
 
+_[existingJdkPath]_::
+  Already existing JDK's path on the filesystem. If path is given, its version will be checked against the incoming one, if not the same, the process will not finish successfully.
```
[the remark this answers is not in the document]
nandorholozsnyak: added could be linked maybe?

src/main/java/dev/jbang/cli/Jdk.java
```diff
@@ -23,10 +25,15 @@
 	public Integer install(
 			@CommandLine.Option(names = { "--force",
 					"-f" }, description = "Force installation even when already installed") boolean force,
-			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version)
+			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version,
+			@CommandLine.Parameters(paramLabel = "existingJdkPath", index = "1", description = "Pre installed JDK path or command to evaluate", arity = "0..1") String path)
```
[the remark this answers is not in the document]
nandorholozsnyak: It is just a small "typo", first I thought we should execute the expression, but cmd will do it, fixing it.

src/main/java/dev/jbang/net/JdkManager.java
```diff
@@ -203,6 +207,40 @@ public static void uninstallJdk(int version) {
...
+		if (!Files.isDirectory(linkedJdkPath)) {
+			throw new ExitException(EXIT_INVALID_INPUT, "Unable to resolve path as directory: " + path);
+		}
+		Optional<Integer> ver = resolveJavaVersionFromPath(linkedJdkPath);
+		if (ver.isPresent() && ver.get() == version) {
+			Util.createLink(jdkPath, linkedJdkPath);
+			Util.infoMsg("JDK " + version + " has been linked to: " + linkedJdkPath);
+		} else {
+			throw new ExitException(EXIT_INVALID_INPUT, "Java version in given path: " + path
+					+ " is not having the requested version " + version + ", please provide a proper path");
```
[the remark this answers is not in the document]
nandorholozsnyak: It may not be infer the X value, cause this path could point to anywhere, it can have no Java inside it, I can do an evaluation and if it does not work then fallback to the original message if it is okay.

nandorholozsnyak:
Mentioned things are fixed, I'm coming soon with the tests.

nandorholozsnyak:
Coverage should be a bit better now.

nandorholozsnyak:
Any cool idea how to fix this error happening with the Windows build?
```
TestJdk > testJdkInstallWithLinkingToExistingJdkPathWhenJBangManagedVersionDoesNotExist(File) FAILED
    java.io.IOException at TempDirectory.java:262

TestJdk > testJdkInstallWithLinkingToExistingJdkPathWhenJBangManagedVersionExistsAndInstallIsForced(File) FAILED
    java.io.IOException at TempDirectory.java:262
^"C:\Users\RUNNER~1\AppData\Local\Temp\junit5745621985287379323\jbang\cache\jdks\11\bin\java.exe^" -Dfoo=bar ^"-Dbar=aap^ noot^ mies^" -classpath ^"D:\a\jbang\jbang\build\classes\java\test\itests\hellojar.jar^" helloworld
java -Dfoo=bar ^"-Dbar=aap^ noot^ mies^" -classpath ^"C:\Users\RUNNER~1\AppData\Local\Temp\junit8100379350[68](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:68)2[69](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:69)4745\jbang\cache\jars\helloworld.java.6c29e59df0efda6a7a9783b1de6b823dc03[71](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:71)000e3c8d32b2b143a430c20900b.jar^" ^"-XX:ArchiveClassesAtExit=C:\Users\RUNNER~1\AppData\Local\Temp\junit8100379350682694[74](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:74)5\jbang\cache\jars\helloworld.java.6c29e59df0efda6a7a9783b1de6b823dc0371000e3c8d32b2b143a430c20900b.jar.jsa^" helloworld

TestJdkManager > testResolveJavaVersionFromPathWhenJavaVersionHasNewVersioning(File) FAILED
    java.io.IOException at TempDirectory.java:262

TestJdkManager > testResolveJavaVersionFromPathWhenJavaVersionHasOldVersioning(File) FAILED
    java.io.IOException at TempDirectory.java:262
```

docs/modules/cli/pages/jbang-jdk-install.adoc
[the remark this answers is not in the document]
quintesse: "linked" is good 👍

quintesse:
> Any cool idea ...

Is that error for real? Because that path looks _very_ wrong! There are URLs and what not inside of it.

nandorholozsnyak:
Hm, on a Windows based laptop the build is fine. Any guesses to the TempDir? 
Which path is not looking good @quintesse ?

quintesse:
This one :-)

```
"C:\Users\RUNNER~1\AppData\Local\Temp\junit8100379350[68](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:68)2[69](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:69)4745\jbang\cache\jars\helloworld.java.6c29e59df0efda6a7a9783b1de6b823dc03[71](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:71)000e3c8d32b2b143a430c20900b.jar^" ^"-XX:ArchiveClassesAtExit=C:\Users\RUNNER~1\AppData\Local\Temp\junit8100379350682694[74](https://github.com/jbangdev/jbang/runs/5324909962?check_suite_focus=true#step:6:74)5\jbang\cache\jars\helloworld.java.6c29e59df0efda6a7a9783b1de6b823dc0371000e3c8d32b2b143a430c20900b.jar.jsa^"
```

nandorholozsnyak:
Sorry I still had no time to finalize the wrong tests, I ran the tests on Linux and on Windows too. I think there were some reasons why this part was not having tests :D @quintesse do you have any tips to resolve the problem? What could be causing the problem on the CI?

quintesse:
Well actually I think the weird path must have been a copy&paste issue on your part, because if I look at the logs it looks normal:

```
TestJdk > testJdkInstallWithLinkingToExistingJdkPathWhenJBangManagedVersionDoesNotExist(File) FAILED
    java.io.IOException at TempDirectory.java:262

TestJdk > testJdkInstallWithLinkingToExistingJdkPathWithDifferentVersion(File) FAILED
    java.io.IOException at TempDirectory.java:262

TestJdk > testJdkInstallWithLinkingToExistingJdkPathWhenJBangManagedVersionExistsAndInstallIsForced(File) FAILED
    java.io.IOException at TempDirectory.java:262
^"C:\Users\RUNNER~1\AppData\Local\Temp\junit4891103843310[71](https://github.com/jbangdev/jbang/runs/5325136154?check_suite_focus=true#step:6:71)9245\jbang\cache\jdks\11\bin\java.exe^" -Dfoo=bar ^"-Dbar=aap^ noot^ mies^" -classpath ^"D:\a\jbang\jbang\build\classes\java\test\itests\hellojar.jar^" helloworld
java -Dfoo=bar ^"-Dbar=aap^ noot^ mies^" -classpath ^"C:\Users\RUNNER~1\AppData\Local\Temp\junit8179139871602089251\jbang\cache\jars\helloworld.java.6c29e59df0efda6a7a9[78](https://github.com/jbangdev/jbang/runs/5325136154?check_suite_focus=true#step:6:78)3b1de6b823dc0371000e3c8d32b2b143a430c20900b.jar^" ^"-XX:ArchiveClassesAtExit=C:\Users\RUNNER~1\AppData\Local\Temp\junit81[79](https://github.com/jbangdev/jbang/runs/5325136154?check_suite_focus=true#step:6:79)139871602089251\jbang\cache\jars\helloworld.java.6c29e59df0efda6a7a9783b1de6b[82](https://github.com/jbangdev/jbang/runs/5325136154?check_suite_focus=true#step:6:82)3dc0371000e3c8d32b2b143a430c20900b.jar.jsa^" helloworld

TestJdkManager > testResolveJavaVersionFromPathWhenReleaseFileHasNoProperJavaVersion(File) FAILED
    java.io.IOException at TempDirectory.java:262

TestJdkManager > testResolveJavaVersionFromPathWhenJavaVersionHasNewVersioning(File) FAILED
    java.io.IOException at TempDirectory.java:262

TestJdkManager > testResolveJavaVersionFromPathWhenJavaVersionHasOldVersioning(File) FAILED
    java.io.IOException at TempDirectory.java:262
```

maxandersen:
I thought we do symbolic linking for jdk install testing elsewhere so what is special about this usecase?

nandorholozsnyak:
> I thought we do symbolic linking for jdk install testing elsewhere so what is special about this usecase?

I do not really see any tests that are installing the JDKs (of course there are but not in the TestJdk class). What I could basically debug is that the creation of hard links on Windows works only if you link one file, not a whole directory to another (like in the TestEdit test cases). 
On my local Windows environment AccessDeniedExceptions are being thrown, but on the CI IOExceptions and not only for the now disabled 2 test cases but another 4. I was assuming that the TempDir annotation and its lifecycle is doing some magic behind it but as I read it's javadoc and it is clear that each method is having its own temporary directory, cause we have no BeforeAll only BeforeEach method in the BaseTest. 
I assume that in the CI when the build is running and the linking is happening after the temporary directory can not "delete" itself, but as I do not see the full stack trace I can not tell if that is the case.

As I read, hard linking folder on Windows required admin privileges or different policies, so this current folder/directory linking might not be the best solution for this particular feature, but on Linux it works fine. 

Should I come up with some other alternative ?

maxandersen:
i'll need to test on my windows setup but I thought we used linking for jdks - but if we dont we might need a "jdklocations.json" instead of 100% file path based solution.
 @quintesse wdyt?

nandorholozsnyak:
I was also thinking on a config file to locate JDKs, it should work everywhere. Let me know if I should rework it

maxandersen:
I think that could be useful.

nandorholozsnyak:
BTW, so if you are using a privileged powershell or command line then the whole feature works on Windows too. The symbolic linking on Windows has been already announced as it could only work with admin privileges only. Knowing this as an already announced limitation on Windows, should it be still reworked to the config file based JDK location?

I just changed the CI build config to see the more information in the logs, I will revert it as soon as I just understand the situation within the CI build.

maxandersen:
symbolic link was supposed to only affect jbang edit; for jdk location that is used for running i think we need to find a way to support it across platforms.

nandorholozsnyak:
Allright, I was not able to reproduce the error that was happening in multiple builds... I'm going to rework it with a config file.

maxandersen:
a reason we didnt do config file before was that the jdk mechanism needs to work in the various bash/powershell/cmd script too - so we need to have that in mind how that will work.

nandorholozsnyak:
@maxandersen 
By the way, I just tested the `jbang jdk default` on Windows, and it is also not working, but it is stated in the documentations as well: https://www.jbang.dev/documentation/guide/latest/javaversions.html#managing-jdks

This config based change could be very "big", I would be totally fine to use admin privileges to set and install JDKs because of that limition, so what is going to be the final decision? Keep it as is, cause it has been stated earlier or should I do the refact on it? I have no Mac to test, but I can imagine it does not have this limitation only Windows.

maxandersen:
hmm okey - if windows already has this issue then maybe better we add this now (with its known limitations) and look at config file based approach separately.

@quintesse wdyt ? you know the nastiness around these

quintesse:
I'll take a look tomorrow. But a quick remark would be that we do indeed use symlinks on all platforms and that has always worked. Using any kind of "index" will make the whole "default" system not work. (You could never use a path like that for PATH or JAVA_HOME, you'd have to change them each time you switch the default JDK).

maxandersen:
Yes, but there are two separate but related features here. 

Jbang jdk install to point to a specific alternate java vm to use when resolving java version. 

Jbang jdk default install location. 

We can do the first via config file and second can keep using symlink and be limited

quintesse:
Sure, we could, but we're already using the linking on Windows and we already accept that in certain cases the linking on Windows isn't ideal. So why introduce a new system just to work around one of those issues and not both? It seems to me that a better investment of time for the future would be to figure out how others are able to reliably make linking work on Windows (eg. Scoop).

quintesse:
Btw, for me this PR works perfectly on Windows 👍 🙂

src/main/java/dev/jbang/cli/Jdk.java
```diff
@@ -23,10 +25,15 @@
 	public Integer install(
 			@CommandLine.Option(names = { "--force",
 					"-f" }, description = "Force installation even when already installed") boolean force,
-			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version)
+			@CommandLine.Parameters(paramLabel = "version", index = "0", description = "The version to install", arity = "1") int version,
+			@CommandLine.Parameters(paramLabel = "existingJdkPath", index = "1", description = "Pre installed JDK path", arity = "0..1") String path)
 			throws IOException {
 		if (force || !JdkManager.isInstalledJdk(version)) {
-			JdkManager.downloadAndInstallJdk(version);
+			if (StringUtils.isNotBlank(path)) {
```
quintesse, line 32: Prefer the usage of https://github.com/jbangdev/jbang/blob/238ef90b0f65aba896b241b06bc8cde796364382/src/main/java/dev/jbang/util/Util.java#L1339_L1349 over the use of external libraries. (This might change in the future but at least that way it would be easy to find all occurrences)

src/test/java/dev/jbang/cli/TestJdk.java
```diff
@@ -32,9 +34,7 @@ void testNoJdksInstalled() throws IOException {
 
 	@Test
 	void testHasJdksInstalled() throws IOException {
-
-		final File testCache = initJBangCacheDir();
-		final File jdkPath = new File(testCache, "jdks");
+		final File jdkPath = new File(jbangTempCacheDir.toFile(), "jdks");
```
quintesse, line 37: Use `JdkManager.getJdksPath()`

src/test/java/dev/jbang/cli/TestJdk.java
```diff
@@ -47,10 +47,106 @@ void testHasJdksInstalled() throws IOException {
...
+			}
+			return null;
+		});
+	}
+
+	@Test
+	void testJdkInstallWithLinkingToExistingJdkPathWhenJBangManagedVersionDoesNotExist(@TempDir File javaDir)
+			throws IOException {
+		initMockJdkDir(javaDir);
+		final File jdkPath = new File(jbangTempCacheDir.toFile(), "jdks");
```
quintesse, line 66: Same here, use `JdkManager.getJdksPath()`

quintesse (approved):
LGTM. The comments I left are mostly clean-up.

quintesse:
Personally I'd merge this PR as-is because it works great, even on Windows as long as you've made the appropriate policy changes (which were already necessary on Windows). We can then think about how to solve the linking issue later.

nandorholozsnyak:
I will fix the findings after work. Thank you.

src/test/java/dev/jbang/cli/TestJdk.java
nandorholozsnyak: Hehe, that is a private method. I can make it public to make it work.

nandorholozsnyak:
Everything should be fixed an the builds should be working.

src/test/java/dev/jbang/cli/TestJdk.java
quintesse: Yeah, probably okay to do that 👍

quintesse (approved):

quintesse:
Good job, thanks @nandorholozsnyak !

maxandersen:
@all-contributors please add @nandorholozsnyak  for code

PR #1256, opened 2022-02-24
feat: Support colors under Windows

knutwannheden:
For various combinations of Windows terminal emulators and shells (e.g. Git Bash) the help displayed by picocli when running `jbang` is garbled with ANSI codes to the point that the help almost can't be deciphered.

According to the [documentation of picocli](https://picocli.info/#_windows) the ANSI support can be enabled by including Jansi and calling `AnsiConsole.systemInstall()`. That is what is being done in this commit.

Note that the Jansi JAR is 215 KB in size.

Please also note that this is not enough for native images. How to use Jansi in GraalVM native images is however also [documented](https://picocli.info/#_jansi_in_graalvm_native_images).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,8 @@ dependencies {
 	implementation 'org.jsoup:jsoup:1.13.1'
 	implementation 'org.codejive:java-properties:0.0.3'
 
+	implementation 'org.fusesource.jansi:jansi:2.4.0'
+
 	implementation 'com.github.jbangdev.jbang-resolver:shrinkwrap-resolver-api:3.1.5-allowpom'
 	implementation 'com.github.jbangdev.jbang-resolver:shrinkwrap-resolver-impl-maven:3.1.5-allowpom'
 
@@ -170,6 +172,8 @@ shadowJar {
 		//exclude(dependency('org.slf4j:slf4j-api:.*'))
 		exclude(dependency('org.slf4j:slf4j-nop:.*'))
 		exclude(dependency('org.jboss.logging:jboss-logging:.*'))
+
+		exclude(dependency('org.fusesource.jansi:jansi:.*'))
 
 
 	}
--- a/src/main/java/dev/jbang/Main.java
+++ b/src/main/java/dev/jbang/Main.java
@@ -3,15 +3,23 @@ package dev.jbang;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.fusesource.jansi.AnsiConsole;
+
 import dev.jbang.cli.JBang;
 
 import picocli.CommandLine;
 
 public class Main {
 	public static void main(String... args) {
+		if (isWindows()) {
+			AnsiConsole.systemInstall();
+		}
 		CommandLine cli = JBang.getCommandLine();
 		args = handleDefaultRun(cli.getCommandSpec(), args);
 		int exitcode = cli.execute(args);
+		if (isWindows()) {
+			AnsiConsole.systemUninstall();
+		}
 		System.exit(exitcode);
 	}
 
@@ -49,5 +57,9 @@ public class Main {
 							.anyMatch(o -> o.startsWith("-i=") || o.startsWith("--interactive=")
 									|| o.startsWith("-c=") || o.startsWith("--code="));
 		return res;
+	}
+
+	private static boolean isWindows() {
+		return System.getProperty("os.name").toLowerCase().startsWith("win");
 	}
 }


knutwannheden:
Someone would have to check that this behaves well on other platforms (e.g. MacOS and Linux).

quintesse:
I just tried on Linux and unfortunately I'm not getting any colors (anymore).

knutwannheden:
Thanks for the feedback! So ot seems like this setup must only be performed on Windows. I will update the code and test it on Windows and Linux.

knutwannheden:
Possibly it is even enough to add Jansi to the classpath and picocli may handle the rest.

maxandersen:
How are you getting garbled results? I tested jbang on all various consoles on windows specifically and not seen this ?

quintesse:
> How are you getting garbled results?

I wanted to say it has happened to me too, but of course now that I want to I'm unable to reproduce it. 🤷

knutwannheden:
@maxandersen I have tried many different combinations on my Windows 10 (and 11) machines.

Using `cmd.exe` and `powershell.exe` it looks OK (but no ANSI magic):
image
Debug output:
```[picocli DEBUG] (ANSI is disabled by default: systemproperty[picocli.ansi]=null, isatty=false, TERM=null, OSTYPE=null, isWindows=true, JansiConsoleInstalled=false, ANSICON=null, ConEmuANSI=null, NO_COLOR=null, CLICOLOR=null, CLICOLOR_FORCE=null)```

When I use Windows Terminal (or also Terminal Preview) it looks the same as above with `cmd.exe` and `powershell.exe`, but when I use it to run `bash.exe` (from the Git installation) it looks like this (regardless of whether it is the MSYS2 or MINGW64 build, which sometimes can make a difference):
image

This is the garbled output I was referring to. The corresponding debug output from picocli environment variable manually):
```[picocli DEBUG] (ANSI is enabled by default: systemproperty[picocli.ansi]=null, isatty=false, TERM=xterm-256color, OSTYPE=null, isWindows=true, JansiConsoleInstalled=false, ANSICON=null, ConEmuANSI=null, NO_COLOR=null, CLICOLOR=null, CLICOLOR_FORCE=1)```

When I open the terminal created by `git-bash.exe` it in fact looks correct:
image

Debug output:
```[picocli DEBUG] (ANSI is enabled by default: systemproperty[picocli.ansi]=null, isatty=false, TERM=xterm, OSTYPE=null, isWindows=true, JansiConsoleInstalled=false, ANSICON=null, ConEmuANSI=null, NO_COLOR=null, CLICOLOR=null, CLICOLOR_FORCE=1)```

Please note that I haven't explicitly set any environment variables like `TERM` or `CLICOLOR_FORCE` in any of the environments.

**Comment by knutwannheden** [comment]
I have now changed the code so that it uses the small `picocli-jansi-graalvm` wrapper (also [documented in picocli](https://picocli.info/#_jansi_in_graalvm_native_images)). This small wrapper offers the advantage that it does nothing when not running under Windows and it should also work using native GraalVM images (I didn't test this however).

Using this commit I now get properly colored output using both `git-bash.exe` and in Windows Terminal using `bash.exe`. In plain `cmd.exe` and `powershell.exe` I still see no ANSI colors, however.

**Comment by quintesse** [comment]
>  In plain cmd.exe and powershell.exe I still see no ANSI colors by default

And what is the picocli debug output in that case?

**Comment by knutwannheden** [comment]
> And what is the picocli debug output in that case?

Then the output looks like this:
```[picocli DEBUG] (ANSI is disabled by default: systemproperty[picocli.ansi]=null, isatty=true, TERM=null, OSTYPE=null, isWindows=true, JansiConsoleInstalled=false, ANSICON=null, ConEmuANSI=null, NO_COLOR=null, CLICOLOR=null, CLICOLOR_FORCE=null)```

when I set `CLICOLOR=1` it looks like this (and the coloring works):
```[picocli DEBUG] (ANSI is enabled by default: systemproperty[picocli.ansi]=null, isatty=true, TERM=null, OSTYPE=null, isWindows=true, JansiConsoleInstalled=false, ANSICON=null, ConEmuANSI=null, NO_COLOR=null, CLICOLOR=1, CLICOLOR_FORCE=null)```

**Comment by quintesse** [comment]
@knutwannheden isn't it strange that `JansiConsoleInstalled=false`? Shouldn't that be `true`?

**Comment by knutwannheden** [comment]
> @knutwannheden isn't it strange that `JansiConsoleInstalled=false`? Shouldn't that be `true`?

Agreed. I could try to debug that. But it is in fact the same in Windows Terminal. Possibly it has to do with the fact that I use the wrapper `picocli-jansi-graalvm` to install Jansi. This is possibly something that could be improved in the debug logging of picocli.

**Comment by knutwannheden** [comment]
@quintesse I did some debugging and found out that the thing with `JansiConsoleInstalled=false` appears to be a picocli issue (see remkop/picocli-jansi-graalvm#3). The latest release of picocli is on Jansi 2.1.0 (while the master is already on Jansi 2.4.0), so this appears to be a bug in Picocli (I will report it). But since I couldn't find any major reported issues in picocli in this area, I am assuming that it works well otherwise.

**Comment by knutwannheden** [comment]
Actually, the `JansiConsoleInstalled=false` issue was caused by the minimizing performed by the Gradle shadow plugin, which I had missed. I have thus excluded the Jansi dependency from the minimization process.

File: build.gradle [hunk 1]
```diff
@@ -35,6 +35,9 @@ dependencies {
 	implementation 'org.jsoup:jsoup:1.13.1'
 	implementation 'org.codejive:java-properties:0.0.3'
 
+	implementation 'org.fusesource.jansi:jansi:2.4.0'
```
maxandersen, line 38: This adds 1.1MB in download - 15% increase in size *just* to enable colors on special case Windows terminal feels overkill to me ;/

is it not just on mingw /msys which are old? basically every other setup works?

build.gradle
```diff
@@ -35,6 +35,9 @@ dependencies {
...
+	implementation 'info.picocli:picocli-jansi-graalvm:1.2.0'
```
maxandersen, line 39: why is it this is needed on top of jansi itself ?

src/main/java/dev/jbang/Main.java
```diff
@@ -6,12 +6,16 @@
...
 import picocli.CommandLine;
+import picocli.jansi.graalvm.AnsiConsole;
 
 public class Main {
 	public static void main(String... args) {
-		CommandLine cli = JBang.getCommandLine();
-		args = handleDefaultRun(cli.getCommandSpec(), args);
-		int exitcode = cli.execute(args);
+		int exitcode;
+		try (AnsiConsole ansi = AnsiConsole.windowsInstall()) {
```
maxandersen, line 14: should it not just fail silently if install does not work?

build.gradle
knutwannheden: Yes, that is very unfortunate. I have however not seen any Windows setup where colors do work, except for `git-bash.exe`, but I don't care much for that terminal. Can you tell me in which Windows aetup this should work? AFAIK I am using the latest version of all tools.

build.gradle
knutwannheden: This second dependency is there to make sure that it also works in GraalVM. But I haven't tested this and since Jansi 2 also added some GraalVM support, it is possible that this dependency isn't required.

src/main/java/dev/jbang/Main.java
knutwannheden: The install is just the resource here which is being closed again. Maybe a bit overkill with System.exit() as the very next thing, but this way it won't get lost if refactored.

build.gradle
knutwannheden: > This adds 1.1MB in download

Hmm... This seems a bit odd. The library is only supposed to be 215 KB in size and has no dependencies on its own. I will take another look at that then.

knutwannheden: I compared the size of the latest `jbang.zip` and it looks like the Jansi dependency adds another 203 KB (~3.6%). I don't like it either, but I am not aware of any alternative (it looks like Quarkus' dev mode Maven mojo also relies on the Jansi bundled with Maven).

I would however still be curious on how I can get colors on Windows without adding Jansi (and not using `git-bash.exe`). I also tried some other terminals (like wezterm), but the result was always the same: ANSI sequences printed to the console.

knutwannheden: Thanks for the info.

> lower left: cygwin - bash, colors

I didn't try Cygwin. I don't know if anybody still uses that these days.

> lower right: windows terminal running wsl with ubuntu, colors

Yes, in Linux under WSL it all works nicely.

> Thus mere lack of color I'm not willing to add 15% "weight" for.

I can understand that. But AFAICT the size increase is not 1.1 MB but more like 203 KiB, which may still be acceptable.

> curious to know what situatons you are seeing garbled output and if you possibly be running 2+ year old windows?

No, I am using a recent Windows 10 build (19044/21H2 which is the build from November 2021) and also the most recent Windows Terminal (Terminal Preview 1.13.10395.0 and Terminal 1.12.10393.0). The difference is that I am running Git's `bash.exe` inside Windows Terminal. I think this is a rather common setup these days for us poor Windows users (I think the Git installer will optionally even create this profile for you) and this is where I have the problems with the ANSI codes.

quintesse: BTW if I set `TERM=xterm` in Windows Terminal with Git Bash then the colors come out fine. Not sure who's wrong here, the picocli terminal detection or WT/Git Bash setting the TERM varaible.

quintesse: BTW2: If I set `CLICOLOR_FORCE=1` for powershell and CMD I get colors as well, both in Windows Terminal and in their "original" windows. So it seems the colors work jsut fine, it's just picocli deciding not to use them?

knutwannheden: ANSI support on Windows is just a mess. If it can be fixed by setting an environment variable, then the fix will be much simpler and leaner.

quintesse: BTW3: I just noticed/remembered that for Linux/Mac we actually always force Picocli to use colors, bypassing its detection mechanism, which means we always output colors even when the user redirects the output to a file. That is kind of heavy-handed, but perhaps we should just do the same for Windows?

maxandersen: Do please list what terminal and shell you are running when seeing these issues ?

quintesse: > Do please list what terminal and shell you are running when seeing these issues ?

I did: "Windows Terminal with Git Bash" :-)

knutwannheden: I have now also tested JBang on my private Windows laptop and there it of course just works... Possibly it has something to do with the Git installation. The installer offers a few options which seem like they could make a difference (see screenshots), but I have now tested many permutations of options (including some obscure environment variables and registry entries) and haven't been able to figure out what causes it to work or not work.

I think it is probably not worth investing more time in this. Maybe someone else who knows more about all this will in the future be able to figure out what is what so that it can be properly fixed or documented (would also be nice for picocli claiming that Jansi is required).

image
image

knutwannheden:
I was never able to track this down. I am not using any outdated versions anywhere, but for some reason the highlighting just doesn't work properly. I am closing this for now. Hopefully someone will be able to figure this one out at some point.

PR #1270, opened 2022-03-01
docs: Make explicit why groovy is needed

jmini:
Make explicit why groovy is needed.

Use java quotes in the line:
`@GrabResolver(name="acme", root="https://maven.acme.local/maven")`

The example does still not compile:
```
classpath_example.java:15: error: GrabResolver is not a repeatable annotation type
@GrabResolver(name="acme", root="https://maven.acme.local/maven")
^
1 error
```

But it is better.


maxandersen:
@all-contributors please add @jmini  for docs

Issue #1267, opened 2022-02-28
Support for maven style version ranges


jmini:
**Is your feature request related to a problem? Please describe.**
I would like to use a version range like in maven. Example `[1.0.0, 1.1.0)`

**Describe the solution you'd like**
Currently only `1.+` is supported. But internally it is turned to a maven range.
As seen in the `[jbang] Resolving dependencies...` lines

**Additional context**
Example:

```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS ch.qos.reload4j:reload4j:[1.2.18,1.2.19)

import static java.lang.System.out;

import org.apache.log4j.Logger;
import org.apache.log4j.BasicConfigurator;

import java.util.Arrays;

class logdemo {

    static final Logger logger = Logger.getLogger(logdemo.class);

    public static void main(String[] args) {
        BasicConfigurator.configure();
        logger.info("Welcome to jbang");

        Arrays.asList(args).forEach(arg -> logger.warn("arg: " + arg));
        logger.info("Hello from Java!");
    }
}
```

Should resolve to `1.2.18.5`.

Currently failing with this error:

```
[jbang] Resolving dependencies...
[jbang] [ERROR] [ERROR] Invalid dependency locator: '1.2.19)'.  Expected format is groupId:artifactId:version[:classifier]
```

jmini:
Probably it is more a bug than a feature.

This is problematic:

https://github.com/jbangdev/jbang/blob/ef362429c1c9a60aec8200b3f3184cbb74160c98/src/main/java/dev/jbang/source/ScriptSource.java#L158-L160

And the corresponding test:
https://github.com/jbangdev/jbang/blob/ef362429c1c9a60aec8200b3f3184cbb74160c98/src/test/java/dev/jbang/source/TestScript.java#L265-L273

The `,` inside a range `[1.0.0, 1.1.0)` should be ignored when it comes to split.

jmini:
Workaround:

```java
///usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS org.apache.groovy:groovy:4.0.0

import static java.lang.System.out;

import org.apache.log4j.Logger;
import org.apache.log4j.BasicConfigurator;

import java.util.Arrays;

import groovy.lang.Grab;
import groovy.lang.Grapes;
import groovy.lang.GrabResolver;

@GrabResolver("mavenCentral")
@Grapes({
[...]

PR #1271, opened 2022-03-01
test: add a test with a version range

jmini:
Add the example discussed in https://github.com/jbangdev/jbang/issues/1267#issuecomment-1054348159 to the test suite.

### DIFF
--- a/itests/classpath_log_grab_with_ranges.java
+++ b/itests/classpath_log_grab_with_ranges.java
@@ -1,0 +1,28 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+
+import org.apache.log4j.Logger;
+import org.apache.log4j.BasicConfigurator;
+
+import java.util.Arrays;
+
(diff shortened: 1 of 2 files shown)


jmini:
@maxandersen, @quintesse what prevents merging this?

This test represents how I am using Jbang with version ranges in the dependency declaration.
I would like to have this in the test-suite so that regressions in this area can be detected directly in the project.

quintesse (approved):
LGTM

quintesse:
@jmini no particular reason, probably just busy :-)
Thanks for your PR!

maxandersen:
its kinda accidental that this works though...I'm fine we merge it to notice it but that exact syntax isn't great to support in //DEPS style...lets see how it goes :)

quintesse:
I agree, but it seemed a good idea to have this test, so we know if something changes and it breaks. Which doesn't have to mean we actually support it :-)

PR #1272, opened 2022-03-02
refactor: Big reorg for handling sources

quintesse:
Lot's of changes here, but basically it boils down to 3 things:

- Introduction of "builders" which allows us to isolate the process of compiling code and turning it into a JAR (and native code). Which will help us in the future to do chains of builds.
- Introduction of "command generators" which allows us to isolate the process of running code (either compiled or as scripts, think Jshell). Not as important as builders but at least there is a bit of symmetry.
- Introduction of "source sets". This is the most important bit where instead of being dependent on the idea that all sources start out with a link to a single file which contains a bunch of //-lines that need to be parsed, which can contain references to other sources and files, etc etc we now abstract that as a `SourceSet`. It contains all the information about a source project; sources, resources, dependencies etc etc. but _without_ defining _how_ that information is obtained. This will allow us in the future to define alternative project types. The new builders use this and are already much more agnostic about JBang-specific scripts. This is a big step but not completely done yet, there are still quite a number of places where assumptions are being made about the project structure but we can weed those out bit by bit in the coming time.


src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -76,14 +34,6 @@ void setJavaVersion(String javaVersion) {
 		this.javaVersion = javaVersion;
 	}
 
-	@CommandLine.Option(names = {
```
maxandersen, line 79: `--cds` seem to have been removed fully for build that does not seem right?

src/main/java/dev/jbang/source/Jar.java
```diff
@@ -24,7 +25,7 @@
  * extracted from the JAR file itself. So all Jars that refer to the same JAR
  * file will contain/return the exact same information.
  */
-public class JarSource implements Source {
+public class Jar implements Input {
```
maxandersen, line 28: Not JarInput as its not a Jar (as the Jar api in jdk?)

src/main/java/dev/jbang/source/RunContext.java
```diff
@@ -328,12 +343,17 @@ public void addJavaAgent(Source src, RunContext ctx) {
...
 	 **/
-	public String resolveClassPath(Source src) {
+	public String resolveClassPath(Input input) {
 		if (mcp == null) {
-			DependencyResolver resolver = new DependencyResolver();
-			updateDependencyResolver(resolver);
-			src.updateDependencyResolver(resolver);
-			mcp = resolver.resolve();
+			if (input instanceof SourceSet) {
+				// TODO this definitely shouldn't be implemented like this!
```
maxandersen, line 349: agreed :)

src/main/java/dev/jbang/source/Jar.java
quintesse: That's the name it had before, but then `SourceSet` should be `SourceSetInput` as well and I didn't like that name.

maxandersen (changes requested):
overall looks good.

--cds on build seems missing.

Not sure I like Jar instead of JarInput, etc. but you might have reason for that change?

i'll play around it more tomorrow.

src/main/java/dev/jbang/cli/BaseBuildCommand.java
quintesse: Well unless I'm misunderstanding the docs, the -X options for CDS only exist for the `java` command, not `javac`? So that would mean it's useless for the `jbang build` command AFAICS.

quintesse: Also, the `cds` value was only referenced from the `Run` class which seems to corroborate that it's not for build-time.

src/main/java/dev/jbang/source/Jar.java
quintesse: Btw, was just thinking that "input" probably isn't a good name. When you `build()` a `SourceSet` you get a `Jar`, so in that case it's obviously not an input, it's an output. So that name might need to change. (It's basically a bunch of methods used for run and info that both a Jar and SourceSet have in common)

src/main/java/dev/jbang/cli/BaseBuildCommand.java
maxandersen: you are right.

src/main/java/dev/jbang/source/Jar.java
quintesse: I settled on `Code`... for now.

quintesse:
Tests have passed yay

maxandersen:
some of the naming i'm not a super fan of but I prefer to live with it until we find better names (if ever :) than not get this refactor in.

this will make great things easier! thanks @quintesse !

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -24,10 +24,7 @@
 import dev.jbang.dependencies.JitPackUtil;
 import dev.jbang.dependencies.MavenRepo;
 import dev.jbang.net.EditorManager;
-import dev.jbang.source.RefTarget;
-import dev.jbang.source.RunContext;
-import dev.jbang.source.ScriptSource;
-import dev.jbang.source.Source;
+import dev.jbang.source.*;
```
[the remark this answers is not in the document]
maxandersen: we don't have a strong rule against it - but prefer to keep it to minimum.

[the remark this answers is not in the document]
quintesse: Honestly I thought Spotless automatically removed all of them.

Issue #1241, opened 2022-02-16
jbang alias list and jbang alias add prints/add empty properties


maxandersen:
**Describe the bug**

`jbang alias add -f . main.java` result in jbang-catalog.json to have empty properties added. Should just not be generated IMO.

But then `jbang alias list` prints:

```
jfxcentral = jfx-central.com as a JavaFX application
             Properties: {}
main@maxandersen/wordle-solve = https://github.com/maxandersen/wordle-solve/blob/HEAD/src/main.java
                                Properties: {}
```

which also is unnecessary as it equates to not have any properties.

quintesse:
Indeed, that's most likely a regression.

grumpyf0x48:
This issue is fixed.

PR #1273, opened 2022-03-03
fix: jbang alias list and jbang alias add prints/add empty properties…

grumpyf0x48:
This PR fixes the issue #1241 that was a regression.


maxandersen (approved):

maxandersen:
thanks @grumpyf0x48 !

PR #1274, opened 2022-03-06
feat: enable jbang tree@jbang.dev and jbang tree@https:/jbang.dev

maxandersen:
some times creating a root jbang-catalog repo can be problematic or the base github user/org is not really related so instead of `jbang cli@awesomefwk` it becomes `jbang cli@l33t1974/awesomefwk`

For a while I had the idea jbang alias should support allowing `jbang main@awesome.fwk` where `awesome.fwk` is treated as https://awesome.fwk instead of looked up in github/gitlab names.

It is only a draft for now but if you build this (with -x test as the test is broken for now) you can do:

`jbang main@jbang.dev` because I created https://jbang.dev/jbang-catalog.json

similar `jbang tree@jbang.dev` works.

imagine, `jbang quarkus@redhat.com` as opposed to `jbang quarkus@RedHatOfficial`  - as an example.

wdyt ?

btw. this also enables `jbang yourscript@https://whatever.path/sdf/sf/jbang-catalog.json`  - not pretty but there for completenes.

even yourscript@file:///path/on/idsk should work but something goes wrong with relative resolve.

### DIFF
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -1,13 +1,10 @@
 package dev.jbang.catalog;
 
-import static dev.jbang.cli.BaseCommand.EXIT_INVALID_INPUT;
-
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
-import dev.jbang.cli.ExitException;
 import dev.jbang.util.Util;
 
 public class ImplicitCatalogRef {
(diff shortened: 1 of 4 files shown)


quintesse (dismissed):
LGTM

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
```diff
@@ -44,7 +47,9 @@ public static ImplicitCatalogRef parse(String name) {
 		}
 		String[] names = parts[0].split("/");
 		if (names.length > 3) {
-			throw new ExitException(EXIT_INVALID_INPUT, "Invalid catalog name '" + name + "'");
+			// throw new ExitException(EXIT_INVALID_INPUT, "Invalid catalog name '" + name +
```
quintesse, line 50: LGTM but I'd remove this commented out code

quintesse (approved):

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
```diff
@@ -44,7 +47,9 @@ public static ImplicitCatalogRef parse(String name) {
...
+			// "'");
```
maxandersen, line 51:
```suggestion
			
```

PR #1283, opened 2022-03-09
feat: Made `--sources script@repo` work

quintesse:
This now only works for `--sources` (`-s`) on the command line, not for `//SOURCES` which needs more changes.

### DIFF
--- a/src/main/java/dev/jbang/source/RunContext.java
+++ b/src/main/java/dev/jbang/source/RunContext.java
@@ -10,6 +10,7 @@ import dev.jbang.catalog.Catalog;
 import dev.jbang.cli.BaseCommand;
 import dev.jbang.cli.ExitException;
 import dev.jbang.dependencies.*;
+import dev.jbang.source.resolvers.AliasResourceResolver;
 import dev.jbang.util.JavaUtil;
 import dev.jbang.util.PropertiesValueResolver;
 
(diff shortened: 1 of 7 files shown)


src/main/java/dev/jbang/source/RunContext.java
```diff
@@ -419,63 +420,52 @@ private ModularClassPath resolveDependency(String dep) {
...
-
 		// Support URLs as script files
 		// just proceed if the script file is a regular file at this point
 		if (resourceRef == null || !resourceRef.getFile().canRead()) {
 			throw new ExitException(BaseCommand.EXIT_INVALID_INPUT,
 					"Script or alias could not be found or read: '" + resource + "'");
 		}
 
+		if (resourceRef instanceof AliasResourceResolver.AliasedResourceRef) {
+			// The resource we found was obtained from an alias which might
```
maxandersen, line 435: should this be moved to method on resourceref or be a interface it implements that allow it to pass over additonal settings?

(not blocker for this pr - just idea for future cleanup)

maxandersen (approved):
lgtm if ci passes.

src/main/java/dev/jbang/source/RunContext.java
quintesse: The method on ResourceRef would be too specific for this one use-case, which was why I discarded that idea. Creating an interface for it would certainly be possible. This part of the code has several inter-dependencies, which is likely the reason I left the alias handling out of the resolvers in the first place. Nothing that can't be solved with more layers of indirection :-)

PR #1287, opened 2022-03-10
feat: allow aliases for `//SOURCES`

quintesse:
You can now use aliases within `//SOURCES` tags.
NB: Only qualified aliases are supported! So the alias needs to have a `@`.

### DIFF
--- a/src/main/java/dev/jbang/catalog/CatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/CatalogRef.java
@@ -52,6 +52,15 @@ public class CatalogRef extends CatalogItem {
 		if (catalog != null) {
 			catalogRef = catalog.catalogs.get(catalogName);
 		}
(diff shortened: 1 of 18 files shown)


maxandersen:
I'm all for having qualified aliases in source files - less ambiguity; but does that also apply when using `jbang -s ...` ?

and is there a technical reason why not or is the reason just to simplify ?

quintesse:
Right now in `-s` you can use the same full range of aliases that you can use with `run`. It's not about simplification (in fact not allowing it for `-s` means having to filter out those cases) but the idea that in the command line you're typing it so you (should) know what aliases you have available locally. But in source code it would be a bit strange to depend on aliases that you might, or might not, have defined locally. But we could disallow it on the command line as well. What would be hard is allowing unqualified aliases in `//SOURCES`.

PR #1291, opened 2022-03-11
feat: don't ask for trust when resource is cached

### LINKED ISSUE #1257: Don't ask to trust a cached resource
When you do `jbang http://some/url`, JBang will ask the user if they trust that URL. If they then answer "just once" and re-run the same command JBang will ask again. This sounds right, the user said "just once" after all, but in actuality this doesn't really make sense for cached resources. IMO JBang should only ask this when actually redownloading the resource (so either when using `--fresh` or when the cache has been cleared).

Btw, not sure how easy this is because the trusting is done at a higher level from the caching. If it's too hard it might not be worth the trouble.

quintesse:
Fixes #1257


maxandersen:
I don't really like it but I can't find a good argument against it so merging it :)

PR #1295, opened 2022-03-11
fix: fixed problem adding multiple repos with same name

### LINKED ISSUE #1293: ConcurrentModificationException when declaring the same repo twice
**Describe the bug**
This is probably a corner case, but I got an error when declaring the same repository twice.

**To Reproduce**
In my case it was once in the source file:
```
//REPOS acme=https://maven.acme.local/maven
```

And once on the command line:
`--repos acme=https://maven.acme.local/maven`

But you can also reproduce it with 2 lines in the source script:
```
//REPOS acme=https://maven.acme.local/maven
//REPOS acme=https://maven.acme.local/maven
```

**Expected behavior**
Maybe a clear error message?
Or if the value is the same, ignore the second declaration?

**JBang version**
0.91.0

**Additional context**
Stacktrace:
```
[jbang] [ERROR] public class java.util.ConcurrentModificationException
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:909)
	at java.util.ArrayList$Itr.next(ArrayList.java:859)
	at org.jboss.shrinkwrap.resolver.impl.maven.MavenWorkingSessionImpl.addRemoteRepo(MavenWorkingSessionImpl.java:315)
	at org.jboss.shrinkwrap.resolver.impl.maven.ConfigurableMavenResolverSystemImpl.withRemoteRepo(ConfigurableMavenResolverSystemImpl.java:59)
	at org.jboss.shrinkwrap.resolver.impl.maven.ConfigurableMavenResolverSystemImpl.withRemoteRepo(ConfigurableMavenResolverSystemImpl.java:36)
	at dev.jbang.dependencies.MavenRepo.apply(MavenRepo.java:32)
	at dev.jbang.dependencies.DependencyUtil.lambda$resolveDependenciesViaAether$1(DependencyUtil.java:152)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at dev.jbang.dependencies.DependencyUtil.resolveDependenciesViaAether(DependencyUtil.java:152)
	at dev.jbang.dependencies.DependencyUtil.resolveDependencies(DependencyUtil.java:122)
	at dev.jbang.dependencies.DependencyUtil.resolveDependencies(DependencyUtil.java:71)
	at dev.jbang.dependencies.DependencyResolver.resolve(DependencyResolver.java:67)
	at dev.jbang.source.SourceSet.getClassPath(SourceSet.java:260)
	at dev.jbang.source.builders.BaseBuilder.compile(BaseBuilder.java:135)
	at dev.jbang.source.builders.BaseBuilder.build(BaseBuilder.java:108)
	at dev.jbang.cli.BaseBuildCommand.buildIfNeeded(BaseBuildCommand.java:49)
	at dev.jbang.cli.Run.prepareArtifacts(Run.java:122)
	at dev.jbang.cli.Run.doCall(Run.java:94)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:93)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:15)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2346)
	at dev.jbang.cli.JBang$3.handle(JBang.java:145)
	at dev.jbang.cli.JBang$3.handle(JBang.java:140)
```

quintesse:
Fixes #1293

### DIFF
--- a/src/main/java/dev/jbang/dependencies/DependencyResolver.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyResolver.java
@@ -65,7 +65,7 @@ public class DependencyResolver {
 
 	public ModularClassPath resolve() {
 		ModularClassPath mcp = DependencyUtil.resolveDependencies(
-				new ArrayList<>(dependencies), new ArrayList<>(repositories),
+				new ArrayList<>(dependencies), new HashSet<>(repositories),
 				Util.isOffline(), Util.isFresh(), !Util.isQuiet());
 		if (artifacts.isEmpty()) {
 			return mcp;
--- a/src/main/java/dev/jbang/dependencies/DependencyUtil.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyUtil.java
@@ -9,14 +9,7 @@ import static dev.jbang.util.Util.infoMsgFmt;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -66,7 +59,7 @@ public class DependencyUtil {
 	private DependencyUtil() {
 	}
 
-	public static ModularClassPath resolveDependencies(List<String> deps, List<MavenRepo> repos,
+	public static ModularClassPath resolveDependencies(List<String> deps, Set<MavenRepo> repos,
 			boolean offline, boolean updateCache, boolean loggingEnabled) {
 		return resolveDependencies(deps, repos, offline, updateCache, loggingEnabled, true);
 	}
@@ -78,7 +71,7 @@ public class DependencyUtil {
 	 * @param loggingEnabled
 	 * @return string with resolved classpath
 	 */
-	public static ModularClassPath resolveDependencies(List<String> deps, List<MavenRepo> repos,
+	public static ModularClassPath resolveDependencies(List<String> deps, Set<MavenRepo> repos,
 			boolean offline, boolean updateCache, boolean loggingEnabled, boolean transitivity) {
 
 		// if no dependencies were provided we stop here
@@ -87,7 +80,7 @@ public class DependencyUtil {
 		}
 
 		if (repos.isEmpty()) {
-			repos = new ArrayList<>();
+			repos = new HashSet<>();
 			repos.add(toMavenRepo("mavencentral"));
 		}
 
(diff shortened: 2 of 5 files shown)


src/main/java/dev/jbang/dependencies/DependencyResolver.java
```diff
@@ -65,7 +65,7 @@ public DependencyResolver addClassPaths(String classPaths) {
 
 	public ModularClassPath resolve() {
 		ModularClassPath mcp = DependencyUtil.resolveDependencies(
-				new ArrayList<>(dependencies), new ArrayList<>(repositories),
```
maxandersen, line 68: this can't be right  - the sequence of maven repositoreis matters.

quintesse: Sure it's right :-)

They're LinkedHashSets, they retain their ordering: https://github.com/jbangdev/jbang/blob/1cf630cd600960f771aaf6e7b9bb14f0283341e8/src/main/java/dev/jbang/dependencies/DependencyResolver.java#L16-L18

maxandersen:
...i'm really not a fan of having the public api say Set when its not true.

Is it really that worse to take in as a List and deduplicate it before/after ?

quintesse:
Not necessarily worse, no. You'd have to make sure that at any point where the list must be _dedupped_ it actually is without duplicates. With LinkedHashSet you make sure it adheres to both requirements, ordered _and_ unique. But I understand your misgivings.

maxandersen:
thinking more about if you have same id with different url it should be an error.

if same id with same url we could potentially let it pass.

maxandersen (changes requested):
need to have it fail instead of silent the error of duplicate ids (with different urls).

quintesse:
Fail? Are you sure? Not simply a warning?
There might be times that it might be hard to dedup those values in a script while it's easy enough for us to do in code...

maxandersen:
A stern warning at least. Probably for any kind of duplication. 

Which also implies we can't make it a set as then we loose the ability to know if duplicates occurs.

quintesse:
I just now read more carefully what you wrote earlier: "same id with different url it should be an error"

So I understand now what you were trying to say and my first reaction was to agree with you, but on the other hand it's so easy to add additional repos that would completely override previous ones that I'm not sure that failing wen a duplicate is detected gives any advantages? What could go wrong if we let users do this?

maxandersen:
Well you could resolve different artifacts than you expect and basically be able to enable a easy man in middle attack.

I still think it should be a hard error if something brings in diffferent url different id. 

Actually thinking about it we probably should allow explicit control of repos ego prevent any kind of tampering.

quintesse:
But aren't you deciding to run that code anyway and couldn't do that code much worse things than overriding a repo url?
Anyway, I'm just discussing this for discussion's sake, I don't mind at all making this a hard error.

maxandersen:
adding same id with same repo is fine, but trying add with same id but different repo should be error.

maxandersen:
this is out of date/not relevant anymore - right @quintesse ?

quintesse:
Out-of-date yes, but relevant... erm... I'm not 100% sure TBH. Perhaps we should check 😅

maxandersen:
closing it to clean up the pr queue as I can't for now think or remember why this should still be an issue.

Issue #1299, opened 2022-03-12
`jbang config set` updates config key with updated value


grumpyf0x48:
**Describe the bug**
When updating an existing key in JBang configuration, the key is replaced with the value to update and the value is unchanged.

**To Reproduce**
1.  jbang config list
2.  jbang config set run.debug 7272
3. jbang config set init.template cli
4. jbang config list
5. jbang config set run.debug 7373
6. jbang config list

**Expected behavior**
Config value should have been replaced with updated value.

**JBang version**
[jbang] jbang version 0.91.0
Cache: /home/java_user/.jbang/cache
Config: /home/java_user/.jbang
Repository:/home/java_user/.m2/repository
0.91.0

**Additional context**
$ jbang config list
init.template = hello
run.debug = 4004
run.jfr = filename={baseName}.jfr
wrapper.install.dir = .

$ jbang config set run.debug 7272
[jbang] Option 'run.debug' set to '7272' in /home/java_user/.jbang/jbang.properties

$ jbang config set init.template cli
[jbang] Option 'init.template' set to 'cli' in /home/java_user/.jbang/jbang.properties

$ jbang config list
init.template = cli
run.debug = 7272
run.jfr = filename={baseName}.jfr
wrapper.install.dir = .

$ jbang config set run.debug 7373
[jbang] Option 'run.debug' set to '7373' in /home/java_user/.jbang/jbang.properties

$ jbang config list
**7373 = 7272**
init.template = cli
run.debug = 4004
run.jfr = filename={baseName}.jfr
wrapper.install.dir = .

grumpyf0x48:
I'm working on a fix. Will make a PR soon.

quintesse:
ooh nasty bug! Looking forward to your fix :+1: :slightly_smiling_face:

quintesse:
Fixed by #1307

PR #1301, opened 2022-03-13
fix: jbang config set updates config key with updated value

grumpyf0x48:
fix: jbang config set updates config key with updated value (#1299)

### DIFF
--- a/itests/config.feature
+++ b/itests/config.feature
@@ -1,0 +1,16 @@
+Feature: config
+
+Scenario: Configuration keys can be updated
+When command('jbang config set run.debug 7272')
+* match exit == 0
+When command('jbang config set init.template cli')
+* match exit == 0
+Then command('jbang config set run.debug 7373')
+* match exit == 0
+Then command('jbang config get run.debug')
+* match exit == 0
+* match out contains "7373"
+Then command('jbang config list')
+* match exit == 0
+* match out contains "run.debug"
+* match out contains "= 7373"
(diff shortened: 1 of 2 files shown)


grumpyf0x48:
Use an intermediate `Properties` object when saving conf as it is already done in load.

Seems that using the initial object has side effects with cursor current position.

quintesse:
@grumpyf0x48 I'm glad you tried to fix this but in fact the problem was in a dependency that had to be fixed upstream: https://github.com/codejive/java-properties/issues/6. When a new release with that fix gets published we can fix it here. Closing this PR.

grumpyf0x48:
Sorry but it's not a try, the test validates it fixes the issue :-)

The question i had was why using another implementation of Properties ? It is not a common practice - especially in only one source file.

quintesse:
It made the test pass, indeed :-)
But unfortunately the underlying issue was more severe so it had to be fixed upstream.

We're using that implementation of Properties because we had a requirement that any formatting and/or comments in the config files had to be maintained. The standard java Properties doesn't do that, you can put comments in a Properties file, but once you read it into memory all of that is gone. This implementation maintains all formatting and comments, which is what we wanted.

grumpyf0x48:
Seen that limitation !

And clearly java.util.Properties probably won't be updated - it's a very old class of JDK.

quintesse:
Hey look, in the end I used (part of) your test :-)

PR #1312, opened 2022-03-16
feat: fall back to TTY when asking for input on Linux/Mac

quintesse:
This means that the GUI that was introduced a while back to be able to
ask the user for feedback when the stdin was redirected will no longer
be used on Linux and Mac, but only on Windows.
The main advantage of this is that zero-install using `curl` will
work when no GUI is available (eg when using containers).

### DIFF
--- a/src/main/java/dev/jbang/util/ConsoleInput.java
+++ b/src/main/java/dev/jbang/util/ConsoleInput.java
@@ -1,22 +1,60 @@
 package dev.jbang.util;
 
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.concurrent.*;
 
-public class ConsoleInput {
+public abstract class ConsoleInput {
 	private final int tries;
 	private final int timeout;
 	private final TimeUnit unit;
 
-	public ConsoleInput(int tries, int timeout, TimeUnit unit) {
+	/**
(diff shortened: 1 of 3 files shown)


maxandersen:
...but the current one already works in a container does it not?

quintesse:
Nope. Not if you do `curl | bash`, that's exactly the reason why the GUI popup was implemented, to be able to ask the user what to do in case the stdin was redirected.

src/main/java/dev/jbang/util/ConsoleInput.java
```diff
@@ -1,41 +1,78 @@
...
+	}
+
+	/**
+	 * Returns a ConsoleInput based upon /dev/tty which only works on Linux and Mac.
+	 */
+	public static ConsoleInput tty(int tries, int timeout, TimeUnit unit) {
+		return new ConsoleInput(tries, timeout, unit) {
+			@Override
+			protected Callable<String> readerTask() throws IOException {
+				return new ConsoleInputReadTask(new FileInputStream("/dev/tty"));
```
maxandersen, line 46: there is not a case where /dev/tty can be unreadable and we need to still use UI?

maxandersen:
had to revert this as "curl -Ls https://sh.jbang.dev | bash -s - app install --fresh --force quarkus@quarkusio" stopped working.

quintesse:
@maxandersen are you sure this PR needed to be reverted as well? Or did you just do it to be safe?

quintesse:
Ah no, I see you later reverted the revert :smile:

maxandersen:
yeah, I reverted it to be safe but later realized it didn't affect the install part so reverted the revert :)

quintesse:
And I was trying to re-apply the patch and it kept failing and I don't immediately understand why :sweat_smile:

PR #1319, opened 2022-03-20
fix: Unable to use {basename} in template file-refs

### LINKED ISSUE #1318: Unable to use {basename} in template file-refs
**Describe the bug**
JBang templates are providing ways to reuse the file input name coming as an input from the user. It can be {filename} and {basename}.
If the input is Example.java (`jbang init Example.java`) the {filename} variable will be Example.java and the {basename} will be Example. That is cool, BUT, look at the following use case:
```
...
"q-aws-lambda-tf": {
      "file-refs": {
        "{filename}": "aws/aws-lambda.java.qute",
        "generate-jar": "aws/generate-jar",
        "generate-native-image": "aws/generate-native-image",
        "application.properties": "aws/application.properties",
        "lambda-aws-{basename}.tf": "aws/lambda-aws.tf.qute"
      },
...
```
If I would like to go with this approach (`jbang init -t=q-aws-lambda-tf Example.java`) the last entry with `"lambda-aws-{basename}.tf": "aws/lambda-aws.tf.qute"` will cause an error for the Init command because its extension is not `.java`, it is `.tf` and the Init commands checks if the {basename} variable is found in a file entry key, then it validates that the entry's key extension matches the input file extension.

Here is the code for the check:
```java
static Path resolveBaseName(String refTarget, String refSource, String outName) {
		String result = refTarget;
		if (dev.jbang.cli.Template.TPL_FILENAME_PATTERN.matcher(refTarget).find()
				|| dev.jbang.cli.Template.TPL_BASENAME_PATTERN.matcher(refTarget).find()) {
			String baseName = Util.base(outName);
			String outExt = Util.extension(outName);
			String targetExt = Util.extension(refTarget);
			if (targetExt.isEmpty()) {
				targetExt = refSource.endsWith(".qute") ? Util.extension(Util.base(refSource))
						: Util.extension(refSource);
			}
			if (!outExt.isEmpty() && !outExt.equals(targetExt)) {
				throw new ExitException(BaseCommand.EXIT_INVALID_INPUT,
						"Template expects " + targetExt + " extension, not " + outExt);
			}
			result = dev.jbang.cli.Template.TPL_FILENAME_PATTERN.matcher(result).replaceAll(outName);
			result = dev.jbang.cli.Template.TPL_BASENAME_PATTERN.matcher(result).replaceAll(baseName);
		}
		return Paths.get(result);
	}
```

In my opinion the two variables (filename and basename) should be two different uses cases, I would like to use the {basename} variable in other entries as well, the filename variable is totally understandable, but the basename should be allowed to use with different extensions.

**To Reproduce**
Above.

**Expected behavior**
Let the {basename} variable be used for other file entries as well.

**JBang version**
0.91

nandorholozsnyak:
- {filename} and {basename} have different validation rules for their usage
- fixes #1318

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -126,9 +126,7 @@ public class Init extends BaseScriptCommand {
 
 	static Path resolveBaseName(String refTarget, String refSource, String outName) {
 		String result = refTarget;
-		if (dev.jbang.cli.Template.TPL_FILENAME_PATTERN.matcher(refTarget).find()
-				|| dev.jbang.cli.Template.TPL_BASENAME_PATTERN.matcher(refTarget).find()) {
-			String baseName = Util.base(outName);
+		if (dev.jbang.cli.Template.TPL_FILENAME_PATTERN.matcher(refTarget).find()) {
 			String outExt = Util.extension(outName);
 			String targetExt = Util.extension(refTarget);
 			if (targetExt.isEmpty()) {
@@ -140,6 +138,9 @@ public class Init extends BaseScriptCommand {
 						"Template expects " + targetExt + " extension, not " + outExt);
 			}
 			result = dev.jbang.cli.Template.TPL_FILENAME_PATTERN.matcher(result).replaceAll(outName);
+		}
+		if (dev.jbang.cli.Template.TPL_BASENAME_PATTERN.matcher(refTarget).find()) {
+			String baseName = Util.base(outName);
 			result = dev.jbang.cli.Template.TPL_BASENAME_PATTERN.matcher(result).replaceAll(baseName);
 		}
 		return Paths.get(result);
--- a/src/test/java/dev/jbang/cli/TestInit.java
+++ b/src/test/java/dev/jbang/cli/TestInit.java
@@ -324,4 +324,35 @@ public class TestInit extends BaseTest {
 
 	}
 
+	@Test
+	void testInitUsingTemplateWithFilenameAndBasename() throws IOException {
+		Path cwd = Util.getCwd();
+		Path javaFileQute = Files.write(cwd.resolve("file1.java.qute"), "Hello World".getBytes());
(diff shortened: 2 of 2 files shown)


nandorholozsnyak:
Can I ask for a review @quintesse @maxandersen ?

quintesse:
I added an extra test that shows where your code causes problems. That test should already have existed of course but it didn't :-)

So it's not that we disagree with what you want to do, it's a very useful feature. But the way templates work is honestly a bit obscure. So right now a template needs at least one of these two things to be true:

- there is at least a single entry named `{filename}=somefile.ext[.qute]` and the name given by user to the `init` command should have the same `.ext` extension as the file _after_ the `=` sign
- there is at least a single entry named `{basename}.ext=somefile.someext[.qute]` and the name given by user to the `init` command should have the same `.ext` extension as the file _before_ the `=` sign

Right now JBang does this check for all files in a template, which is not really what you want because it prevents a template author from doing exactly what you want to do.

But we still need to make sure that _at least one file follows one of these rules_!

And right now your code only follows the first rule.

The code would have to be adapted in such a way that it checks that the first file it encounters that matches those rules is enough. So having:

```
{filename}=somefile.java
{basename}.md=somefile.md
```

is correct, but so is:

```
{basename}.java=somefile.java
{basename}.md=somefile.md
```

And in both cases JBang has to check that typing `init myfile.java` is correct and so is `init myfile`, but typing `init myfile.foo` or `init myfile.md` should throw an error.

quintesse (changes requested):
See comment I left on PR

quintesse:
Perhaps the best thing to do would be to do is to split this code into two steps: https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/cli/Init.java#L67-L76

The first step checks if the correct rules are followed and throws an error when it finds a problem. And then the second step would be the current code without the error checking.

nandorholozsnyak:
Wow.

Thanks for the input, let me dig deeper and find a good solution for this as soon as I have time.

maxandersen:
replaced with #2646

PR #1320, opened 2022-03-22
feat: "zero" install now lives up to its name (#1219)

quintesse:
Rebased against main to see if we can reproduce what went wrong.


maxandersen:
what was this about? close or reengage?

PR #1326, opened 2022-04-04
chore: cleaner edit code for easier refactor/improve

maxandersen:
Just a few changes to make it easier to
update edit. 

should not break/change things besides the cleanup removing what seems to me like redundant calls and introduce a generic watch method with a callback.


quintesse (approved):
LGTM

PR #1327, opened 2022-04-04
fix: `edit` now sets up correct Java environment

quintesse:
The `edit` command now sets `JAVA_HOME` and adds the Java bin folder
to the `PATH` to make sure the editor/IDE uses the correct Java version.

Especially useful if the code has a higher Java version than the one on the user's PATH.

The problem I can see is that perhaps the editor/IDE doesn't like the version specified in the code.
Wdyt @maxandersen ?

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -2,6 +2,7 @@ package dev.jbang.cli;
 
 import static dev.jbang.Settings.CP_SEPARATOR;
 import static dev.jbang.source.builders.BaseBuilder.escapeOSArguments;
+import static dev.jbang.util.Util.searchPath;
 import static dev.jbang.util.Util.verboseMsg;
 import static java.lang.System.out;
 
@@ -25,6 +26,7 @@ import dev.jbang.dependencies.DependencyUtil;
 import dev.jbang.dependencies.JitPackUtil;
 import dev.jbang.dependencies.MavenRepo;
 import dev.jbang.net.EditorManager;
+import dev.jbang.net.JdkManager;
 import dev.jbang.source.*;
 import dev.jbang.util.TemplateEngine;
 import dev.jbang.util.Util;
@@ -73,14 +75,14 @@ public class Edit extends BaseScriptCommand {
 
 		SourceSet ss = (SourceSet) code;
 		File project = createProjectForLinkedEdit(ss, ctx, false);
-		String projectPathString = Util.pathToString(project.getAbsoluteFile().toPath());
 		// err.println(project.getAbsolutePath());
 
 		if (!noOpen) {
-			openEditor(project, projectPathString);
+			openEditor(ss, ctx, project);
 		}
 
 		if (!live) {
+			String projectPathString = Util.pathToString(project.getAbsoluteFile().toPath());
 			out.println(projectPathString); // quit(project.getAbsolutePath());
 		} else {
 			watchForChanges(code, () -> {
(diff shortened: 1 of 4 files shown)


src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -101,7 +102,11 @@ public Integer doCall() throws IOException {
 					cmd = new String[] { "cmd", "/c", editorCommand };
 				}
 				verboseMsg("Running `" + String.join(" ", cmd) + "`");
-				new ProcessBuilder(cmd).start();
+				ProcessBuilder pb = new ProcessBuilder(cmd);
+				Path javaHome = JdkManager.getCurrentJdk(ctx.getJavaVersionOr(ss));
+				pb.environment().put("JAVA_HOME", javaHome.toString());
+				pb.environment().put("PATH", javaHome.resolve("bin") + File.pathSeparator + System.getenv("PATH"));
```
maxandersen, line 108: what if java is found in /usr/bin won't this be wrong ?

maxandersen:
I could probably work - you are right the main reason coul be the IDE does not like the right java version.

i would say rebase it and we just try...

src/main/java/dev/jbang/cli/Edit.java
quintesse: I'll re-check but AFAIK `getCurrentJdk()` always returns a valid Path to the JDK that was found, not just JBang's managed JDKs.

quintesse:
Done

src/main/java/dev/jbang/cli/Edit.java
quintesse: But you were right, that case had to be handled separately.

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -168,7 +171,23 @@ private boolean openEditor(File project, String projectPathString) throws IOExce
 				cmd = new String[] { "cmd", "/c", editorCommand };
 			}
 			verboseMsg("Running `" + String.join(" ", cmd) + "`");
-			new ProcessBuilder(cmd).start();
+			ProcessBuilder pb = new ProcessBuilder(cmd);
+			Path javaHome = JdkManager.getCurrentJdk(ctx.getJavaVersionOr(ss));
+			if (javaHome == null) {
+				// If no Java home was returned it means we're using the default
```
maxandersen, line 177: if we are using the default javac on path - why need to set it ?

maxandersen:
Testing this out I'm discovering some inconsistencies:

I can't specify a certain java version using `jbang edit` (i.e. `jbang edit --java 17` does not work)
nor can I specify `jbang init --java 17` to create a file referencing `//JAVA 17` so the only way to do `jbang edit` with 
a different java version is to edit the file first....

No matter if we do this change in this PR...I would say we would wanna enable use of `--java N` on init and edit (and other commands too)

maxandersen:
using vscodium this change *feels* right as it will when using run have the right java in use.

with idea I see no change - it just stays at java 11 - do you see the same @quintesse ?

quintesse:
> with idea I see no change - it just stays at java 11 - do you see the same @quintesse ?

Yeah, not sure if that is because I've got it installed with IntelliJ's own JDK (the `jbr` folder). Do you have that too?

src/main/java/dev/jbang/cli/Edit.java
quintesse: Basically just to be sure that the program we're about to start will really use the correct JDK. There are many examples of applications using the value of `JAVA_HOME` over what's available in the `PATH`. This way we'd make sure that all values point to the same place.

maxandersen:
gonna reject this as it is not reliable for most IDE's outside vscode.

quintesse:
Indeed, this should be handled in the generated build/config files. In the gradle build file for example and in IDE-specific files where appropriate.

PR #1329, opened 2022-04-05
fix: some errors in alias documentation page

grumpyf0x48:
fix: some errors in alias documentation page:

1. Incorrect description for https alias usage
2. Missing local path information
3. Missing quote

### DIFF
--- a/docs/modules/ROOT/pages/alias_catalogs.adoc
+++ b/docs/modules/ROOT/pages/alias_catalogs.adoc
@@ -47,7 +47,7 @@ The full format is `<alias>@<user/org>(/repository)(/branch)(~path)` or `<alias>
 |Command | Description
 
 |`jbang hello@acme.corp`
-|`hello` alias found in `https://acme.corp/jbang-catalog.json` if available or the default branch searched on github, gitlab and bitbucket in that order.
+|`hello` alias found in `https://acme.corp/jbang-catalog.json` if available.
 
 |`jbang hello@acme.corp/a/path/to/jbang-catalog.json`
 |`hello` alias found in `https://acme.corp/a/path/to/jbang-catalog.json` if available.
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/alias_catalogs.adoc
```diff
@@ -47,7 +47,7 @@ The full format is `<alias>@<user/org>(/repository)(/branch)(~path)` or `<alias>
 |Command | Description
 
 |`jbang hello@acme.corp`
-|`hello` alias found in `https://acme.corp/jbang-catalog.json` if available or the default branch searched on github, gitlab and bitbucket in that order.
```
maxandersen, line 50: Why remove that info? Jbang does do that search.

maxandersen (changes requested):

docs/modules/ROOT/pages/alias_catalogs.adoc
grumpyf0x48: Really ? I though the '.' used indicated a catalog on a https site instead of of a Git backed service.

I should have missed something :-)

quintesse: I think you're actually right @grumpyf0x48 , that looks like a copy&paste error (or _I_ don't understand how it works either hehe)

maxandersen: well, the code just have it all setup as a chain and can be seen here: https://github.com/jbangdev/jbang/blob/2dceffb51c799b6cba9bac8d5a449fd25b5e7810/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java#L72

basically we use https:// prefix OR a dot in host name to decide if https lookup should be done ....if it finds something all good; but if it fails it gets treated as a git repo lookup.

grumpyf0x48:
So to sum up, this line should remain unchanged ?

Anyway, i think for someone who does not know the internals of JBang, it is a bit puzzling to hear talking about gitlab or github for an https site ! WDYT ?

quintesse:
> Anyway, i think for someone who does not know the internals of JBang, it is a bit puzzling to read about gitlab or github for an https site ! WDYT ?

Well the thing is that it's not an https site. The line just says that if you type `alias@XXXX` then JBang will try the following options in the given order:

1. `https://XXXX/jbang-catalog.json`
2. `https://github.com/XXXX/jbang-catalog/jbang-catalog.json`
3. `https://gitlab.com/XXXX/jbang-catalog/jbang-catalog.json`
4. `https://bitbucket.org/XXXX/jbang-catalog/jbang-catalog.json`

But I do agree that the way it's worded is not very clear, We probably tried to be as short and as concise as possible but it became confusing instead :slightly_smiling_face:

quintesse:
Wdyt @maxandersen ? Should we leave it unchanged or should we instead extend it? The text would get longer but would be more exact and it would perhaps make it easier to understand what is going on.

maxandersen:
I'm fine updating if you got suggestions to improve doc or code

PR #1337, opened 2022-04-18
feat: better error handling for //FILES

maxandersen:
this PR introduces a better error handling when //FILES cannot resolve files


maxandersen:
the number of files changed is just high as I added notion of description to resolvers to give some context.

Not 100% convinced we need to include it always (maybe only when --verbose?)

maxandersen:
whoops - some tests not passing ...investigating

maxandersen:
originally pr had "fix" to make **/*.java include root files but realized that is not right...if one wants that one uses **.java (no /).

which also makes the code much cleaner (zero changes needed for that :)

quintesse (dismissed):
LGTM

quintesse:
Good PR. The only thing I'm unsure about is this one:

> originally pr had "fix" to make **/*.java include root files but realized that is not right...if one wants that one uses **.java (no /).

Is that common behaviour? I always assumed that `**/` meant 0, 1 or more folders. So `**/*.java` would indeed include root items.

maxandersen:
Its at least how globmatcher in the jdk behaves. And it kinda makes sense that \ means at least one directory. 

I thought same as you but I think I thought that as in traditional java you always have layers of dira :)

quintesse:
> Its at least how globmatcher in the jdk behaves

Seems you are right. And it does make sense, but it seems it's different from other glob patterns I have used in the past.

Indeed, if I look at Gradle it says they follow Ant's globbing rules and looking at the Ant docs they do specifically state that `**/X` also matches `X`.

Issue #447, opened 2020-10-15
Jbang in WINDOWS under corporate Proxy with auth


khorfox:
I tried a simple helloword with dependencies (work correctly at home) on my corporate environment. 
I tried to set the proxy via environment variables (http.proxy): not work.
I tried to modify the jbag.cmd and passing JAVA_OPTIONS with proxy, user, password: not work.
There's a way to set corporate proxy and authentication ?
The stack trace is:
[jbang] Resolving dependencies...
[jbang]     Resolving com.github.lalyos:jfiglet:0.0.8...Done
[jbang]     Resolving info.picocli:picocli:4.5.0...[jbang] [ERROR] Could not resolve dependency
dev.jbang.ExitException: Could not resolve dependency
        at dev.jbang.DependencyUtil.lambda$resolveDependenciesViaAether$3(DependencyUtil.java:145)
        at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:267)
        at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at dev.jbang.DependencyUtil.resolveDependenciesViaAether(DependencyUtil.java:152)

maxandersen:
If you set JAVA_OPTIONS it should work as that's picked up by Java itself - jbang isn't deliberately blocking that. Afaik it should just use proxy settings. 

Not sure how to mimick your setup so I can test.

maxandersen:
Hmmm. I does look like shrink wrap resolver isn't being nice citizen but instead listening to proxy settings in a settings.xml. 

We probably need to add a setting to override this.

maxandersen:
okey - now at laptop so could check what is happening.

jbang uses shrinkwrap resolvers and that library does not honor java proxy settings via properties but it *do* read ~/.m2/settings.xml for settings like additional repos and proxies. 

[...]

PR #1339, opened 2022-04-20
test: added docker compose file for testing proxy

quintesse:
This allows you to easily start up a proxy and a container with a shell
and JBang pre-installed that is only allowed to connect to the proxy.
Includes a README with some instructions.

See #447


quintesse:
So this is _not_ a solution of any kind, it's just a way to easily test JBang with a proxy!

PR #1346, opened 2022-04-27
chore: add explicit cast

gastaldi:
Because `java.nio.file.FileSystems#newFileSystem(java.nio.file.Path, java.util.Map<java.lang.String,?>)` was introduced in JDK 13+, using only `null` without casting leads to
compilation errors when building in JDK 17 for example


quintesse (approved):
LGTM.

PR #1348, opened 2022-04-29
fix: add sources to export

### LINKED ISSUE #1344: jbang build or export with --sources=... without a source/main should be able to build jar
jbang build with --sources should be able to build and export jar.

atm it fails complaining about missing sourceref and exported local jar is empty

andreaTP:
Resolves #1344 (I hope 🤞 )

Still not "super-clean" but it allows the use-case to work 🎉 
https://github.com/andreaTP/playing-with-jbang

### DIFF
--- a/src/test/java/dev/jbang/cli/TestRun.java
+++ b/src/test/java/dev/jbang/cli/TestRun.java
@@ -49,7 +49,6 @@ import javax.xml.parsers.ParserConfigurationException;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.collection.IsCollectionWithSize;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


maxandersen:
Nice. That actually looks like all there is to it. Will give it some testing and merge asap if nothing found.

quintesse:
We'd probably have to test what happens now if you use `jbang run` on code that doesn't have a main. Specifically the kind of error that the user will get and if that's what we want.

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -371,11 +370,13 @@ void testHelloWorldGAVWithNoMain() throws IOException {
...
 		assertThat(code.getResourceRef().getFile().toString(), matchesPattern(".*\\.m2.*codegen-4.5.0.jar"));
 
-		ExitException e = Assertions.assertThrows(ExitException.class,
-				() -> code.cmdGenerator(ctx).generate());
+		String result = code.cmdGenerator(ctx).generate();
+		assertThat(result, matchesPattern("^.*java(.exe)?.*"));
+		assertThat(ctx.getMainClassOr(code), nullValue());
 
-		assertThat(e.getMessage(), startsWith("no main class"));
+		assertThat(code.isJar(), equalTo(true));
```
quintesse, line 377: This can't be correct, because it would mean that trying to run a script that has no main will not result in any error for the user.

I can understand not getting an error for the `build` command, but the same can't be true for `run`.

quintesse (changes requested):

src/test/java/dev/jbang/cli/TestRun.java
maxandersen: correct - i've restored this and then build/export still works afaics

maxandersen:
@all-contributors please add @andreaTP  for code

src/test/java/dev/jbang/cli/TestRun.java
andreaTP: @quintesse @maxandersen , I would like to follow up on this detail, in the example I'm building up it's perfectly legit to `run` in interactive mode without a `main` class.

Ideally, those 2 `jbang` invocations should be only one:
https://github.com/andreaTP/playing-with-jbang/blob/758f87d9a20471e5caeaafacd88e060138a3f3f7/k8s-shell.sh#L25-L35

Would you agree to make this Main mandatory only under the conditions:
 - it's a `run` command
 - it's not interactive

quintesse: @andreaTP good point, I agree

maxandersen: Yes. Agree.

PR #1355, opened 2022-05-02
Allow no main if not non-interactive(batch) run

andreaTP:
Follow up on this discussion:
https://github.com/jbangdev/jbang/pull/1348#discussion_r862709540

### DIFF
--- a/src/main/java/dev/jbang/cli/Run.java
+++ b/src/main/java/dev/jbang/cli/Run.java
@@ -109,6 +109,7 @@ public class Run extends BaseBuildCommand {
 
 	RunContext getRunContext() {
 		RunContext ctx = super.getRunContext();
+		ctx.setRunCommand(true);
 		ctx.setArguments(userParams);
 		ctx.setJavaOptions(javaRuntimeOptions);
 		ctx.setInteractive(interactive);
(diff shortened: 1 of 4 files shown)


quintesse (changes requested):
In general this works perfectly, but I think we can simplify this by simply using the status of the `ctx.isInteractive()` which means we wouldn't need to add the `isRunCommand()`.

Why? Because `isInteractive()` is never `true` for builds (the option doesn't exist for it) so it can only ever be `true` when doing a `run`.

quintesse (approved):
Sorry, wasn't thinking straight. We also need to skip the check when doing builds of course.

So I'll approve this because it works.

But I wonder if we shouldn't rename the `isRunCommand()` to something like `isMainRequired()` and then make it default to `false` and set it to `! isInteractive()` for the `run` command.

andreaTP:
> But I wonder if we shouldn't rename the isRunCommand() to something like isMainRequired() and then make it default to false and set it to ! isInteractive() for the run command.

I like this proposal!

maxandersen:
+1 for requiring main. 

Also need to check how interactive run deal with its main generation. It generates a top level method to run main. That should be skipped if no main class but not sure it's guarded atm.

PR #1356, opened 2022-05-03
feat: Set Maven update policy to ALWAYS when --fresh is used

### LINKED ISSUE #1347: Forcing dependencies update (like maven "-U")

When jbang downloads artifact, it seems to be bound to the updatePolicy configured for maven, which is daily by default (see https://maven.apache.org/ref/3.8.5/maven-settings/settings.html#releases)

In maven you have the `-U` (or `--update-snapshots`) flag to force the update of dependencies.

A way (maybe the existing command `--fresh`) to force jbang to be sure to get the latest version.

**Describe alternatives you've considered**

Run `mvn -U dependency:get -Dartifact=<my dep>` before running `jbang` to be sure to have resolved the latest version.

jmini:
Fixes #1347

### DIFF
--- a/src/main/java/dev/jbang/dependencies/DependencyUtil.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyUtil.java
@@ -119,8 +119,8 @@ public class DependencyUtil {
 		}
 
 		try {
-			List<ArtifactInfo> artifacts = resolveDependenciesViaAether(depIds, repos, offline, loggingEnabled,
-					transitivity);
+			List<ArtifactInfo> artifacts = resolveDependenciesViaAether(depIds, repos, offline, updateCache,
+					loggingEnabled, transitivity);
 
 			ModularClassPath classPath = new ModularClassPath(artifacts);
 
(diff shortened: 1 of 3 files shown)


quintesse (commented):
This certainly looks okay, but given the fact that calling --fresh on something that has lots of dependencies might have a large negative effect (in download times) I wonder if we can't be more intelligent about this.

For example perhaps we could only apply it if a `-SNAPSHOT` dependency is detected?

maxandersen:
I have same concern as you @quintesse that would be best if we only have snapshot forced fresh but I did a few trial runs enabling shrinkwrap logging and afaics it won't re-download artifacts if there are no changes to the artifacts.

quintesse:
> it won't re-download artifacts if there are no changes to the artifacts.

Ok, that's good. We can always change it if there's some situation where this becomes a problem.

quintesse (approved):
LGTM. Let's merge this and see if it needs improving in the future.

jmini:
> For example perhaps we could only apply it if a `-SNAPSHOT` dependency is detected?

My use-case is to get the latest released version of an internal CLI tool.

Something like what you have in the docs ([Running jar](https://www.jbang.dev/documentation/guide/latest/usage.html#running-jars)), but with `RELEASE` instead of a fixed version

```
jbang --main picocli.codegen.aot.graalvm.ReflectionConfigGenerator info.picocli:picocli-codegen:RELEASE
```

When there is a new release, the maven resolver is checking if the latest version was downloaded only once a day.
With Maven you can force maven to check this with `-U`. With jbang, before this change, there was no option.

So I am not interested with the `-SNAPSHOT` check at all.

quintesse:
Makes sense, thanks for the explanation @jmini 👍

maxandersen:
Ah, has not thought about effect of RELEASE - that actually makes perfect sense and seems to strike a good balance and Afaics if you don't use snapshot nor release you are not impacted as for normal versioned artifacts no remote download happens Afaics. 

Good change.

PR #1359, opened 2022-05-11
fix: can now use globbing for `--sources`

quintesse:
Besides that it's now also possible to use absolute paths together
with globbing.

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -188,46 +188,63 @@ public class Util {
 	}
 
 	/**
-	 * Explodes filepattern found in baseDir returnin list of relative Path names.
+	 * Explodes filePattern found in baseDir returning list of relative Path names.
 	 *
 	 * TODO: this really should return some kind of abstraction of paths that allow
 	 * it be portable for urls as wells as files...or have a filesystem for each...
 	 * 
 	 * @param source
 	 * @param baseDir
-	 * @param filepattern
+	 * @param filePattern
 	 * @return
 	 */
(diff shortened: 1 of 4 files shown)


src/main/java/dev/jbang/util/Util.java
```diff
@@ -217,17 +217,29 @@ public static List<String> explode(String source, Path baseDir, String filepatte
 			// not a pattern thus just as well return path directly
 			results.add(filepattern);
 		} else {
-			// it is a non-url letls try locate it
+			// it is a non-url let's try to locate it
+			final Path bd;
+			final boolean useAbsPath;
+			if (Paths.get(filepattern).isAbsolute()) {
```
maxandersen, line 223: what is this trying to do?

quintesse: The "it's now also possible to use absolute paths together with globbing." part from the commit message ;-)

For the tests it's easier if you can just do `--sources /full/path/to/some/test/files*` so I made that work. (Absolute paths were already supported so it would be strange not to allow globbing in those paths)

quintesse:
Always Windows! Why Windows, why?

src/main/java/dev/jbang/util/Util.java
maxandersen: this should get tested with C:, D: paths on Windows

maxandersen: this should get tested with C:, D: paths on Windows

quintesse:
Merging this, if it causes any problems will hopefully notice before the next release :-)

PR #1362, opened 2022-05-13
fix: move away from java.io.File to Path/Files

maxandersen:
Moves away from java.io.File to use Path and Masters

simplifies  and remove a bunch of unnecessary toPath transforms we have


maxandersen:
fyii, @quintesse if tests paths i would like to merge this as touches all path stuff

quintesse (approved):
Love it 👍

maxandersen:
Cool. I'll just do a manual check on windows to check haven't by accident started produce non windows paths.

PR #1363, opened 2022-05-13
feat: //FILES now supports globbing and new --resources option

### LINKED ISSUE #1338: //FILES should support glob patterns
from beginning I deliberaly made it so //FILES was a 1-to-1 mapping to allow use it on remote urls.

But more and more I try use jbang for quarkus/ui apps I can't see it be useful unless we allow use of globpatterns, i.e. `//FILES templates/=**.qute.html` to include all files named qute.html in all subdirs.

tried doing a quickfix but atm RefTarget assumes it only resolves to a single file...so need some new notion.

quintesse:
The new `--resources` (`-r`) option allows the user to set additional
files for the script (`--files` wasn't available anymore).

Fixes #1338


src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -8,17 +8,7 @@
...
-import static org.hamcrest.Matchers.hasItem;
-import static org.hamcrest.Matchers.hasKey;
-import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.hasXPath;
-import static org.hamcrest.Matchers.instanceOf;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.matchesPattern;
-import static org.hamcrest.Matchers.not;
-import static org.hamcrest.Matchers.startsWith;
+import static org.hamcrest.Matchers.*;
```
quintesse, line 11: Not sure why this happens. Shouldn't the `spotlessApply` expand this automatically?

quintesse:
So perhaps the only thing I'm not 100% happy with is the fact that `-f` (and `--file`, singular) are already taken so I had to use `--resources`. The `-r` was also already taken, but as a shortcut for `--jfr` which I am assuming isn't much used anyway so a shortcut isn't really necessary.

But of course you might disagree. Which would either mean not giving `--resources` a shortcut, or try to rename the current `--file`/`-f`, or come up with some other name besides `--resources`. I'm open to suggestions :-)

maxandersen:
-r for --jfr we can fix/change. 

Where are we using --file? (On mobile atm)

src/test/java/dev/jbang/cli/TestRun.java
maxandersen: no, spotless apply doesn't unroll import.

maxandersen:
`--file/-f` is for alias/catalogs/config not build/run/export etc.

`--files` would be okey would it not? can you imagine there would be conflicts?

quintesse:
--file/-f is is for _all_ commands because the config is always read, regardless of the command, so it's part of the base/root JBang command,

Now could/should that have been called `--config` instead of `--file`? Yes, perhaps, but is that something we can change now? That seems more "dangerous" (ie potentially breaking people's scripts) than taking away `-r` from `--jfr`.

quintesse:
Hmm wait, that _is_ actually called `--config`, so what was the conflict I was getting yesterday? ... let me check...

quintesse:
Ok, that's weird I just changed it to `--files` and `-f` and things just worked... I'm sure I got conflicts yesterday! Anyway, added a commit to the PR for that. Much nicer this way. Internally the code still calls them "resources", but that's in line with existing code, so that's perfectly okay.

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -46,6 +46,9 @@ public class Edit extends BaseScriptCommand {
 	@CommandLine.Option(names = { "-s", "--sources" }, description = "Add additional sources.")
 	List<String> sources;
 
+	@CommandLine.Option(names = { "-r", "--resources" }, description = "Add additional resources.")
```
maxandersen, line 49: should be --files?

src/main/java/dev/jbang/cli/BaseBuildCommand.java
```diff
@@ -20,6 +20,9 @@ public abstract class BaseBuildCommand extends BaseScriptCommand {
 	@CommandLine.Option(names = { "-s", "--sources" }, description = "Add additional sources.")
 	List<String> sources;
 
+	@CommandLine.Option(names = { "-f", "--files" }, description = "Add additional files.")
```
maxandersen, line 23: should have "converter=CommaSeparatedConverter.class"

maxandersen:
> Ok, that's weird I just changed it to `--files` and `-f` and things just worked... I'm sure I got conflicts yesterday!

because you didn't change it for Alias.java :)

maxandersen:
pushed commit that makes "--files" the only way to set this.

Not 100% convinced that is the right way but at least then "-f" does not conflict on Alias.

maxandersen:
BUT now ./bootstrap.sh works!

maxandersen:
`--files src/main/resources/jbang` gives NPE:

```
[jbang] [ERROR] public class java.lang.NullPointerException
	at dev.jbang.source.SourceSet.lambda$getStableId$1(SourceSet.java:320)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1384)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.StreamSpliterators$WrappingSpliterator.forEachRemaining(StreamSpliterators.java:313)
	at java.util.stream.Streams$ConcatSpliterator.forEachRemaining(Streams.java:743)
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:647)
	at dev.jbang.util.Util.getStableID(Util.java:977)
	at dev.jbang.source.SourceSet.getStableId(SourceSet.java:322)
	at dev.jbang.source.SourceSet.getJarFile(SourceSet.java:312)
	at dev.jbang.source.builders.BaseBuilder.build(BaseBuilder.java:67)
	at dev.jbang.cli.BaseBuildCommand.buildIfNeeded(BaseBuildCommand.java:54)
```

quintesse:
> --files src/main/resources/jbang gives NPE:

This is now fixed. In fact it not only doesn't give an error but is treated as if you specified `--files src/main/resources/jbang/**`

maxandersen:
> > --files src/main/resources/jbang gives NPE:
> 
> 
> 
> This is now fixed. In fact it not only doesn't give an error but is treated as if you specified `--files src/main/resources/jbang/**`

Great the npe is fixed but is that the right semantics ? The pattern matcher requires normally a /** to trigger deep traverse ?

maxandersen:
Note I'm not convinced I'm right here just feel like that is copying too much...

quintesse:
Could be, but to me it "feels right". If I'm saying `--sources src/main/java --files src/main/resources` then I probably want to include everything. In fact , in what case would you _not_want that? And then it's easy to just put an `/*` at the end to make it explicit you only want a single directory level.

And I'm not sure if your use of the word "copying" was on purpose, but just to be sure: nothing is being copied here of course, it's just adding paths to the compiler.

maxandersen:
Eh? //FILES are copied into the build dir not just added to compiler dirs, right?

quintesse:
Ah copied in that way. I thought you meant like sometimes files get copied from their original place to the cache first. But yes, all resource files get copied to the build dir.

generated/BuildConfig.java
```diff
@@ -0,0 +1,9 @@
+package dev.jbang;
```
maxandersen, line 1: this shouldn't be committed/added?

maxandersen (changes requested):
jbang.jar in root should not be added/committed.

maxandersen:
`jbang export portable --files=wonka=notthere/itests/**/*.svg  --force t.java
[jbang] [ERROR] Problem looking for **/*.svg in /Users/max/code/personal/jbangdev/jbang/notthere/itests
```

is not informative.

it is failing because notthere folder does not exist. Should be stated without having to use --verbose.

generated/BuildConfig.java
quintesse: Oops!

quintesse:
> is not informative.
> 
> it is failing because notthere folder does not exist. Should be stated without having to use --verbose.
> ```

I'm not saying this couldn't be improved, but no changes were made to this error message, it was already like this before (added by you, no less 😉 ).

quintesse:
Ok, added the exception's string tot the message, hopefully that makes it a bit clearer.

PR #1369, opened 2022-05-19
feat: added --available to jdk list

quintesse:
It's now possible to list the JDK versions available for installation.


src/main/scripts/jbang
```diff
@@ -200,7 +199,7 @@ if [[ -z "$JAVA_EXEC" ]]; then
       fi
       mkdir -p "$TDIR/jdks"
       echo "Downloading JDK $javaVersion. Be patient, this can take several minutes..." 1>&2
-      jdkurl="https://api.foojay.io/disco/v2.0/directuris?distro=$distro&javafx_bundled=false&libc_type=$libc_type&archive_type=$file_type&operating_system=$os&package_type=jdk&version=$javaVersion&release_status=$release&architecture=$arch&latest=available"
```
maxandersen, line 203: why need to remove this ?

maxandersen:
what does the output mean?

```
jbang jdk list
Installed JDKs:
  8
  11
  13
  14
  15
  16
  17 <
  18
  19
```

```
jbang jdk list --available
Available JDKs:
  8 *
  9
  10
  11 *
  12
  13 *
  14 *
  15 *
  16 *
  17 <
  18 *
  19 *
```

whats the diff between <, * and blank ?

src/main/scripts/jbang
quintesse: It's only the `release_status=` that gets removed. If you don't set it it already does what you'd normally want it to do: it will use a GA if it's available and only use an EA when that's the only option. Being able to force an install of an EA over a GA seems like a useless option, especially without being able to exactly specify which version you'd want to install. (But if we ever get that option we'd still not need to use this parameter)

quintesse:
We can easily make this just plain words or give a short explanation, eg:

```
jbang jdk list --available
Available JDKs (*=installed, <=default):
  8 *
  9
  10
  11 *
  12
```

maxandersen:
> The less-than means "default". In the "available" list the asterisk means "installed". In the "installed" list the `*` is left out because it would be superfluous, all items in the list are already installed.

okey, but it makes one wonder what the difference mean. Can't we find one set of symbols that mean the same no matter how user list them?

like:
```
jbang jdk list --available
Available JDKs:
  8 
  9 ?
  10 ?
  11 
  12 ?
  13 
  14 
  15 
  16 
  17 <
  18 
  19 
```

making ? indicate its available but not installed ?

quintesse:
> indicate its available but not installed ?

We could of course, but it seems strange to indicate a negative state.
We'd have one symbol for "installed and default" and another for "not installed", it just doesn't feel very logical.
If it's confusing we can remove them all together, or only keep the "default" one.

quintesse:
For now I changed the output to include the explanation: `Installed JDKs (<=default):` and `Available JDKs (*=installed, <=default):`

maxandersen:
> For now I changed the output to include the explanation: `Installed JDKs (<=default):` and `Available JDKs (*=installed, <=default):`

that works! thanks.

PR #1373, opened 2022-06-09
refactor: Build mixin, SourceSet and others

quintesse:
Bunch of refactors, the most important being the introduction of a `BuildMixin`

The commit "refactor: using less Optional" is in preparation for the build stages stuff because it needs to (de)serialize to/from JSON and Optionals don't serialize well.


quintesse:
@maxandersen this is mostly just a bunch of refactorings, cleaning things up for features to come. If it looks good to you we can merge it soon.

src/main/java/dev/jbang/cli/Alias.java
```diff
@@ -96,7 +96,7 @@ public Integer doCall() {
 		RunContext ctx = getRunContext();
 		Code code = ctx.forResource(scriptMixin.scriptOrFile);
 		if (name == null) {
-			name = CatalogUtil.nameFromRef(ctx.getOriginalRef());
+			name = CatalogUtil.nameFromRef(scriptMixin.scriptOrFile);
```
maxandersen, line 99: is that an intentional change?

src/main/java/dev/jbang/cli/Build.java
```diff
@@ -17,7 +17,9 @@ public Integer doCall() throws IOException {
 		RunContext ctx = getRunContext();
 		Code code = ctx.forResource(scriptMixin.scriptOrFile);
 
-		buildIfNeeded(code, ctx);
+		if (code.needsBuild(ctx)) {
```
maxandersen, line 20: why not have this in the method to avoid the 2-3x repetition?

src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -280,26 +280,27 @@ Path createProjectForLinkedEdit(SourceSet ss, RunContext ctx, boolean reload) th
 		Util.mkdirs(srcDir);
 
 		Path srcFile = srcDir.resolve(name);
-		Util.createLink(srcFile, originalFile.toPath());
+		Util.createLink(srcFile, originalFile);
 
-		for (Source source : ss.getSources()) {
+		for (ResourceRef sourceRef : ss.getSources()) {
```
maxandersen, line 285: sources are now resourceRefs ?

src/main/java/dev/jbang/cli/KeyValueConsumer.java
```diff
@@ -20,17 +19,15 @@ public void consumeParameters(Stack<String> args, CommandLine.Model.ArgSpec argS
 		Matcher m = p.matcher(arg);
 		if (m.matches()) {
 
-			Map<String, Optional<String>> kv = argSpec.getValue();
+			Map<String, String> kv = argSpec.getValue();
```
maxandersen, line 22: how come not optional anymore ? how do you know if it was specified but no value ?

src/main/java/dev/jbang/dependencies/DependencyResolver.java
```diff
@@ -49,7 +49,8 @@ public DependencyResolver addArtifacts(List<ArtifactInfo> artifacts) {
 	}
 
 	public DependencyResolver addClassPath(String classPath) {
-		return addArtifact(DependencyCache.findArtifactByPath(new File(classPath)));
+		// WARN need File here because it's more lenient about paths than Path!
```
maxandersen, line 52: interesting - got example where Path fails but File is okey?

src/main/java/dev/jbang/source/Source.java
```diff
@@ -456,6 +458,45 @@ public List<Source> collectSources(ResourceResolver siblingResolver) {
...
+
+	public SourceSet createSourceSet(ResourceResolver resolver) {
+		SourceSet ss = new SourceSet(this);
+		ss.setDescription(getDescription().orElse(null));
+		ss.setGav(getGav().orElse(null));
+		return updateSourceSet(ss, resolver);
+	}
+
+	@Nonnull
+	public SourceSet updateSourceSet(SourceSet ss, ResourceResolver resolver) {
```
maxandersen, line 473: what does this do ? javadoc it ?

src/main/java/dev/jbang/source/SourceSet.java
```diff
@@ -12,94 +12,56 @@
...
-	public static SourceSet forSource(Source mainSource) {
-		return new SourceSet(mainSource, ResourceResolver.forResources());
-	}
-
-	public static SourceSet forSource(Source mainSource, ResourceResolver resolver) {
-		return new SourceSet(mainSource, resolver);
-	}
+	public static final String ATTR_PREMAIN_CLASS = "Premain-Class";
+	public static final String ATTR_AGENT_CLASS = "Agent-Class";
+	public static final String ATTR_STAGE_PREBUILD_CLASS = "JBang-Stage-Prebuild-Class";
```
maxandersen, line 40: this does not look like a refactoring ?

maxandersen (changes requested):
looks ok, just a few questions and some leaking of future code?

src/main/java/dev/jbang/cli/Alias.java
quintesse: Yes, it just doesn't make sense to put `scriptOrFile` into the `RunContext` to use it again a couple of lines down and nowhere else.

src/main/java/dev/jbang/cli/Build.java
quintesse: It's basically because I'm not sure we want this to be too "black box". You would have no proper way to detect if a build is possibly going to be done or not. But it could definitely be done.

src/main/java/dev/jbang/cli/Edit.java
quintesse: Yes, "sources" (as in //SOURCES or --sources) are now being stored as simple ResourceRefs, just like the "resources" (//FILES and --files) are FileRefs (which are a ResourceRef + an optional alias). Before they were full-blown `Source` instances which a) are not always necessary so you're doing all the parsing for nothing and b) it's not future-proof with the idea of us wanting to create new project types.

src/main/java/dev/jbang/cli/KeyValueConsumer.java
quintesse: Just like before but using `null` instead of `Optional.empty()`. The example you giove would have been `Optional.empty()` and `Optional("")`, now it is `null` and `""`. I just needed this to be easily (de)serializable to/from JSON. It's easier to make this change than to add more custom (de)serializers. Also, all the code became easier because we were doing `value.orElse(null)` all over the place.

src/main/java/dev/jbang/dependencies/DependencyResolver.java
quintesse: Yes, the `Class-Path` element of the MANIFEST uses an URI-like format for paths which, on WIndows, turns `C:\foo\bar` into `/C:/foo/bar` then when you try to read it again FIle will happily parse it while Path complains that `:` is not a valid char. Perhaps this should be handled in some other way, but for now I decided just to keep using FIle in this case.

src/main/java/dev/jbang/source/SourceSet.java
quintesse: Oops, I hadn't noticed that one. A line that shouldn't be there.

src/main/java/dev/jbang/cli/Build.java
quintesse: Ah yes, now that I try to change it I remember why it's like this :-)

A `Builder` takes a `SourceSet` and turns it into a `Jar`, which is executable. But in case of JShell scripts it's actually the `SourceSet` itself which can be executed directly so no build is needed.

So right now we can do:

```java
Code code = ...
if (code.needsBuild()) {
   // It's java, kotlin or groovy code here
   code = code.builder().build();
   // Now it's a Jar
} else {
   // It's Jshell code here
}
// execute "code", whatever type it is
```

So of course we can change `Builder.build()` to return `Code` and just make it a no-op to try to build a `SourceSet` containing JShell code ands simply return the `SourceSet`.

But it feels a bit weird to do that. We'd be making the API slightly worse just to save a couple of lines of code. 
I had it like this at some point but decided against it and rewrote it to what it is now. But we can always change our minds again :-)

quintesse: Ok, the changes aren't too big so I added an extra commit with these changes. Let's see if it's nicer or not.

quintesse: Well actually to get things to work correctly I did have to make a bunch more changes. But I'm actually pretty happy with it. I also added new feature where you can force a source to be of any type, not just Jshell. For that purpose I added a `--source-type` option and deprecated `--jsh` (adding a bunch of `--kotlin`, `--groovy` options was out of the question of course).

src/main/java/dev/jbang/cli/Run.java
```diff
@@ -75,7 +76,10 @@ public Integer doCall() throws IOException {
 			if (!literalScript.get().isEmpty()) {
 				script = literalScript.get();
 				if (scriptOrFile != null) {
-					userParams.add(0, scriptOrFile);
+					List<String> args = new ArrayList<>();
+					args.add(scriptOrFile);
+					args.addAll(ctx.getArguments());
```
maxandersen, line 81: is Arguments and useParams the same ? I thought userParam had to come before the arguments?

quintesse: I took a closer look and yes RunContext.arguments and Run.userParams are the same. In the old code the extra script argument would get added to the start of Run.userParams and then that would be copied to RunCOntext.arguments. In the new code RunContext.arguments has already been initialized with Run.userParams and this code then adds an extra argument to the start. Again, the behaviour should be exactly the same (no tests failed at least, but we know that isn't always a good indication)

PR #1381, opened 2022-06-19
docs: Updating Integration documentation (#1380)

### LINKED ISSUE #1380: Documentation about Integration is not up-to-date
**Describe the bug**
The documentation for the Integration part is not up-to-date, if you create an integration point based on that docs, it will not work.

https://www.jbang.dev/documentation/guide/latest/integration.html

In the docs now:
```java

/**
*
* @param param build dir directory which will be made into a jar when build is done
* @param pomFile location of pom.xml representing the projects dependencies
* @param dependencies list of GAV to Path of artifact/classpath dependencies
* @param nativeImage true if --native been requested
* @return Map<String, Object> map of returns; special keys are "native-image" which is a and "files" to
*          return native-image to be run and list of files to get written to the output directory.
*
*/
Map<String, Object> postBuild(Path builddir, Path pomFile, List<Map.Entry<String, Path>> dependencies,
            boolean nativeImage) {
}
```

But in reallity this should in the docs, I mean the parameter list:
```java
public static Map<String, Object> postBuild(Path appClasses,
                                                Path pomFile,
                                                List<Map.Entry<String, String>> repositories,
                                                List<Map.Entry<String, Path>> dependencies,
                                                List<String> comments,
                                                boolean nativeImage) {
} 
```

And in JBang the following code snippet is happening:
```java
Method method = clazz.getDeclaredMethod("postBuild", Path.class, Path.class, List.class, List.class,
				List.class, boolean.class);
```

**Expected behavior**
Change the documentation for the Integration

**JBang version**
❯ jbang --version
0.95.0

**Additional context**
@maxandersen I saw some tickets about the Integration API/SPI "refactor" or rework. Is it process somewhere? Any ideas for it?
Maybe a Maven dependency that would include an interface and a wrapper class that can include the already existing method parameters inside that and it would be much cleaner probably to implement an integration point.

nandorholozsnyak:
Fixes #1380

### DIFF
--- a/docs/modules/ROOT/pages/integration.adoc
+++ b/docs/modules/ROOT/pages/integration.adoc
@@ -52,6 +52,6 @@ public class JBangIntegrationExample {
 }
 ----
 
-Since version 0.92.0 the full path to the JBang script could be acquired by querying the `jbang.source` system property - `System.getProperty("jbang.source")`
+Since version 0.92.0 the full path to the JBang script can be acquired by querying the `jbang.source` system property - `System.getProperty("jbang.source")`
 
 Still very experimental and bound to change. Example of its use can be found in https://github.com/quarkusio/quarkus/pull/11542[Quarkus].
--- a/docs/modules/ROOT/pages/integration.adoc
+++ b/docs/modules/ROOT/pages/integration.adoc
@@ -27,18 +27,31 @@ Any classes listed in this file will be loaded and jbang will expect and call th
 
 [source,java]
 ----
-/**
-*
-* @param param build dir directory which will be made into a jar when build is done
-* @param pomFile location of pom.xml representing the projects dependencies
-* @param dependencies list of GAV to Path of artifact/classpath dependencies
-* @param nativeImage true if --native been requested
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/integration.adoc
```diff
@@ -27,18 +27,31 @@ Any classes listed in this file will be loaded and jbang will expect and call th
...
+                                                Path pomFile,
+                                                List<Map.Entry<String, String>> repositories,
+                                                List<Map.Entry<String, Path>> dependencies,
+                                                List<String> comments,
+                                                boolean nativeImage) {
+    }
+}
 ----
 
+Since version 0.92.0 the full path to the JBang script could be acquired by querying the `jbang.source` system property - `System.getProperty("jbang.source")`
```
quintesse, line 55: I'd use "can" instead of "could"

nandorholozsnyak: Fixed.

quintesse (approved):

PR #1384, opened 2022-06-22
feat: introduced `build.jbang`

quintesse:
Is basically a file that only supports //-lines, no source. You add the source by using a //SOURCES line.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -185,6 +185,9 @@ test {
 	useJUnitPlatform()
 	//testLogging.showStandardStreams = true
 
+	jacoco {
+		enabled = false
+	}
 }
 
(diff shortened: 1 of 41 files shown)


quintesse:
It's just a very basic implementation right now. There's no validation yet, so you could for example create a `build.jbang` file without a `//SOURCES` line and that would probably fail horribly.

maxandersen:
should we make this feature enabled only when --preview specified as we discussed? 

especially since the format of build.jbang most likely isn't going to stay at being // based.

quintesse:
Ok, will merge this as-is and any improvements/refactorings will be made in future PRs.

PR #1389, opened 2022-06-27
fix: fix problem with JDK paths containing spaces

### LINKED ISSUE #1388: Problem with win10/MinGW when JAVA_HOME contains space
See https://github.com/quarkusio/quarkus/issues/26345

quintesse:
Fixes #1388

### DIFF
--- a/src/main/scripts/jbang.cmd
+++ b/src/main/scripts/jbang.cmd
@@ -54,7 +54,7 @@ if "!JAVA_EXEC!"=="" (
       echo powershell -NoProfile -ExecutionPolicy Bypass -NonInteractive -Command "%~dp0jbang.ps1 jdk install %JBANG_DEFAULT_JAVA_VERSION%"
       if !ERRORLEVEL! NEQ 0 ( exit /b %ERRORLEVEL% )
       rem Set the current JDK
-      !JAVA_EXEC! -classpath "%jarPath%" dev.jbang.Main jdk default "%javaVersion%"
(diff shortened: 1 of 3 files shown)


src/main/scripts/jbang.cmd
```diff
@@ -66,7 +66,7 @@ set JBANG_RUNTIME_SHELL=cmd
 2>nul >nul timeout /t 0 && (set JBANG_STDIN_NOTTY=false) || (set JBANG_STDIN_NOTTY=true)
 set "CMD=!JAVA_EXEC!"
 SETLOCAL DISABLEDELAYEDEXPANSION
-%CMD% > "%tmpfile%" %JBANG_JAVA_OPTIONS% -classpath "%jarPath%" dev.jbang.Main %*
+"%CMD%" > "%tmpfile%" %JBANG_JAVA_OPTIONS% -classpath "%jarPath%" dev.jbang.Main %*
```
garodriguezlp, line 69: Hi @quintesse, 

It looks like this breaks those who install `jbang` using `chocolatey`. Or I guess anyone using it on Windows with a `JAVA_HOME` path with spaces, which is the most common case if the JDK has been installed with `chocolatey` as well.

Here's some debugging of the issue 
image

And here is it back working once I remove the quotes
image

This is my modified version [jbang.cmd.txt](https://github.com/jbangdev/jbang/files/9214969/jbang.cmd.txt)

quintesse: Hmm, so perhaps that fix should only have been made for the *nix shell script.

PR #1398, opened 2022-07-07
fix: don't add `Class-Path` by default to JARs

### LINKED ISSUE #1396: native-image doesn't like Windows paths in JARs
In JBang we generate JARs that contain paths to their dependencies. Now although this does work (as long as you don't move the resulting JAR to a different OS) it:

a) does not seem to adhere to [the spec](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#class-path-attribute) which specifies that the MANIFEST Class-Path should only contain relative paths/URLs

b) the Graal native-image does _not_ like the absolute Windows paths at all and exits with an error that the path contains illegal characters (`:`)

We might try to see if we can use `file:` URLs instead. According to the spec they are allowed, although still not absolute, but we might be able to get away with it. Otherwise we might want to revisit the idea that we add a `Class-Path` to our JARs by default (perhaps make it an option of `export` instead?).

quintesse:
We used to add absolute paths for all dependencies to the MANIFEST.MF
file inside the JAR files Jbang creates. But this is actually not
according to spec, paths may only be relative, but it often just
worked fine. Except that Graal's `native-image` transpiler doesn't
lke that at all. So now we don't generate those `Class-Path` entries
anymore by default, only when using `export local` and `export portable`
will the `Class-Path` be set.

Fixes #1396

### DIFF
--- a/src/main/java/dev/jbang/cli/Export.java
+++ b/src/main/java/dev/jbang/cli/Export.java
@@ -3,23 +3,21 @@ package dev.jbang.cli;
 import static dev.jbang.cli.Export.handle;
 import static dev.jbang.source.builders.BaseBuilder.getImageName;
 import static dev.jbang.util.JavaUtil.resolveInJavaHome;
-import static dev.jbang.util.Util.downloadFile;
 
-import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


maxandersen (approved):

maxandersen:
+1, but kind of saddened no tests needed updating for this :/

Issue #1401, opened 2022-07-11
Disco API v2 is returning 404 when trying to download jdks


andreaTP:
**Describe the bug**
Starting jbang with a JDK non directly available always fails with:
```
Unable to download or install JDK version
```

Since `discoapi` is failing for v2.

**To Reproduce**
- `jbang init main.java`
- add as a second line `//JAVA 19` and make sure that Java 19 is not available on the current PATH
- jbang --verbose main.java
fails with:

```
[jbang] jbang version 0.95.0
[jbang] Build required as /Users/aperuffo/.jbang/cache/jars/main.java.88cf3be1c151ba55f47e473664fb78ccd7d2851aa37a038646157510ca0e773c.jar not readable or not found.
[jbang] System Java version detected as 11
[jbang] System Java version 11 incompatible, using JBang managed version 19
[jbang] Downloading JDK 19. Be patient, this can take several minutes...
[jbang] Downloading https://api.foojay.io/disco/v2.0/directuris?distro=temurin&javafx_bundled=false&libc_type=libc&archive_type=tar.gz&operating_system=mac&package_type=jdk&version=19&architecture=x64&latest=available
[jbang] [ERROR] Required Java version not possible to download or install. You can run with '--java 11' to force using the default installed Java.
[jbang] Deleting folder /Users/aperuffo/.jbang/cache/jars/main.java.88cf3be1c151ba55f47e473664fb78ccd7d2851aa37a038646157510ca0e773c.jar.tmp
[jbang] [ERROR] Unable to download or install JDK version 19
dev.jbang.cli.ExitException: Unable to download or install JDK version 19
        at dev.jbang.net.JdkManager.downloadAndInstallJdk(JdkManager.java:153)
        at dev.jbang.net.JdkManager.getInstalledJdk(JdkManager.java:114)
        at dev.jbang.net.JdkManager.getCurrentJdk(JdkManager.java:107)
        at dev.jbang.util.JavaUtil.resolveInJavaHome(JavaUtil.java:163)
        at dev.jbang.source.builders.JavaBuilder.getCompilerBinary(JavaBuilder.java:16)
        at dev.jbang.source.builders.BaseBuilder.compile(BaseBuilder.java:133)
        at dev.jbang.source.builders.BaseBuilder.build(BaseBuilder.java:108)
        at dev.jbang.cli.BaseBuildCommand.buildIfNeeded(BaseBuildCommand.java:48)
[...]

PR #1402, opened 2022-07-11
Fix retrieval of download link for ea Java releases

andreaTP:
This is tentatively fixing #1401 , but I would like some guidance:

- are you ok with keeping `latest_ga` in code?
- should we include this (or similar) fix in the `scripts`?
- do we need any kind of test?

### DIFF
--- a/src/main/java/dev/jbang/net/JdkManager.java
+++ b/src/main/java/dev/jbang/net/JdkManager.java
@@ -22,6 +22,7 @@ import dev.jbang.util.UnpackUtil;
 import dev.jbang.util.Util;
 
 public class JdkManager {
+	private static final int LATEST_GA = 18;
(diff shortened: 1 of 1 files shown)


quintesse:
Hi @andreaTP , thanks for the PR, but this is definitely not how I'd like to see this work. IMO the API should have a way that will always return a result, without having to know and specify the status. I'm saying this because the EA status could change from one day to the next and it would be a maintenance nightmare if we would have to publish a new version of Jbang just for that. (And people that don't update regularly would have problems downloading JDKs)

andreaTP:
@quintesse thanks a lot for commenting, fortunately, this has been fixed in https://github.com/foojayio/discoapi/issues/60 and doesn't need this workaround at all 🎉

quintesse:
Awesome! :-)

Issue #1409, opened 2022-07-18
 ConcurrentModificationException with --repos from the cli while running info tools


jmini:
**Describe the bug**
When you use `--repos mavencentral` in the command line, you get a `ConcurrentModificationException` when you run `info tools`

> **Note**
> Our use case is not use `mavencentral` but our company mirror of maven central. But for the reproducer I thought it is easier to use `mavencentral`, because everybody has it.

This is similar to https://github.com/jbangdev/jbang/issues/1377 but the command is not the same `info tools` instead of `run`.

**To Reproduce**

Create a small jbang script `log.java`:

```java
///usr/bin/env jbang "$0" "$@" ; exit $?

//DEPS log4j:log4j:1.2.17

import static java.lang.System.out;

import org.apache.log4j.Logger;
import org.apache.log4j.BasicConfigurator;

import java.util.Arrays;

class log {

    static final Logger logger = Logger.getLogger(log.class);

    public static void main(String[] args) {
        BasicConfigurator.configure();
        logger.info("Welcome to jbang");

        Arrays.asList(args).forEach(arg -> logger.warn("arg: " + arg));
        logger.info("Hello from Java!");
    }
}
```

Run it with:
```
 jbang --verbose info tools --fresh --repos mavencentral log.java
```

Usage of `--fresh` make the error appear all the time.
Without `--fresh` it depends (and I could not yet figure out when it isn't working)

jbang is failing with:
```
[jbang] [ERROR] public class java.util.ConcurrentModificationException
```
(see the verbose output at the bottom)

**Expected behavior**
It should work.

**JBang version**

* `0.91.0` 🟢 --> works
* `0.92.0` 🔴 --> first version where this is broken
* `0.92.1` 🔴 --> broken
* `0.92.2` 🔴 --> broken
* `0.93.0` 🔴 --> broken
* `0.94.0` 🔴 --> broken
* `0.95.0` 🔴 --> broken
* `0.96.0` 🔴 --> still broken with `info tools`, 🟢 `run` is working.

**Additional context**
Verbose output:
```
[jbang] jbang version 0.96.0
[jbang] System Java version detected as 8
[jbang] System Java version matches requested version 8
[jbang] Repositories: mavencentral=https://repo1.maven.org/maven2/, mavencentral=https://repo1.maven.org/maven2/
[jbang] Resolving dependencies...
[jbang] [ERROR] public class java.util.ConcurrentModificationException
java.util.ConcurrentModificationException
[...]

PR #1410, opened 2022-07-18
fix: concurrent modification of repos

maxandersen:
this makes https://github.com/jbangdev/jbang/issues/1409 go away but i'm not happy that it is necessary.

it indicates double resolving is occurring as mavenrepo gets added when it should not.

@quintesse suggestions or you think this is valid?


quintesse:
I'd have to look into this more closely tomorrow. But the first thing we'd have to check is if this still happens with the refactor of the Project PR.

maxandersen:
was fixed by other means.

PR #1422, opened 2022-08-03
chore: add deprecation message for `export --native`

quintesse:


src/main/java/dev/jbang/cli/ExportMixin.java
```diff
@@ -61,4 +66,12 @@ private Path getOutputPath() {
 		outputPath = cwd.resolve(outputPath);
 		return outputPath;
 	}
+
+	public void validate() {
+		scriptMixin.validate();
+		if (nativeImage) {
+			throw new IllegalArgumentException(
+					"Use of the `-n` and `--native` flags is deprecated, use `jbang export native` instead.");
```
maxandersen, line 74:
```suggestion
					"The `-n` and `--native` flags have been removed, use `jbang export native` instead.");
```

maxandersen: its removed not just deprecated.

PR #1427, opened 2022-08-10
feat: add jbang-vscode to codium + add usability vscode settings

maxandersen:
jbang-vscode is still in preview mode but it is too good to not be enabled imo.

plus i realized vscodium defaults are slightly bad so added a few default settings
such as autosave on, zoom by trackpad/mousewheel, linked editing and method name guesing.

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -36,7 +36,7 @@ import picocli.CommandLine;
 @CommandLine.Command(name = "edit", description = "Setup a temporary project to edit script in an IDE.")
 public class Edit extends BaseCommand {
 
-	static String[] knownEditors = { "code", "eclipse", "idea", "netbeans" };
+	static String[] knownEditors = { "codium", "code", "eclipse", "idea", "netbeans" };
 
 	@CommandLine.Mixin
 	ScriptMixin scriptMixin;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -226,12 +226,39 @@ private static void setupEditor(Path editorBinPath, Path dataPath) throws IOExce
...
 
+		Path settingsjson = dataPath.resolve("user-data/User/settings.json");
+
+		if (!Files.exists(settingsjson)) {
+			verboseMsg("Setting up some good default settings at " + settingsjson);
+			Files.createDirectories(settingsjson.getParent());
+
+			String vscodeSettings = "{\n" +
+			// better than breadcrumbs
+					"    \"editor.experimental.stickyScroll.enabled\": true,\n" +
```
fbricon, line 237: doesn't work well with Java code. Try a file with a copyright header, or annotations

fbricon:
will it still generate the eclipse project files?

maxandersen:
> will it still generate the eclipse project files?

for now yes; but my thinking is that in futureif you do jbnag edit <folder> it_wont, but for jbang edit xyz.java it will continue to do the linked trick as it works more places out of box.

fbricon:
that means if you open a jbang script it'll first be imported as an eclipse project. Dependency validation will require clicking on the Synchronize JBang code lens

maxandersen:
yes, but users using codium from command line will enjoy it.

PR #1445, opened 2022-09-02
feat: Added `//MANIFEST` option for source files

### LINKED ISSUE #1444: Allow adding entries to MANIFEST.MF
Some tools like PicoCLI provide ways to extract its version from `MANIFEST.MF`. It would be cool if jbang would provide a way to add additional entries to the JAR's manifest :).

_Originally posted by @helpermethod in https://github.com/jbangdev/jbang/discussions/1432_

quintesse:
Fixes #1444


docs/modules/ROOT/pages/running.adoc
```diff
@@ -67,6 +67,19 @@ Since Java 9 https://docs.oracle.com/javase/9/tools/java.htm#GUID-3B1CE181-CD30-
 
 For Java 8 and if you want to set explicitly only for `jbang` you can also add flags by setting `JBANG_JAVA_OPTIONS` and `JBANG_JAVAC_OPTIONS` respectively.
 
+
+== Adding entries to `MANIFEST.MF`
+
+If you want to set custom entries in the `MANIFEST.MF` file of the Jar file generated by Jbang then you can
```
maxandersen, line 73:
```suggestion
If you want to set custom entries in the `MANIFEST.MF` file of the Jar file generated by JBang then you can
```

quintesse:
Fixed

itests/helloworld.java
```diff
@@ -1,6 +1,7 @@
 ///usr/bin/env jbang "$0" "$@" ; exit $?
 // //DEPS <dependency1> <dependency2>
 //JAVA_OPTIONS -Dfoo=bar "-Dbar=aap noot mies"
+//MANIFEST foo bar=baz baz=${bazprop:nada}
```
maxandersen, line 4: split this to two lines to test that actually works.

quintesse: wdym? to test that having multiple lines works you mean?

maxandersen: also if you do not add -Dbazprop the build fails.

That is somewhat of a breaking change.

but feels right we should fail - might just want to make the error message more informative.

maxandersen: > wdym? to test that having multiple lines works you mean?

yeah

quintesse: wdym it fails? where does it fail? how? I don't see that here (and it shouldn't happen).

maxandersen: On command Line.

if you had //JAVAAGENT key=${wonka} in a file it would pas before but today it fails as wonka key does not exist. 

More an observation and that the error message does not tell you *where* the missing key is.

maxandersen (approved):
lgtm

itests/helloworld.java
quintesse: Aaah ok, but that seems like an unlikely value to have in your source file if properties are not supported.

And being able to tell _where_ is actually pretty hard right now. It would probably need an entire refactor of the parsing code. Unless there's some clever trick I'm missing.

PR #1448, opened 2022-09-03
fix: handle quotes better for compiler call

maxandersen:
trying to fix kotlinc.bat bad behavior  (or rather windows lack of safe default quoting)
as discussed in https://github.com/jbangdev/jbang/issues/1447#issuecomment-1236013251

needs testing cross all OS versions.

### DIFF
--- a/src/main/java/dev/jbang/source/builders/BaseBuilder.java
+++ b/src/main/java/dev/jbang/source/builders/BaseBuilder.java
@@ -133,7 +133,7 @@ public abstract class BaseBuilder implements Builder {
 		if (!Util.isBlankString(path)) {
 			optionList.addAll(Arrays.asList("-classpath", escapeOSArgument(path, shell)));
 		}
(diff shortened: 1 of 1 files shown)


src/main/java/dev/jbang/source/builders/BaseBuilder.java
```diff
@@ -131,9 +131,9 @@ public IntegrationResult compile() throws IOException {
 		optionList.addAll(prj.getMainSourceSet().getCompileOptions());
 		String path = prj.resolveClassPath().getClassPath();
 		if (!Util.isBlankString(path)) {
-			optionList.addAll(Arrays.asList("-classpath", path));
+			optionList.addAll(Arrays.asList("-classpath", escapeOSArgument(path, shell)));
```
quintesse, line 134: I left a bunch of comments on #1447 about some tests I did and I think this fix only works for Windows but breaks things on Linux/Mac.

maxandersen:
closing as not fully working. #1447 does it better.

PR #1456, opened 2022-09-16
feat: Switched Shrinkwrap for Maven/Aether API

### LINKED ISSUE #1077: //REPOS example not working for private gitlab repo
So, I am trying to create a example on how to use a library which is there in my organization's private repo. I stumbled across jbang, thought if I can create an example project in jbang, it will be easier for other team mates to try. Here is how I am using it:

```java
//REPOS <repo id>=https://<company domain>/api/v4/projects/10125/packages/maven
//DEPS com.example:demo:1.1.3

public class hello {
        out.println("Hello World");
    }
}
```

If you notice, I am not using mavencentral here, just to keep things simple. But still, its not working for me:

```java
Caused by: org.jboss.shrinkwrap.resolver.api.NoResolvedResultException: Unable to collect/resolve dependency tree for a resolution due to: Could not find artifact com.example:demo:1.1.3 in <repo id> (https://<company domain>/api/v4/projects/10125/packages/maven), caused by: Could not find artifact com.example:demo:1.1.3 in <repo id> (https://<company domain>/api/v4/projects/10125/packages/maven)
	at org.jboss.shrinkwrap.resolver.impl.maven.MavenWorkingSessionImpl.wrapException(MavenWorkingSessionImpl.java:503)
	at org.jboss.shrinkwrap.resolver.impl.maven.MavenWorkingSessionImpl.resolveDependencies(MavenWorkingSessionImpl.java:242)
	at org.jboss.shrinkwrap.resolver.impl.maven.MavenStrategyStageBaseImpl.using(MavenStrategyStageBaseImpl.java:70)
	at org.jboss.shrinkwrap.resolver.impl.maven.MavenStrategyStageBaseImpl.withTransitivity(MavenStrategyStageBaseImpl.java:52)
	at org.jboss.shrinkwrap.resolver.impl.maven.MavenStrategyStageBaseImpl.withTransitivity(MavenStrategyStageBaseImpl.java:39)
	at dev.jbang.dependencies.DependencyUtil.lambda$resolveDependenciesViaAether$3(DependencyUtil.java:194)
```

I installed jbang using sdkman. I think I am using the latest version 0.83.0. Let me know if you need any more info.

quintesse:
The code takes into account the user's `settings.xml`, so things like repository authentication and proxies should now work.

Fixes #1077


quintesse:
Ok, there's still an issue with the new code not taking into account the `JBANG_REPO` setting.

maxandersen:
running this I get these failed tests locally:

```
TestExport > testExportPortableNoclasspath() FAILED
    java.lang.AssertionError at TestExport.java:67

TestExport > testExportWithClasspath() FAILED
```

maxandersen:
nice with 0.5 MB smaller distro though ;)

❯ ls -l ~/.jbang/bin/jbang.jar
.rw-r--r-- 6.4M max 18 Jul 21:12 /Users/max/.jbang/bin/jbang.jar
❯ ls -l build/install/jbang/bin/jbang.jar
.rw-r--r-- 5.8M max 17 Sep 11:17 build/install/jbang/bin/jbang.jar

quintesse:
> running this I get these failed tests locally:

You might have to investigate yourself because all tests run fine for me locally and they've passed online as well.

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
```diff
@@ -0,0 +1,459 @@
...
+
+	private Settings newEffectiveSettings() {
+		DefaultSettingsBuilderFactory factory = new DefaultSettingsBuilderFactory();
+		DefaultSettingsBuilder builder = factory.newInstance();
+
+		SettingsBuildingRequest settingsBuilderRequest = new DefaultSettingsBuildingRequest();
+		settingsBuilderRequest.setSystemProperties(System.getProperties());
+
+		// if we have a root folder, don't read settings at all
+		if (withUserSettings && (localFolderOverride == null || settingsXml != null)) {
```
JakobFels, line 274: This if lead to my settings.xml not being read at all

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
```diff
@@ -0,0 +1,459 @@
...
+		DefaultAuthenticationSelector authenticationSelector = new DefaultAuthenticationSelector();
+		for (Server server : settings.getServers()) {
+			AuthenticationBuilder auth = new AuthenticationBuilder();
+			if (server.getUsername() != null)
+				auth.addUsername(server.getUsername());
+			if (server.getPassword() != null)
+				auth.addPassword(server.getPassword());
+			if (server.getPrivateKey() != null)
+				auth.addPrivateKey(server.getPrivateKey(), server.getPassphrase());
+			authenticationSelector.add(server.getId(), auth.build());
```
JakobFels, line 328: For custom headers, which for example gitlab requires, there still needs to be some processing done

```
if (server.getConfiguration() != null) {
	Xpp3Dom dom = (Xpp3Dom) server.getConfiguration();
	for (int i = dom.getChildCount() - 1; i >= 0; i--) {
		Xpp3Dom child = dom.getChild(i);
		if ("httpHeaders".equals(child.getName())) {
			HashMap<String, String> headers = new HashMap<>();
			Xpp3Dom[] properties = child.getChildren("property");
			for (Xpp3Dom property : properties) {
				headers.put(property.getChild("name").getValue(), property.getChild("value").getValue());
			}
			session.setConfigProperty(ConfigurationProperties.HTTP_HEADERS + "." + server.getId(), headers);
		}
	}
}
```

This will configure the aether http connector to send these headers

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
quintesse: Indeed. I've simplified it and should now pass.

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
quintesse: @JakobFels [here](https://github.com/jbangdev/jbang/issues/1077#issuecomment-1237721810) you commented with a [link to code](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java#L238-L263) that's very different from this. Do you know which one is the correct code? (And out of curiosity, where did you get this new code from?)

quintesse: @JakobFels can you try the new build? It uses the code you posted above. Let's see if that works...

JakobFels: With these additions jbang could resolve an artifact hosted at a gitlab maven registry.

> @JakobFels [here](https://github.com/jbangdev/jbang/issues/1077#issuecomment-1237721810) you commented with a [link to code](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java#L238-L263) that's very different from this. Do you know which one is the correct code? (And out of curiosity, where did you get this new code from?)

I stepped through parts of the code to find out where the headers are read from config.
The relevant code is [here](https://github.com/apache/maven-resolver/blob/master/maven-resolver-transport-http/src/main/java/org/eclipse/aether/transport/http/HttpTransporter.java#L153).
I then tried the linked code from before, until I noticed that the properties are different (aether.connector.wagon vs aether.connector.http).
The current result was my attempt to read these settings.

The first link was me poking at maven code to see where maven itself pulls these settings from, which seems to have been the wrong place (I still don't quite understand when maven sets these up.)

maxandersen:
>You might have to investigate yourself because all tests run fine for me locally and they've passed online as well.

turned out our tests was broken and tested `lib` was missing from current directory rather than the temporary dir we exported to. Thus it failed locally for me as I had a 'lib' dir.

pushed a commit that fixes the tests.

maxandersen:
on app created with `jbang init -t qrest wonka.java` that states:

```
//DEPS io.quarkus:quarkus-bom:${quarkus.version:2.11.2.Final}@pom
```

 running jbang wonka.java I got this output:

```
jbang wonka.java
[jbang]           io.quarkus:quarkus-resteasy:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-vertx-http:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-core:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-ide-launcher:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-development-mode-spi:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-bootstrap-runner:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-security-runtime-spi:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-credentials:jar:2.11.2.Final
[jbang]           io.quarkus.arc:arc:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-mutiny:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-smallrye-context-propagation:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-vertx-http-dev-console-runtime-spi:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-vertx:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-netty:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-vertx-latebound-mdc-provider:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-resteasy-server-common:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-arc:jar:2.11.2.Final
[jbang]           io.quarkus:quarkus-resteasy-common:jar:2.11.2.Final
[jbang] Dependencies resolved
[jbang] Building jar...
[jbang] Running external post build for io.quarkus.launcher.JBangIntegration
[jbang] Post build with io.quarkus.launcher.JBangIntegration
Sep 19, 2022 10:14:45 PM org.jboss.threads.Version <clinit>
INFO: JBoss Threads version 3.4.2.Final
Sep 19, 2022 10:14:46 PM io.quarkus.deployment.QuarkusAugmentor run
INFO: Quarkus augmentation completed in 1559ms
```

why is it not showing resolve of io.quarkus:quarkus-bom:2.11.2.Final@pom?

and do we really need to show the individual downlaods unless in verbose? 
actually what does each line signifies ? it does not seem to be printed when downloading more as that is what will be resolved?

maxandersen:
running `jbang jreleaser-snapshot@jreleaser` it seems stuck at "resolving dependencies..." ...then suddenly prints out list of dependencies.

as if we are printing what we resolve AFTER downlaod completed. It should be when resolution/downloads starts.

quintesse:
> why is it not showing resolve of io.quarkus:quarkus-bom:2.11.2.Final@pom?

Well, that's actually because I'm filtering out @pom entries, if not every line would be duplicated, once for the `@pom` and another one for the `@jar`. In this case you're right it should show it, but see below.

> and do we really need to show the individual downlaods unless in verbose?

We don't _need_ to but it's definitely the easiest to implement. I'm just getting a bunch of events that I'm printing out.
At that point of printing there's no concept of what artifacts were passed as primary dependencies and which are secondary sub dependencies.

> actually what does each line signifies ? it does not seem to be printed when downloading more as that is what will be resolved?

Not sure what you're trying to say here. It's printing each artifact that needs to be downloaded at the moment it starts downloading it.

> as if we are printing what we resolve AFTER downlaod completed. It should be when resolution/downloads starts.

It's definitely printing before. I'll have to investigate that one to see if something special is happening.

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
quintesse: > I still don't quite understand when maven sets these up

yeah documentation is severely lacking, so it's down to looking at the code.

Btw, will you have time to look at the latest build to see if it works for you?

quintesse:
> running jbang jreleaser-snapshot@jreleaser it seems stuck at "resolving dependencies..." 

Not sure what's going on tbh, but the old code has the same problem. The only difference is is that because in the old code we just blindly print out the list of dependencies while in the new code we actually get notified when a download starts. So I think the wait is for something else before any downloads are started.

In fact it might be because it's using jitpack.io, perhaps that's the slow part...

quintesse:
> In fact it might be because it's using jitpack.io, perhaps that's the slow part...

This does indeed seem to be the case. It's downloading a _lot_ of `maven_metadata.xml` files for each artifact. Some of these take a very long time, even though they're quite small. And because you're running the snapshot version they get downloaded _every_ time.

quintesse:
Ok, logging should now be very similar to what it was before except for the improvement that artifacts will be logged in order they are handled by Maven. And `--verbose` will show all dependencies, not just the ones mentioned in the source code.

src/main/java/dev/jbang/dependencies/DependencyUtil.java
```diff
@@ -111,9 +81,10 @@ public static ModularClassPath resolveDependencies(List<String> deps, List<Maven
 				repos.stream().map(m -> m.toString()).collect(Collectors.joining(", "))));
 
 		String depsHash = String.join(CP_SEPARATOR, depIds);
-		if (!transitivity) { // the cached key need to be different for non-transivity
-			depsHash = "notransitivity-" + depsHash;
-		}
+		// if (!transitivity) { // the cached key need to be different for
```
maxandersen, line 84: why don't we need this anymore ?

maxandersen:
i tried latest, its behavior is better but:

a) the indent seems overly big now. Like could just be 2 spaces not the 10+ now
b) i still dont see printing for when download starts. could it be we are not flushing the stream so it does not print until buffer reaches certain size?

src/main/java/dev/jbang/dependencies/DependencyUtil.java
quintesse: The methods was only ever called with `true` so it obviously wasn't used anymore. I commented it out exactly for this reason: that we might take a closer look. Btw that code has been like that for quite some time now, it's not a recent change.

quintesse:
> a) the indent seems overly big now. Like could just be 2 spaces not the 10+ now

I agree, but it's what it was before :-)
Btw, our indents tend to use 3 spaces, shall we use that here as well?

> b) i still dont see printing for when download starts. could it be we are not flushing the stream so it does not print until buffer reaches certain size?

Not sure what you mean. Can you give an example if what you're executing, what you're seeing and what you expected to see?

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
JakobFels: > Btw, will you have time to look at the latest build to see if it works for you?

The current build works for my setup, artifacts hosted on gitlab resolve properly.

maxandersen:
with latest fixes this looks perfect!

maxandersen (approved):

PR #1468, opened 2022-09-27
fix: use `//GAV` for `export mavenrepo`

### LINKED ISSUE #1465: //GAV not taking into account when doing jbang export mavenrepo
put //GAV dk.xam:hello and jbang export mavenrepo hello.java says:

```
[WARN] Cannot export to maven publish as no group specified. Add --group=<group id> and run again.
```

the pom.xml has it included so something is amiss.

quintesse:
Fixes #1465


quintesse:
Btw, I also made two minor text changes where before it said "maven publish" it now says "maven repository" which seems to me less confusing given that the command is named "mavenrepo" (I guess at some point the idea was to name it "mavenpublish").

maxandersen:
this works beautifully! thx!

PR #1469, opened 2022-09-30
Refactor RunContext into ProjectBuilder and other stuff

quintesse:
Large part of this PR was to get to the point where `RunContext` wouldn't need to be passed around anymore. Basically with `Project` and its `SourceSet` we had a nice model for dealing with application sources and how they can be built (`Builder`) and executed (`CmdGenerator`), but there was always that `RunContext` around that was just a dumping ground for any options or values that didn't fit anywhere else. This PR fixes that. In fact the refactor was so successful that `RunContext` didn't need to be passed around anymore was only used to create `Project` instances, well that's a builder pattern so that's when `RunContext` became `ProjectBuilder`.

Getting rid of those long `RunContext` tentacles also made clear that the `Code` - `Project` - `Jar` hierarchy was just overly complex. So all that got folded into `Project`. It also made it necessary to move the definition of Agents into the Project.

The decision to split the build process (now named `AppBuilder`) into separate "build steps" was perhaps better left to a separate PR, but after some more refactoring it was too much work to extract those changes into a separate PR, so I left them in. Those changes also brought in some small extra features that suddenly became easy to implement.


maxandersen:
Interesting!!

src/main/java/dev/jbang/source/Project.java
```diff
@@ -119,12 +125,23 @@ public Project addRuntimeOptions(@Nonnull Collection<String> options) {
...
 
+	@Nonnull
+	public List<Project> getJavaAgents() {
+		return Collections.unmodifiableList(javaAgents);
+	}
+
+	@Nonnull
+	public Project addJavaAgents(List<Project> javaAgents) {
+		this.javaAgents = javaAgents != null ? javaAgents : Collections.emptyList();
+		return this;
```
filippor, line 136: For add method I expect to add something not to replace the collection

quintesse:
@maxandersen we might think about merging this PR as it stands now. It's not 100% perfect yet but it's enough to be able to implement the `build.jbang` file the way you imagined it: without the `//` prefix. Other theoretical future project types should be much easier too.

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -69,52 +81,50 @@ static class ScriptInfo {
...
+					String cp = prj.resolveClassPath().getClassPath();
 					if (cp.isEmpty()) {
 						resolvedDependencies = Collections.emptyList();
 					} else {
 						resolvedDependencies = Arrays.asList(cp.split(CP_SEPARATOR));
 					}
 
-					if (code.asJar() != null && code.asJar().getBuildJdk() > 0) {
-						javaVersion = Integer.toString(code.asJar().getBuildJdk());
+					// TODO remove if everything okay
```
maxandersen, line 112: hmm - what did that code ensure in the past? when jdk was stated inside the jar it would use that instead of default?

PR #1479, opened 2022-10-11
feat: enable -g by default + allow -g=vars,lines,source,none

### LINKED ISSUE #1478: expose/enable compile with debug
turns out debugging with jbang compiled does not work great as we dont add `-g` on javac  by default.

```
 -g                           Generate all debugging info
  -g:{lines,vars,source}       Generate only some debugging info
  -g:none                      Generate no debugging info
```

we could add `-g` in a first fix, but might wanna have a way for users to specify: `none,lines,vars,source`.

Possibly: `jbang -g=none whatever` would be sufficient....just feels `-g` is quite non-informative.

maxandersen:
Fixes #1478

### DIFF
--- a/docs/modules/ROOT/pages/debugging.adoc
+++ b/docs/modules/ROOT/pages/debugging.adoc
@@ -27,6 +27,10 @@ This will make it use port 4321 and make it listen on all ('*') network interfac
 
 NOTE: Be sure to put a breakpoint in your IDE/debugger before you connect to make the debugger actually stop when you need it.
 
+== Debug info
+
+JBang since version 0.100 compiles with debug info enabled by default.
+
 == Debug jbang itself
 
 Java itself will add `JAVA_TOOL_OPTIONS` which will apply to `jbang` too.
--- a/docs/modules/ROOT/pages/debugging.adoc
+++ b/docs/modules/ROOT/pages/debugging.adoc
@@ -29,7 +29,7 @@ NOTE: Be sure to put a breakpoint in your IDE/debugger before you connect to mak
 
 == Debug info
 
-JBang since version 0.100 compiles with debug info enabled by default.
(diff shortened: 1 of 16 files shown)


quintesse:
Looks just fine, but could we wait merging this until after the new project refactor has been merged? It's probably a lot easier to adapt this to that PR than the other way around.

maxandersen:
> Looks just fine, but could we wait merging this until after the new project refactor has been merged? It's probably a lot easier to adapt this to that PR than the other way around.

Sure ... whats missing for that then ? :)

fbricon:
what happens with `-g:none` and script contains `//JAVAC_OPTIONS -g`

fbricon:
Doesn't seem to work with the CI build:

fbricon:
Actually it works. JBang didn't know it needed to recompile the class with the new debug flags. Making a simple change in the source files invalidates the cache and it's properly recompiled with debug info

maxandersen:
Ah. Good point about caching. 

@quintesse should we get caching to include these settings ?

quintesse:
We probably should, yes. 👍

quintesse:
@maxandersen I split the PR into 2 parts: 1 that simply always adds `-g` en 2 where I added `--compile-options` (alias `-C`).

Thing is adding the second part I noticed that the existing `--java-options` is plural while we only allow a single option to be passed. Which is probably wise because trying to allow multiple ones might become a quoting nightmare. But it does mean the name is not really correct.

I added the alias `-R` for `--java-options` which is what JShell uses (R for Runtime).

I wondered if in this case we should just get rid of the long option `--compile-options` and just leave `-C` and at the same time deprecate `--java-options` and prefer `-R` instead. Wdyt?

maxandersen:
I'm not a fan of single letter only options. 

Could we have "-R", "--runtime-option", "--java-options" (the latter just for backwards compatibility)

and "-C", "--compile-option"  ?

quintesse:
Of course!

quintesse:
Done.

PR #1484, opened 2022-10-13
fix: the Aether resolver now uses packing types

### LINKED ISSUE #1482: regression 0.99 with aether does not honor package types
as reported in https://twitter.com/rotnroll666/status/1579765915503452160 by @michael-simons  we no longer seem to honor package type, i.e. @test-jar cannot be resolved.

seems like we don't mimic stereo types from maven: https://github.com/apache/maven/blob/master/maven-resolver-provider/src/main/java/org/apache/maven/repository/internal/MavenRepositorySystemUtils.java#L96

or from shrinkwrap resolvers: https://github.com/shrinkwrap/resolver/blob/master/maven/api-maven/src/main/java/org/jboss/shrinkwrap/resolver/api/maven/PackagingType.java#L35

this a confirmed regression building [quarkusio/quarkus/independent-projects](https://github.com/quarkusio/quarkus/tree/main/independent-projects) with `mvn -Dquickly install`

and do:

```
docker run -v `pwd`:/ws -v ~/.m2/:/root/.m2 --workdir=/ws jbangdev/jbang-action:0.99.0 --fresh --verbose info tools --deps io.quarkus:quarkus-bootstrap-core:999-SNAPSHOT:tests@test-jar mvnw.cmd
```

vs

```
docker run -v `pwd`:/ws -v ~/.m2/:/root/.m2 --workdir=/ws jbangdev/jbang-action:0.98.0 --fresh --verbose info tools --deps io.quarkus:quarkus-bootstrap-core:999-SNAPSHOT:tests@test-jar mvnw.cmd
```

quintesse:
Fixes #1482

### DIFF
--- a/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
@@ -36,6 +36,7 @@ import org.eclipse.aether.DefaultRepositorySystemSession;
 import org.eclipse.aether.RepositoryEvent;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.ArtifactType;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.CollectRequest;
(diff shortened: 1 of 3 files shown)


quintesse:
Confirmed by user, merging.

Issue #1476, opened 2022-10-06
JDK lookups should include sdkman installed JDKs


jeffmaury:
**Is your feature request related to a problem? Please describe.**
When a specific version of Java is required for a script, the JDK installed by sdkman should be looked

**Describe the solution you'd like**
I had Java 17.0.3 installed by sdkman and ran https://raw.githubusercontent.com/maxandersen/gh-jtriage/main/jtriage.java but jbang says Downloading JDK 16.
Not mentioning permission issues on Windows

**Describe alternatives you've considered**

**Additional context**

quintesse:
@jeffmaury can you tell us exactly which JDK you have installed using sdkman?

And if possible can you see if that installed JDK has a `release` file in its root folder? And if so could you paste its contents?

maxandersen:
@jeffmaury is your case you don't have that java 17 from sdkman in path and you expect jbang to go search on local disk for alternative sdk installs?

If yes then I'm not convinced that will provide a consistent behavior as sdkman has no notion of ordering of its jdks. 

You can if you want tell jbang about specific version of jdk to use:

"jbang jdk install 17 `sdk home java 17.0.4.1-tem`"

jeffmaury:
> @jeffmaury can you tell us exactly which JDK you have installed using sdkman?
> 
> And if possible can you see if that installed JDK has a `release` file in its root folder? And if so could you paste its contents?
> 
> Edit: Oh, and to be sure, if you type `java -version` in a terminal you get that 17..0.3 version, right?

I was running JBang with Java11.
I had:

- 8.0.302-tem
- 11.0.16.1-tem
- 17.0.3-tem
- 21.3.0.r11-grl

installed with sdkman.
17.0.3-tem has the following `release` file:
```

IMPLEMENTOR="Eclipse Adoptium"
IMPLEMENTOR_VERSION="Temurin-17.0.3+7"
JAVA_VERSION="17.0.3"
JAVA_VERSION_DATE="2022-04-19"
LIBC="default"
[...]

PR #1487, opened 2022-10-19
docs: Explain how you can install an existing jdk

jmini:
Update the documentation to mention what is discussed in https://github.com/jbangdev/jbang/issues/1476#issuecomment-1271208199

It is possible to use `jdk install` and provide the path of the installed JDK.
This can also be combined with SDKMAN!


maxandersen:
Thanks! good addition!

PR #1492, opened 2022-10-24
feat: add Java 19 support for native image

linux-china:
- Add `--enable-preview` and `--add-modules=jdk.incubator.concurrent` for Java 19 native build
- Update Kotlin default version to `1.7.20`
- Update Groovy default version to `3.0.13`

### DIFF
--- a/src/main/java/dev/jbang/source/buildsteps/NativeBuildStep.java
+++ b/src/main/java/dev/jbang/source/buildsteps/NativeBuildStep.java
@@ -31,6 +31,12 @@ public class NativeBuildStep implements Builder<Project> {
 		List<String> optionList = new ArrayList<>();
 		optionList.add(resolveInGraalVMHome("native-image", project.getJavaVersion()));
 
+		final int actualVersion = JavaUtil.javaVersion(project.getJavaVersion());
+		if (actualVersion == 19) {
+			optionList.add("--enable-preview");
+			optionList.add("--add-modules=jdk.incubator.concurrent");
+		}
+
 		optionList.add("-H:+ReportExceptionStackTraces");
 
 		optionList.add("--enable-https");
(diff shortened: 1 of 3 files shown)


src/main/java/dev/jbang/source/buildsteps/NativeBuildStep.java
```diff
@@ -31,6 +31,12 @@ public Project build() throws IOException {
 		List<String> optionList = new ArrayList<>();
 		optionList.add(resolveInGraalVMHome("native-image", project.getJavaVersion()));
 
+		final int actualVersion = JavaUtil.javaVersion(project.getJavaVersion());
+		if (actualVersion == 19) {
+			optionList.add("--enable-preview");
+			optionList.add("--add-modules=jdk.incubator.concurrent");
```
quintesse, line 37: Why is this option necessary exactly?

maxandersen: Same question here - why would we enable this by default for kotlin? We dont enable it by default in java

maxandersen: Sorry thought this was in kotlin part, most definitely Should not be actively enabled for native by default. Need to add explicit support for it to enable for native.

linux-china: > Since virtual threads are a preview feature in JDK 19, they are also a preview feature for GraalVM Native Image. They can be enabled with the new option --enable-preview.

https://docs.oracle.com/en/graalvm/enterprise/22/docs/release-notes/

quintesse: Ok, understood. In that case we really need https://github.com/jbangdev/jbang/issues/1493 instead of adding a bunch of hard-coded options.

quintesse: Created PR #1494 for #1493

maxandersen: @linux-china can you redo this PR without the enable preview /add modules flags?

PR #1494, opened 2022-10-25
feat: added support for specifying native options

### LINKED ISSUE #1493: Support `--native-options` and `//NATIVE_OPTIONS`
In #1397 `--native-options` was mentioned but it wasn't implemented in the PR. So opening this issue to track proper support for passing native options to the native image tool.

quintesse:
Added both `//NATIVE_OPTIONS` and `--native-options`

Fixes #1493


maxandersen:
the pr also introduces RUNTIME_OPTIONS and COMPILE_OPTIONS.

Good to make them slightly more generic but not a fan of longer names.... wdyt about "RUN_OPTIONS"?

I can't come up with shorter COMPILE_OPTIONS so I would probably personally just keep using JAVAC_OPTIONS but fine we have consistent naming :)

quintesse:
> the pr also introduces RUNTIME_OPTIONS and COMPILE_OPTIONS.

Ah yes, because in the previous PR we added `--compile-option` and `--runtime-option` I thought it best to add aliases for the code tags as well. But indeed, I forgot to mention them.

`JAVAC_OPTIONS` still works, but at least it means we don't have to add `KOTLINC_OPTIONS` and `GROOVYC_OPTIONS` and so forth for every language. (Although we could still do so and just internally name them all `compileOptions`, so if you prefer that it's certainly doable!)

And I don't mind `//RUN_OPTIONS` if you prefer that.

maxandersen:
naah - i'll live with RUNTIME_OPTIONS now that I see it read out.

Issue #1504, opened 2022-11-04
Regression with GitHub repositories authentication


gsmet:
With JBang 0.97.0, I got the public GitHub repositories authentication working by:
- pushing the GITHUB_TOKEN secret to the action
- injecting it as an env var
- name the GitHub repository where my action is `github`
- the default `~/.m2/settings.xml` for the `github` pushes the secret as the password:

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <servers>
    <server>
      <id>github</id>
      <username>${env.GITHUB_ACTOR}</username>
      <password>${env.GITHUB_TOKEN}</password>
    </server>
  </servers>
</settings>
```

With the latest version, this setup no longer works and I get a 401 Unauthorized error when JBang tries to get the artifact from the GitHub (not so) public repository.

The setup of the action:
https://github.com/gsmet/github-action-playground/blob/main/action.yml
(see how I installed JBang 0.97.0 or latest via SDKMAN!)

The workflow using the action:
https://github.com/gsmet/github-action-playground-target/blob/main/.github/workflows/publish-image.yml#L39-L44

A working workflow using 0.97.0 (only the Linux build is supposed to work as my dirty SDKMAN! install only works on Linux):
https://github.com/gsmet/github-action-playground-target/actions/runs/3395475566/jobs/5645406665

A failing workflow using latest:
https://github.com/gsmet/github-action-playground-target/actions/runs/3395508036/jobs/5645479380

maxandersen:
@quintesse something we missed when moving from shrinkwrap to aether raw ?

maxandersen:
Might just be we don't interpolate that ${} value.

quintesse:
That's definitely possible. But handling of the settings.xml file is done internally by Aether itself (we just tell it what file to read). But perhaps it needs some kind of special invocation to enable it. I've done a quick search but haven't been able to find anything. I'd need to spend more time on this.

maxandersen:
I can confirm that env variables does get interpolated, so it is something different.

maxandersen:
ok, figured out the root cause.

We explicitly create RemoteRepos where we skip what the Session has of auth settings meaning it completely ignores the auth settings.

[...]

PR #1505, opened 2022-11-05
feat: honor JBANG_DOWNLOAD_VERSION in bootstrap setup

maxandersen:
allows to do:

`curl -Ls https://sh.jbang.dev | JBANG_DOWNLOAD_VERSION=0.99.2 bash -s - app setup`

Similar for Windows but haven't tested it - thus marking this draft for now.

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -144,11 +144,15 @@ elif [ -f "$abs_jbang_dir/.jbang/jbang.jar" ]; then
   jarPath=$abs_jbang_dir/.jbang/jbang.jar
 else
   if [[ ! -f "$JBDIR/bin/jbang.jar" || ! -f "$JBDIR/bin/jbang" ]]; then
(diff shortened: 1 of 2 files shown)


maxandersen:
also added consistent printing of url if download fails.

maxandersen:
rationale is to make it easy to do regression testing and not requiring having to fallback to use sdkman or similar as seen in #1504

quintesse (approved):
LGTM

maxandersen:
Did you test it on windows? :)

quintesse:
Testing is for wussies ;-)
Should work now.

quintesse:
Btw, I don't how to do the same thing you showed in the issue description on Windows. There's no way on Windows (AFAIK) to set a variable and run a command at the same time, so you'd need to run two separate commands.

maxandersen:
Yes. Env setup on windows requires seperate step.

PR #1506, opened 2022-11-06
fix: apply auth from global settings to repositories and use ~/.m2/se…

### LINKED ISSUE #1504: Regression with GitHub repositories authentication

maxandersen:
…ttings.xml as global rather than user settings

fixes #1504

### DIFF
--- a/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
@@ -155,11 +155,13 @@ public class ArtifactResolver {
 		settings = newEffectiveSettings();
 		session = newSession(system, settings);
 		configureSession(session, settings);
+		List<RemoteRepository> partialRepos;
 		if (builder.repositories != null) {
-			repositories = builder.repositories.stream().map(this::toRemoteRepo).collect(Collectors.toList());
+			partialRepos = builder.repositories.stream().map(this::toRemoteRepo).collect(Collectors.toList());
 		} else {
-			repositories = newRepositories();
+			partialRepos = newRepositories();
 		}
+		repositories = system.newResolutionRepositories(session, partialRepos);
 	}
 
 	public List<ArtifactInfo> resolve(List<String> depIds) {
@@ -302,6 +304,7 @@ public class ArtifactResolver {
 		return new RemoteRepository.Builder(repo.getId(), "default", repo.getUrl())
 																					.setPolicy(getUpdatePolicy())
 																					.build();
+
 	}
 
 	private static Dependency toDependency(Artifact artifact) {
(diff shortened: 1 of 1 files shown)


maxandersen:
not fully tested yet but this should in *theory* fix it - at least now the auth handler is configured with settings.xml username/passwords.

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
```diff
@@ -155,11 +155,13 @@ private ArtifactResolver(Builder builder) {
...
 		configureSession(session, settings);
+		List<RemoteRepository> partialRepos;
 		if (builder.repositories != null) {
-			repositories = builder.repositories.stream().map(this::toRemoteRepo).collect(Collectors.toList());
+			partialRepos = builder.repositories.stream().map(this::toRemoteRepo).collect(Collectors.toList());
 		} else {
-			repositories = newRepositories();
+			partialRepos = newRepositories();
 		}
+		repositories = system.newResolutionRepositories(session, partialRepos);
```
quintesse, line 164: Great find! Where did you encounter this solution?

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
```diff
@@ -354,13 +357,13 @@ private Settings newEffectiveSettings() {
...
-				if (Files.exists(userSettings)) {
-					settingsFile = userSettings.toAbsolutePath();
+				Path globalSettings = Paths.get(System.getProperty("user.home"), ".m2", "settings.xml");
+				if (Files.exists(globalSettings)) {
+					settingsFile = globalSettings.toAbsolutePath();
 				}
 			}
 			if (settingsFile != null) {
-				settingsBuilderRequest.setUserSettingsFile(settingsFile.toFile());
+				settingsBuilderRequest.setGlobalSettingsFile(settingsFile.toFile());
```
quintesse, line 366: What exactly is the difference between using global and user?

quintesse (approved):
LGTM

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
maxandersen: Found by doing find references until I found where repos auth gets set. 

So...pure luck? :)

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
maxandersen: In our specific currrent situation probably not much but thought might as well just treat home dir settings as global settings if we ever want to introduce using user settings That can override global settings.

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
quintesse: Well obviously a bit more than luck. A bit like stumbling around in the dark but at least you brought one of those tiny keychain flashlights 😄

PR #1508, opened 2022-11-07
fix: Fixes dependent source discovery

### LINKED ISSUE #1502: Automatic source discovery fails if script contains //DEPS
**Describe the bug**
Given a Foo.java script and a companion Bar.java at the same level and no explicit `//SOURCES Bar.java` directive in Foo.java, running `jbang Foo.java` will yield different results depending on the presence of `//DEPS` directive or no.

**To Reproduce**
1. Have a Bar.java file containing 
```java
public class Bar{}
```
3. Have a Foo.java file containing
```java
public class Foo extends Bar {
        System.out.println("hello JBang");
    }
}
```
4. Running `jbang Foo.java` yields:
> [jbang] Building jar...
hello JBang
5. Now just add a `//DEPS` directive to Foo.java, like:
```java
//DEPS com.github.lalyos:jfiglet:0.0.9
public class Foo extends Bar {
        System.out.println("hello JBang");
    }
}
```
6. Running `jbang Foo.java` now yields:
> [jbang] Building jar...
/Users/fbricon/Dev/souk/jbangmain/Foo.java:4: error: cannot find symbol
public class Foo extends Bar {
                         ^
  symbol: class Bar
1 error

**Expected behavior**
JBang should be able to compile the same file, regardless whether `//DEPS` directives are present or not.

**JBang version**
>[jbang] jbang version 0.99.1
Cache: /Users/fbricon/.jbang/cache
Config: /Users/fbricon/.jbang
Repository:/Users/fbricon/.m2/repository
0.99.1

**Additional context**
Adding `//SOURCES Bar.java` prevents compilation failures when `//DEPS` is present

quintesse:
Parent directories for sources are now added to the compiler's source path so any types referenced in them that are available in the same directories can be found even when they are not explicitly mentioned in any `//SOURCES` lines. This also fixes the problem where adding a `//DEPS` line would cause the compiler to be unable to find those same types.

Fixes #1502

### DIFF
--- a/src/main/java/dev/jbang/source/generators/JshCmdGenerator.java
+++ b/src/main/java/dev/jbang/source/generators/JshCmdGenerator.java
@@ -44,8 +44,10 @@ public class JshCmdGenerator extends BaseCmdGenerator<JshCmdGenerator> {
 		List<String> optionalArgs = new ArrayList<>();
 
 		String requestedJavaVersion = getProject().getJavaVersion();
-		String javacmd;
-		javacmd = JavaUtil.resolveInJavaHome("jshell", requestedJavaVersion);
+		if (requestedJavaVersion == null) {
+			requestedJavaVersion = "9+";
+		}
+		String jshcmd = JavaUtil.resolveInJavaHome("jshell", requestedJavaVersion);
 
 		// NB: See https://github.com/jbangdev/jbang/issues/992 for the reasons why we
 		// use the -J flags below
@@ -91,7 +93,7 @@ public class JshCmdGenerator extends BaseCmdGenerator<JshCmdGenerator> {
 			Util.warnMsg("Java Flight Recording not possible when running via jshell.");
 		}
 
-		fullArgs.add(javacmd);
+		fullArgs.add(jshcmd);
 		addAgentsArgs(fullArgs);
 
 		fullArgs.addAll(jshellOpts(project.getRuntimeOptions()));
--- a/src/test/java/dev/jbang/cli/TestRun.java
+++ b/src/test/java/dev/jbang/cli/TestRun.java
@@ -572,7 +572,7 @@ public class TestRun extends BaseTest {
 
 		String result = prj.cmdGenerator().generate();
 
-		assertThat(result, startsWith("jshell"));
+		assertThat(result, matchesPattern("^.*jshell(.exe)?.*"));
 		assertThat(result, not(containsString("  ")));
 		assertThat(result, containsString("helloworld.jsh"));
 		assertThat(result, not(containsString("--source 11")));
@@ -662,7 +662,7 @@ public class TestRun extends BaseTest {
 
 		String result = prj.cmdGenerator().generate();
 
-		assertThat(result, startsWith("jshell "));
+		assertThat(result, matchesPattern("^.*jshell(.exe)?.*"));
 		assertThat(result, (containsString("classpath_example.java")));
 //		assertThat(result, containsString(" --source 11 "));
 		assertThat(result, not(containsString("  ")));


src/main/java/dev/jbang/source/buildsteps/CompileBuildStep.java
```diff
@@ -42,10 +43,24 @@ protected Project compile() throws IOException {
...
+			optionList.add("-classpath");
+			optionList.add(path);
 		}
 		optionList.addAll(Arrays.asList("-d", compileDir.toAbsolutePath().toString()));
 
+		// add -sourcepath for all source folders
+		List<String> srcDirs = project	.getMainSourceSet()
+										.getSources()
+										.stream()
+										.map(x -> x.getFile().getParent().toAbsolutePath())
```
maxandersen, line 55: if I read this right it is adding blindly folders parent - not taking into consideration the package hierarchy - thus this would add the wrong root afaics?

quintesse: Yes, that's what I realized too, it's also why it's still Draft 😅 

But that unfortunately means that I need to rethink the Project handling. I might need to add the idea of source folders to it, which is somewhat unfortunate because it adds complexity which until now wasn't necessary.

maxandersen: Could we do it on just one file per unique folder. To not require it run on multiple files again and again.

quintesse: I'm actually not so much worried about performance but more that the extraction of the information might be ambiguous. We're just doing a very simple match on lines that start with "package" so even a block comment that has the word "package" might match and throw things off. That was something that never bothered me before because at most it would cause some minor problems with the `edit` command, but now it would affect the actual running of the code and any mistake would result in fatal errors. So I'm somewhat hesitant suddenly requiring this information.

maxandersen:
fyi, the jshell tests fails locally for me too thus there is something causing jshell code to fail.

maxandersen:
i.e.

```
jbang helloworld.jsh
Unexpected exception reading startup: java.lang.RuntimeException: java.lang.IllegalArgumentException: error: --source-path requires an argument
java.lang.RuntimeException: java.lang.IllegalArgumentException: error: --source-path requires an argument
	at jdk.compiler/com.sun.tools.javac.api.JavacTool.getTask(JavacTool.java:198)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskPool.getTask(JavacTaskPool.java:177)
	at jdk.jshell/jdk.jshell.TaskFactory.runTask(TaskFactory.java:206)
	at jdk.jshell/jdk.jshell.TaskFactory.parse(TaskFactory.java:140)
	at jdk.jshell/jdk.jshell.TaskFactory.parse(TaskFactory.java:238)
	at jdk.jshell/jdk.jshell.Eval.sourceToSnippets(Eval.java:197)
	at jdk.jshell/jdk.jshell.Eval.eval(Eval.java:133)
	at jdk.jshell/jdk.jshell.JShell.eval(JShell.java:493)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.processSource(JShellTool.java:3563)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.processSourceCatchingReset(JShellTool.java:1310)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.processInput(JShellTool.java:1208)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.run(JShellTool.java:1181)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.startUpRun(JShellTool.java:1148)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.resetState(JShellTool.java:1104)
	at jdk.jshell/jdk.internal.jshell.tool.JShellTool.start(JShellTool.java:938)
	at jdk.jshell/jdk.internal.jshell.tool.JShellToolBuilder.start(JShellToolBuilder.java:261)
	at jdk.jshell/jdk.internal.jshell.tool.JShellToolProvider.main(JShellToolProvider.java:120)
Caused by: java.lang.IllegalArgumentException: error: --source-path requires an argument
	at jdk.compiler/com.sun.tools.javac.file.BaseFileManager.handleOption(BaseFileManager.java:234)
	at jdk.jshell/jdk.jshell.MemoryFileManager.handleOption(MemoryFileManager.java:343)
	at jdk.compiler/com.sun.tools.javac.api.ClientCodeWrapper$WrappedJavaFileManager.handleOption(ClientCodeWrapper.java:265)
	at jdk.compiler/com.sun.tools.javac.main.Arguments.doProcessArgs(Arguments.java:390)
	at jdk.compiler/com.sun.tools.javac.main.Arguments.processArgs(Arguments.java:347)
	at jdk.compiler/com.sun.tools.javac.main.Arguments.init(Arguments.java:246)
	at jdk.compiler/com.sun.tools.javac.api.JavacTool.getTask(JavacTool.java:185)
	... 16 more
Caused by: com.sun.tools.javac.main.Option$InvalidValueException: error: --source-path requires an argument
	at jdk.compiler/com.sun.tools.javac.main.OptionHelper.newInvalidValueException(OptionHelper.java:95)
	at jdk.compiler/com.sun.tools.javac.main.Option.handleOption(Option.java:1081)
	at jdk.compiler/com.sun.tools.javac.file.BaseFileManager.handleOption(BaseFileManager.java:232)
	... 22 more
```

so looks like needing to not set source path for jshell based launches?

maxandersen:
is this still relevant @quintesse ?

PR #1513, opened 2022-11-12
only add -g for javac invocations.

### LINKED ISSUE #1512: Universally adding '-g' to the compiler invocation breaks kotlin scripts
kotlinc doesn't support a `-g` flag so kotlin scripts can't be compiled.

evanchooly:
Only add -g for javac invocations so kotlin scripts can be compiled.

fixes #1512

### DIFF
--- a/itests/helloworld.kt
+++ b/itests/helloworld.kt
@@ -1,0 +1,11 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+// //DEPS <dependency1> <dependency2>
+//NATIVE_OPTIONS -O1
+//RUNTIME_OPTIONS -Dfoo=bar "-Dbar=aap noot mies"
+//MANIFEST foo bar=baz baz=${bazprop:nada}
+
+class helloworld {
+    fun main(vararg args: String) {
+        println("Hello " + if (args.length>0) args[0] else "World"));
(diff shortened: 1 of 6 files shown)


quintesse (changes requested):
Hi @evanchooly , great try, but unfortunately that's not the way to fix it 🙂 
If you want to improve it you're more than welcome!
The actual change, which I should have implemented in the first place, should be made here: https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/source/sources/JavaSource.java#L30
(you can take a look at this as an example of how: https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/source/sources/GroovySource.java#L39-L41)

quintesse:
Also, if you could add a test, similar to this one: https://github.com/jbangdev/jbang/blob/main/src/test/java/dev/jbang/source/TestBuilder.java#L32-L50 , where you assert that a "-g" does **not** exist for Kotlin, I would be eternally grateful 🙂

evanchooly:
done.  that should be closer to correct.

quintesse (approved):

quintesse:
Merged. Thanks for your patch @evanchooly !

PR #590, opened 2020-12-06
fix: use environment variable GITHUB_TOKEN while accessing GitHub API…

jeffmaury:
… (gist)

Fixes #168

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

PR #1516, opened 2022-11-17
feat: support basic-auth for loading scripts

micmeyer:
Adds support for basic-auth when loading a script.

Usage:
export JBANG_BASIC_AUTH_USERNAME=<username>
export JBANG_BASIC_AUTH_PASSWORD=<password>
jbang https://<company domain>/.....

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -63,6 +63,8 @@ public class Util {
 	public static final String JBANG_JDK_VENDOR = "JBANG_JDK_VENDOR";
 	public static final String JBANG_RUNTIME_SHELL = "JBANG_RUNTIME_SHELL";
 	public static final String JBANG_STDIN_NOTTY = "JBANG_STDIN_NOTTY";
+	public static final String JBANG_AUTH_BASIC_USERNAME = "JBANG_AUTH_BASIC_USERNAME";
+	public static final String JBANG_AUTH_BASIC_PASSWORD = "JBANG_AUTH_BASIC_PASSWORD";
 
 	public static final Pattern patternMainMethod = Pattern.compile(
 			"^.*(public\\s+static|static\\s+public)\\s+void\\s+main\\s*\\(.*",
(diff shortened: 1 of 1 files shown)


maxandersen:
hmm - I understand the need but would it make more sense we used maven style config for this ? basically use auth settings in /.m2/pom.xml ?

micmeyer:
Being able to store credentials in a file would definitely be nice. From a security point of view the credentials would have to be encrypted tough.

Maven and Ansible use a "master password" for this:
- Maven: https://blog.sonatype.com/2009/10/maven-tips-and-tricks-encrypting-passwords/
- Ansible: https://docs.ansible.com/ansible/latest/user_guide/vault.html

With the "tooling"  that is needed for users to encrypt individual passworts (Maven) or a whole file (Ansible) this would become quite a big feature.

I liked the solution that was chosen for the GITHUB_TOKEN (https://github.com/jbangdev/jbang/pull/590) since one doesn't have to store unencrypted passwords in a file.

micmeyer:
@maxandersen I would really like to see support for Basic-Auth in JBang. This would allow us to have branch-specific  scrips for setting up a workspace.
Any chance that this pull request will be merged? Or are you planning to add this functionality in combination with a bigger refactoring that supports encrypted configuration properties/files?

maxandersen:
I want something like this in too but just considering how. I know that the single global env variable won't work in the long run though. Thus im being hesitant to just add it straight up.

quintesse:
@maxandersen I do feel the same way as you , but after having thought about it I'm not coming up with alternatives that are much better, I also imagine that we won't support many more different authentication schemes so we'd probably have something like `BEARER` and `TOKEN` and that's it, right? It also seems that most services like GitHub or anything container based tend to depend on passing secrets using environment variables. Given all that I'm leaning towards applying this.

NB: The only thing I would like to see changed is the naming of the variables: `JBANG_AUTH_BASIC_USERNAME` and `JBANG_AUTH_BASIC_PASSWORD`, so future extensions would be likewise named, eg: `JBANG_AUTH_BEARER`, `JBANG_AUTH_TOKEN`, etc.

maxandersen:
merging with the understanding we'll want to broaden this in future. we'll treat the JBANG_AUTH as fallback values on hosts without specific setting in future.

PR #1517, opened 2022-11-19
feat: mastodon url support

maxandersen:
Twitter blocked no-javascript access but mastodon does not have this issue.

Thus this PR enables you to do this: 

`jbang https://fosstodon.org/@jbangdev/109367735752497165`

it will match any URL that matches `https://.*/@(\w+)/([0-9]+)` which afaics are fairly specific to mastodon style urls.


maxandersen:
That's my concern too ... what we could do is that if the content does not actually parse as html or don't have that specific element we look for we just return the raw string Wdyt?

quintesse:
Yeah that sounds like a good compromise 👍

PR #1518, opened 2022-11-21
chore: upgraded to Gradle 7

quintesse:


quintesse:
In the end I used that plugin you found that can generate the `BuildConfig` class. It was the simplest change.

maxandersen:
add line to ignore generated sources and now merging.

build.gradle
```diff
@@ -94,6 +94,7 @@ spotless {
 		importOrder 'java', 'javax', 'org', 'com', 'dev.jbang', ''
 		removeUnusedImports()
 		eclipse().configFile "misc/eclipse_formatting_nowrap.xml"
```
quintesse, line 96: Aaah, I had put my exclude in the wrong section. Duh.

maxandersen: Yeah. I was getting upset why it wasn't working ...until i noticed there was multiple sections :)

PR #1525, opened 2022-12-20
fix: make integrations work again

quintesse:
- properly passing tags to the integrations again
- fixed that the integrations were running externally in too many cases where the current JVM was matching but wasn't used.
- passing properties to external integrations


maxandersen (approved):
LGTM - makes the world sane again ;)

Issue #1044, opened 2021-10-19
jbang export maven/gradle project


maxandersen:
would be nice for completeness sake that we had a way to convert a jbang app to maven/gradle.

something like:

`jbang export maven .` or `jbang export gradle .`

somewhat related to #1042

tadayosi:
@maxandersen What's the current status of the feature?  I'd really like to have this feature. FYI, Apache Camel already has it for Camel apps:
https://camel.apache.org/manual/camel-jbang.html#_creating_projects

maxandersen:
just waiting for that contribution back to jbang :)

PR #1533, opened 2022-12-31
rough cut of export project

maxandersen:
initial cut of `jbang export project --group=org hello.java`

still very rough.


maxandersen:
#1044 does this and more :)

PR #1539, opened 2023-01-11
feat: adding auto download for sources via edit and JBANG_DOWNLOAD_SOURCES=true

maxandersen:
adding option to have jbang download sources for dependencies.

approach is a bit a naive and will only work if all //DEPS resolves properly but simply
takes the resolved set and then one by one try resolve matching sources classifier.


maxandersen:
for now we'll change this to use JBANG_DOWNLOAD_SOURCES env var to turn on/off.

in future when we have the resolver pluggable/configurable we can expose a more explicit flag.

quintesse (approved):
LGTM. We do this quick & dirty solution and then do it properly once we factor out that resolver, as we discussed.

fbricon:
you merged too soon ;-)

jbang-eclipse doesn't rely on `jbang edit`, but calls `jbang info tools`, that's when I expect sources to be downloaded

maxandersen:
Jbang edit will do it automatically. 

Set JBANG_DOWNLOAD_SOURCES=true to have it happen on any command.

PR #1542, opened 2023-01-17
Allow JREs

quintesse:
This PR allows JREs to be used by Jbang. This should make it possible to use much of Jbang's functionality without having a JDK installed. Except for the most important of course, which is being able to `build`.

But this will let you for example list the available JDKs and decide which one you want to install instead Jbang first installing a default version, like 11, before it will allow you do do anything else.


maxandersen:
rebase?

quintesse:
Still not marked as ready :-)
I'm just not 100% happy with it, even though it's a pretty nice feature.

bartoszpop:
@quintesse any chances to have it completed? When running JBang in a Docker image, it is reasonable to use a headless JRE to reduce the image size.

quintesse:
This is still something I want to implement, yes. Especially when the extraction of the JDK management code into its own module has completed (see https://github.com/jbangdev/jbang/pull/1983). It won't be by merging this PR though, so I'm going to open an issue for it (#1994) and will close this PR.

PR #1555, opened 2023-02-03
feat: implemented proper cache control

quintesse:
All remote requests now use `ETag` and `Last-Modified` information.
This should make sure that all locally cached information should be as up-to-date as possible and it should not be necessary to use `--fresh` that often anymore.

The config option `cache-evict` can be set to a non-zero value if the user wants to be more efficient and not generate requests for all remote information that's already in the cache. For now it's set to 0 by default meaning that jbang always checks all locally cached files for up-to-dateness.

And finally any time that a resource resolve fails and the resource reference looks like it could be an alias or a remote URL we'll now retry with the `cache-evict` set to a very low value thereby making sure we'll be updating any catalogs that have been updated since the last time we retrieved them.

NB: if you quickly scan the commits you'll see talk about using `HEAD` requests. That is no longer the case in the final code. This was changed to use `If-Modified-Since` and `If-None-Match` headers and handling `304 Not Modified` HTTP result codes with regular `GET` requests instead.


quintesse:
@maxandersen created this especially for you :-)
No more adding `--fresh` everywhere.
It _does_ make things slower though because we'll be doing many more requests, even if they are simple `304 Not Modified` requests.
Personally I used `jbang config set cache-evict P1D` to make jbang only check once a day (per resource) which makes things almost as fast as before, so that seems to be a good compromise.

maxandersen:
Could we make it so if alias fail to resolve alias in a catalog then it would still check if changed so could still "just work" and then we could Till have default be once a day?

quintesse:
@maxandersen done. Any failed resource retrieval from a (potential) remote resource will now trigger a retry with cache eviction. This should be mergeable now.

maxandersen:
good PR and looks on first looks a-ok.

two things though:
1) some docs for the options would be nice 
2) i had expected --verbose to show some requests?

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -276,11 +277,27 @@ public Project build(String resource) {
...
+			ref = resolver.resolve(resource);
+		} catch (ExitException ee) {
+			if (ee.getStatus() != BaseCommand.EXIT_INVALID_INPUT || !retryCandidate) {
+				throw ee;
+			}
+		}
+		if (ref == null && retryCandidate) {
+			// We didn't get a result and the resource looks like something
+			// that could be an alias or a remote URL, so we'll try again
+			// with the cache evict set to 0, forcing Jbang to actually check
```
maxandersen, line 296:
```suggestion
			// with the cache evict set to 0, forcing JBang to actually check
```

maxandersen:
- set default cache to P12H to have likelihood for updates happens but don't bog down normal usage.
- add verbose logging for the URL connect so we can debug user reports :)

we agree to not print cache updates as this will be jbangs new default for alias handling - users can use --fresh as always or add sha/tag/branch to their alias, i.e. `jbang kill@quarkusio//f294e3f7cf76f3fa7b154da29e1b32a3ba65b3e2` or `jbang quarkus@quarkusio//3.0.0.Alpha3` (note the "//" double slash to avoid having to use full catalog name)

PR #1567, opened 2023-02-16
feat: added support for modules

quintesse:
There is now a new source tag `//MODULE <name>` which signifies that the code should be compiled as a module with the given name. The dependencies defined are automatically added as required modules. This of course requires at least Java 9 so the minimal version for scripts with the `//MODULE` tag is automatically set to 9.


quintesse:
@maxandersen this actually works, I implemented it because I needed some quick and dirty tests dealing with modules and wanted to use Jbang. So I spent a couple of days on this, which got rid of the "quick" in "quick and dirty", but this is the result :slightly_smiling_face: 

Now I don't know that much about modules so there might be all kinds of things wrong with this implementation, but for example the jetty.java script in the examples works simply by adding a `//MODULE` tag and updating the dependencies to the latest Jetty version (which supports modules).... well that and I turned it into a .java file instead of .jsh. Haven't thought about what `//MODULE` means in the context of a jshell file, or if that's even possible.

quintesse:
Ah yes, and the build file is probably a mess. Gradle... don't get me started :wink:

src/main/java/dev/jbang/source/buildsteps/CompileBuildStep.java
```diff
@@ -111,6 +133,43 @@ protected Path generatePom() throws IOException {
...
+												.collect(Collectors.toSet());
+			// Now filter out the resolved artifacts that are root dependencies
+			// and get their names
+			List<String> moduleNames = project	.resolveClassPath()
+												.getArtifacts()
+												.stream()
+												.filter(a -> deps.contains(a.getCoordinate()))
+												.map(ArtifactInfo::getModuleName)
+												.collect(Collectors.toList());
+			// Finally create a module-info file with the name of the module
```
maxandersen, line 158: what happens if user provides their own module-info file?

maxandersen:
got test/example on something needing this? ...and how are dependencies generally converted to modules in moduleinfo? not all GAV's has module names do they?

src/main/java/dev/jbang/source/buildsteps/CompileBuildStep.java
quintesse: No idea :-)

quintesse:
> got test/example on something needing this? ...and how are dependencies generally converted to modules in moduleinfo? not all GAV's has module names do they?

No, no tests yet, I need to figure out how to tell JUnit not to test something when on Java 8.

And dependencies that are not modules are skipped and not added as a `require`.

maxandersen:
I mean you said you wrote some 'thing with it that worked ? :) Just curious what it was to have something to try it with?

maxandersen:
this needs some docs

aalmiray:
> > got test/example on something needing this? ...and how are dependencies generally converted to modules in moduleinfo? not all GAV's has module names do they?
> 
> No, no tests yet, I need to figure out how to tell JUnit not to test something when on Java 8.
> 
> And dependencies that are not modules are skipped and not added as a `require`.

Somoun @sormuras

sormuras:
> No, no tests yet, I need to figure out how to tell JUnit not to test something when on Java 8.

https://junit.org/junit5/docs/current/user-guide/#writing-tests-conditional-execution-jre

```java
@Test
@DisabledOnJre(JAVA_8)
void notOnJava8() {
    // ...
}
```

quintesse:
Ok, this is ready for some hammering and some merging :slightly_smiling_face:

quintesse:
Oh and if someone could look at the build... it works in Gradle, but most likely not in the best possible way. And in IntelliJ you'll always see a compile error because it thinks it should use Java 8. I have no idea how to tell it that there is a single file that has to be compiled with Java 9. I tried fooling around with (IntelliJ) modules, but no luck.

docs/modules/ROOT/pages/running.adoc
```diff
@@ -73,6 +73,39 @@ Since Java 9 https://docs.oracle.com/javase/9/tools/java.htm#GUID-3B1CE181-CD30-
...
+
+This will make Jbang treat the code a being a module with the given name, and it will compile it as such.
+Any declared dependencies are automatically marked as being required.
+
+You can also set this on the command line by using the `--module=<name>` option on the `jbang build` command.
+The option also exists for `jbang run` where it will treat anything to be run as a module.
+
+== Setting main class
+
+Normally Jbang will figure out by itself what your code's main class is, but in case you want to override
```
maxandersen, line 93:
```suggestion
Normally JBang will figure out by itself what your code's main class is, but in case you want to override
```

docs/modules/ROOT/pages/running.adoc
```diff
@@ -73,6 +73,39 @@ Since Java 9 https://docs.oracle.com/javase/9/tools/java.htm#GUID-3B1CE181-CD30-
...
+== Module support (EXPERIMENTAL)
+
+You can mark code as being part of a module by adding the following line:
+
+[source, java]
+----
+//MODULE <module-name>
+----
+
+This will make Jbang treat the code a being a module with the given name, and it will compile it as such.
```
maxandersen, line 85:
```suggestion
This will make JBang treat the code a being a module with the given name, and it will compile it as such.
```

PR #1575, opened 2023-02-21
fix: Symbolic link permission issues on Windows

### LINKED ISSUE #1574: Investigate Symbolic/Hard link permission issues on Windows
Setting up WIndows to be able to use symbolic and hard links can be somewhat of a challenge.

We should be able to give the user straightforward instruction on what to do.
Right now our documentation is sorely lacking in that aspect.

quintesse:
Fixes #1574

### DIFF
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -346,21 +346,24 @@ No one would want to do that (right?) but now you know.
 
 Some JBang commands need to create symbolic links when running on Windows.
 For example, this is required for Managing JDKs or editing the files with the `edit` command.
-To resolve the issue, on Windows 10 you will need to enable symbolic linking for your user account.
 
-* Review security considerations and best practices https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links[here].
-* Enable the `Create symbolic links` group policy setting for your user/group
-  by following the guidelines on the https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/user-rights-assignment[User Rights Assignment] page.
+If you encounter issues on Windows related to the creation of symbolic links follow
+these instructions:
 
-You might also need elevated privileges to create symbolic links
-If the JBang process has no full administrative rights, the execution may result in an error like:
+1. From Windows 10 onwards you can turn on "Developer Mode", this will automatically
+enable the possibility to create symbolic links. Read here how to enable this mode:
+https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development[Enable your device for development]. On Windows 11 this might already
+be enabled by default.
 
-```
-java.nio.file.FileSystemException: <CURRENT_JDK>: A required privilege is not held by the client.
-```
+2. If you're using a Java version equal to or newer than 13 then you're good to go.
+This Java version already works correctly. Make sure that JBang is actually using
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/pages/usage.adoc
```diff
@@ -346,21 +346,24 @@ No one would want to do that (right?) but now you know.
...
+1. From Windows 10 onwards you can turn on "Developer Mode", this will automatically
+enable the possibility to create symbolic links. Read here how to enable this mode:
+https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development[Enable your device for development]. On Windows 11 this might already
+be enabled by default.
 
-```
-java.nio.file.FileSystemException: <CURRENT_JDK>: A required privilege is not held by the client.
-```
+2. If you're using a Java version equal to or newer than 13 then you're good to go.
+This Java version already works correctly. Make sure that Jbang is actually using
```
maxandersen, line 359:
```suggestion
This Java version already works correctly. Make sure that JBang is actually using
```

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -346,21 +346,24 @@ No one would want to do that (right?) but now you know.
...
-If the JBang process has no full administrative rights, the execution may result in an error like:
+1. From Windows 10 onwards you can turn on "Developer Mode", this will automatically
+enable the possibility to create symbolic links. Read here how to enable this mode:
+https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development[Enable your device for development]. On Windows 11 this might already
+be enabled by default.
 
-```
-java.nio.file.FileSystemException: <CURRENT_JDK>: A required privilege is not held by the client.
-```
+2. If you're using a Java version equal to or newer than 13 then you're good to go.
```
maxandersen, line 358: I guess we got the first real reason to default to download java 17 if nothing else installed?

src/main/scripts/jbang.ps1
```diff
@@ -39,6 +39,15 @@ if ([System.Enum]::GetNames([System.Net.SecurityProtocolType]) -notcontains 'Tls
     break
 }
 
+$DevModRegistryPath = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock"
+if (!(Test-Path -Path $DevModRegistryPath) -or (Get-ItemProperty -Path `
+        $DevModRegistryPath -Name AllowDevelopmentWithoutDevLicense -ErrorAction `
+        SilentlyContinue).AllowDevelopmentWithoutDevLicense -ne 1) {
+    [Console]::Error.WriteLine("WARNING: Windows Developer Mode is not enabled on your system, this is necessary");
+    [Console]::Error.WriteLine("for Jbang to be able to function correctly, see this page for more information:");
```
maxandersen, line 47:
```suggestion
    [Console]::Error.WriteLine("for JBang to be able to function correctly, see this page for more information:");
```

docs/modules/ROOT/pages/usage.adoc
quintesse: Well I would hope that wasn't the _only_ reason :rofl:

quintesse: But yeah, I was thinking that if we still wouldn't want to go too high then _at least_ 14. But then we'd miss out on all those sweet new features, so yeah I'm all for 17 LTE :smile:

PR #1577, opened 2023-02-22
feat: added support for remote-file arguments

### LINKED ISSUE #1043: Generic remote-file argument support for `run`
It was a bit hard to come up with a good title, but I was basically thinking about something like this:

Imagine that you have a script that reads a local file:

```
jbang run readFile.java subdir/file.ext
```

But now imagine that you want to have it read a remote file, but the app doesn't support that.
But JBang already has all the code necessary to download and cache files locally.
So why not have some special syntax that lets JBang handle that? For example:

```
jbang run readFile.java @https://example.org/subdir/file.ext
```

We'd look for any arguments starting with `@http://` or `@https://` and replace them with paths to cached versions of those files.
To escape this we could do `@@http://`, which would then be replaced with the literal `@http://`.

quintesse:
Fixes #1043

### DIFF
--- a/docs/modules/ROOT/pages/running.adoc
+++ b/docs/modules/ROOT/pages/running.adoc
@@ -100,3 +100,18 @@ You can activate a javaagent using `--javaagent=<agent>[=<options>]` where agent
 It can also be a jbang script itself where you have put `//JAVAAGENT` to activate agent packaging.
 
 You can create a basic agent using `jbang init -t agent myagent.java` to get started.
+
+== Remote File Arguments
+
+The `run` command has a special feature where any of the arguments for the code to be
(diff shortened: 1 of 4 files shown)


quintesse:
Sorry, couldn't resist, it's just oo useful a feature not to implement it :grin:

maxandersen:
Maaaan...I agree this is such a tempting thing to do. But boy it feels wrong too :)

quintesse:
@maxandersen the new version supports what you were looking for with `--javaagent`.
I have also changed the special character to `%`.

maxandersen:
Interesting. Docs are out of sync?

quintesse:
Oops, updated and extended docs

maxandersen:
this is too good to not do.

PR #1578, opened 2023-02-22
fix: all default values can be set by config

### LINKED ISSUE #1576: Not all @Option(defaultValue) have been added to the defaut config
There are some PicoCLI options that have been added after the move to using `Configuration` that are using `defaultValue`.
Those should be removed and added to the default config file instead.

(My own fault btw)

quintesse:
Some default values were not settable by config, an oversight.
Also added the possibility to globally set default values for all commands that support the same option.

Fixes #1576

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -14,22 +14,27 @@ endif::[]
 
 toc::[]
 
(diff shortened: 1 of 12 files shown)


maxandersen:
What does:
"Also added the possibility to globally set default values for all commands that support the same option." Refer to?

quintesse:
> What does: "Also added the possibility to globally set default values for all commands that support the same option." Refer to?

See the commit where I changed the docs

PR #1579, opened 2023-02-23
fix: fixed issue with PowerShell quoting

quintesse:
Turns out the regular expression was formatted wrong.


quintesse:
Nasty bug @maxandersen ☹️ 
Nobody must be using JBang with Powershell or they would have noticed almost immediately.

PR #1583, opened 2023-02-23
feat: set Java 17 as the default for new installations

### LINKED ISSUE #1582: JBang default to run with Java 17?
### Discussed in https://github.com/jbangdev/jbang/discussions/1088

<div type='discussions-op-text'>

<sup>Originally posted by **maxandersen** November 22, 2021</sup>
Currently jbang defaults to Java 11 if no other java found.

Should we move it up to Java 17 ? 

Reasons for: 
  records and multi line strings are super useful when writing scripts and utilities

Reasons against:
gradle still seem to suck when running under Java 17 so jbang edit experience aint great
many(?) libraries still have challenges with Java 17 due to the enforced locked down jdk.

Because if reasons against i'm leaning on staying on Java 11 but I wish someone could 
convince me otherwise.

Thoughts?</div>

quintesse:
Fixes #1582


maxandersen:
perfect.

PR #1589, opened 2023-02-24
feat: many run flags for aliases and `--manifest`

quintesse:
Added most remaining run flags to aliases.
Finished implementing --manifest option.


quintesse:
Merging this because most of it is just adding fields to Alias and finishing a cli flag that turned out to be incomplete.
One thing that could be considered a feature is that the new remote-files can now also be used by aliases (before they only worked when specified directly on the command line). Minor but can save a lot of typing :-)

PR #1590, opened 2023-02-27
Update to gradle 8

quintesse:
Wondering if we shouldn't make sure we're on the latest version of all the build tools. So far things seem to work.


maxandersen:
generated jar has no differences thus +1

PR #1591, opened 2023-02-27
fix: agents shouldn't be compiled as sub projects

quintesse:


quintesse:
Simple change. Merging.

PR #1593, opened 2023-02-28
Added `export jlink`

quintesse:


src/main/java/dev/jbang/cli/Export.java
```diff
@@ -378,3 +385,85 @@ private Path getFatjarOutputPath() {
 		return outputPath;
 	}
 }
+
+@Command(name = "jlink", description = "Exports a minimized JDK distribution")
+class ExportJlink extends BaseExportCommand {
+
+	@CommandLine.Parameters(index = "1..*", arity = "0..*", description = "Parameters to pass on to the jlink command")
+	public List<String> params = new ArrayList<>();
```
maxandersen, line 393: we normally  treat arguments that comes after the script as args to the script not to the underlying tools.

should this not be a dedicated arg instead?

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -239,11 +235,38 @@ ProjectBuilder createProjectBuilder() {
 @CommandLine.Command(name = "tools", description = "Prints a json description usable for tools/IDE's to get classpath and more info for a jbang script/application. Exact format is still quite experimental.")
 class Tools extends BaseInfoCommand {
 
+	@CommandLine.Option(names = {
```
maxandersen, line 238: not a fan of --select the name; but i don't have a better one - lets at least get docs for these changes added.

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -267,24 +295,13 @@ public Integer doCall() throws IOException {
 	}
 }
 
-@CommandLine.Command(name = "modulepath", description = "Prints module-path used for this application using operating system specific path separation.")
-class ModulePath extends BaseInfoCommand {
```
maxandersen, line 271: was it your intention to remove modulepath?

src/main/java/dev/jbang/util/ModuleUtil.java
```diff
@@ -1,13 +1,37 @@
...
+		try {
+			URL url = new URL("jar:" + file.toUri().toURL() + "!/module-info.class");
+			try (InputStream s = url.openStream()) {
+				return true;
+			}
+		} catch (Exception ex) {
+		}
+		try {
+			// TODO This is a very specific test, we should do better
+			URL url = new URL("jar:" + file.toUri().toURL() + "!/META-INF/versions/9/module-info.class");
```
maxandersen, line 26: what is this trying to do ? detect multi-jarred modules?

maxandersen:
when I run: `jbang export jlink env@jbangdev`

I get:

```
[jbang] [ERROR] Module code is missing a 'package' statement
[jbang] Run with --verbose for more details`
```

I reckon that's an error message from jlink complaining about missing packages - could we improve that and give user some more actionable advice? I.e. "jlink does not work with the default empty package name. Add a package and try again" or similar?

maxandersen:
is it expected to get no main class declared? 

```
jbang export jlink sqlsh@maxandersen
[jbang] [WARN] Export might fail because some dependencies are not full modules: org.apache.calcite:calcite-plus:jar:1.33.0, org.apache.calcite:calcite-core:jar:1.33.0, org.locationtech.jts:jts-core:jar:1.19.0, org.locationtech.jts.io:jts-io-common:jar:1.19.0, com.googlecode.json-simple:json-simple:jar:1.1.1, org.locationtech.proj4j:proj4j:jar:1.2.2, com.google.errorprone:error_prone_annotations:jar:2.11.0, org.yaml:snakeyaml:jar:1.33, com.google.uzaygezen:uzaygezen-core:jar:0.2, com.jayway.jsonpath:json-path:jar:2.7.0, net.minidev:json-smart:jar:2.4.7, net.minidev:accessors-smart:jar:2.4.7, com.yahoo.datasketches:sketches-core:jar:0.9.0, com.yahoo.datasketches:memory:jar:0.9.0, commons-codec:commons-codec:jar:1.13, net.hydromatic:aggdesigner-algorithm:jar:6.0, commons-lang:commons-lang:jar:2.4, commons-logging:commons-logging:jar:1.1.3, org.apache.commons:commons-dbcp2:jar:2.6.0, org.apache.commons:commons-pool2:jar:2.6.1, org.apache.commons:commons-lang3:jar:3.8, org.apache.commons:commons-math3:jar:3.6.1, commons-io:commons-io:jar:2.11.0, org.codehaus.janino:commons-compiler:jar:3.1.8, org.codehaus.janino:janino:jar:3.1.8, org.apache.calcite:calcite-linq4j:jar:1.33.0, net.hydromatic:quidem:jar:0.10, org.apache.calcite.avatica:avatica-core:jar:1.23.0, org.apache.calcite.avatica:avatica-metrics:jar:1.23.0, com.google.protobuf:protobuf-java:jar:3.21.9, org.apache.httpcomponents.client5:httpclient5:jar:5.1.3, org.apache.httpcomponents.core5:httpcore5-h2:jar:5.1.3, org.apache.httpcomponents.core5:httpcore5:jar:5.1.3, org.checkerframework:checker-qual:jar:3.12.0, com.google.guava:guava:jar:31.1-jre, com.google.guava:failureaccess:jar:1.0.1, com.google.guava:listenablefuture:jar:9999.0-empty-to-avoid-conflict-with-guava, com.google.code.findbugs:jsr305:jar:3.0.2, com.google.j2objc:j2objc-annotations:jar:1.3, com.teradata.tpcds:tpcds:jar:1.2, com.google.code.findbugs:annotations:jar:2.0.3, io.airlift:airline:jar:0.7, javax.inject:javax.inject:jar:1, io.prestosql.tpch:tpch:jar:1.0, net.hydromatic:chinook-data-hsqldb:jar:0.2, net.hydromatic:tpcds:jar:0.4, org.apache.calcite.avatica:avatica-server:jar:1.23.0, javax.servlet:javax.servlet-api:jar:4.0.1, org.eclipse.jetty:jetty-http:jar:9.4.44.v20210927, org.eclipse.jetty:jetty-io:jar:9.4.44.v20210927, org.eclipse.jetty:jetty-security:jar:9.4.44.v20210927, org.eclipse.jetty:jetty-server:jar:9.4.44.v20210927, org.eclipse.jetty:jetty-util:jar:9.4.44.v20210927, org.hsqldb:hsqldb:jar:2.5.2
[jbang] [WARN] No launcher will be generated because no main class is defined. Use '--main' to set a main class
[jbang] [ERROR] Unable to export Jdk distribution.
```

maxandersen:
basically - I got nothing that works :) should we add a `jbang init -t module hello.java` template or even a `jbang init -Dmodule hello.java`

maxandersen:
I would assume this works:

```
//MODULE 
package blah;

public class hello {

        out.println("Hello World");
    }
}
```

but fails requiring module identifier...shouldn't that default to `hello` based on hello.java ?

maxandersen:
manage to get something: 

```
jbang export jlink hello.java
[jbang] Building jar...
[jbang] Exported to /Users/manderse/code/maxandersen/jbang-catalog/hello-jlink
```

should we print info on what was just generated?, i.e.

```
jbang export jlink hello.java
[jbang] Building jar...
[jbang] Exported to /Users/manderse/code/maxandersen/jbang-catalog/hello-jlink
[jbang] Run using hello-jlink/bin/hello
```

src/main/java/dev/jbang/cli/Export.java
quintesse: In this case it's the intention that _all_ extra parameters are passed directly on to the underlying `jlink` command. This way you have the full feature set of that command without us having to add all those options.

quintesse: Btw, the treating parameters after the script as arguments to the script is only true the `run` and `alias` commands (and `alias` is basically a delayed `run` command), so IMO we're free to re-interpret that for other commands that are not run-related.

src/main/java/dev/jbang/cli/Info.java
quintesse: It's a _very_ minor and technical feature and we don't even mention `jbang info` outside the CLI docs. So I'm not sure what kind of docs do you want for this that won't already be in the generated CLI docs?

src/main/java/dev/jbang/cli/Info.java
quintesse: Yup, I added it when I added the module support without really understanding the difference between class-path and module-path in Java. I thought a module path would only contain modules and the class-path the rest of the dependencies that are not modules. But that's not the case.

To explain it a bit more: in Java choosing to use the class-path or the module-path is basically like a switch that tells Java to go into module mode or not. If you put everything on the class-path it will just work like old-fashioned unmodulerized Java. If you put things on the module-path it will switch to module mode and everything has to be modules or they will be automatically turned into modules.

In JBang that's now controlled by the `--module` parameter, if it exists we put the dependencies on the module-path, if not we use the class-path. But in both cases they will have the exact same list of dependencies. So it just didn't make sense to have a new command that returned the exact same information.

src/main/java/dev/jbang/util/ModuleUtil.java
quintesse: Not multi-jarred, but multi-versioned. JARs that can run on, for example, Java 8 and Java 9 but have slightly different code for each. In this case I've found JARs that have everything compiled with Java 8 but have a single module-info.class file compiled with Java 9

quintesse:
> I.e. "jlink does not work with the default empty package name. Add a package and try again" or similar?

Well, it's not so much jlink, as Java modules in general. So that's a limitation of the Java language.
So I'm not against adding better messages of course, but I simply thought it's not our job to teach people how to use Java 😆

quintesse:
> basically - I got nothing that works :)

😄 

Yeah, you need at least some code that _could_ be a module. The first rule is that it _must_ have a package.
And for jlink there is indeed an additional requirement that the dependencies you use must _all_ support modules fully.

quintesse:
> but fails requiring module identifier...shouldn't that default to hello based on hello.java ?

It could, yes, it just didn't seem terribly obvious, although I guess if someone would have to guess it would be the first thing they think of.

quintesse:
> [jbang] Run using hello-jlink/bin/hello

Sure, we could do that

quintesse:
> is it expected to get no main class declared?

It isn't. Thanks!

src/main/java/dev/jbang/cli/Export.java
maxandersen: we somewhat did it for jbang edit too - but with the new edit that won't hold up.

still - this thing of passing directly to jlink feels wrong to me.... can you give an example of how/what it would look like? what arguments would you want to put there?

src/main/java/dev/jbang/cli/Info.java
quintesse: Well, like I said "I'm not sure what kind of docs do you want for this that won't already be in the generated CLI docs"?

src/main/java/dev/jbang/cli/Export.java
quintesse: Well this is what jlink supports:

```
      --add-modules <mod>[,<mod>...]    Root modules to resolve
      --bind-services                   Link in service provider modules and
                                        their dependences
  -c, --compress=<0|1|2>                Enable compression of resources:
                                          Level 0: No compression
                                          Level 1: Constant string sharing
                                          Level 2: ZIP
      --disable-plugin <pluginname>     Disable the plugin mentioned
      --endian <little|big>             Byte order of generated jimage
                                        (default:native)
  -h, --help, -?                        Print this help message
      --ignore-signing-information      Suppress a fatal error when signed
                                        modular JARs are linked in the image.
                                        The signature related files of the
                                        signed modular JARs are not copied to
                                        the runtime image.
      --launcher <name>=<module>[/<mainclass>]
                                        Add a launcher command of the given
                                        name for the module and the main class
                                        if specified
      --limit-modules <mod>[,<mod>...]  Limit the universe of observable
                                        modules
      --list-plugins                    List available plugins
  -p, --module-path <path>              Module path
      --no-header-files                 Exclude include header files
      --no-man-pages                    Exclude man pages
      --output <path>                   Location of output path
      --post-process-path <imagefile>   Post process an existing image
      --resources-last-sorter <name>    The last plugin allowed to sort
                                        resources
      --save-opts <filename>            Save jlink options in the given file
  -G, --strip-debug                     Strip debug information
      --suggest-providers [<name>,...]  Suggest providers that implement the
                                        given service types from the module path
  -v, --verbose                         Enable verbose tracing
      --version                         Version information
      @<filename>                       Read options from file
```

Almost any of those options could be useful to a user and jlink even has some options that are not in this list, you can add `-J` for example to pass options to the underlying JVM for example. And the command can be extended with modules (exactly what I'm using it for) that add their own options to the command dynamically.

So we can't really predict what flags the user will need. Which is why passing on anything seems like the most useful option.

Why does passing those options feel bad to you?

maxandersen: mkay.

I'll accept it. 

reason i dont like it is that the command line args for the app is useful...but I realize we could if/when that becomes relevant for some commands we can do like git does and uses -- to separate between branch/flags and additional arguments/files.

src/main/java/dev/jbang/cli/Info.java
maxandersen: well to actually list it in the main docs that you can do `jbang info jarpath GAV`

quintesse: Aah, ok, I thought you meant I had to mention that `--select` option somehow. But you're talking about `info` in general.So sure, I agree, but that would be unrelated to this PR :-)

PR #1595, opened 2023-03-19
feat: jbang init --edit to have editor open directly

maxandersen:

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -39,6 +39,9 @@ public class Init extends BaseCommand {
 			"--force" }, description = "Force overwrite of existing files")
 	public boolean force;
 
+	@CommandLine.Option(names = { "--edit" }, description = "Open editor on the generated file(s)")
+	public boolean edit;
+
 	@CommandLine.Option(names = { "-D" }, description = "set a system property", mapFallbackValue = "true")
 	public Map<String, Object> properties = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


maxandersen:
quick'n'dirty...

src/main/java/dev/jbang/cli/Init.java
```diff
@@ -116,13 +119,22 @@ public Integer doCall() throws IOException {
...
-				+ "' or edit it using 'jbang edit --open=[editor] "
-				+ renderedScriptOrFile + "' where [editor] is your editor or IDE, e.g. '"
-				+ Edit.knownEditors[new Random().nextInt(Edit.knownEditors.length)]
-				+ "'. If your IDE supports JBang, you can edit the directory instead: 'jbang edit . '"
-				+ renderedScriptOrFile + ". See https://jbang.dev/ide");
-
+		if (edit) {
+			info("File initialized. Opening editor for you. You can also now run it with 'jbang "
+					+ renderedScriptOrFile);
+			// TODO: quick hack that gets the job of opening editor done; but really should
```
quintesse, line 125: To me this actually isn't a hack but a completely valid way of doing things. An API _might_ be nice but could very well be overkill for something that (so far at least) isn't meant to be used as a library.

quintesse (approved):
Love it :-)

src/main/java/dev/jbang/cli/Init.java
maxandersen: its just thtat it prints out you can run with -b which is not the case for this ;)

PR #1602, opened 2023-03-28
fix: trim trailing whitespaces in //MODULE directive

### LINKED ISSUE #1601: `//MODULE ` with a trailing space causes an error
**Describe the bug**
jbang-vscode provides completion for `//MODULE`, but it inserts a space at the end, so the user can type the module name directly. However, if you don't specify a custom module name, JBang will error with:
```
[jbang] [ERROR] //MODULE line has wrong format, should be '//MODULE [identifier]'
```

**To Reproduce**
1. Create a JBang file in a package, eg. foo/Hello.java
2. Add the `//MODULE ` directive (with trailing space(s))
3. Run `jbang foo/Hello.java` or `jbang info tools foo/Hello.java`
4. See `[jbang] [ERROR] //MODULE line has wrong format, should be '//MODULE [identifier]'` error

**Expected behavior**
Since `//MODULE` defaults to an automatic module name, JBang should be more robust and trim the `//MODULE` directive  line before causing an error.

**JBang version**
0.105.2

fbricon:
Fixes #1601

### DIFF
--- a/src/main/java/dev/jbang/source/TagReader.java
+++ b/src/main/java/dev/jbang/source/TagReader.java
@@ -32,9 +32,6 @@ public abstract class TagReader {
 	protected final String contents;
 	protected final Function<String, String> replaceProperties;
 
-	// Cached values
-	private List<String> lines;
-
 	private static final String REPOS_COMMENT_PREFIX = "REPOS ";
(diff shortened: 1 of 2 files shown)


maxandersen:
thanks!

PR #1603, opened 2023-03-28
feat: support use of ChatGPT in init

maxandersen:
Adds a new feature to init where `gptcontent` will be initialized with response
from ChatGPT to use in templates.

It requires presence of OPENAI_API_KEY environment variable and use of `--preview` flag.

Allowing you to do something like this:

```bash
export OPENAI_API_KEY=sk-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
jbang --preview init -t gpt almiray.java "write me a hello world using StringUtils from commons-lang3"
```

and it will generate:

```java
//usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS org.apache.commons:commons-lang3:3.10

import org.apache.commons.lang3.StringUtils;

public class almiray {
        System.out.println(StringUtils.capitalize("hello world!"));
    }
}
```

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -85,7 +85,7 @@ public class Init extends BaseCommand {
 		if (Util.isPreview()) {
 			Util.infoMsg("JBangGPT Preview activated");
 			Util.warnMsg(
-					"The result can vary greatly. Sometimes it works - others its just for inspiration or a good laugh.");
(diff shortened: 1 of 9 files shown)


maxandersen:
it adds no extra dependencies so for now considering it a easteregg in time for April 1st :)

maxandersen:
tests failure due to gist rate limit. merged.

PR #1604, opened 2023-03-31
refactor: Put MIMA in action

cstamas:
Re-use MIMA in ArtifactResolver, and properly handle repoSys shutdown.

About MIMA https://github.com/maveniverse/mima


maxandersen:
related to https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/maven.203.2E9.20.28and.20past.29.20compatability/near/345682355

quintesse:
Personally I'd be happy to use something made by people who actually understand Maven, because what we did with JBang was just cobble different things from different sources and different versions together, without understanding things fully, and just pray it works.

cstamas:
Also, am currently playing on "two fronts", since MIMA 1.0.1 release added some new stuff (like https://github.com/maveniverse/mima/commit/5719b7b6f6843c60a5cc3b4ff82d6bff76c8a82c) so my goal is to make MIMA provide "fully configured" miniMaven, while JBang only have to care about it's own specific overrides. (like JBANG_REPO is)...

quintesse:
> Settings.getLocalMavenRepo() never returns null, it's confidence is infinite

Well exactly because it returns a "factory default" it never returns `null`, that should be correct, right?

cstamas:
> Well exactly because it returns a "factory default" it never returns `null`, that should be correct, right?

No, that is not correct. User may have settings.xml in default place (`~/.m2/settings.xml`) that says that Maven local repo is actually at `/foo/bar/repo`.

cstamas:
Just FTR I pushed this change to different branch as am getting a bit lost, or am just digging too deep:
https://github.com/cstamas/jbang/compare/mima...cstamas:jbang:mima-step-1?expand=1

Goals:
* manage rootContext single instance (create when needed, close it when done, TBD yet)
* the context is "customized" per invocation, but am unsure can params change (ie. logging or offline), they seems mutable but per JBang invocation they seem unchanged?
* this makes resolver _reuse_ session. caches, and whole component graph, so it should be faster to resolve, especially bigger trees

quintesse:
> No, that is not correct. User may have settings.xml in default place (~/.m2/settings.xml) that says that Maven local repo is actually at /foo/bar/repo.

Sure, but that would still be a result, right? I mean the settings.xml can't "unset" the local repo, or can it? I do understand that the implementation is wrong for the reasons you mention, I'm just wondering if it not returning `null` is an issue or not.

quintesse:
> the context is "customized" per invocation, but am unsure can params change (ie. logging or offline), they seems mutable but per JBang invocation they seem unchanged?

Indeed, right now JBang doesn't change those parameters during an invocation, but I wrote that code to be flexible in that aspect "just in case".

src/main/java/dev/jbang/source/Project.java
```diff
@@ -26,6 +28,10 @@
 public class Project {
 	@Nonnull
 	private final ResourceRef resourceRef;
+
+	@Nonnull
```
quintesse, line 32: We'll have to come up with a different solution for this because Project is a model object and cannot contain anything that is not directly related top project management.

quintesse: I haven't looked at it in-depth yet, but I'm guessing this would need to be reversed, perhaps have some kind of factory that given a Project can return a RootContext.

cstamas:
FYI, will pick up this once resolver 1.9.10 and maven 3.9.2 are out.

quintesse:
> FYI, will pick up this once resolver 1.9.10 and maven 3.9.2 are out.

This was on my TODO-SOON list (like one of these days), are you saying it would be better to wait for those new versions?

cstamas:
Yes, i'd wait, we are not in hurry :smile: 
Latest MIMA released (1.1.0) uses Resolver 1.9.8 that suffers from eventing bug (IMHO it would not affect use case in jbang), while resolver release 1.9.9 vote was canceled today by me, as another bug was found. All this is fixed, and will release latest MIMA, so MIMA, Maven 3.9.2 and Maven 4-alpha-7 will be aligned re resolver versions.

quintesse:
@cstamas a question: what is the object that we should try to cache/not create too often and that we have to make sure to close?
I would assume it was the `Context`, but in the latest commit I see that an additional context is created using `customize()` and that one isn't being closed. Is that because it depends on the "parent" and it's enough to only close that one?

I'm just trying to figure out where exactly to fit the creation of the `Context` into our code. We have two situations where we do lookups and they use slightly different configurations and have different lifecycles.

Knowing how best to create/cache will help me select the best way forward

cstamas:
Yes, ideall a `Context` should be used in try-with-resource, nested if you "derive"/customize context. The topmost "root" context should be kept open as long as app/jvm runs.Personally I'd make ArtifactResolver closeable, and close of it would clode the (derived/customized) context as well.

cstamas:
Tomorrow vote for resolver 1.9.10 ends, will do a mima release and pick this up. 

If you want to toy with it, look at latest "static" addition (I think MIMA version I use here did not yet have it), it is equivalent of deprecated ServiceLocator runtime (no guice/sisu involved, much lighter) https://github.com/maveniverse/mima/tree/main#runtimes

cstamas:
It will work if you make it as before "A bit bigger commit". Actually, I'd revert that commit fully, as I went to completely wrong direction. So for start, doing in as it was before it would be ok (and using static uber  runtime) and see the startup penalty.... and later we could improve by refactoring?

cstamas:
Sisu is bad candidate for "minification", so i pushed:
* change from sisu to static runtime
* updated version of mima (as version used in this PR had no static runtime yet)

cstamas:
Locally 7 tests fail, but mostly due "unexpected" (junit tmp) location of local repo/resolved file, while otherwise assertions looks they should pass...

quintesse:
Ok, I'll take a look at those tests. Thanks!

build.gradle
```diff
@@ -189,7 +182,11 @@ shadowJar {
 		//exclude(dependency('org.slf4j:slf4j-api:.*'))
 		exclude(dependency('org.slf4j:slf4j-nop:.*'))
 		exclude(dependency('org.jboss.logging:jboss-logging:.*'))
+		exclude(dependency('eu.maveniverse.maven.mima:.*'))
+		exclude(dependency('eu.maveniverse.maven.mima.runtime:.*'))
+		exclude(dependency('org.apache.maven.resolver:.*'))
```
cstamas, line 187: These lines now with static-uber runtime could be removed, as minimize removed almost all of resolver (due implicit uses via JSR330 that minimize was unaware of).

build.gradle
```diff
@@ -189,7 +182,11 @@ shadowJar {
...
 	}
+	append 'META-INF/sisu/javax.inject.Named'
```
cstamas, line 189: Also unneded: already merged in static-uber runtime (unless we bring in some other Sisu stuff) -- is safe to be left as is

build.gradle
cstamas: Hm, given that mima.runtime goes in via ServiceLocator, maybe not.... (unsure does minimize plugin "knows" about ServiceLocator pattern, as it does not know about Sisu index either).

quintesse:
> Locally 7 tests fail, but mostly due "unexpected" (junit tmp) location of local repo/resolved file, while otherwise assertions looks they should pass...

@cstamas it seems this is because MIMA isn't honoring the `maven.repo.local` system property

quintesse:
We could simply add that to `Settings.getJBangLocalMavenRepoOverride()` of course, I'm just mentioning this in case this is something that should be handled by MIMA or not...

cstamas:
yup, not honoring `maven.repo.local` (or honoring it ONLY as a user property) was a bug, doing 1.1.2 release in few (as it turned out there were more system prop and env related issues): https://github.com/maveniverse/mima/commit/46651eb7ebab132d2428abe0b5386e9bb283aabd
1.1.2 sync pending to central (20ish minutes), and then `maven.repo.local` should be picked up as well

cstamas:
1.1.2 synced to central

w/ 1.1.2 my local test results:
```
BUILD SUCCESSFUL in 2m 54s
26 actionable tasks: 14 executed, 12 up-to-date
[cstamas@urnebes jbang (mima *)]$ 
```

pushing changes...

quintesse:
Looks great! 👍

cstamas:
ITs show there are still problems due minimize:
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/slf4j/spi/LoggerFactoryBinder
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:800)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:698)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:621)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:579)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at org.slf4j.LoggerFactory.bind(LoggerFactory.java:150)
	at org.slf4j.LoggerFactory.performInitialization(LoggerFactory.java:124)
	at org.slf4j.LoggerFactory.getILoggerFactory(LoggerFactory.java:417)
	at org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:362)
	at org.slf4j.LoggerFactory.getLogger(LoggerFactory.java:388)
	at eu.maveniverse.maven.mima.context.Runtimes.<init>(Runtimes.java:16)
	at eu.maveniverse.maven.mima.context.Runtimes.<clinit>(Runtimes.java:14)
	at dev.jbang.dependencies.ArtifactResolver.<init>(ArtifactResolver.java:148)
	at dev.jbang.dependencies.ArtifactResolver.<init>(ArtifactResolver.java:47)
	at dev.jbang.dependencies.ArtifactResolver$Builder.build(ArtifactResolver.java:117)
	at dev.jbang.dependencies.DependencyUtil.resolveDependencies(DependencyUtil.java:103)
	at dev.jbang.dependencies.DependencyResolver.resolve(DependencyResolver.java:80)
	at dev.jbang.source.Project.resolveClassPath(Project.java:235)
	at dev.jbang.source.buildsteps.CompileBuildStep.compile(CompileBuildStep.java:74)
	at dev.jbang.source.buildsteps.CompileBuildStep.build(CompileBuildStep.java:58)
	at dev.jbang.source.buildsteps.CompileBuildStep.build(CompileBuildStep.java:41)
	at dev.jbang.source.AppBuilder.build(AppBuilder.java:98)
	at dev.jbang.source.AppBuilder.build(AppBuilder.java:22)
	at dev.jbang.cli.Run.doCall(Run.java:95)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:145)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:21)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at dev.jbang.cli.JBang$3.handle(JBang.java:147)
	at dev.jbang.cli.JBang$3.handle(JBang.java:142)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
Caused by: java.lang.ClassNotFoundException: org.slf4j.spi.LoggerFactoryBinder
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 40 more
```

As I see, there is (commented out) entry for slf4j-api (as SPI is there), but when I locally uncomment that line, SPI is still missing, so unsure what am I missing?

quintesse:
Confirmed, seeing the same issue

cstamas:
Am completely lost... tried clean again w/ minimize for sl4j and SPI package is still stripped off... So it's either me (I hope) or gradle minimize is having some issue? I cannot figure how to NOT have it removed :(

cstamas:
As a desperate move completely disabled minimize, just to be able to run all ITs:
https://gist.github.com/cstamas/5cd98da21384f92449414ced744a57cb

cstamas:
The one failed feature:
```
21:01:33.224 assertion failed: path: $, actual: '[jbang] Resolving dependencies...
[jbang] Dependencies resolved
[jbang] Building jar...
', expected: '[jbang] Resolving dependencies...
[jbang]    log4j:log4j:1.2.17
[jbang] Dependencies resolved
[jbang] Building jar...
', reason: not equal
dependencies.feature:6 - path: $, actual: '[jbang] Resolving dependencies...
[jbang] Dependencies resolved
[jbang] Building jar...
', expected: '[jbang] Resolving dependencies...
[jbang]    log4j:log4j:1.2.17
[jbang] Dependencies resolved
[jbang] Building jar...
', reason: not equal
```

But this seems like a sideeffect? As dependencies were already resolved? unsure...

cstamas:
Manually redone what feature test did: https://gist.github.com/cstamas/912c8f7e70a0451fc174e1aaf6a6e8d5

To me it seems JBang output changed? As it does not emit artifact anymore, while in --verbose it does log it as resolved (but a bit differently)

quintesse:
I didn't change the output, no. Could it be that the events that are emitted are slightly different when using MIMA?
I'll check on monday to see what the difference is between running the old version and with MIMA.

cstamas:
It should not: MIMA just construct Resolver, it does not "interfere" once created it.... but resolver changed from 1.8 to 1.9... but again, am unaware of ANY eventing change...

quintesse:
Btw, I think it's best not to run the testsuite.sh script. It messes with your environment. I normally run `jbang karate.java *.feature`.

cstamas:
It sets $PATH but I usually keep changing (closing, reopening) Terminals anyway... and as for local repo, "regular nuking of local repo is part of good maven hygiene" (and keep a good MRM running close to you) :smile:

quintesse:
So what I see is that the dependencies are only logged if they're a) being downloaded or b) --verbose is on. So possibly if you're running that test multiple times and it doesn't set up a different Maven repo it might not give correct output.

cstamas:
Will doublecheck eventing/logging tomorrow, am off for now

cstamas:
I think figured logging: the private final loggingEnabled was NOT YET set while the setupSessionLogging queried it. Fixed by removing this member, and instead invoke setupSessionLogging only when needed

cstamas:
And it fixed it, thefailing karate test now does this:
```
[cstamas@blondie itests (mima *%)]$ jbang classpath_log.java 
[jbang]    log4j:log4j:1.2.17
[jbang] Dependencies resolved
[jbang] Building jar...
[cstamas@blondie itests (mima *%)]$ 
```

cstamas:
So, we are back to minimize, as to make all this work I had to completely disable minimize... so why is org.slf4j.spi package completely removed, even if we ask the plugin not to?

cstamas:
As I see minimize issues https://github.com/johnrengelman/shadow/issues/638 or https://github.com/johnrengelman/shadow/issues/744 and some others, this may be some "peculiarity" of the minimize plugin. Hence, am pushing my local change to disable it instead.

cstamas:
So with these changes (w/ minimization disabled) all have no everything pass OK
```
[cstamas@blondie itests (mima)]$ ./test_suite.sh 
Using /tmp/test_suite.sh.Wiq as scratch dir
Cleaning JBANG_CACHE
Testing with /home/cstamas/Worx/jbangdev/jbang/build/install/jbang/bin/jbang
...........[jbang] Resolving dependencies...
[jbang]    log4j:log4j:1.2.17
[jbang] Dependencies resolved
[jbang] Building jar...
.........
all 20 jbang tests passed in 17.979s.
Deleted scratch dir /tmp/test_suite.sh.Wiq
[cstamas@blondie itests (mima)]$ ./itests.sh 
...
12:51:14.968 [1683370274968]  DEBUG com.intuit.karate - command: [sh, -c, jbang wrapper install]
12:51:15.312 [1683370274968]  DEBUG com.intuit.karate.shell.Command - command complete, exit code: 0 - [sh, -c, jbang wrapper install]
12:51:15.312 [1683370274968]  DEBUG com.intuit.karate.shell.Command - console readers complete
12:51:15.316 [pool-1-thread-1]  INFO  com.intuit.karate.Runner - <<pass>> feature 22 of 22: wrapper.feature
---------------------------------------------------------
feature: wrapper.feature
report: ../build/karate/surefire-reports/wrapper.json
scenarios:  5 | passed:  5 | failed:  0 | time: 3.2242
---------------------------------------------------------

HTML report: (paste into browser to view) | Karate version: 2.0.0
file:/home/cstamas/Worx/jbangdev/jbang/itests/../build/karate/surefire-reports/karate-summary.html
===================================================================

Karate version: 2.0.0
======================================================
elapsed: 100.12 | threads:    1 | thread time: 98.55 
features:    22 | ignored:    0 | efficiency: 0.98
scenarios:   74 | passed:    74 | failed: 0
======================================================

[cstamas@blondie itests (mima *%)]$
```

quintesse (approved):
Great job @cstamas . I've tested it locally and everything seems to work fine. I also did some very unscientific performance tests and I basically didn't see any difference between the current jbang version and the one using MIMA, so that's good!
Let's see what @maxandersen thinks. Ok to merge from me.

cstamas:
Thanks @quintesse 

Re performance, standalone-static wires things directly up, so it should be "on par" with SL of not faster (due some "poor man DI logic in there"). Where perf may count is lack of minimization in this PR, but here am totally lost why gradle minimize breaks things.

In my initial attempt with standalone-sisu I understood (implicit relations via JSR330 that minimize plugin is not aware of), but in case of standalone-static and really lost...

quintesse:
Ah yes, the minimize issue might be something that could prevent the merge, now that I think of it. It makes the size of the .jar go up from ~5MB to ~8MB, which is not a lot of course, but we might need to take a look into that issue before deciding to merge or not. Wdyt @maxandersen ?

cstamas:
As this was my first "mixed" use of gradle and maven, have to share my experience:
* as expected, I wanted local round trips, working on MIMA, rebuild and install it using mvn into local repo as snapshot
* wanted gradle pick it up (so modified locally the build to factor in Maven local repo), rinse repeat
* but alas, it seems gradle picked it up only the FIRST TIME and from there on, it did not detect any changes in my local repo, missed to pick up reinstalled snaphots
* I spent nice amount of time in WTF state, until I figured out, that jbang distro still contains unmodifiedMIMA classes (after one roundtrip: change mima, mvn install, gradle build)
* as I did not dig into gradle cache layout, "solved" it by nuking gradle env, but alas, that made everything painfully slow (I think I never drank so much coffee in my life)
* when I hit this minimize issue locally, I simply gave up :smile: 

Figured Gradle is hazardous for my health :smile:

quintesse:
Hmm no idea why the test_suite keeps failing. It Works On My Machine (tm).

quintesse:
> Figured Gradle is hazardous for my health

Hahaha, yeah, I think Max starting using Gradle for JBng as a test, but AFAIK he's not really happy with it. I never liked Maven very much and had actually hoped I would like Gradle more, but after using it for some time I must say I'm not really enjoying it either :slightly_smiling_face:

cstamas:
> Hahaha, yeah, I think Max starting using Gradle for JBang as a test, but AFAIK he's not really happy with it. I never liked Maven very much and had actually hoped I would like Gradle more, but after using it for some time I must say I'm not really enjoying it either slightly_smiling_face

Not to brag, but this above is what I keep hearing from people "coming back to Maven" from Gradle :smile: 

So, what about a PR  for that?

quintesse:
> So, what about a PR for that?

That's something you'd have to ask @maxandersen , it's his baby after all :slightly_smiling_face:

cstamas:
re size: for some reason I cannot make gradle minimize NOT remove the `org.slf4j.spi` package (yes, dependency slf4j-api where these are is added to minimize exclusions, see commented out minimize config, but to no avail)

re "custom builds" will check how releases are made for jbang (did not dig into it). Is it documented maybe?

aalmiray:
@cstamas JReleaser performs its magic regardless of the interface used to configure and invoke it. This means everything that can be done on the CLI can also be done via `jreleaser-maven-plugin`.

cstamas:
Could someone please approve the CI? --verbose added to at least see what is happening

cstamas:
Um, so something is off: was my commit applied to ITs at all? JBang did run without `--verbose` but https://github.com/jbangdev/jbang/pull/1604/commits/204ad423a268c7ce0c1898bbd1257724f7d2cd94

So am confused here

quintesse:
Does the `assert` perhaps redirect any error output? Because all `--verbose` output in JBang goes to stderr.

cstamas:
Right, locally w/ --verbose I have:
```
[cstamas@urnebes itests (mima)]$ ./test_suite.sh 
Using /tmp/test_suite.sh.Jo5 as scratch dir
Cleaning JBANG_CACHE
Testing with /home/cstamas/Worx/jbangdev/jbang/build/install/jbang/bin/jbang
...........[jbang] Resolving dependencies...
[jbang]    log4j:log4j:1.2.17
[jbang] Dependencies resolved
[jbang] Building jar...
.........
all 20 jbang tests passed in 25.469s.
Deleted scratch dir /tmp/test_suite.sh.Jo5
[cstamas@urnebes itests (mima)]$
```

So output is really same... and as I see, output of err goes to /dev/null?

cstamas:
Re "cannot create context from scratch" interesting (and probably related) issue:
https://github.com/moditect/layrry/pull/218

Seems MIMA detects Maven-env laid out by GH Action... And this is where it fails, as it cannot decrypt encrypted settings.xml for `github`. 

Would really like to see err output from jbang...

cstamas:
Update to 2.1.0 https://github.com/maveniverse/mima/releases/tag/release-2.1.0

Several issues fixed (among other the "eager" decrypt-if-not-die behaviour).

cstamas:
@quintesse @maxandersen can someone please approve the workflow?

cstamas:
bah:
```
java.io.IOException: Server returned HTTP response code: 500 for URL: https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7%2B7/OpenJDK17U-jdk_x64_linux_hotspot_17.0.7_7.tar.gz
```

Seems GH coughed?

cstamas:
bah:
```
java.io.IOException: Server returned HTTP response code: 500 for URL: https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7%2B7/OpenJDK17U-jdk_x64_linux_hotspot_17.0.7_7.tar.gz
```

Seems GH coughed?

cstamas:
GH seems still having issues, the one failed job is "canceled"

quintesse:
Finally it passed 👍

cstamas:
Cool, thanks! 
But, there is still the minimize issue to look into...

cstamas:
Help with taming minimize welcome!

So far tried:
* uncomment config as is (undo related changes in this PR): then it stripped off full SLF4J (only org.slf4j.impl package was present in jbang.jar), moreover it stripped off the service loader as well (the one it is supposed to merge), so ended up with jbang.jar (4.5MB) as:
```
[cstamas@urnebes bin (mima *)]$ ./jbang --verbose foo.java
[jbang] [0:321] jbang version 0.106.3.15
[jbang] [0:328] Resolving resource ref: foo.java
[jbang] [0:332] Resolved resource ref as: foo.java (cached as: /home/cstamas/Worx/jbangdev/jbang/build/install/jbang/bin/foo.java)
[jbang] [0:362] Build required as /home/cstamas/.jbang/cache/jars/foo.java.1d6a09febb57800b207cc02d3d6ec4a2347391de63bcfbc6e33f2e501aeec80d/foo.jar not readable or not found.
[jbang] [0:366] Looking for JDK: 0
[jbang] [0:369] Using JDK: 11 (11.0.19, current, /home/cstamas/.sdkman/candidates/java/11.0.19-tem)
[jbang] [0:370] Resolving artifact(s): junit:junit:4.13.2
[jbang] [0:372] Repositories: mavencentral=https://repo1.maven.org/maven2/
[jbang] [0:400] Resolving dependencies...
[jbang] [0:411] Deleting folder /home/cstamas/.jbang/cache/jars/foo.java.1d6a09febb57800b207cc02d3d6ec4a2347391de63bcfbc6e33f2e501aeec80d/classes
[jbang] [0:414] Deleting folder /home/cstamas/.jbang/cache/jars/foo.java.1d6a09febb57800b207cc02d3d6ec4a2347391de63bcfbc6e33f2e501aeec80d/generated
Exception in thread "main" java.util.ServiceConfigurationError: eu.maveniverse.maven.mima.context.Runtime: Provider eu.maveniverse.maven.mima.runtime.standalonestatic.StandaloneStaticRuntime not found
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:589)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(ServiceLoader.java:1212)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1221)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1265)
	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1300)
	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1385)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:132)
	at eu.maveniverse.maven.mima.context.Runtimes.getRuntime(Runtimes.java:33)
	at dev.jbang.dependencies.ArtifactResolver.<init>(ArtifactResolver.java:148)
	at dev.jbang.dependencies.ArtifactResolver.<init>(ArtifactResolver.java:47)
	at dev.jbang.dependencies.ArtifactResolver$Builder.build(ArtifactResolver.java:116)
	at dev.jbang.dependencies.DependencyUtil.resolveDependencies(DependencyUtil.java:103)
	at dev.jbang.dependencies.DependencyResolver.resolve(DependencyResolver.java:80)
	at dev.jbang.source.Project.resolveClassPath(Project.java:235)
	at dev.jbang.source.buildsteps.CompileBuildStep.compile(CompileBuildStep.java:74)
	at dev.jbang.source.buildsteps.CompileBuildStep.build(CompileBuildStep.java:58)
	at dev.jbang.source.buildsteps.CompileBuildStep.build(CompileBuildStep.java:41)
	at dev.jbang.source.AppBuilder.build(AppBuilder.java:98)
	at dev.jbang.source.AppBuilder.build(AppBuilder.java:22)
	at dev.jbang.cli.Run.doCall(Run.java:95)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:145)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:21)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at dev.jbang.cli.JBang$3.handle(JBang.java:148)
	at dev.jbang.cli.JBang$3.handle(JBang.java:143)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
```

* then tried to add two MIMA deps to exclusion, then it "minimized" required Maven classes (that ARE present in minimize-excluded standalone-static-uber), jbang.jar (6.6MB) was like:
```
[cstamas@urnebes bin (mima *)]$ ./jbang init foo.java
[jbang] File initialized. You can now run it with 'jbang foo.java' or edit it using 'jbang edit --open=[editor] foo.java' where [editor] is your editor or IDE, e.g. 'codium'. If your IDE supports JBang, you can edit the directory instead: 'jbang edit . 'foo.java. See https://jbang.dev/ide
[cstamas@urnebes bin (mima *)]$ edit foo.java 
[cstamas@urnebes bin (mima *)]$ ./jbang foo.java
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/maven/artifact/versioning/InvalidVersionSpecificationException
	at org.apache.maven.model.building.DefaultModelBuilderFactory.newInstance(DefaultModelBuilderFactory.java:202)
	at eu.maveniverse.maven.mima.runtime.standalonestatic.RepositorySystemSupplier.getModelBuilder(RepositorySystemSupplier.java:515)
	at eu.maveniverse.maven.mima.runtime.standalonestatic.RepositorySystemSupplier.get(RepositorySystemSupplier.java:623)
	at eu.maveniverse.maven.mima.runtime.standalonestatic.StandaloneStaticRuntime.createRepositorySystem(StandaloneStaticRuntime.java:37)
	at eu.maveniverse.maven.mima.runtime.standalonestatic.StandaloneStaticRuntime.create(StandaloneStaticRuntime.java:29)
	at dev.jbang.dependencies.ArtifactResolver.<init>(ArtifactResolver.java:148)
	at dev.jbang.dependencies.ArtifactResolver.<init>(ArtifactResolver.java:47)
	at dev.jbang.dependencies.ArtifactResolver$Builder.build(ArtifactResolver.java:116)
	at dev.jbang.dependencies.DependencyUtil.resolveDependencies(DependencyUtil.java:103)
	at dev.jbang.dependencies.DependencyResolver.resolve(DependencyResolver.java:80)
	at dev.jbang.source.Project.resolveClassPath(Project.java:235)
	at dev.jbang.source.buildsteps.CompileBuildStep.compile(CompileBuildStep.java:74)
	at dev.jbang.source.buildsteps.CompileBuildStep.build(CompileBuildStep.java:58)
	at dev.jbang.source.buildsteps.CompileBuildStep.build(CompileBuildStep.java:41)
	at dev.jbang.source.AppBuilder.build(AppBuilder.java:98)
	at dev.jbang.source.AppBuilder.build(AppBuilder.java:22)
	at dev.jbang.cli.Run.doCall(Run.java:95)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:145)
	at dev.jbang.cli.BaseCommand.call(BaseCommand.java:21)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at dev.jbang.cli.JBang$3.handle(JBang.java:148)
	at dev.jbang.cli.JBang$3.handle(JBang.java:143)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
Caused by: java.lang.ClassNotFoundException: org.apache.maven.artifact.versioning.InvalidVersionSpecificationException
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 29 more=
```

maxandersen:
I'll try take a look.

cstamas:
No luck so far. What I tried:
* update shade plugin to latest 8.1.1
* updated to mima 2.2.0 (has this PR https://github.com/maveniverse/mima/commit/e48f8615bf7416485ebdd77afdab875981e1cdcb) to allow integrator to provide "binary compatible" slf4j.
* making runtime-static-uber scope to "implementation", assumed minimize is confused if one runtime class refs another runtime class

To no avail, at least the version ex is stripped, making whole resolver defunct.

cstamas:
FYI MIMA 2.3.0 is out with bunch of new features https://github.com/maveniverse/mima/releases/tag/release-2.3.0

quintesse:
Thanks @cstamas for keeping us up-to-date. We talked about this today again in our weekly meeting and have promised ourselves to look at this again ASAP because we really want to release JBang with MIMA :-)

maxandersen:
sorry for delay on this - i've looked at it a bit and its weird. --debug reveals it removes all the classes as its deemed the classes are unused - even when explicitly including the dependency.

documentation around shadowjar is quite limited when it comes how to control it...

maxandersen:
okey - pushed a one line fix after i realized you want to exclude not include in minimization :) 

The resulting jar is still slightly bigger 6.1 instead of 5.6 but better than 9MB.

any idea if there are things we can strip out?

cstamas:
Probably are, but am unsure what. I'd leave it as is for now and we can work out some improvements later down on the road

maxandersen:
Seems missing aether still :/

cstamas:
Is minimizing disabled? What strips off aether to make it missing? :thinking:

maxandersen:
No. Trying to have 
Minimized active. I haven't figured out why it thinks aether is unused as it's quite obviously is :)

cstamas:
Back from vacation, Maven Resolver 1.9.14, Maven Indexer 7.0.4 and Maven 3.9.4 done (last is on vote from today), will try to dig more into this one as well.... thanks for work and support so far! (also a new mima release with new resolver + maven bits will happen in the mean time, once maven is out)....

cstamas:
Changes:
* updated MIMA to latest (mvn 3.9.4 + resolver 1.9.15)
* using non-uber mima for now (as seems gradle had issues with it)
* completed classpath with slf4j bits (as mime leaves this to user): slf4j-api and jcl-over-slf4j (needed by httpClient)
* moved off deprecated MIMA methods

Minimize still misbehaves, strips off complete resolver impl package and mima static provider...

cstamas:
Seems latest commit fixed minimize, for me jbang.jar is now 7.3 MB

quintesse:
> OTOH, now after several builds I have UT errors from GH gist

Yeah, unfortunately that's "normal". We haven't been able to figure out yet why sometimes GH seems to apply a very low rate limit. Almost as if it is running the actions as an anonymous user.

A bit unfortunate that the JAR has grown even larger than before though, (From 6.9MB last time to 7.3MB)

cstamas:
Yeah, minimize doesn't work as expected, so i just excluded all. Still unsure why, but switched from uber to "normal" deps with high hopes, but... Meh

maxandersen:
if this also fix proxy issues lets try and get this in. shadow shavings be damned.

@cstamas can you resolve the conflict?

I'll try test the proxy setups and if all good lets make it happen.

cstamas:
Sure, but first we need to see Grenaa river estuary, given DMI is shouting about some warnings... will do later this evening

maxandersen:
Stay safe. This one is bigger than any flooding I've seen in DK before

cstamas:
The family was blown away :) (literally...). Ozzy (the labrador) did not dare to leave the car.

Resolver conflict, and pushed changes: 
* updated to latest MIMA 2.3.5 (Maven 3.9.5 + Resolver 1.9.16)
* removed maven-embedded runtime (unsure why was it here in the first place, JBang will not run inside of maven-plugin, will it?)
* removed append for sisu index

maxandersen:
We have jbang maven plugin but I believe it runs as efterlader proces there. 

Would it be problematic if we wanted to rub from inside maven plugin at some point?

maxandersen:
And good to hear family safe. My flagpole snapped 30 km away from where is worst and it have not even peaked yet

cstamas:
> Would it be problematic if we wanted to rub from inside maven plugin at some point?

Nope, just return maven-embedded runtime.

maxandersen:
> > Would it be problematic if we wanted to rub from inside maven plugin at some point?
> 
> 
> 
> Nope, just return maven-embedded runtime.

So for now we remove it to counter the fat gained by shadowjar not working :)

maxandersen:
so first proxy test not looking as there is improvement.

setting http proxy like documented in misc/proxy and use tinyprox jbang fetches catalog etc. but jbang is not using the proxy when resolving.

cstamas:
Will take a peek, it may be MIMA issue (does not have proxy tests either). 

But my current impediment is that my Beast (desktop) damaged during transport, so am stuck on my fallback laptop...

maxandersen:
I'll take look later. For now it doesnt look like mima is worse or better than what is in main.

cstamas:
MIMA opens path toward Resolver 2, last night did some testing (re http2 transport) https://gist.github.com/cstamas/30b817042f8c1210673eebfb593558cd

cstamas:
I just realized how the "proxy issue" is problematic here: resolver will NOT obey any properties (as it uses it's own config), so users can configure proxy in their settings.xml User Maven settings file. This will make Resolver use proxy. But, jbang is unaware of that (so jbang will not, and other way around).

So, unsure how JBang is configured to use proxy, BUT, added support for "settings mixins", and created small CLI example how "things should happen in case like JBang":
https://github.com/maveniverse/mima/pull/29/files#diff-c3024e5cb1661064967eb2988525536f54f821ce0c4e998dd17289b2e3c736eaR175-R189

This means, that IF (am unsure here) JBang has "configure proxy" capability, and is aware proxy is must, it can just create this mixin for MIMA and it will apply it, just like it happens in case of that CLI example. Essentially, jbang "drives".

Of course, question still remains what if user settings.xml have proxy defined, but JBang is invoked without one?

cstamas:
So, with "settings mixin" JBang (or any integrating app) is able to force use of HTTP proxies (assuming the app is configured to use proxy), and I think this is good.

What should happen the other way around? JBang is invoked, and user has settings.xml with HTTP proxy configured?

maxandersen:
Cool. 

Generelly it would be most local wins. So:

No props, no settings.xml: no proxy

Local props, no settings xml: use local props for proxy

Local props, settings.xml: use local props overrides what is in settings.xml

No props, settings.xml: use settings.xml

I guess what would be nice is jbang downloading direct http would honor the same. Is that what your code example does ? (Don't have laptop with me so can't test yet)

cstamas:
I don't quite get your intent, so let me draft it more detailed:

| jbang proxy | settings.xml proxy | result | How |
|-----|-----|-----|------|
| not configured | not configured | not used |  |
| configured | not configured | used (jbang + resolver use jbang configured proxy) | mixin |
| not configured | configured | used (in resolver only?) | maybe somehow pick up active proxy? (as there can be only one... proxy) |
| configured | configured | used (jbang proxy overrides settings.xml proxy?) | mixin is no good here, maybe identify the active proxy in settings.xml, disable it **and** inject jbang (using mixin)? |

So maybe best would be for MIME to provide some "loophole" to inspect effective settings.xml? And based on seen, do some action? Or just expose settings.xml for possible _alteration_ before applied? This would work in "standalone" runtime, but what in case of embedded case (ie. Maven Plugin? or then proxy settings of jbang does not apply?)

The mixing really solved the "push jbang config into resolver" case, so will try to add two-way door here...

cstamas:
If this above looks ok for you, will merge and release MIMA 2.4.0 and update it here.

maxandersen:
> If this above looks ok for you, will merge and release MIMA 2.4.0 and update it here.

yes sounds good.

One clarification - would the proxy mixin need to be fully configured (all hosts and ports) or can it just mutate the existing settings? (if not feasible not a problem if jbang can read the current settings it can merge)

cstamas:
The settings mixin literally can be any "incomplete" bit of settings.xml that will maven settings merger merge before using settings.

For example, to add a proxy to settings, all you need is mixin like this:
```
<settings>
  <proxies>
    <proxy>
      <id>myproxy</id>
      <active>true</active>
      <protocol>http</protocol>
      <host>proxy.somewhere.com</host>
      <port>8080</port>
      <username>proxyuser</username>
      <password>somepassword</password>
      <nonProxyHosts>*.google.com|ibiblio.org</nonProxyHosts>
    </proxy>
  </proxies>
</settings>
```

So in this manner, yes, the mixin needs to contain "fully configured" [Maven Settings Proxy](https://maven.apache.org/settings.html#proxies) as no interpolation or any other post-processing is applied to it.

And this is literally what MIMA CLI showcases:
https://github.com/maveniverse/mima/blob/main/cli/src/main/java/eu/maveniverse/maven/mima/cli/CommandSupport.java#L181-L195

When MIMA detects that mixin contains proxy, it will disable any other possibly active proxy in settings and merge the mixin. Hence, if your mixin contains active=false proxy definition, you end up with fully disabled Resolver proxy, but if you add real active=true proxy definition, you basically override Resolver proxy settings with mixin.

Maven settings.xml proxies are actually a list (of any size), it can contain multiple proxy definitions, but only `0..1` are allowed to be active (in reality, first found active is used). It was envisioned, when this model was defined, to have multiple proxy definitions like "home", "office-hq", "office-grenaa" etc and user just "tick" the needed one.

Personally, I apply similar pattern, but by having _several_ settings.xml files and just symlinking them to `.m2/settings.xml` as needed, as I want to keep separate not only proxy (am not), but also mirrorsOf, local repository, profiles w/ repo definitions and many more (so I have `settings-apache.xml`, `settings-oss.xml`, `settings-$work.xml` and so on).

Due these above (list of proxies, proxy selection), Context now expose the `HTTPProxy` "in effect" that you can use to pick-up resolver config, but as it is exposed on Context, it raises a bit of "chicken and egg" problem (to inspect it, you need Context, but when you have Context, it is "too late"). One way to overcome that could be creating a "temp" context, inspect it, toss it away (close), and create new Context + mixin as you want.

Also, created issue https://github.com/maveniverse/mima/issues/39 for use case that I believe is very similar you mention, but there are some technical related issues to it: MIMA context module is "maven free" (it's classpath does not contain any Maven classes, only Resolver, it uses Resolver -api and -util classes only) and this is on purpose. Hence the mixin is represented as `Object` as well on ContextOverrides (but nothing stops integrating app, like JBang to have Maven models on classpath). So, "most direct" way to express this in current state of art would be to have something like `Object Function<Object>` that in reality should be `Settings Function<Settings>`, but again, context does not have Maven classes on classpath... will come up with something soon.

maxandersen:
Ok. So is this ready to go in and have those changes in? Just need to wire up
Proxy handling ?

cstamas:
Yes, I just updated to MIMA 2.4.0 (and one source change needed), did not change anything re proxy....

maxandersen:
whoops - things broke? 

```
DependencyResolverTest > testResolveDependenciesAltRepo(File) FAILED
    dev.jbang.cli.ExitException at DependencyResolverTest.java:126
        Caused by: org.eclipse.aether.resolution.DependencyResolutionException at DependencyResolverTest.java:126
            Caused by: org.eclipse.aether.collection.DependencyCollectionException at DependencyResolverTest.java:126
                Caused by: org.eclipse.aether.resolution.VersionRangeResolutionException at DependencyResolverTest.java:126

DependencyResolverTest > testResolveDependenciesWithAether() FAILED
    dev.jbang.cli.ExitException at DependencyResolverTest.java:115
        Caused by: org.eclipse.aether.resolution.DependencyResolutionException at DependencyResolverTest.java:115
            Caused by: org.eclipse.aether.collection.DependencyCollectionException at DependencyResolverTest.java:115
                Caused by: org.eclipse.aether.resolution.VersionRangeResolutionException at DependencyResolverTest.java:115

DependencyResolverTest > testResolveJavaFXWithAether() FAILED
    dev.jbang.cli.ExitException at DependencyResolverTest.java:108
        Caused by: org.eclipse.aether.resolution.DependencyResolutionException at DependencyResolverTest.java:108
            Caused by: org.eclipse.aether.resolution.ArtifactResolutionException at DependencyResolverTest.java:108
                Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException at DependencyResolverTest.java:108
```

pretty sure it passed earlier ?

maxandersen:
the dependencies it fails on in case @cstamas can spot if somehow special:

org.openjfx:javafx-base:jar:mac-aarch64:18.0.2
log4j:log4j:jar:[1.2,)

cstamas:
Pushed fix, locally passes OK: in 2.4 I changed the `.repositories()` semantics on builder (the ContextOverride), as before it was impossible to configure MIMA without remote repositories (and sometimes you do want that). Also, added the new "op" for handling explicitly provided list of remote repositories when considering remote repositories discovered from settings.xml as well (you usually want to prepend, to "override").

Depending on intent here (I just guessed), in case `partialRepos` is null, maybe more correct would be to NOT invoke ContextOverrides.Builder.repositories() method at all... (and rely on defaults)

cstamas:
Btw, resolver 2.0.0-alpha-2 just released, it introduces new HTTP/2 transport based on JDK 11 java.net.http.HttpClient (that is naturally dependency-less) and boosts around 20% in download perf when target (like Central) supports HTTP/2 (it does). Speed is pretty much same as "apache" transport when target is HTTP/1.1 (like some Maven Repo Manager....).

MIMA will soon start supporting resolver 2.0.0 (probably in parallel with 1.9..x line) and once JBang upgrades to MIMA, following resolver updates will be trivial (just update MIMA).

For JBang this may be interesting, as it drops requirement for ASF httpClient, jcl-over-slf4j...

maxandersen:
Can that be done in a way so when run on java 8 it will still fallback?

maxandersen:
btw. about: 

"Also, there is new "op" what to do with provided list,
when repositories discovered from settings.xml are considered
as well.
"

did we ever consider repos from settings.xml ? that wouldn't be what I expect...unless its purely for the case where no repos specified in jbang?

cstamas:
> Can that be done in a way so when run on java 8 it will still fallback?

Yes, mvn4 does exactly that:
* it it runs on java8, transport defaults to "apache" (as with mvn3)
* if it runs on java11+ transport defaults to "jdk" and is HTTP/2
* there is also new "jetty" transport, other alternative that knows HTTP/2 (and soon even HTTP/3)

The "jdk" transport is actually multi-release JAR that is "no op" on Java8:
https://github.com/apache/maven-resolver/tree/master/maven-resolver-transport-jdk-parent

cstamas:
> did we ever consider repos from settings.xml ? that wouldn't be what I expect...unless its purely for the case where no repos specified in jbang?

MIMA always did consider reposes from settings.xml, and I assumed this is what you want... IF that is not the case, OP.REPLACE needs to be set and then jbang "drives" (and reposes from settings.xml are neglected/replaced with those specifified in repositories).

MIMA mimics what Maven does in a way: reposes specified in contextOverrides are like POM specified RemoteReposes, and are merged (prepended) to those discovered in settings.xml. But, for flexibility, I added the "op" that can APPEND and REPLACE as well... if JBang does not want to care for settings reposes, it should use op REPLACE then...

maxandersen:
It's mainly for repos - default in jbang have been nothing spécified- we add central. If you specify something you control all the repos. 

I wonder if we should have a "magic" settings.xml repo name that lets user place it. Or a flag that says honor settings.xml...

cstamas:
I'd consider wrong to not observe settings.xml defined reposes, but I can understand it can be overwhelming and why would someone want to ignore those. So, yes, exposing this on jbang (honor or do not honor reposes in settings.xml) with sensible default (i guess as current intent: ignore) would make sense.

maxandersen:
Rationale to not honor it is reproducibility/consistency to ensure that a jbang script will run the same by default. 

Totally get that it would be nice to honor settings repositories - but that I feel should be optin not opt out.

maxandersen:
> > did we ever consider repos from settings.xml ? that wouldn't be what I expect...unless its purely for the case where no repos specified in jbang?
> 
> MIMA always did consider reposes from settings.xml, and I assumed this is what you want... IF that is not the case, OP.REPLACE needs to be set and then jbang "drives" (and reposes from settings.xml are neglected/replaced with those specifified in repositories).
> 
> MIMA mimics what Maven does in a way: reposes specified in contextOverrides are like POM specified RemoteReposes, and are merged (op.PREPEND, the default) to those discovered in settings.xml. But, for flexibility, I added the "op" that can op.APPEND and op.REPLACE as well... if JBang does not want to care for settings reposes, it should use op REPLACE then...

...i'm being blind - where is it you say we should set this in this PR to not have repos from settings.xml included?

cstamas:
Here you are:
https://github.com/jbangdev/jbang/pull/1604/commits/ff51b8a89aad6f0c8349377fb13ccd3665e3b523

(sadly spotless behaves very strange).
Make sure partialReposes ALWAYS have repo (at least central if user does not set any), and set Op to REPLACE (to ignore settings.xml discovered reposes).

itests/test_suite.sh
```diff
@@ -52,7 +52,7 @@ rm -rf ~/.jbang/cache
 echo Testing with `which jbang`
 
 ## init ##
-assert "jbang init $SCRATCH/test.java"
+assert "jbang --verbose init $SCRATCH/test.java"
```
maxandersen, line 55: why this change?

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -521,7 +521,7 @@ void testAliasWithRepo(@TempDir File output) throws IOException {
 				"  \"aliases\": {\n" +
 				"    \"aliaswithrepo\": {\n" +
 				"      \"script-ref\": \"dummygroup:dummyart:0.1\",\n" +
-				"      \"repositories\": [ \"http://dummyrepo\" ]\n" +
+				"      \"repositories\": [ \"https://dummyrepo\" ]\n" +
```
maxandersen, line 524: is http now allowed anymore or ?

maxandersen (approved):
just minor questions - all big important stuff looks resolved now!

maxandersen:
Lets do this - thank you!

@all-contributors please add @cstamas for code

quintesse:
Yay! 🎉

itests/test_suite.sh
cstamas: IMHO can be undone, I think that was done early in this PR, probably had some issue so needed debug logs for resolver?

src/test/java/dev/jbang/cli/TestRun.java
cstamas: what is allowed (only HTTPS or HTTP) or not is matter of setup/config. For example if you have something like this in your settings.xml (user or system): https://github.com/apache/maven/blob/master/apache-maven/src/assembly/maven/conf/settings.xml#L160-L166 then HTTP is blocked.

PR #1623, opened 2023-05-16
feat: add support for --enable-preview

### LINKED ISSUE #1622: enable preview should be more consistently available
trying out java 21 with jshell turns out harder than wanted:

```
JBANG_JDK_VENDOR=openjdk jbang --java 21 -i -R --enable-preview -C --enable-preview
jshell> STR."test"
|  Error:
|  string templates are a preview feature and are disabled by default.
|    (use --enable-preview to enable string templates)
|  STR."test"
|      ^
```

this fails as `--enable-preview` is not passed to jshell.

Given `--enable-preview` now is a common flag; should we have that as a first-class thing that will consistently enable-preview across jshell, javac, java ?

maxandersen:
Why:

 * --enable-preview is a common flag to enable preview features in Java 9 and up
    but is hard to use with jbang as it is not passed on to the java,javac
    and jshell command when running a script.
 * workaround thus far been to pass -C=--enable-preview and -R=--enable-preview
    but that does not work for jshell

 * introducing `--enable-preview` flag to jbang which will pass it on to
    java, javac and jshell as appropriate
 * `//PREVIEW` in source files is possible
 * aliases can specify enable preview too.

Fixes #1622

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ plugins {
 	id "com.github.breadmoirai.github-release" version "2.4.1"
 	id "io.toolebox.git-versioner" version "1.6.5"
 	id 'org.gradle.crypto.checksum' version '1.4.0'
-	id "com.diffplug.spotless" version "6.15.0"
+	id "com.diffplug.spotless" version "6.13.0" // can't go higher to still work with Java 8
 	id 'com.github.gmazzo.buildconfig' version "3.1.0"
 	id "com.dorongold.task-tree" version "2.1.1"
 	id "com.geoffgranum.gradle-conventional-changelog" version "0.3.1"
--- a/src/main/java/dev/jbang/catalog/Alias.java
+++ b/src/main/java/dev/jbang/catalog/Alias.java
@@ -43,6 +43,8 @@ public class Alias extends CatalogItem {
 	public final String debug;
 	public final Boolean cds;
 	public final Boolean interactive;
+	@SerializedName(value = "enable-preview")
+	public final Boolean enablePreview;
 	@SerializedName(value = "enable-assertions")
 	public final Boolean enableAssertions;
 	@SerializedName(value = "enable-system-assertions")
@@ -69,6 +71,7 @@ public class Alias extends CatalogItem {
 
 	public Alias() {
 		this(null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null,
+				null,
 				null, null, null, null, null, null, null);
 	}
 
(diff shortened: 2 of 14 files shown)


maxandersen:
realizing it would be nice //ENABLE_PREVIEW or simply //PREVIEW is possible to put in sources.

This PR does not cover kotlin,groovy etc. but intent would be if they have similar flag they would pass that through.

maxandersen:
> realizing it would be nice //ENABLE_PREVIEW or simply //PREVIEW is possible to put in sources.

updated with `//PREVIEW` flag

quintesse (dismissed):
LGTM

maxandersen:
hmm - test runs reveals a problem I didn't see locally. enable-preview requiring using --source or --release with a version number.

not sure  how to handle that best....

quintesse:
Ahh.... well the other option is that inside the `JshCommandGenerator` we specifically check for the existence of a `--enable-preview` flag in the runtime options and we then also add it to the jshell command itself (instead of only to the `-R` flag)

maxandersen:
pushed update that set `--source <requestedJavaVersion int>` for `javac`  calls.

maxandersen:
> Ahh.... well the other option is that inside the `JshCommandGenerator` we specifically check for the existence of a `--enable-preview` flag in the runtime options and we then also add it to the jshell command itself (instead of only to the `-R` flag)
> 
> This would also simplify the PR substantially

not following?

quintesse:
> > Ahh.... well the other option is that inside the `JshCommandGenerator` we specifically check for the existence of a `--enable-preview` flag in the runtime options and we then also add it to the jshell command itself (instead of only to the `-R` flag)
> > This would also simplify the PR substantially
> 
> not following?

I was suggesting that if someone does a `//JAVA_OPTIONS --enable-preview` or `jbang -R --enable-preview` we could detect that it has that preview option and simply add it to the `jshell` command as well.

It has the advantage that if someone is using the standard Java flags it will still work. With this new `jbang --enable-preview` flag it is now the _only_ option that will work for jshell. So it creates this slightly weird situation where `jbang -R --enable-preview` will work of it's a .java file you're running but not if you're running a .jsh file.

And it would be simpler because we wouldn't need to add any flags (`--enable-preview`) and tags (`//PREVIEW`).

maxandersen:
the issue with picking up -R --enable-preview and apply that everywhere would break or at least mix up semantics and we would still need --enable-preview as a flag as `-R --enable-preview` can't automatically be assumed to imply `-C --enable-preview -source N` should be done.

maxandersen:
merging this as javac at least is fine having multiple --enable-preview so its a future optimization to avoid conflicts.

PR #1630, opened 2023-06-07
fix: Now showing integration errors when verbose is on

quintesse:


maxandersen:
Using the self classloader so scripts can be their own integration? That's...not sure - either crazy or brilliant :)

quintesse:
> Using the self classloader so scripts can be their own integration? That's...not sure - either crazy or brilliant :)

🙂 

But why not? You might want to achieve something that you just can't do with regular code. I mean, the alternative is really hard: you would first need to turn the integration into a dependency published on Maven central before being able to use it. 

Also, we support `//SOURCES` that are URLs, so you could imagine some kind of nice integration that you just include in your script and it automagically does something useful.

But the real reason is that I was looking for an easy way to test integrations 😉

maxandersen:
very weird issue ...first time I ran with this --verbose resulted in the final output to not be shown.

```
jbang init -t qcli t.java
jbang --verbose t.java
```

shows all the output but didn't actually run or at least not printed "Hello World!!"

but I can't reliably reproduce it.

wondering if there are a possible race condition on out put streams?

src/main/java/dev/jbang/spi/IntegrationManager.java
```diff
@@ -267,6 +278,7 @@ public static void main(String... args) {
 		IntegrationInput input = parser.fromJson(new InputStreamReader(System.in), IntegrationInput.class);
 		ClassLoader old = Thread.currentThread().getContextClassLoader();
 		PrintStream oldout = System.out;
+		Util.setVerbose(input.verbose);
```
maxandersen, line 281: shouldn't this not be reversed afterwards?

maxandersen:
also - i dont notice a difference.. got example to test with?

src/main/java/dev/jbang/spi/IntegrationManager.java
quintesse: No, in the case JBang itself is running the code this line is a no-op (it will be setting it to its current value) while in the case of external code the app will exit afterwards anyway.

quintesse:
> also - i dont notice a difference.. got example to test with?

Something has to go wrong, that's the only thing that should change. So instead of swallowing errors you should now be able to see them.

quintesse:
> wondering if there are a possible race condition on out put streams?

Perhaps, I haven't ever seen it myself. But there's nothing in this PR that fools around with streams, so it might be an existing issue.

maxandersen:
needs rebasing

quintesse:
Done

PR #1632, opened 2023-06-07
fix: detect version from JAVA_RUNTIME_VERSION and don't use JAVA_VERS…

### LINKED ISSUE #1631: cant use nightly build of jdk with jbang
follow instructions at https://github.com/openjdk/jdk/blob/master/doc/building.md

then:

```
jbang jdk install --force 21 `pwd`
[jbang] [ERROR] Java version in given path: /Users/manderse/code/openjdk/jdk/build/macosx-aarch64-server-release/jdk is 2023 which does not match the requested version 21
```

the release file contains this:
```
cat release
SOURCE=".:git:4ffc8cc2169c"
IMPLEMENTOR="N/A"
JAVA_VERSION_DATE="2023-09-19"
JAVA_RUNTIME_VERSION="21-internal-adhoc.manderse.jdk"
OS_NAME="Darwin"
OS_ARCH="aarch64"
LIBC="default"
```

```
 bin/java --version
openjdk 21-internal 2023-09-19
OpenJDK Runtime Environment (build 21-internal-adhoc.manderse.jdk)
OpenJDK 64-Bit Server VM (build 21-internal-adhoc.manderse.jdk, mixed mode)
```

I tried doing ` jbang jdk install --force 2023 `pwd`` which works but then when using `//PREVIEW` I get:

```
jbang --enable-preview -m main UnnamedClassMainMethodLauncher.java
[jbang] Building jar...
error: invalid source release: 2023
Usage: javac <options> <source files>
use --help for a list of possible options
```

maxandersen:
fixes #1631

### DIFF
--- a/src/main/java/dev/jbang/util/JavaUtil.java
+++ b/src/main/java/dev/jbang/util/JavaUtil.java
@@ -143,7 +143,7 @@ public class JavaUtil {
 	public static Optional<String> readJavaVersionStringFromReleaseFile(Path home) {
 		try (Stream<String> lines = Files.lines(home.resolve("release"))) {
 			return lines
-						.filter(l -> l.startsWith("JAVA_VERSION=") || l.startsWith("JAVA_RUNTIME_VERSION"))
+						.filter(l -> l.startsWith("JAVA_VERSION=") || l.startsWith("JAVA_RUNTIME_VERSION="))
 						.map(JavaUtil::parseJavaOutput)
 						.findAny();
 		} catch (IOException e) {
--- a/src/main/java/dev/jbang/util/JavaUtil.java
+++ b/src/main/java/dev/jbang/util/JavaUtil.java
@@ -143,7 +143,7 @@ public class JavaUtil {
 	public static Optional<String> readJavaVersionStringFromReleaseFile(Path home) {
 		try (Stream<String> lines = Files.lines(home.resolve("release"))) {
 			return lines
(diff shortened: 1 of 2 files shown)


src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -143,7 +143,7 @@ public static Optional<String> resolveJavaVersionStringFromPath(Path home) {
 	public static Optional<String> readJavaVersionStringFromReleaseFile(Path home) {
 		try (Stream<String> lines = Files.lines(home.resolve("release"))) {
 			return lines
-						.filter(l -> l.startsWith("JAVA_VERSION"))
+						.filter(l -> l.startsWith("JAVA_VERSION=") || l.startsWith("JAVA_RUNTIME_VERSION"))
```
quintesse, line 146: Isn't it better to have both tests the same way? (Both ending in `=`)

maxandersen, line 146:
```suggestion
						.filter(l -> l.startsWith("JAVA_VERSION=") || l.startsWith("JAVA_RUNTIME_VERSION="))
```

quintesse:
So, the thinking behind this is correct... I'm just worried that there might be RELEASE files that have lines like `JAVA_VERSION = XX` (see spaces).

maxandersen:
JAVA_RUNTIME_VERSION is part of java 21 https://github.com/openjdk/jdk/pull/12813

PR #1634, opened 2023-06-08
feat: scripts can now implement integrations

quintesse:
Needs #1630 to be applied first

### DIFF
--- a/itests/integration/IntegrationTest.java
+++ b/itests/integration/IntegrationTest.java
@@ -1,0 +1,23 @@
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class IntegrationTest {
+    public static Map<String, Object> postBuild(Path temporaryJar,
+                                                Path pomFile,
+                                                List<Map.Entry<String, String>> repositories,
+                                                List<Map.Entry<String, Path>> dependencies,
+                                                List<String> comments,
+                                                boolean nativeImage) {
+        System.out.println("Integration... (out)");
+        System.err.println("Integration... (err)");
+        if (System.getProperty("failintegration") != null) {
+            throw new RuntimeException("Failing integration...");
+        } else {
+            System.out.println("Integration OK (out)");
+            System.err.println("Integration OK (err)");
+            return Collections.emptyMap();
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


quintesse:
So the discussion to merge this or not was about the issue that user files are now part of the class path when running the post build and if this could result in problems or not. Right now it only runs with all the dependencies specified by the user code, but not the code itself. (Which is why currently it isn't possible to write integration code within the user code).

A possible issue might be that the classloader would hold a lock on the very resources that the integration might try to change, something that especially on Windows might cause problems.

maxandersen:
with #1641 this should be doable without "polluting" the integration class path. 

But it does not yet work because the integration assumes all deps have a GAV...

quintesse:
This is an old PR but I still think it could be useful. The changes are minimal.

The only thing is that the new IT tests fail, it seems because the capturing if input and output doesn't work as expected.

maxandersen:
The GAV issue mentioned [here](https://github.com/jbangdev/jbang/pull/1634#issuecomment-1626899080) is no longer relevant — `IntegrationManager` now handles deps without a GAV by falling back to the filename as key:

```java
if (art.getCoordinate() != null) { // skipping dependencies that does not have a GAV
    deps.put(art.getCoordinate().toCanonicalForm(), art.getFile());
} else {
    deps.put(art.getFile().getFileName().toString(), art.getFile());
}
```

Also cherry-picked and rebased onto current main. Tests now pass (needed `--fresh` to avoid cached builds skipping integration execution). Branch: `pr-1634-cherry`

maxandersen:
**Status update:** Cherry-picked onto current main and pushed to `pr-1634-cherry`. All 6 integration tests pass (including 4 new in-script integration tests).

**Summary of changes:**
- `IntegrationManager.getClassLoader()` now includes the compile output dir, so scripts can implement their own integrations via `META-INF/jbang-integration.list`
- Keeps main's improvements: try-with-resources for the classloader, `@NonNull` (jspecify)
- The prerequisite PRs (#1630, #1641) are already merged

**Previously raised concerns — all resolved:**
- **GAV issue** (maxandersen, July 2023): was already fixed in May 2022 via #1350 — deps without a GAV fall back to filename as key
- **Test failures** (quintesse, June 2025): caused by missing `--fresh` flag — cached builds skip integration execution entirely. Fixed by adding `--fresh` to all 4 new tests, consistent with the existing integration tests in the same file

Ready for review as a fresh PR from `pr-1634-cherry`, or this PR can be retargeted to that branch.

PR #1641, opened 2023-06-28
feat: Added support for script dependencies

quintesse:
You can now do:

```
//DEPS local_file.java http://www.example.com/remote_file.java
```

And those files will be compiled, if needed, and added as dependencies on the classpath.

### DIFF
--- a/src/main/java/dev/jbang/dependencies/JitPackUtil.java
+++ b/src/main/java/dev/jbang/dependencies/JitPackUtil.java
@@ -11,17 +11,17 @@ import java.util.stream.Stream;
 
 public class JitPackUtil {
 	private static final Pattern GITHUB_TREE_PATTERN = Pattern.compile(
-			"^https?://github.com/(.+?)/(.+?)(/tree/(.+?)(/(.+?))?)?/?$");
+			"^https?://github.com/([^/]+?)/([^/]+?)(/tree/([^/]+?)(/(.+?))?)?/?$");
 	private static final Pattern GITUB_COMMIT_PATTERN = Pattern.compile(
-			"^https?://github.com/(.+?)/(.+?)/commit/(.+)$");
+			"^https?://github.com/([^/]+?)/([^/]+?)/commit/(.+)$");
 	private static final Pattern GITLAB_TREE_PATTERN = Pattern.compile(
-			"^https?://gitlab.com/(.+?)/(.+?)(/-/tree/(.+?)(/(.+?))?)?/?$");
+			"^https?://gitlab.com/([^/]+?)/([^/]+?)(/-/tree/([^/]+?)(/(.+?))?)?/?$");
 	private static final Pattern GITLAB_COMMIT_PATTERN = Pattern.compile(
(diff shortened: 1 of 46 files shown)


maxandersen:
Awesome. Does swizzled urls work too? Ie. For gist and fully qualified catalogs?

src/main/java/dev/jbang/dependencies/JitPackUtil.java
```diff
@@ -11,17 +11,17 @@
 
 public class JitPackUtil {
 	private static final Pattern GITHUB_TREE_PATTERN = Pattern.compile(
-			"^https?://github.com/(.+?)/(.+?)(/tree/(.+?)(/(.+?))?)?/?$");
+			"^https?://github.com/([^/]+?)/([^/]+?)(/tree/([^/]+?)(/(.+?))?)?/?$");
```
maxandersen, line 14: Why are these patterns changing?

quintesse: Because they are too lenient, they match URLs that are not valid JitPack URLs

quintesse: For example it will match https://github.com/jbangdev/jbang/blob/main/itests/SankeyPlotTest.java , even though that's obviously neither a "tree" nor a "commit".

maxandersen:
first quick test with this beauty:

```
//DEPS info.picocli:picocli:4.6.3
//DEPS https://github.com/jbangdev/jbang-catalog/blob/main/httpd.java


import picocli.CommandLine;
import picocli.CommandLine.Parameters;

import java.util.concurrent.Callable;

@Command(name = "hello", mixinStandardHelpOptions = true, version = "hello 0.1",
        description = "hello made with jbang")
class hello implements Callable<Integer> {

    @Parameters(index = "0", description = "The greeting to print", defaultValue = "World!")
    private String greeting;

        int exitCode = new CommandLine(new hello()).execute(args);
        System.exit(exitCode);
    }

    @Override
    public Integer call() throws Exception { // your business logic goes here...
        System.out.println("Hello again " + greeting);

        httpd.main(new String[] { "--port=8080" });
        return 0;
    }
}
```

at first glance worked - even vscode picked up the new deps.

a few things:

```
[jbang] Building jar...
[jbang]    info.picocli:picocli:4.6.3
[jbang] Dependencies resolved
[jbang] Building jar...
```

The Building jar should have some indication of what its building now.

If touch just the hello.java I still get:

```
jbang hello.java
[jbang] Building jar...
[jbang] Building jar...
Hello again World!
```

shouldn't there only be one Building Jar in that case ?

maxandersen:
```
//JAVA 21+
//PREVIEW
//DEPS hello.java

void main() {
    System.out.println("Hello, world!");
}
```

instant stack overflow :)

maxandersen:
btw. `//DEPS other.jar` works. sweet!

maxandersen:
`//DEPS httpd@jbangdev` does *not* work...part of me says it shouldn't but then again....its pretty damn nice.

PR #1654, opened 2023-07-19
feat: introduce clean docker multi java version build

maxandersen:
Why:

 * jbang-action is hardcoded to use a specific java version
   requiring users to download new java version
 * nicer if can have a docker images that has the java version
   built in

 * move jbang-action to a separate distribution (might consider dropping it fully)
 * add a jbang-docker repo that builds a docker image with jbang and java
   and publishes under jbangdev/jbang:<version>-java-<java major version>.


src/jreleaser/distributions/jbang/docker/Dockerfile.tpl
```diff
@@ -8,30 +9,16 @@ LABEL {{.}}
...
 
-ENV PATH="${PATH}:/{{distributionArtifactName}}/bin"
-
-ADD ./entrypoint /bin/entrypoint
-
-ENV SCRIPTS_HOME /scripts
-ENV JBANG_VERSION {{projectVersion}}
-ENV JBANG_PATH=/jbang/bin
-
-VOLUME /scripts
```
maxandersen, line 30: might want to keep /scripts

maxandersen:
merging this to see if can make it all work :) wish the next release luck!

PR #1660, opened 2023-08-02
fix: Fixes NPE when running in container

### LINKED ISSUE #1653: console read NPE's in container with no terminal associated
I'm trying to make a new docker image for jbang that is simpler than current jbang-action and hit this issue:

```
[jbang] [0:469] Type in your choice and hit enter. Will automatically select option (0) after 30 seconds.
[jbang] [0:472] [ERROR] public class java.lang.NullPointerException
        at dev.jbang.util.ConsoleInput.readLine(ConsoleInput.java:71)
        at dev.jbang.util.Util.askInput(Util.java:1774)
        at dev.jbang.source.resolvers.RemoteResourceResolver.fetchScriptFromUntrustedURL(RemoteResourceResolver.java:69)
        at dev.jbang.source.resolvers.RemoteResourceResolver.resolve(RemoteResourceResolver.java:42)
        at dev.jbang.source.resolvers.CombinedResourceResolver.lambda$resolve$0(CombinedResourceResolver.java:27)
```

simply running `jbang quarkusinfo@maxandersen` from within a container with no terminal.

we should not NPE.

quintesse:
Fixes #1653

### DIFF
--- a/src/main/java/dev/jbang/util/ConsoleInput.java
+++ b/src/main/java/dev/jbang/util/ConsoleInput.java
@@ -1,13 +1,18 @@
 package dev.jbang.util;
 
-import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.concurrent.*;
 
 public abstract class ConsoleInput {
 	private final int tries;
 	private final int timeout;
 	private final TimeUnit unit;
+
+	private static final Path TTY = Paths.get("/dev/tty");
 
 	/**
 	 * Will either return a ConsoleInput that enables reading a line from the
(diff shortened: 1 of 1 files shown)


src/main/java/dev/jbang/util/ConsoleInput.java
```diff
@@ -68,11 +84,9 @@ public String readLine() {
 					input = result.get(timeout, unit);
 					break;
 				} catch (ExecutionException | IOException e) {
-					e.getCause().printStackTrace();
+					Util.verboseMsg("/dev/tty is not available");
```
maxandersen, line 87: It won't still make sense to print exception if in verbose mode ?

maxandersen (changes requested):
Looks good but many print exception on verbose?

src/main/java/dev/jbang/util/ConsoleInput.java
quintesse: Sure

maxandersen (approved):

PR #1657, opened 2023-07-21
fix: we now only add `-g` flag to Java source types

quintesse:
Before it would get added to Groovy and Kotlin as well.

Fixes #1656

PR #1670, opened 2023-08-29
chore: some changes for Groovy and Kotlin: Groovy default version to 4, support Groovy 5

linux-china:
- Update to Groovy default version to 4.0.14
- Support Groovy 5
- Update to Kotlin default version 1.8.22

### DIFF
--- a/src/main/java/dev/jbang/source/sources/GroovySource.java
+++ b/src/main/java/dev/jbang/source/sources/GroovySource.java
@@ -40,6 +40,7 @@ public class GroovySource extends Source {
 	protected List<String> getRuntimeOptions() {
 		List<String> gopts = Collections.singletonList("-Dgroovy.grape.enable=false");
 		List<String> opts = tagReader.collectOptions("JAVA_OPTIONS", "RUNTIME_OPTIONS");
+		opts.remove("-g");
 		return Util.join(gopts, opts);
 	}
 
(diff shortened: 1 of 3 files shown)


src/main/java/dev/jbang/source/sources/GroovySource.java
```diff
@@ -40,14 +40,15 @@ protected List<String> getNativeOptions() {
 	protected List<String> getRuntimeOptions() {
 		List<String> gopts = Collections.singletonList("-Dgroovy.grape.enable=false");
 		List<String> opts = tagReader.collectOptions("JAVA_OPTIONS", "RUNTIME_OPTIONS");
+		opts.remove("-g");
```
quintesse, line 43: This is definitely not the way to handle this :-)

But I'm pretty sure this was already fixed by #1657

maxandersen: yeah - this should not be needed.

linux-china:
> How does user run groovy 3 based scripts with this change ?

//GROOVY 3.0.19

src/main/java/dev/jbang/source/sources/GroovySource.java
linux-china: Done, and delete this line.

maxandersen:
@linux-china sorry missed your comment. could you in this PR then please update the docs feeding https://www.jbang.dev/documentation/guide/latest/usage.html#running-groovy-groovy-experimental and document the option to use //GROOVY for specific versions?

maxandersen:
i've added the docs so we can get this out.

PR #1686, opened 2023-09-28
fix: alias settings not being properly applied

quintesse:


quintesse:
@maxandersen ready to be merged. The fix is small , most of this PR is tests.

maxandersen:
What is it fixing?

quintesse:
The fact that some alias settings were not being applied. Like the one from your `gython` script that says Java 21 is required. It failed on my machine because I don't have 21 installed and JBang didn't offer to install it.

PR #1692, opened 2023-10-05
fix: handle debug output messing up jbangdev

### LINKED ISSUE #1572: handle debug output messing up jbangdev
if you run with `JBANG_JAVA_OPTIONS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5555`

an error occurs like:

```
/Users/manderse/code/jbangdev/jbang/build/install/jbang/bin/jbang: line 230: exec: Listening: not found
```

because the  output contains the following:
```
Listening for transport dt_socket at address: 5555
/Users/manderse/.jbang/cache/jdks/11/bin/java '-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5555' -classpath /Users/manderse/.jbang/cache/jars/env.java.a9698542287f63ccc8120cf14c41e464b3d015626675bbe06bcca02faaea47d1/env.jar:/Users/manderse/.m2/repository/info/picocli/picocli/4.2.0/picocli-4.2.0.jar:/Users/manderse/.m2/repository/de/vandermeer/asciitable/0.3.2/asciitable-0.3.2.jar:/Users/manderse/.m2/repository/de/vandermeer/ascii-utf-themes/0.0.1/ascii-utf-themes-0.0.1.jar:/Users/manderse/.m2/repository/de/vandermeer/skb-interfaces/0.0.1/skb-interfaces-0.0.1.jar:/Users/manderse/.m2/repository/org/apache/commons/commons-lang3/3.4/commons-lang3-3.4.jar:/Users/manderse/.m2/repository/org/antlr/ST4/4.0.8/ST4-4.0.8.jar:/Users/manderse/.m2/repository/org/antlr/antlr-runtime/3.5.2/antlr-runtime-3.5.2.jar:/Users/manderse/.m2/repository/org/antlr/antlr4/4.5.1/antlr4-4.5.1.jar:/Users/manderse/.m2/repository/org/slf4j/slf4j-api/1.7.12/slf4j-api-1.7.12.jar:/Users/manderse/.m2/repository/de/vandermeer/char-translation/0.0.2/char-translation-0.0.2.jar:/Users/manderse/.m2/repository/org/jline/jline/3.16.0/jline-3.16.0.jar env xyz
```

I wonder if we could strip out first line if it starts with "Listening" or we should simply just execute the *last* line ?

essobedo:
fixes #1572 

## Motivation

The verbosity of the JVM may prevent the debug mode from working properly with an error of type `jbang: line 230: exec: Listening: not found`

## Modifications:

* Keep the last line of the output to avoid being affected by the verbosity of the JVM.

## Result:

```
$ export JBANG_JAVA_OPTIONS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005" 
$ jbang hello.java Max!
Listening for transport dt_socket at address: 5005
Hello Max!
```

**NB:** This fix doesn't cover Windows OS

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -221,6 +221,11 @@ export JBANG_STDIN_NOTTY=$([ -t 0 ] && echo "false" || echo "true")
 output=$(CLICOLOR_FORCE=1 "${JAVA_EXEC}" ${JBANG_JAVA_OPTIONS} -classpath "${jarPath}" dev.jbang.Main "$@")
 err=$?
 if [ $err -eq 255 ]; then
+  nl=$'\n'
+  if [[ $output =~ $nl  ]]; then
+    ## keep the last output line to avoid being polluted by the potential verbosity of the JVM
+    output=$(echo "$output" | tail -1)
+  fi
   eval "exec $output"
 elif [ -n "$output" ]; then
   echo "$output"


essobedo:
@maxandersen Hi 👋 , WDYT of this small fix?

quintesse (commented):
I like this second version better than the first, but I'm not thrilled about the fact that:

a) this means we can never have multi-line output
b) we have to cross our fingers and hope the order of output is always the same
c) we leave Windows users hanging

It's the reason why I hadn't approved this PR yet.
Not sure how @maxandersen feels about this though.

essobedo:
@quintesse Thank you for your feedback, I understand your point of view, my 2 cents

> a) this means we can never have multi-line output 
Maybe, I missed something but according to what I understood from the code, when the exit code is 255, the goal is to generate a command to launch the application and the command is a single-line
> b) we have to cross our fingers and hope the order of output is always the same
This JVM output message will always be the first message as the custom Java code is not yet launched at this stage 
> c) we leave Windows users hanging
Do you mean that you rather prefer not having any fix instead of having a partial fix?

maxandersen:
Im currently on same page as @quintesse; mainly because we don't have a windows solution. 

I'm wondering if we should prefix line(s) to be executed with a string, I.e. "exec:" to indicate what to execute to make it more reliable and work reliably and consistent on all platforms?

quintesse:
> I'm wondering if we should prefix line(s) to be executed with a string, I.e. "exec:" to indicate what to execute to make it more reliable and work reliably and consistent on all platforms?

The problem has always been that .bat files have very crappy support for anything related to string manipulation. So finding lines that start with a word and stripping it off without in some way introducing new issues with encodings and escapes, etc is really hard

essobedo:
Never mind sorry for the noise

maxandersen:
@essobedo no need to close. Trying to figure out if way to fix it while keeping behavior consistent.

Issue #1694, opened 2023-10-13
Where does JBang look for proxy information when downloading JDKs?


aghasemi:
Hi,

My JBang installation works fine with proxy when downloading Maven dependencies, but not when downloading JDKs. The proxy is specified in both `~/.m2/settings.xml` and `http_proxy` (and related) environment variables. Is there a third place JBang expects to see proxy? 

Many thanks

gwentey:
Same issue here...

maxandersen:
Hmm. For http download it uses standard java httpd so the system properties should work?

aghasemi:
http_proxy or HTTP_PROXY?

maxandersen:
jbang -Dhttp.proxyHost=127.0.0.1 -Dhttp.proxyPort=3128

quintesse:
Also important to know which OS you're using. Not all OSes use the same system to set up proxies.

gwentey:
jbang -Dhttp.proxyHost=127.0.0.1 -Dhttp.proxyPort=3128
with this we can't put credentials (username@password)

maxandersen:
I don't have access to a proxy that requires authentication. What do you normally do for java apps to have this work?

maxandersen:
Tried `Dhttp.proxyHost=myproxy -Dhttp.proxyPort=myport -Dhttp.proxyUser=myuser -Dhttp.proxyPassword=mypass -Dhttps.proxyHost=myproxy -Dhttps.proxyPort=myport -Dhttps.proxyUser=myuser -Dhttps.proxyPassword=mypass`

?

Note: afaik none of this is jbang specific. Just standard java proxy config.

quintesse:
Those options don't seem to work for me (see #1695)

gwentey:
I tried options don’t work too for me

DyspC:
Hello, 

I was looking into an issue like this when downloading remote scripts but looking into the code it appears the JBangJdkProvider and RemoteResourceResolver use the same utility class so I'll report here

https://github.com/jbangdev/jbang/blob/d8b8bb4613f6b07b6601f148a56c2f9faf5d301c/src/main/java/dev/jbang/util/Util.java#L800

The download process uses the `URL::openConnection` method that IIRC does not use any other proxy than the one it's fed 

Hardcoding it and rebuilding jbang locally fixes it for me so it would be great if this proxy was somehow configurable

```java
		URL url = new URL(fileURL);
[...]

PR #1695, opened 2023-10-15
test: updated docker-compose file for testing proxies

quintesse:
We can now test with proxies that require authentication.


quintesse:
Created this for #1694 , to test the suggested options, but they don't seem to work for me, Still I thought the update could be relevant for future tests.

PR #1698, opened 2023-10-21
fix: allow to mark http url as trusted

ritikverma2000:
I have added http inside trustedSource.startsWith .

image

### DIFF
--- a/src/main/java/dev/jbang/net/TrustedSources.java
+++ b/src/main/java/dev/jbang/net/TrustedSources.java
@@ -85,7 +85,7 @@ public class TrustedSources {
 			}
 
 			URI parsedTrustedSource;
-			if (trustedSource.startsWith("http://")) {
+			if (trustedSource.startsWith("https://")||trustedSource.startsWith("http://")) {
 				parsedTrustedSource = new URI(trustedSource);
 				if (!url.getScheme().equals(parsedTrustedSource.getScheme())) {
 					continue;
--- a/src/main/java/dev/jbang/net/TrustedSources.java
+++ b/src/main/java/dev/jbang/net/TrustedSources.java
@@ -85,7 +85,7 @@ public class TrustedSources {
 			}
 
 			URI parsedTrustedSource;
-			if (trustedSource.startsWith("https://")) {
+			if (trustedSource.startsWith("http://")) {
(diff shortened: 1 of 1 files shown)


src/main/java/dev/jbang/net/TrustedSources.java
```diff
@@ -85,7 +85,7 @@ public boolean isURLTrusted(URI url) throws URISyntaxException {
 			}
 
 			URI parsedTrustedSource;
-			if (trustedSource.startsWith("https://")) {
+			if (trustedSource.startsWith("http://")) {
```
maxandersen, line 88: Now it won't work with https so that does not work as fix.

maxandersen (changes requested):

src/main/java/dev/jbang/net/TrustedSources.java
ritikverma2000: Okay I can just add a or condition in here then @maxandersen what do you suggest.

maxandersen: Yes. That seems reasonable. 

I highly expect the tests to fail with the current suggestion so I suggest you run the tests, then see what fails and add a test that test for http based URL. Then fix it so all the tests passes.

ritikverma2000: @maxandersen we have this method already present. 
image

maxandersen: Yes. So add a test that uses http so we can verify it failed before and fixed now with the change.

maxandersen:
The build is failing because formatting not applied and no tests added to verify this actually works. 

Please run `gradle spotlessApply build clean` to have formatting applied and have tests pass. See CONTRIBUTING.MD for details.


Thanks.

ritikverma2000:
> The build is failing because formatting not applied and no tests added to verify this actually works.
> 
> Please run `gradle spotlessApply build clean` to have formatting applied and have tests pass. See CONTRIBUTING.MD for details.
> 
> Thanks.

Sure @maxandersen

maxandersen:
Thans @ritikverma2000 I fixed the formatting so builds will work.

PR #1700, opened 2023-10-25
docs: add docs for proxies and add settings.xml in proxy setup

maxandersen:
realized our docs were missing proxy config and our own testing missed the settings.xml settings.

should still make it better and require less setup as discussed in #1604 but this is a start.

### DIFF
--- a/docs/modules/ROOT/pages/configuration.adoc
+++ b/docs/modules/ROOT/pages/configuration.adoc
@@ -50,5 +50,55 @@ classpath:/jbang.properties
    wrapper.install.dir = .
 ```
 
+== Proxy configuration
 
+If you use proxies there are a few settings to apply to make JBang *fully* work with proxies. Unfortunately Java treats proxy setup a little bit different based on OS and environment.
+
+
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/configuration.adoc
```diff
@@ -50,5 +50,48 @@ classpath:/jbang.properties
...
 ```
 
+== Proxy configuration
 
+If you use proxies there are a few settings to apply to make JBang *fully* work with proxies.
+
+First the proxy settings for bash/zsh scripting part (needed if jbang nor java is not already installed). This is done by setting the `http_proxy` and `https_proxy` environment variables. For example:
+
+```bash
+http_proxy=http://proxyhost:8888
```
quintesse, line 60: I wonder if we need to explain what to do on Windows. But I'm not sure how to test. I have Windows, but not sure how to simulate a proxy.

maxandersen: Good point - i reckon on Windows use system proxies actually Works.

maxandersen: I've updated with some more system context.

PR #1707, opened 2023-11-22
feat: added support for `%{deps:...}` substitutions in arguments

### LINKED ISSUE #1703: support generating classpaths embedded on command line
I find that sometimes cli's out there need a classpath and it can be quite tedious to type out.

I wonder if like done in #1043 where %{url} fetches content and make it available as a file we could support %{deps://DEPS syntax} to generate classpath.

i.e.

`jbang sqlline@maxandersen -cp %{deps:org.hsqldb:hsqldb:RELEASE,..} other args`

quintesse:
Fixes #1703


maxandersen:
this works remarkably well :)

quintesse:
Yea, although I only implemented this for the `%{ ... }` and not the simpler `%...` form... and now I'm wondering if that was wise. Perhaps we should just support `%deps:...` as well?

PR #1738, opened 2024-01-23
feat: treat empty catalog as jbanghub/jbang-catalog aliases

maxandersen:
prototype for https://github.com/jbangdev/jbang/discussions/1733

lets you do:

`jbang sqlline@/sqlline`

and

`jbang h2@/`

basically treating "empty catalog" as a referring to "jbanghub"

### DIFF
--- a/src/main/java/dev/jbang/catalog/CatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/CatalogRef.java
@@ -52,9 +52,11 @@ public class CatalogRef extends CatalogItem {
 		if (catalog != null) {
 			catalogRef = catalog.catalogs.get(catalogName);
 		}
-		if (catalogRef == null && Util.isValidPath(catalogName)) {
+		// check if local file matches first.
+		if (catalogRef == null && !catalogName.equals("")
+				&& Util.isValidPath(catalogName)) {
 			Path p = Util.getCwd().resolve(catalogName);
-			if (!p.getFileName().toString().equals(Catalog.JBANG_CATALOG_JSON)) {
+			if (p.getFileName() != null && !p.getFileName().toString().equals(Catalog.JBANG_CATALOG_JSON)) {
 				p = p.resolve(Catalog.JBANG_CATALOG_JSON);
 			}
 			if (Files.isRegularFile(p)) {
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -41,6 +41,9 @@ public class ImplicitCatalogRef {
 		if (Util.isURL(name)) {
 			return null;
 		}
+		if (name.startsWith("/")) {
+			name = Catalog.JBANG_DEFAULT_CATALOG + name;
+		}
 		String[] parts = name.split("~", 2);
 		String path;
 		if (parts.length == 2) {
--- a/src/main/java/dev/jbang/catalog/Alias.java
+++ b/src/main/java/dev/jbang/catalog/Alias.java
@@ -2,7 +2,9 @@ package dev.jbang.catalog;
 
 import static dev.jbang.cli.BaseCommand.EXIT_INVALID_INPUT;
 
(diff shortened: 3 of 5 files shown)


maxandersen:
hmm - test pass but have some other bugs :)

maxandersen:
fixed it so it now works from command line - lets see if all other tests passes.

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
```diff
@@ -52,7 +52,11 @@ public static ImplicitCatalogRef parse(String name) {
 		if (names.length > 3) {
 			return null;
 		}
-		String org = names[0];
+		String org = null;
+		if (names.length > 2) {
+			org = names[0];
+		}
+		org = "jbangapp"; // TODO: should global default be controllable by config?
```
quintesse, line 59: I think we should just make it a constant, just like `Catalog.JBANG_CATALOG_JSON`

maxandersen:
Should we accept h2@ to be equal to h2@/ and h2@jbangapp ?

quintesse (dismissed):

maxandersen:
ive updated so: h2@ is treated as blank catalog and thus maps to h2@jbangapp  and sqlline@/sqlline works too

should we error on h2@/  or be lenient ? (we do already accept `env@jbangdev/` before so I guess yes for consistency.

src/main/java/dev/jbang/catalog/CatalogRef.java
```diff
@@ -52,9 +52,10 @@ static CatalogRef get(String catalogName) {
 		if (catalog != null) {
 			catalogRef = catalog.catalogs.get(catalogName);
 		}
-		if (catalogRef == null && Util.isValidPath(catalogName)) {
+		if (catalogRef == null && (!catalogName.startsWith("/") || !catalogName.equals(""))
```
maxandersen, line 55: this is causing the test failure...we are for some reason resolving local paths first to support `somealias@/var/path/locally/ondisk` ?

@quintesse what usecase is that for ?

maxandersen: https://github.com/jbangdev/jbang/pull/1363 is what introduced `bar@/some/file/on/disk`

maxandersen: I've updated the code to make it so paths that exist on disk will be honorored.

But I'm not too happy about it.

its only needed for when you use absolute paths...relative paths would be different.

I could also let it run through first and then if fails lookup in jbanghub as a fallback - to avoid scanning multiple times.

maxandersen: could accept file:// instead of raw path ?

Because as it is now a local dir will actually override an implicit catalog. 

Jbang myapp@jbanghub and you have jbanghub locally will override jbanghub catalog.

quintesse: > could accept file:// instead of raw path ?

We could do that, yes. It does go somewhat against the fact that generally speaking a catalog reference is also a `name@resourceref` thing, so almost anything you should be able to pass to for example a `//FILE` tag was possible to use in a catalog reference. We'd now be limiting that. Now, I don't think anyone was actually using this, so we most likely won't be breaking anything, but just be aware that we _will_ be making a breaking change.

quintesse: Ok, I just checked the URI spec, it seems that `file:` doesn't allow relative paths, so in that respect using `file:` is not a good replacement. And that _does_ seem somewhat wrong. You'd be forced to always use absolute paths.

maxandersen: hmm - I thought `file:/./mydir` was allowed but yeah does not seem to be "official" syntax.

URI.create("file:./pom.xml").getSchemeSpecficiPart() does let you fetch it though... but yeah not great.

maxandersen:
i believe we don't have a good usecase for this anymore - if we do we can reopen.

PR #1743, opened 2024-02-02
Catalog command improvements

### LINKED ISSUE #1678: "Script or alias could not be found or read" should not be issued if "jbang alias list" lists it
**Describe the bug**

"Script or alias could not be found or read" is output even if the script is shown at `jbang alias list`

**To Reproduce**

1. `jbang catalog add https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/jbang-catalog.json`
2. ```cmd
   > jbang --verbose alias list
   Picked up JAVA_TOOL_OPTIONS: -Duser.language=en
   [jbang] [0:240] jbang version 0.110.1
   [jbang] [0:255] Reading catalog from C:\Users\koppor\.jbang\jbang-catalog.json
   [jbang] [0:286] Reading catalog from classpath:/jbang-catalog.json
   [jbang] [0:299] Using cached file C:\Users\koppor\.jbang\cache\urls\32d7e3573295d223682ea94b393cf5ad399c3fbc872fe793866bbb25cb937610\jbang-catalog.json for remote https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/jbang-catalog.json
   [jbang] [0:300] Reading catalog from https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/jbang-catalog.json
   [jbang] [0:302] Obtained catalog from https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/jbang-catalog.json
   hello-nbbrd = Sandbox - Bill of Materials
              (https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/hello_nbbrd.java)
   heylogs = Keep-a-changelog tool - Bill of Materials
           (https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/heylogs.java)
   sdmx-dl = Download SDMX data - CLI
           (https://raw.githubusercontent.com/nbbrd/jbang-catalog/master/sdmx_dl.java)
   ```
4. ```cmd
   > jbang --verbose heylogs
   Picked up JAVA_TOOL_OPTIONS: -Duser.language=en
   [jbang] [0:238] jbang version 0.110.1
   [jbang] [0:254] Resolving resource ref: heylogs
   [jbang] [0:261] Reading catalog from C:\Users\koppor\.jbang\jbang-catalog.json
   [jbang] [0:289] Reading catalog from classpath:/jbang-catalog.json
   [jbang] [0:291] Retry using cache-evict: heylogs
   [jbang] [0:294] [ERROR] Script or alias could not be found or read: 'heylogs'
   dev.jbang.cli.ExitException: Script or alias could not be found or read: 'heylogs'
         at dev.jbang.source.ProjectBuilder.resolveChecked(ProjectBuilder.java:244)
         at dev.jbang.source.ProjectBuilder.build(ProjectBuilder.java:218)
         at dev.jbang.cli.Run.doCall(Run.java:73)
         at dev.jbang.cli.BaseCommand.call(BaseCommand.java:145)
         at dev.jbang.cli.BaseCommand.call(BaseCommand.java:21)
         at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
         at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
         at dev.jbang.cli.JBang$3.handle(JBang.java:148)
         at dev.jbang.cli.JBang$3.handle(JBang.java:143)
         at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
   [jbang] [0:299] If you believe this a bug in jbang, open an issue at https://github.com/jbangdev/jbang/issues
   ```

**Expected behavior**

`heylogs` executed

Alternatively, a hint which command to execute to make the alias **really** available.

**JBang version**

```cmd
> jbang version --verbose
Picked up JAVA_TOOL_OPTIONS: -Duser.language=en
[jbang] [0:238] jbang version 0.110.1
Cache: C:\Users\koppor\.jbang\cache
Config: C:\Users\koppor\.jbang
Repository:C:\Users\koppor\.m2\repository
0.110.1
```

quintesse:
fix: properly listing the catalog names (fixes #1678)
feat: added --force options to catalog add commands (related to the above)
feat: added import feature to catalog refs

The last feature is something that we discussed (not sure if there's an issue for it, couldn't find it): an option that imports the items defined in a catalog and makes them available to the user without having to specify the full name, eg. using `@somecatalog`.

Right now this only works for local catalogs, so if you add a catalog reference with `jbang catalog add --import maxandersen` it will make all aliases, templates and catalog refs in that catalog available to the user without having to specify `@maxandersen`.

NB: The reason for the local catalog limitation (for now?) is to prevent a cascade. Otherwise you could import a catalog that references a bunch of other catalogs and imports them all and suddenly you have a gazillion aliases and templates available. But perhaps this is a restriction that could be relaxed in the future.

Btw, with this feature you could almost implement something like the jbanghub with very little changes. If you import the Hub's catalog and if it would have a list of catalog references, like for example "sqlline", then we would have a simple way to manage "named catalogs". You could do `jbang foo@sqlline` and it would work.

### DIFF
--- a/src/main/java/dev/jbang/catalog/CatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/CatalogRef.java
@@ -1,11 +1,9 @@
 package dev.jbang.catalog;
 
 import static dev.jbang.cli.BaseCommand.EXIT_INVALID_INPUT;
-import static dev.jbang.cli.BaseCommand.EXIT_UNEXPECTED_STATE;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Optional;
 
 import com.google.gson.annotations.SerializedName;
@@ -28,38 +26,18 @@ public class CatalogRef extends CatalogItem {
 		this.importItems = importItems;
 	}
 
-	/**
-	 * Creates a CatalogRef for the given catalog
-	 *
-	 * @param catalogRef File path, full URL or implicit Catalog reference to a
-	 *                   Catalog.
-	 * @return A CatalogRef object
-	 */
-	public static CatalogRef createByRefOrImplicit(String catalogRef) {
-		if (Util.isAbsoluteRef(catalogRef) || Files.isRegularFile(Paths.get(catalogRef))) {
-			Catalog cat = Catalog.getByRef(catalogRef);
-			return new CatalogRef(catalogRef, cat.description, null, cat);
-		} else {
-			Optional<String> url = ImplicitCatalogRef.resolveImplicitCatalogUrl(catalogRef);
-			if (!url.isPresent()) {
-				throw new ExitException(EXIT_UNEXPECTED_STATE,
-						"Unable to locate catalog: " + catalogRef);
-			}
-			Catalog cat = Catalog.getByRef(url.get());
-			return new CatalogRef(url.get(), cat.description, null, cat);
-		}
-	}
-
 	public static CatalogRef get(String catalogRefString) {
 		String[] parts = catalogRefString.split("@", 2);
 		if (parts[0].isEmpty()) {
 			throw new RuntimeException("Invalid catalog ref name '" + catalogRefString + "'");
 		}
-		CatalogRef catalogRef;
+		CatalogRef catalogRef = null;
 		if (parts.length == 1) {
-			catalogRef = getLocal(catalogRefString);
+			if (!Util.isRemoteRef(catalogRefString)) {
+				catalogRef = getLocal(catalogRefString);
+			}
 			if (catalogRef == null) {
-				catalogRef = getImplicit(catalogRefString);
+				catalogRef = getRemote(catalogRefString);
 			}
 		} else {
 			if (parts[1].isEmpty()) {
@@ -82,21 +60,31 @@ public class CatalogRef extends CatalogItem {
 				p = p.resolve(Catalog.JBANG_CATALOG_JSON);
 			}
 			if (Files.isRegularFile(p)) {
-				catalogRef = createByRefOrImplicit(p.toString());
+				Catalog cat = Catalog.getByRef(p.toString());
+				catalogRef = new CatalogRef(p.toString(), cat.description, null, cat);
 			}
 		}
 		return catalogRef;
 	}
 
-	private static CatalogRef getImplicit(String catalogRefString) {
+	private static CatalogRef getRemote(String catalogRefString) {
 		CatalogRef catalogRef = null;
-		Util.verboseMsg("Local catalog '" + catalogRefString + "' not found, trying implicit catalogs...");
-		Optional<String> url = ImplicitCatalogRef.resolveImplicitCatalogUrl(catalogRefString);
-		if (url.isPresent() && !url.get().equals(catalogRefString)) {
-			// Store the resolved URL for future reference
-			Catalog implicitCatalog = Catalog.getByRef(url.get());
-			catalogRef = CatalogUtil.addCatalogRef(Settings.getUserImplicitCatalogFile(), catalogRefString,
-					url.get(), implicitCatalog.description, null);
+		Optional<String> url;
+		if (Util.isRemoteRef(catalogRefString) && Util.isURL(catalogRefString)) {
+			url = Optional.of(catalogRefString);
+		} else {
+			Util.verboseMsg("Local catalog '" + catalogRefString + "' not found, trying implicit catalogs...");
+			url = ImplicitCatalogRef.resolveImplicitCatalogUrl(catalogRefString);
+		}
+		if (url.isPresent()) {
+			if (url.get().equals(catalogRefString)) {
+				catalogRef = new CatalogRef(url.get(), null, null, null);
+			} else {
+				// Store the resolved URL for future reference
+				Catalog implicitCatalog = Catalog.getByRef(url.get());
+				catalogRef = CatalogUtil.addCatalogRef(Settings.getUserImplicitCatalogFile(), catalogRefString,
+						url.get(), implicitCatalog.description, null);
+			}
 		}
 		return catalogRef;
 	}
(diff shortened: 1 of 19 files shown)


maxandersen:
functionallity wise looks good. 2 questions:

1) not sure I follow what "only local" means ...got an example of where using remote vs local?

2) the imported catalog still "pollutes" my default catalog names available. I didn't expect that. i.e. I put this in a jbang-catalog:
```
"catalogs": {
    "jbanghubby": {
      "catalog-ref": "https://github.com/jbanghub/jbang-catalog/blob/HEAD/jbang-catalog.json",
      "import": true
    }
  },
```

now `jbang mima` works - great, but `jbang mima@jbanghubby` works too... meaning that added catalogs wether implicit or explicit can "rewire" what catalogs I have on my system...that is not good from a security and predictability wise.

i.e. lets say someone adds this to their public catalog:

```
"catalogs": {
    "jbangdev": {
      "catalog-ref": "https://github.com/jbanghub/jbang-catalog/blob/HEAD/jbang-catalog.json",
      "import": true
    }
  },
```

now `jbang env@jbangdev` is failing but imagine i add a env command that just did some random operation.

This side-effect is not caused by this change but. a problem we have already today and I honestly think we need to fix/break that behavior - at least by default.

quintesse:
Sure:

1. it only works for catalogs found on your file system, it doesn't work for remote catalogs. Or in other words you _must_ at some point have used an `--import` flag while adding a catalog manually. It's not that you added a catalog and then they decide to add an "import: true" and suddenly you get a whole bunch of extra stuff. (On the other hand, if you _did_ use `--import` on an untrusted catalog they could decide to add a whole bunch of stuff you don't want).
2. "now jbang mima works - great, but jbang mima@jbanghubby works too..", sure, the catalog still has an explicit name, nothing you can do about that. The `--import` just means that you can now _also_ use all the items in the catalog without the fully qualified name.

The overriding... well, yes, that could probably happen. I can think of 2 solutions: make jbangedv/jbanghub special (not my favorite) or prioritize "toplevel" catalogs over "sub" catalogs. That way any catalogs that you explicitly added or used yourself will always have preference.

quintesse:
I've updated the code to prioritize local catalogs over imported catalog refs. Still need to fix an integration test though.

quintesse:
@maxandersen this is now in a testable state again. It won't show aliases/templates recursively anymore. And you can now also do things like `jbang catalog add microsoft@jbanghub` (you could do `jbang catalog add --import jbanghub` but then you'd get _all_ catalog refs and perhaps you're only interested in the MS one).

While creating this PR I started wondering if we should actually take into account the catalogs that get added to `implicit-catalog.json`. Right now whenever you do anything like `jbang script@somecatalog` and `somecatalog` is some implicit catalog reference we store it in that file. But what happens is that immediately all the aliases and templates (and catalog refs) in that file start showing up in the `alias list` and `template list` commands. And I've started wondering if that's actually a good idea. It's at least surprising, I think.

maxandersen:
Totally agree we shouldn't list from implicit catalogs by default.

maxandersen:
`j! catalog add --import jbanghub` gives me `Unknown option: '--import'` ...

src/main/java/dev/jbang/catalog/CatalogRef.java
```diff
@@ -18,11 +18,14 @@ public class CatalogRef extends CatalogItem {
 	@SerializedName(value = "catalog-ref", alternate = { "catalogRef" })
 	public final String catalogRef;
 	public final String description;
+	@SerializedName(value = "import")
+	public final Boolean importItems;
```
maxandersen, line 22: doe that need to be a tri-state Boolean?

quintesse:
> not sure if this is expecte:
...
> result in duplicated catalog list - one from the first one then the other set.
> 
> `jbang catalog remove jbanghub` remove them all.

Could you try and see if only running `jbang catalog add --import --force jbanghub` gives you the same result? Because then it might be expected. Because the `add` will import the jbanghub "alias", import or not. The import just has the _added_ effect of making its contents available locally as well. That would also explain why "both" disappear after a remove.

> I have ~14 jbang-catalog.json files in the cache.
>
> I'm assuming that is also cause of why jbang catalog list is slow.
>
> Is that expected?

I'll look into it because we definitely don't want that, no.

> does that need to be a tri-state Boolean?

It doesn't _need_ to be, no. I simply tend to like it because it gives us a way to check if the value was actually added by the user or not. But if you prefer I can change it to a simple `boolean`.

src/main/java/dev/jbang/catalog/CatalogRef.java
maxandersen: if good for deciding if set by user or not its all good.

maxandersen:
> Could you try and see if only running `jbang catalog add --import --force jbanghub` gives you the same result?

it does ...which seems wrong imo. the duplication is "noise" compared to what --import is supposed to do (make the aliases available)

quintesse:
@maxandersen this is now in a testable state again. Let me know what you think.

maxandersen:
jbang catalog remove karatelabs seem to update the cached version so it stays removed until clear cache.

maxandersen:
@quintesse areyø ypu able to look Into this one or should I just make it error if it tried to edit file in cache?

quintesse:
> @quintesse areyø ypu able to look Into this one or should I just make it error if it tried to edit file in cache?

Sorry, got side-tracked, I will fix this first thing tomorrow!

quintesse:
Ok, fixed

src/main/java/dev/jbang/catalog/CatalogUtil.java
```diff
@@ -65,7 +79,11 @@ public static Alias addAlias(Path catalogFile, String name, Alias alias) {
 	public static void removeNearestAlias(String name) {
 		Catalog catalog = Alias.findNearestCatalogWithAlias(Util.getCwd(), name);
 		if (catalog != null) {
-			removeAlias(catalog, name);
+			if (catalog.catalogRef.isURL() && Util.isRemoteRef(catalog.catalogRef.getOriginalResource())) {
+				Util.warnMsg("Unable to remove alias because it is imported from a remote catalog");
```
maxandersen, line 83:
```suggestion
				Util.warnMsg("Unable to remove alias " + name + " because it is imported from a remote catalog");
```

maxandersen (changes requested):
looks good but should be okey to list the name in the exception should it not ?

src/main/java/dev/jbang/catalog/CatalogUtil.java
quintesse: Done

PR #1756, opened 2024-02-22
feat: extended syntax for catalog references

### LINKED ISSUE #1755: Allow `catalogref@catalog` syntax in list commands
The idea is that you can do things like `jbang alias list alibaba@jbanghub`

quintesse:
Fixes #1755


maxandersen:
seems to work as expected :)

PR #1770, opened 2024-03-13
feat: added jbanghub to builtin catalog

quintesse:
This is the simplest solution where we add the jbanghub catalog ref to the "builtin" catalog. It does mean however that a user could never remove it (they could override it, for example to make it non-importing, but they could not remove it).

### DIFF
--- a/src/main/resources/jbang-catalog.json
+++ b/src/main/resources/jbang-catalog.json
@@ -1,4 +1,11 @@
 {
+  "catalogs": {
+    "jbanghub": {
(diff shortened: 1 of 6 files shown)


src/main/resources/jbang-catalog.json
```diff
@@ -1,4 +1,11 @@
 {
+  "catalogs": {
+    "jbanghub": {
+      "catalog-ref": "https://raw.githubusercontent.com/jbanghub/jbang-catalog/main/jbang-catalog.json",
+      "description": "JBangHub - Unleashing the Java community",
```
quintesse, line 5: @maxandersen we probably need a different description here

maxandersen (approved):

PR #1771, opened 2024-03-13
fix: fixes for handling more 3xx codes and not asking for trust twice

quintesse:


maxandersen (approved):

src/main/java/dev/jbang/net/TrustedSources.java
```diff
@@ -193,6 +204,24 @@ protected String getJSon(Collection<String> rules) {
 		return trustedsources;
 	}
 
+	public void addTemporary(String trust) {
+		Util.infoMsg("Adding temporary " + trust);
```
maxandersen, line 208: this shows up weird..how is this different from "trust once" option ?

maxandersen (changes requested):
the "adding temporary url" message is not great - need to find better way for that to be clearer what is being added temporarily and for how long.

src/main/java/dev/jbang/net/TrustedSources.java
quintesse: This _is_ the "trust once" option, thing is that the "trust once" was literally that, it would only trust once. So if you have some code that tries to access the same URL twice, for example when retrying, then you're actually accessing the same URL multiple times... which would result in the trust question to be asked multiple times for the same URL!

This is because there was no "memory" for the "trust once" option. So with this change we will keep track of all the URLs that the user answered "1" to for the duration of the JBang execution, they will be forgotten once JBang exits.

maxandersen: Ok - then let's use the same terminology. Something like:

"Trusting for this one run: URL" or similar. 

Or we rename the question to be Trusting temporarily - or Trussting for run or something similar. 

Just to reduce confusion.

quintesse: Will do 👍

quintesse:
@maxandersen how about these new messages?

maxandersen:
Good ones!

maxandersen (approved):

PR #1772, opened 2024-03-18
feat: add riscv64 support

sxa:
Title says it all. Now that Temurin has a GA JDK21 and ea versions on other openjdk release lines this will add support for Linux/riscv64.

Related: I was tempted to modify the default `JBANG_DEFAULT_JAVA_VERSION` for this platform to 21 instead of 17 in https://github.com/jbangdev/jbang/blob/826755c427070d2b69e88256a164658160bbedc6/src/main/scripts/jbang#L10C45-L10C46

This is because 21 is currently the only "GA" level at Adoptium. The 17 that is pulled down by default to bootstrap the process is an early access release. I'll leave it to reviewers to decide if that is a concern but I may raise a separate issue to consider bumping it to 21. I'll note that the obvious potential concern with doing that is that there are platforms (arm32) which is not available from Temurin in a GA JDK21 version. All other platforms should now be available.

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -417,7 +417,7 @@ public class Util {
 	}
 
 	public enum Arch {
-		x32, x64, aarch64, arm64, ppc64, ppc64le, s390x, unknown
+		x32, x64, aarch64, arm64, ppc64, ppc64le, s390x, riscv64, unknown
 	}
 
 	public enum Shell {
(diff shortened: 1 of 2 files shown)


maxandersen:
We are considering raising the default to download to be java 21 but yeah do open separate issue on it. 

This fix looks fine to me but something is causing break on windows. @quintesse you able to spot something?

sxa:
> This fix looks fine to me but something is causing break on windows. @quintesse you able to spot something?

Windows checks LGTM now but I fixed a syntax error in the shell script.

quintesse:
> This fix looks fine to me but something is causing break on windows. @quintesse you able to spot something?

Weird, perhaps it was a temporary GH problem? Because this change shouldn't affect Windows at all and everything's green now.

maxandersen (approved):

PR #1783, opened 2024-04-02
Make 'app list' only return executables

foreverigor:
Currently `jbang app list` prints out all the files in the bin directory. This PR fixes this by filtering the files by the executable flag (scripts in PATH are picked up by the shell only if they're executable anyways)

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -322,6 +322,7 @@ class AppList extends BaseCommand {
 	private static List<AppOut> listCommandFiles() {
 		try (Stream<Path> files = Files.list(Settings.getConfigBinDir())) {
 			return files
+						.filter(Files::isExecutable)
 						.sorted()
 						.map(AppOut::new)
 						.filter(distinctByKey(AppOut::getName))


foreverigor:
Do I need to change the commit name? If required I'll do that

quintesse:
> Do I need to change the commit name? If required I'll do that

If you could, that would be awesome :-)

quintesse (approved):
Seems like a simple and possibly useful enough change.

foreverigor:
> Btw, what is your specific use-case that you need this fix? _Theoretically_ there shouldn't be any non-executable files in that directory, right?

Theoretically, yes. But in practice any random file will show up, like macOS .DS_Store files (that's how I stumbled on this). Also I could see someone pointing this directory to a different, 'unclean' one where also any file would be seen in the output

quintesse:
Btw, for future improvement we should probably change this to use `Util.isExecutable()` so it will work for Windows too.

foreverigor:
> `Util.isExecutable()` so it will work for Windows too.

TIL, I wasn't aware of this distinction

quintesse:
> TIL, I wasn't aware of this distinction

Yeah, I should have mentioned it, but I forgot. No worries.

Things is that on Windows `Files.isExecutable()` always returns `true` because Windows doesn't really have a flag or anything that determines if something is executable or not, there's just a list of extensions that can be considered executable but the JDK function doesn't deal with that. So we wrote our own that looks at the extension on Windows and uses `File.isExecutable` for everything else.

PR #1794, opened 2024-04-05
chore: made `version --verbose` more verbose

quintesse:
Added a couple of more informational lines that can help us debug issues that people might have with JBang.


maxandersen (approved):

maxandersen:
@quintesse seems like integrations tests is broken because of this. probably more content than expected?

quintesse:
> @quintesse seems like integrations tests is broken because of this. probably more content than expected?

No, the error seems to be:

```
Error:  [ERROR] Could not resolve dependencies: Failed to collect dependencies at com.github.intuit.karate:karate-netty:jar:e2882c4
```

Let's re-run and see if that fixes things....

maxandersen:
Ohh.. Missed that:/

quintesse:
It built correctly this time

maxandersen (approved):

PR #1797, opened 2024-04-13
chore: update deps

cstamas:
And add gradle-versions-plugin (w/ tweaks as OOTB is broken).


maxandersen:
What is that versions plugin for?

cstamas:
Adds this task that tells you what you can update (like m-versions-p):
```
./gradlew dependencyUpdates
```

PR #1800, opened 2024-04-29
feat: introduce type 'fatjar'

### LINKED ISSUE #1778: Catalog script-ref for uber artifacts
I have a CLI artifact that is uber of main one (that is a maven-plugin), with classifier "cli". Naturally, cli JAR is self contained. I want to have catalog that tells jbang "I need this artifact ONLY" (like resolve file vs resolve transitively file). The catalog entry is https://github.com/maveniverse/jbang-catalog/blob/main/jbang-catalog.json#L20

I think this is pretty common, that if artifact pointed at is "sub artifact" (have classifier), Maven would pick up POM, but IMO for jbang "use case" this is not needed. JBang could be smart, and resolve transitively if no classifier present, or resolver artifact along if classifier present, as this seems to me as pretty common situation (main JAR w/ CLI uber jar).

Currently toolbox works, but I have warnings as slf4j-simple is twice present on classpath (once in uber CLI and once as resolved dependency):
```
[cstamas@angeleyes toolbox (main)]$ jbang toolbox@maveniverse
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/cstamas/.m2/repository-oss/eu/maveniverse/maven/plugins/toolbox/0.1.0/toolbox-0.1.0-cli.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/cstamas/.m2/repository-oss/org/slf4j/slf4j-simple/1.7.36/slf4j-simple-1.7.36.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.SimpleLoggerFactory]
Toolbox 0.1.0 (MIMA Runtime 'standalone-static' version 2.4.8)
=======
          Maven version 3.9.6
...
```

Somehow to express that i want 'resolve file' and not 'resolve transitively' operation here.

**Describe alternatives you've considered**
I tried to use direct HTTPS URL to Central, like https://repo.maven.apache.org/maven2/eu/maveniverse/maven/plugins/toolbox/0.1.0/toolbox-0.1.0-cli.jar for `scriprt-ref`, and it did work, kinda, but asked me to "trust" Maven Central, which is kinda nonsense, as when script-ref is G:A:V it resolves from there without asking for "trust".

**Additional context**
...

cstamas:
A new ArtifactType that is also present in Maven4. The 'fatjar' type denotes a Java JAR file that is
"self sufficient", contains all the dependencies. This makes resolver stop resolving. Usable when for example there is an `uber` or `cli` classified artifact, as in other cases resolver would pick up POM and (probably wrongfully) resolve it.

Fixes #1778

Alows something like this
```
$ jbang eu.maveniverse.maven.plugins:toolbox:0.1.9:cli@fatjar
```

Note: Maven4 introduced this type, but Maven3 has no idea about it. JBang uses new feature of MIMA that allows one to define own ArtifactTypes. See https://github.com/apache/maven/pull/1459

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -103,8 +103,8 @@ dependencies {
 	implementation "org.slf4j:jcl-over-slf4j:1.7.30"
 	implementation "org.jboss:jandex:2.2.3.Final"
 
-	implementation "eu.maveniverse.maven.mima:context:2.4.4"
-	runtimeOnly    "eu.maveniverse.maven.mima.runtime:standalone-static:2.4.4"
+	implementation "eu.maveniverse.maven.mima:context:2.4.12"
+	runtimeOnly    "eu.maveniverse.maven.mima.runtime:standalone-static:2.4.12"
 
 	testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:5.10.1"
 	testImplementation "org.junit.jupiter:junit-jupiter:5.10.1"
@@ -235,6 +235,7 @@ compileJava9Java {
 shadowJar {
 	minimize() {
 		//exclude(dependency('org.slf4j:slf4j-api:.*'))
+		exclude(dependency('eu.maveniverse.maven.mima:context:.*'))
 		exclude(dependency('eu.maveniverse.maven.mima.runtime:standalone-static:.*'))
 		exclude(dependency('org.slf4j:jcl-over-slf4j:.*'))
 		exclude(dependency('org.slf4j:slf4j-nop:.*'))
--- a/docs/modules/ROOT/pages/dependencies.adoc
+++ b/docs/modules/ROOT/pages/dependencies.adoc
@@ -78,6 +78,18 @@ is the equivalent of the following Maven dependency declaration:
 ```
 ====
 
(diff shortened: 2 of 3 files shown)


quintesse:
Nice!

maxandersen:
nice, it does not really fix #1778 does it ? users will have to have published the artifact with that qualifer would they not ?

maxandersen:
ah no, its a "magical" type modifier. 

I guess if maven4 supports it we can too.

dsyer:
Might be good to have a test? There are "fat jars" already in use in existing tests (e.g. `TestRun.testHelloWorldGAVWithAMainViaAlias`) that could be modified?

maxandersen:
I'm really not fan of the 'fatjar' name but since maven4 went with that and had multiple issues popup that this solves I'll cave ;) 

I've pushed a commit that documents the feature as experimental.

PR #1804, opened 2024-05-28
feat: we can now create junctions on Windows

### LINKED ISSUE #1793: jbang jdk default <version> fails on Windows
**Describe the bug**
Follow-up to https://github.com/jbangdev/jbang/issues/1574 by @quintesse 

This is when using "Git Bash" - and particularly the msys bash inside of it. 

This is what I see:

```
Ben@t480-ben dotfiles [master]$ jbang jdk default 21
[jbang] Creation of symbolic link failed C:\Users\Ben\.jbang\currentjdk -> C:\Users\Ben\.jbang\cache\jdks\21
[jbang] Now try creating a hard link instead of symbolic.
[jbang] Creation of hard link failed C:\Users\Ben\.jbang\currentjdk -> C:\Users\Ben\.jbang\cache\jdks\21
[jbang] [ERROR] Failed to create link C:\Users\Ben\.jbang\currentjdk -> C:\Users\Ben\.jbang\cache\jdks\21
```

**To Reproduce**
I'm running bash on Windows 10. `uname -a` reports `MSYS_NT-10.0-19045 t480-ben 3.4.7-25de8b84.x86_64 2023-08-28 21:32 UTC x86_64 Msys`

I reproduced this in Powershell IIRC but my main way of working in Windows is inside cygwin / msys.

I execute via the Windows Terminal - I configure a profile like this after installing Git via `winget install Git.Git`. I'm using `/Git/usr/bin/bash.exe` which has the cygwin dll included, but not sure that matters.

```
            {
                "commandline": "%PROGRAMFILES%/Git/usr/bin/bash.exe -i -l",
                "guid": "{00000000-0000-0000-ba54-000000000002}",
                "icon": "%PROGRAMFILES%/Git/mingw64/share/git/git-for-windows.ico",
                "name": "Bash",
                "startingDirectory": "%USERPROFILE%"
            },
````

```
scoop install jbang
jbang jdk install 21
jbang jdk default 21
```

**Expected behavior**
jbang jdk default 21 should succeed

**JBang version**
```
$ jbang version --verbose
[jbang] [0:425] jbang version 0.116.0
Cache: C:\Users\Ben\.jbang\cache
Config: C:\Users\Ben\.jbang
Repository: C:\Users\Ben\.m2\repository
0.116.0
```

**Additional context**
I don't think requiring Developer Mode is good - many people developing in Windows are doing it for security reasons and it's hard convince a security department to enable Developer Mode.

As noted earlier in the linked issue, you can't hard link to directories. But there's another option - use junctions instead. fnm did that recently https://github.com/Schniz/fnm/pull/519 - see also https://github.com/kubernetes-sigs/krew/issues/843

An example of creating a junction from my initial dotfiles setup script: `powershell.exe New-Item -ItemType Junction -Force -Path $(cygpath -m "$HOME/scoop/persist/vscode/data/user-data/User") -Target $(cygpath -m "$SCRIPT_DIR"/config/vscode/User)`

### PR DESCRIPTION
**Pull request by quintesse**
Fixes #1793

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,8 +18,11 @@ plugins {
 	id 'maven-publish'
 }
 
-//remove this to see all the missing tags/parameters.
-javadoc.options.addStringOption('Xdoclint:none', '-quiet')
+javadoc {
+	options.encoding = 'UTF-8'
+	//remove this to see all the missing tags/parameters.
+	options.addStringOption('Xdoclint:none', '-quiet')
+}
 
 repositories {
 	mavenCentral()
--- a/docs/modules/ROOT/pages/javaversions.adoc
+++ b/docs/modules/ROOT/pages/javaversions.adoc
@@ -55,10 +55,6 @@ You can change the default JDK by running:
 
 Running it without an argument will return the version of the JDK that is currently set as the default.
 
-NOTE: On Windows you might need elevated privileges to create symbolic links. If you don't have permissions then
-running the above command  will result in an error. To use it https://stackoverflow.com/a/24353758[enable symbolic links]
-for your user or run your shell/terminal as administrator to have this feature working.
-
 When you `uninstall` a JDK by running:
 
   jbang jdk uninstall 12
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -348,15 +348,15 @@ No one would want to do that (right?) but now you know.
 == Usage on Windows
 
 Some JBang commands need to create symbolic links when running on Windows.
-For example, this is required for Managing JDKs or editing the files with the `edit` command.
+For example, this is required for editing the files with the `edit` command.
 
 If you encounter issues on Windows related to the creation of symbolic links follow
 these instructions:
 
 1. From Windows 10 onwards you can turn on "Developer Mode", this will automatically
 enable the possibility to create symbolic links. Read here how to enable this mode:
-https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development[Enable your device for development]. On Windows 11 this might already
-be enabled by default.
+https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development[Enable your device for development].
+On Windows 11 this might already be enabled by default.
 
 2. If you're using a Java version equal to or newer than 13 then you're good to go.
 This Java version already works correctly. Make sure that JBang is actually using
@@ -369,4 +369,3 @@ is no  other option than setting the correct privileges for your user by enablin
 the `Create symbolic links` group policy setting. See the instruction on this page
 for more information on how to do this:
 https://superuser.com/a/105381[Permission to make symbolic links in Windows].
-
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: build.gradle [hunk 1]
```diff
@@ -233,13 +238,13 @@ compileJava9Java {
 }
 
 shadowJar {
-	minimize() {
+	/*minimize() {
```
maxandersen, line 241: why not minimize anymore?

quintesse: Oops, sorry, that was a test that shouldn't have been committed...

src/main/java/dev/jbang/util/WinJunction.java
```diff
@@ -0,0 +1,141 @@
+/*
+ * Copyright 2011 杨博 (Yang Bo)
+ * Translated from Scala to Java 2024 (Tako Schotanus)
```
maxandersen, line 3: where is this from?

src/main/java/dev/jbang/util/WinJunction.java
```diff
@@ -0,0 +1,141 @@
...
+public class WinJunction {
+
+	private static final int sizeOfReparseDataHead = 8;
+	private static final int sizeOfMountPointReparseHead = 8;
+	private static final int IO_REPARSE_TAG_MOUNT_POINT = 0xA0000003;
+	private static final int FSCTL_SET_REPARSE_POINT = 0x000900a4;
+
+	private static final String PathPrefix = "\\??\\";
+
+	public interface Kernel32 extends StdCallLibrary {
```
maxandersen, line 43: will this work in newer java versions too without warning/errors?

src/main/java/dev/jbang/util/WinJunction.java
quintesse: https://github.com/Atry/scala-junction/blob/master/src/main/scala/com/dongxiguo/junction/Junction.scala

src/main/java/dev/jbang/util/WinJunction.java
quintesse: Why wouldn't it? Is there something here I'm not seeing?

maxandersen: i might remember wrong but just have recollection that jna library is't fully supported across java 8 to Java 23+ but it might just be a rumour.

quintesse:
Updated PR with simplified code using `cmd /c mklink /j ....` to create junctions.

The `createLink()` will now also automatically distinguish between links to files and directories and use junctions for directories on Windows, while still using sym links for files (and possibly failing and explaining the user what to do to fix it).

jcrben:
@quintesse @maxandersen thoughts on merging this?

maxandersen:
@quintesse didnt we have some concern with this ? 

I can definitely give it a go again but feel like there was still a higher permission level needed still or was that fixed?

quintesse:
@maxandersen no, on Windows it will use an external command (that should exist on all modern Windows versions) to create a junction when dealing with links to folders. Only when we use links to files (as for the `edit` command) will we fall back to creating symbolic links which still need the same privileges as before. So worst case it does the same as what JBang does right now.

PR #1811, opened 2024-06-20
feat: export an additional env var pointing to the location of jbang

dsyer:
It's *really* hard (impossible?) to find the location of the script that launched the JVM from within jbang. It would be really useful to be able to do that so you can run jbang again with some logic to precompute the command line. This works really simply:

```
$ jbang --interactive
WARNING: Using incubator modules: jdk.incubator.vector, jdk.incubator.foreign
|  Welcome to JShell -- Version 17.0.7
|  For an introduction type: /help intro

jshell> System.getenv("JBANG")
$1 ==> "/home/dsyer/.sdkman/candidates/jbang/current/bin/jbang"
```

Edit: I don't know how to do this for Windows users. Anyone got an idea?

### DIFF
--- a/itests/shellenv.feature
+++ b/itests/shellenv.feature
@@ -1,0 +1,13 @@
+Feature: env variables
+
+Scenario: JBANG_RUNTIME_SHELL available
+When command('jbang env@jbangdev jbang')
+Then match err contains "^JBANG_RUNTIME_SHELL"
+
+Scenario: JBANG_STDIN_NOTTY available
+  When command('jbang env@jbangdev jbang')
+  Then match err contains "^JBANG_STDIN_NOTTY"
+
+Scenario: JBANG_LAUNCH_CMD available
+  When command('jbang env@jbangdev jbang')
+  Then match err contains "^JBANG_LAUNCH_CMD"
--- a/itests/shellenv.feature
+++ b/itests/shellenv.feature
@@ -1,13 +1,13 @@
 Feature: env variables
 
 Scenario: JBANG_RUNTIME_SHELL available
-When command('jbang env@jbangdev jbang')
-Then match err contains "^JBANG_RUNTIME_SHELL"
+When command('jbang env@jbangdev JBANG')
+Then match out contains "JBANG_RUNTIME_SHELL"
 
 Scenario: JBANG_STDIN_NOTTY available
-  When command('jbang env@jbangdev jbang')
-  Then match err contains "^JBANG_STDIN_NOTTY"
+  When command('jbang env@jbangdev JBANG')
+  Then match out contains "JBANG_STDIN_NOTTY"
 
 Scenario: JBANG_LAUNCH_CMD available
-  When command('jbang env@jbangdev jbang')
-  Then match err contains "^JBANG_LAUNCH_CMD"
+  When command('jbang env@jbangdev JBANG')
+  Then match out contains "JBANG_LAUNCH_CMD"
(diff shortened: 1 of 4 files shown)


maxandersen:
I'm not totally against exporting this variable assuming we find way it works on windows but Thus far we haven't exposed this as its not guaranteed to work.

i.e. the same app could have been run with pure java or native exc launched.

I would prefer some more internal name like _JBANG_LAUNCH_SCRIPT_ or similar.

dsyer:
I don't care about the name really. How about `JBANG_LAUNCHER`? I get that it's not guaranteed to work if the launcher isn't used, but at least it provides a way to launch `jbang` in that case. I can't think of anything that would work completely generically.

maxandersen:
okey, so I pushed code that should work on osx, linux and windows (need ci to verify the latter one).

 I also checked what other env vars we actually expose and added tests for those ...BUT I can't actually spot where we use them.. @quintesse can we remove those?

itests/shellenv.feature
```diff
@@ -0,0 +1,13 @@
+Feature: env variables
+
+Scenario: JBANG_RUNTIME_SHELL available
```
maxandersen, line 3: @quintesse  what is the use of this variable?

itests/shellenv.feature
```diff
@@ -0,0 +1,13 @@
...
+When command('jbang env@jbangdev jbang')
+Then match err contains "^JBANG_RUNTIME_SHELL"
+
+Scenario: JBANG_STDIN_NOTTY available
```
maxandersen, line 7: @quintesse  what is the use of this variable?

itests/shellenv.feature
```diff
@@ -0,0 +1,13 @@
...
+  When command('jbang env@jbangdev jbang')
+  Then match err contains "^JBANG_STDIN_NOTTY"
+
+Scenario: JBANG_LAUNCH_CMD available
```
maxandersen, line 11: I actually prefer if we call this `_JBANG_LAUNCH_CMD` but kept with `JBANG_` as we use that above already.

Reason beeing we actually have a few JBANG_* env var used for input - so wanted _JBANG to signal its not input just temporarily available.

itests/shellenv.feature
quintesse: The runtime scripts tell JBang what shell (eg. *nix sh/bash/etc or CMD or Powershell) it was started from. There's just no good way to otherwise detect what kind of command line we should be generating.

itests/shellenv.feature
quintesse: This tells JBang that we don't have a console/terminal connected to stdin. Java can't detect this by itself.

dsyer: I believe this is a pretty good approximation, in case you ever need it:

```java
boolean isAnsi = System.console()!=null && 
  System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("win");
```

itests/shellenv.feature
dsyer: I said I didn't care about the name, but I don't like starting env var names with non-alpha chars, so I vote for `JBANG_LAUNCH_CMD`.

itests/shellenv.feature
quintesse: @dsyer that would only work on *nix systems and we need it to work for Windows too.

maxandersen: But where in code are we using that env var?

quintesse: See `Util.haveConsole()` which is used by `ConsoleInput.get()` which is used by `Util.askInput()` to detect if we can ask the user for input on the console or if we should use a GUI.

maxandersen: kk

maxandersen:
@all-contributors please add @dsyer for code

PR #1824, opened 2024-08-02
feat: when no jdk installable show possible reasons/erros

maxandersen:
in https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/JBang.20failure/near/455708888 @gsmet had this happen:

```
Error:  [ERROR] JDK version is not available for installation: 21
Use 'jbang jdk list --available' to see a list of JDKs available for installation
Error: Process completed with exit code 2.
```

as it happens in github he can't really easily re-run with verbose to get info why failed.

This PR tries without having to reorganize the whole call chain and print out last possible exception for each provider.

### DIFF
--- a/src/main/java/dev/jbang/net/jdkproviders/JBangJdkProvider.java
+++ b/src/main/java/dev/jbang/net/jdkproviders/JBangJdkProvider.java
@@ -69,7 +69,8 @@ public class JBangJdkProvider extends BaseFoldersJdkProvider {
 			result.sort(Jdk::compareTo);
 			return Collections.unmodifiableList(result);
 		} catch (IOException e) {
-			Util.verboseMsg("Couldn't list available JDKs", e);
+			Util.errorMsg("Could not list available JDK's using foojay.io: " + e.getMessage(), e);
+			// Util.verboseMsg("Couldn't list available JDKs", e);
 		}
 		return Collections.emptyList();
 	}


quintesse:
I think you added a bit more to this PR than necessary :-)

maxandersen:
> I think you added a bit more to this PR than necessary :-)

whoops. fixed. i hope.

quintesse:
> whoops. fixed. i hope.

Yes!

Now for the PR itself, let me think about it, because you've just given the providers ephemeral state and that wasn't really the idea (configuration state, okay, anything else, hmmm 🤔 ).

src/main/java/dev/jbang/net/jdkproviders/JBangJdkProvider.java
```diff
@@ -69,6 +69,7 @@ public List<Jdk> listAvailable() {
 			result.sort(Jdk::compareTo);
 			return Collections.unmodifiableList(result);
 		} catch (IOException e) {
+			setLastError(e);
```
maxandersen, line 72: just print warn message always ? skip the above.

maxandersen:
updated this to just have jdk provider that uses foojay to print explicit message if errror happens. should suffice.

quintesse (approved):
LGTM

PR #1826, opened 2024-08-03
feat: linuxdistro jdk provider

### LINKED ISSUE #1825: add /usr/lib/jvm as a optional jdk location to search for preinstalled javass
from https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/JBang.20failure/near/455967437 @sanne felt sad his already installed jdks via fedora dnf was not being found.

fedora (and possible other linux distros) install jvm's into `/usr/lib/jvm` which has subdirs for both jre's and jdk's but i think our base path provider already ensures it only picks up jdks.

thus should be fairly easy adding a UsrLibJvm provider...other is `/etc/alternatives` but here its intermixed with tons of others so might affect performance. Maybe have a Linux or Fedora provider that looks in /usr/lib/jvm first - then /etc/alternatives ?

maxandersen:
adds `linuxdistro` provider which for now searches `/usr/lib/jvm` for jdks.

It is not enabled by default as it is not possible to safely determine the version and build of the jdk
present or needed to be installed if mising.

Meaning to get it user need to run `jbang config set jdkproviders all` or `jbang config set jdkproviders linuxdistro` if only want this.

You can see which providers are finding jdks by running `jbang jdk list --show-details`

Fixes #1825 

@quintesse I see that `jbang jdk list --show-details` list what looks like all dirs even though javac is not available. is that expected?

### DIFF
--- a/src/main/java/dev/jbang/net/JdkProvider.java
+++ b/src/main/java/dev/jbang/net/JdkProvider.java
@@ -121,6 +121,8 @@ public interface JdkProvider {
 
 	default String name() {
 		String nm = getClass().getSimpleName();
+		// TODO: this 11 is here assuming it ends in "JdkProvider" - dont make that
+		// broken assumption
 		return nm.substring(0, nm.length() - 11).toLowerCase();
 	}
 
--- a/src/main/java/dev/jbang/net/JdkProvider.java
+++ b/src/main/java/dev/jbang/net/JdkProvider.java
@@ -121,9 +121,11 @@ public interface JdkProvider {
 
 	default String name() {
 		String nm = getClass().getSimpleName();
-		// TODO: this 11 is here assuming it ends in "JdkProvider" - dont make that
-		// broken assumption
-		return nm.substring(0, nm.length() - 11).toLowerCase();
+		if (nm.endsWith("JdkProvider")) {
+			return nm.substring(0, nm.length() - 11).toLowerCase();
+		} else {
+			return nm.toLowerCase();
+		}
 	}
 
 	/**
--- a/src/main/java/dev/jbang/net/jdkproviders/LinuxDistroJdkProvider.java
+++ b/src/main/java/dev/jbang/net/jdkproviders/LinuxDistroJdkProvider.java
@@ -1,5 +1,6 @@
 package dev.jbang.net.jdkproviders;
 
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -36,5 +37,23 @@ public class LinuxDistroJdkProvider extends BaseFoldersJdkProvider {
 	@Override
 	public boolean canUse() {
 		return Files.isDirectory(JDKS_ROOT);
+	}
+
+	@Override
+	protected boolean acceptFolder(Path jdkFolder) {
+		return super.acceptFolder(jdkFolder) && !isSameFolderSymLink(jdkFolder);
+	}
+
+	// Returns true if a path is a symlink to an entry in the same folder
+	private boolean isSameFolderSymLink(Path jdkFolder) {
+		Path absFolder = jdkFolder.toAbsolutePath();
+		if (Files.isSymbolicLink(absFolder)) {
+			try {
+				Path realPath = absFolder.toRealPath();
+				return Files.isSameFile(absFolder.getParent(), realPath.getParent());
+			} catch (IOException e) {
+				/* ignore */ }
+		}
+		return false;
 	}
 }
--- a/src/main/java/dev/jbang/net/jdkproviders/LinuxDistroJdkProvider.java
+++ b/src/main/java/dev/jbang/net/jdkproviders/LinuxDistroJdkProvider.java
@@ -1,0 +1,40 @@
+package dev.jbang.net.jdkproviders;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * This JDK provider is intended to detects JDKs that have been installed in
+ * standard location of the users linux distro.
+ *
+ * For now just using `/usr/lib/jvm` as apparently fedora, debian, ubuntu and
+ * centos/rhel use it.
+ *
+ * If need different behavior per linux distro its intended this provider will
+ * adjust based on identified distro.
+ *
+ */
+public class LinuxDistroJdkProvider extends BaseFoldersJdkProvider {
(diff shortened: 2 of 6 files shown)


quintesse:
> It is not enabled by default as it is not possible to safely determine the version and build of the jdk
present

Why not? Is there no RELEASE file in those JDKs?

> or needed to be installed if mising.

Not sure what you mean by this?

> is that expected?

It shows what the providers tell it to show. Providers shouldn't return anything without doing sanity checks.

maxandersen:
> Why not? Is there no RELEASE file in those JDKs?

I mean, once these are enabled we no longer have a consistent java version when using jbang by default.

Today its always, honor your configured path, otherwise fallback to .jbang managed or foojay fetched jdks.

if we enable the others like sdkman/linux alternatives you will get whatever order of providers are enabled and used on your system and when jbang perform installs it might not be the one selected next time you run - even in the same shell.

unless jbang managed ones trumph everything else I reckon.

quintesse:
> I mean, once these are enabled we no longer have a consistent java version when using jbang by default.

Ah yes, indeed. Although I'd say that given what you mention next:
 
> Today its always, honor your configured path, otherwise fallback to .jbang managed or foojay fetched jdks.

the fact that we first use what's on the path already means that we can never be sure what JDK we're running.

> if we enable the others like sdkman/linux alternatives you will get whatever order of providers are enabled and used on your system and when jbang perform installs it might not be the one selected next time you run - even in the same shell.
> 
> unless jbang managed ones trumph everything else I reckon.

jbang managed ones do indeed have priority (after what's on your PATH that is, like mentioned above)

quintesse:
Btw, talking about native providers... you run MacOS, can't you make a provider for JDKs installed on Macs? :-)

src/main/java/dev/jbang/net/JdkProvider.java
```diff
@@ -121,6 +121,8 @@ default Jdk createJdk(@Nonnull String id, @Nullable Path home, @Nonnull String v
 
 	default String name() {
 		String nm = getClass().getSimpleName();
+		// TODO: this 11 is here assuming it ends in "JdkProvider" - dont make that
```
quintesse, line 124: Why not? It's a reasonable default implementation, isn't it? If your implementation doesn't follow that scheme you can simply override this method and return the value that you desire.

quintesse:
> > is that expected?
> 
> It shows what the providers tell it to show. Providers shouldn't return anything without doing sanity checks.

I see now what you were trying to do. You're using the `BaseFolderJdkProvider` which assumes that the only folders that it will encounter are always valid JDKs. But we could add a simple filter for `./RELEASE`.

maxandersen:
> Btw, talking about native providers... you run MacOS, can't you make a provider for JDKs installed on Macs? :-)

macs doesn't install native jdks anymore do they ?

src/main/java/dev/jbang/net/JdkProvider.java
maxandersen: I had a name called LinxDistroProvider and it showed up as 'linuxdist' which confused me - will update one that is a bit more explicit.

maxandersen:
> the fact that we first use what's on the path already means that we can never be sure what JDK we're running.

we can be sure the user gets the JDK that is defined in his terminal/session like any other common cli command.

maxandersen:
> But we could add a simple filter for `./RELEASE`.

not sure where to best add it as some of the code already does use it for naming ...but other seems to use filenames. lets do that in another pr.

quintesse:
> > the fact that we first use what's on the path already means that we can never be sure what JDK we're running.
> 
> we can be sure the user gets the JDK that is defined in his terminal/session like any other common cli command.

What I mean is that _we_ cannot make assumptions about the JDK we're running, either we're running a JBang-managed JDK or ... something else that we know very little about. So what's so different about running another JDK that's available on the system (as long as we can verify it's the correct versions and it's a JDK)? In what way are managed JBang JDKs "better" than what the user has on their system? (I'm of course assuming modern JDKs from any vendor are fully functional)

quintesse:
> macs doesn't install native jdks anymore do they ?

Dunno, I don't use Macs so I have no idea how people install JDKs there :-)

src/main/java/dev/jbang/net/JdkProvider.java
```diff
@@ -121,9 +121,11 @@ default Jdk createJdk(@Nonnull String id, @Nullable Path home, @Nonnull String v
 
 	default String name() {
 		String nm = getClass().getSimpleName();
-		// TODO: this 11 is here assuming it ends in "JdkProvider" - dont make that
-		// broken assumption
-		return nm.substring(0, nm.length() - 11).toLowerCase();
+		if(nm.endsWith("JdkProvider")) {
+			return nm.substring(0, nm.length() - 11).toLowerCase();
+		} else {
+			return nm;
```
quintesse, line 127: Should probably do .toLowerCase() here

quintesse (approved):
It will work as-is. The only thing we could still do is the filtering for the existence of the RELEASE file. But we could also leave that for a future PR if there aren't any real issues.

quintesse (changes requested):
Actually, I've just tested this myself and I do see the problem now. There's a whole bunch of duplication going on. I think what we should do is add a `protected boolean acceptFolder(Path jdkFolder) { return true; }` to the `BaseFoldersJdkProvider`  which subclasses can override. In the case of the LinuxDistro provider I'd then check that the folders are _not_ symlinks, and a `bin/javac` exists (in my case all folders are JREs)

NB: we could make this test the default implementation as well, because realistically _all_ JDKs should be able to pass these tests.

maxandersen:
@quintesse is that not done better in separate PR ?

quintesse:
> @quintesse is that not done better in separate PR ?

Well , right now this PR would add a dozen unusable entries on my system. That's not really something we want, right?

maxandersen:
@quintesse i don't see them so if you can push update to this pr with what you wanna see fixed lets do that.

quintesse:
Will do, but on PTO now, Have to see when I have a bit of time :-)

src/main/java/dev/jbang/net/jdkproviders/LinuxDistroJdkProvider.java
```diff
@@ -0,0 +1,59 @@
...
+	@Nonnull
+	@Override
+	protected Path getJdksRoot() {
+		return JDKS_ROOT;
+	}
+
+	@Nullable
+	@Override
+	protected String jdkId(String name) {
+		return name + "-nixdistro";
```
quintesse, line 34: Is there a way you chose a different name here than the name of the provider itself? ("linuxdistro" vs "nixdistro")

PR #1842, opened 2024-09-27
fix: use main-SNAPSHOT as best-effort fallback for jitpack

### LINKED ISSUE #1817: Incorrect locator generated for `//DEPS` links to GitHub repos
https://www.jbang.dev/documentation/guide/latest/dependencies.html#using-links-to-git-sources

says that `https://github.com/jbangdev/jbang` is converted to the locator `com.github.jbangdev:jbang:HEAD-SNAPSHOT`. I'm seeing it being converted to a `jbang:-SNAPSHOT` locator, which can't resolve.

I wondered about just dropping the line in the docs, but it's possibly almost as easy to correct the locator generation to do what the docs say. 

For example, 

```
//DEPS https://github.com/holly-cummins/github-api/
```

gives 

```
Run jbangdev/jbang-action@v0.111.0
...
[jbang]    com.github.holly-cummins:github-api:-SNAPSHOT
[jbang]    info.picocli:picocli:4.2.0
Error:  [ERROR] Could not resolve dependencies: Could not find artifact com.github.holly-cummins:github-api:jar:-SNAPSHOT 
...
```

I see the same thing locally with jbang 0.92.2

Tree references with an explicit commit, like `//DEPS https://github.com/holly-cummins/github-api/tree/gg287262` do generate a correct locator.  If I add a `tree` reference to head, it works, as long as `SNAPSHOT` is mentioned explicitly: 

```
//DEPS https://github.com/holly-cummins/github-api/tree/main#:SNAPSHOT
```

Without the `SNAPSHOT`, it doesn't work:

```
//DEPS https://github.com/holly-cummins/github-api/tree/main
```

gives a locator of `com.github.holly-cummins:github-api:jar:main`, which isn't valid. (That's maybe ok, because the docs don't say it should work, I was just trying it as a workaround.)

maxandersen:
https://github.com/jbangdev/jbang/issues/1817 made me bite the bullet and have github urls default to use `main-SNAPSHOT` so it wont fail on new stuff. The whole reason I hesitated was that jitpack in past would accept HEAD-SNAPSHOT to pickup whatever is the default name on github (main or master). But that stop working some years ago - back then I hoped it was just a glitch..but since not fixed i'm now at least making it work for new things.

also updated the docs and added more tests as things were slightly incorrect.

Fixes #1817

### DIFF
--- a/docs/modules/ROOT/pages/dependencies.adoc
+++ b/docs/modules/ROOT/pages/dependencies.adoc
@@ -112,11 +112,14 @@ name of the module by appending `#name-of-module` to the URL.
 And finally if the link you provide is to a specific branch of the project then you need to append
 `#:SNAPSHOT` to the URL. (If you have both a branch and a module name then use `#name-of-module:SNAPSHOT`)
 
+Note: if no branch specified `main-SNAPSHOT` will be used and if you list `main` or `master` they will both
+be treated as if you want the snapshot version.
+
 .Examples of links and their resulting locator:
 |===
 |Link | Locator
 |https://github.com/jbangdev/jbang
-|com.github.jbangdev:jbang:HEAD-SNAPSHOT
+|com.github.jbangdev:jbang:main-SNAPSHOT
 
 |https://github.com/jbangdev/jbang/tree/v1.2.3
 |com.github.jbangdev:jbang:v1.2.3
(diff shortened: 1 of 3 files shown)


maxandersen:
wrong branch pushed - replaced by #1843

PR #1851, opened 2024-10-24
feat: jbang <gav> should honor the gav's dependency mgmt

maxandersen:
fix issue where `jbang g:a:v` is not honoring dependency mgmt in its pom.

/cc @cstamas

### DIFF
--- a/src/main/java/dev/jbang/dependencies/DependencyUtil.java
+++ b/src/main/java/dev/jbang/dependencies/DependencyUtil.java
@@ -52,11 +52,16 @@ public class DependencyUtil {
 	private DependencyUtil() {
 	}
 
-	public static ModularClassPath resolveDependencies(List<String> deps, List<MavenRepo> repos,
+	public static ModularClassPath resolveDependencies(List<String> deps, List<MavenRepo> repos, boolean offline,
+			boolean updateCache, boolean loggingEnabled, boolean downloadSources) {
+		return resolveDependencies(null, deps, repos, offline, updateCache, loggingEnabled, downloadSources);
+	}
+
+	public static ModularClassPath resolveDependencies(String root, List<String> deps, List<MavenRepo> repos,
 			boolean offline, boolean updateCache, boolean loggingEnabled, boolean downloadSources) {
 
 		// if no dependencies were provided we stop here
-		if (deps.isEmpty()) {
+		if (deps.isEmpty() && root == null) {
 			return new ModularClassPath(Collections.emptyList());
 		}
 
@@ -68,6 +73,8 @@ public class DependencyUtil {
 		}
 
 		// Turn any URL dependencies into regular GAV coordinates
+		// TODO: we should also check if the root is a URL and turn it into a GAV if
+		// possible?
 		List<String> depIds = deps
 									.stream()
 									.map(JitPackUtil::ensureGAV)
@@ -81,6 +88,11 @@ public class DependencyUtil {
 				repos.stream().map(MavenRepo::toString).collect(Collectors.joining(", "))));
 
 		String depsHash = String.join(CP_SEPARATOR, depIds);
+
+		if (root != null) {
+			depsHash = root + CP_SEPARATOR + CP_SEPARATOR + depsHash; // using double ;; to separate from if using non
+																		// root resolve
+		}
 
 		List<ArtifactInfo> cachedDeps = null;
 		if (!updateCache) {
@@ -106,7 +118,7 @@ public class DependencyUtil {
 																	.logging(loggingEnabled)
 																	.downloadSources(downloadSources)
 																	.build()) {
-			List<ArtifactInfo> artifacts = resolver.resolve(depIds);
+			List<ArtifactInfo> artifacts = resolver.resolve(root, depIds);
 
 			ModularClassPath mcp = new ModularClassPath(artifacts);
 
--- a/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
@@ -175,13 +175,27 @@ public class ArtifactResolver implements Closeable {
 	}
 
 	public List<ArtifactInfo> resolve(List<String> depIds) {
+		return resolve(null, depIds);
+	}
+
+	public List<ArtifactInfo> resolve(String root, List<String> depIds) {
 		context.repositorySystemSession().getData().set("depIds", depIds);
 		try {
 			Map<String, List<Dependency>> scopeDeps = depIds.stream()
 															.map(coord -> toDependency(toArtifact(coord)))
 															.collect(Collectors.groupingBy(Dependency::getScope));
 
-			List<Dependency> deps = scopeDeps.get(JavaScopes.COMPILE);
+			Dependency rootDep = root != null ? toDependency(toArtifact(root)) : null;
+
+			List<Dependency> deps = new ArrayList<>();
+
+			// if (rootDep != null) {
+			// deps.add(rootDep);
+			// }
+			if (!scopeDeps.isEmpty()) {
+				deps.addAll(scopeDeps.get(JavaScopes.COMPILE));
+			}
+
 			List<Dependency> managedDeps = null;
 			if (scopeDeps.containsKey("import")) {
 				// If there are any @pom artifacts we'll apply their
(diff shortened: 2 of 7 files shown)


src/main/java/dev/jbang/dependencies/DependencyUtil.java
```diff
@@ -68,6 +73,8 @@ public static ModularClassPath resolveDependencies(List<String> deps, List<Maven
 		}
 
 		// Turn any URL dependencies into regular GAV coordinates
+		// TODO: we should also check if the root is a URL and turn it into a GAV if
```
quintesse, line 76: I don't think there should be any URLs at this point. They should all be GAvs already.

quintesse:
So in general this looks good, but I'm worried that we might be overlooking places where the code only expects dependencies and now suddenly we have to check if a root is defined or not. For example, simply doing a usage search for `SourceSet.getDependencies()` all the 3 results it returns from the main code will need to be changed: `info`? Definitely. `edit`: Also. and the code that generates module info?? Oh yes, things will definitely break if that isn't updated.

So I'm wondering if this really is the correct way. Are we perhaps assuming too much? Because this seems like a very technical Maven detail that shouldn't affect us.

Hear me out.

So, for getting things to work we definitely need to set the root, no doubt about that.

But we set the root because of some Maven internal reason where if you don't have a POM file (which would be the root when building a Maven project, right?) then you need to select a proper root manually. But we **never** have a POM file! Doesn't this mean that we'd also see this problem if we do `//DEPS g.a.v` of the problematic artifact from within a source file? Wouldn't that also select the wrong jackson version? (You should have given a proper reproducible example in the issue description, then I could have tested it instead of just hypothesizing :-P )

maxandersen:
@quintesse you aren't wrong about your comments - the question is what is the "right" way to resolve it. What I've implemented here is closest to how Maven 3 behaves. that if you add a dependency to a project then the managed dependencies of those dependencie are ignored and only the roots managed dependencies is trusted.

As I understand it Maven 4 is changing that so they are listened to but the root (possibly order of dependencies) has priority.

@cstamas wdyt? what would be the "right"/"best" way?

quintesse:
@maxandersen indeed, I understand that it's the closest way, but we're suddenly creating a divergence depending on how you include things. Suddenly `jbang g.a.v.` becomes different from `//DEPS g.a.v.` and I don't think that should be the case. Both situations should in principle result in the same artifacts being resolved, right? (Otherwise the original issue could still occur, depending on where it is included, which would be very confusing for users... and us :-) )

cstamas:
This "demo" (or reproducer) shows what is changing in Maven4: https://github.com/cstamas/MRESOLVER-614
Basically "root vs non root" will not matter anymore. This is really just a Maven3 quirk that is present and was present for Maven2 _backward compat_! (best seen in Resolver 2.x as it was refactored):
https://github.com/apache/maven-resolver/blob/2f112ff3ff45c153bd1714120c7c9a225b3ab3f2/maven-resolver-util/src/main/java/org/eclipse/aether/util/graph/manager/ClassicDependencyManager.java#L87-L95

quintesse:
@cstamas ok, but do you have a suggestion what would be the best way forward for JBang right now? Because while this PR fixes part of the problem it won't fix it in all cases and it also introduces this concept of "root" in our code where I would really prefer not to know anything about it.

Is there a solution where we can resolve (har har)  this in the code that given a list of top-level dependencies (and no root level pom or anything like that) and get a workable set of artifacts? I mean in the case of `jbang g.a.v.` we can say "well, we just pick g.a.v. as the root and be done with it", but what is the "root" in a .java file that has a `//DEPS g1.a1.v1 g2.a2.v2 g3.a3.v3` line? Do we just pick the first one? Is that smart? Does it even make sense?

cstamas:
Well, IMO when doing `jbang script`, then script _is root_ (as that is equivalent to have POM with dependencies, no?).

So IMO something like this:
```
// jbang --deps a,b,c src.java -> root is src.java
// jbang --deps a src.java (bad) -> root is sec.java, --deps is "injected" (like it would be //DEPS)
// jbang gav -> root is gav
// jbang --deps a,b,c gav (bad) -> root is gav, but --deps a,b,c should be injected
// should keep track of possible root
```

Or something along those lines... 

Something similar is modeled in toolbox:
https://github.com/maveniverse/toolbox/blob/main/shared/src/main/java/eu/maveniverse/maven/toolbox/shared/ResolutionRoot.java

That is "a top" of CollectRequest. load=true means "it is some resolvable GAV somewhere" (this is jbang gav case), while load=false means "it is like local project/POM", so you need to fill in deps, depMgt etc (this js jbang script.java case).

maxandersen:
The problem is that our current code code is assuming one list in expanded list out. 

If there was a higher abstraction akin to a "collectrequest" it wouldn't have as much impact on jbang code base.

quintesse:
(Sorry for the long message)

> Well, IMO when doing `jbang script`, then script _is root_ (as that is equivalent to have POM with dependencies, no?).
> 
> So IMO something like this:
> 
> ```
> // jbang --deps a,b,c src.java -> root is src.java
> ...
> ```

Well sure, except for the fact that it isn't. We can't do `collectRequest.setRoot("src.java")` or anything similar. There is no POM file that serves as root. There's just a script that we can _think_ of as root, but thinking is not the same as being an actual root.

So like @maxandersen mentioned, we suddenly need to keep track of how JBang was started to be able to turn `jbang --deps a,b,c g.a.v.` into:

```
collectRequest
  .setRoot(g.a.v.)
  .setDependencies(List.of(a, b, c))
```

while turning `jbang --deps g.a.v,a,b,c src.java` into:

```
collectRequest
  .setRoot(null)
  .setDependencies(List.of(g.a.v., a, b, c))
```

While in both cases we want the exact same artifacts to be added to the classpath!

There _is_ no POM so there are no managed dependencies that could be set or anything else that could potentially affect the resolving process (at least AFAIK).

So we'd rather have a single process that doesn't require us to know how JBang was started but where we can simply say: "Please resolve g.a.v., a, b, c for us". Is that possible?

cstamas:
Yes, you could consider the list of GAVs as "list of roots". So resolve each as root, and just (carefully) merge the resulting list of lists (of artifacts)

quintesse:
For example, could we create a "dummy root" that would always serve as the root of whatever list of dependencies we pass?

(I don't know the internals of the Maven resolve process, but I'm guessing the root serves the purpose to make sure that the versions of everything below the root coincide, while things outside might either result in duplicate artifacts being present or versions not coinciding?)

quintesse:
> Yes, you could consider the list of GAVs as "list of roots". So resolve each as root, and just (carefully) merge the resulting list of lists (of artifacts)

And how does one merge "carefully"? :-D

quintesse:
And doesn't that also mean that we'd have to do multiple collect requests? THat could become pretty expensive if there are a lot of dependencies, right?

cstamas:
Use ordered set (to retain classpath order but also uniqueness)

cstamas:
> And doesn't that also mean that we'd have to do multiple collect requests? THat could become pretty expensive if there are a lot of dependencies, right?

Yes, but would also be the most correct currently. Another approach is to go over all dep GAVs and "lift" depMgt to root, and that is doable with latest MIMA as it can serve you effective models... will do PR for that, but currently am on resolver issue and prepping a release 2.0.3 (that was canceled once already due very same issues but for transitive dep Mgr mvn4)

cstamas:
Re resolver: @maxandersen or @quintesse could any of you point me at the GAV that startled all this issue (just to vreify current resolver on it as well)

cstamas:
Just some "status report": we are nearing release of next Resolver 2.x and Maven 4 beta... and spent time fixing exactly this issue (in perspective of Maven4):
https://github.com/apache/maven-resolver/pull/595
Related "doco" if someone is interested in details:
https://gist.github.com/cstamas/a65f042d14ee8e58c1a6d3c17e3fecaf
ITs now contain this very example as well, making sure we do not miss it:
https://github.com/apache/maven-integration-testing/pull/398

quintesse:
> Re resolver: @maxandersen or @quintesse could any of you point me at the GAV that startled all this issue (just to vreify current resolver on it as well)

@maxandersen showed me but I don't remember how to reproduce exactly. Could you help out Max?

quintesse:
> Just some "status report": we are nearing release of next Resolver 2.x and Maven 4 beta... and spent time fixing exactly this issue

So perhaps it makes sense to wait for Resolver v2 instead of fixing this now in JBang in a way we're not really happy with?

cstamas:
No, it can be done in Resolver 1 as well (see Toolbox). For example I don't think "multiple collect request" are such a big problem (especially if you contrast "correctness" vs "speed"), just let me get some time post votes/releases...

cstamas:
Proposed fix instead this one: https://github.com/jbangdev/jbang/pull/1853

cstamas:
This is the ` Another approach is to go over all dep GAVs and "lift" depMgt to root,` and you had all in place as you already lifted BOMs, so just apply the same thing to all deps.

maxandersen:
https://github.com/jbangdev/jbang/pull/1853 is the better fix.

PR #1853, opened 2024-10-29
fix: fix dependency resolution re depMgmt

cstamas:
Issue is related how Maven3 works, as JBang creates "fake" root and adds direct dependencies to it. This in Maven3 implies that depMgmt entries of added dependencies are NOT applied.

Example: `jbang io.quarkus:quarkus-tls-registry:jar:3.15.1` will use Jackson 2.16... instead of wanted 2.17...

Fix: I missed that depMgmt of BOMs is already slurped up, so basically just do the same to **all deps**, while BOMs are added first (to prevail). ~~Also, no need to "apply" depMgt and any sort of magic, just let Resolver it's job.~~ There is need for magic, as unlike Maven, JBang allows use of GA and will deduce V from BOM.

Example now uses Jackson 2.17... as can be seen in gist w/ PR applied locally:
https://gist.github.com/cstamas/024db87af357b658492a9d9b60857bfa

### DIFF
--- a/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
@@ -193,6 +193,7 @@ public class ArtifactResolver implements Closeable {
 				List<Dependency> mdeps = boms	.stream()
 												.flatMap(d -> getManagedDependencies(d).stream())
 												.collect(Collectors.toList());
+				deps = deps.stream().map(d -> applyManagedDependencies(d, mdeps)).collect(Collectors.toList());
 				managedDeps.addAll(0, mdeps);
 			}
 
@@ -303,6 +304,21 @@ public class ArtifactResolver implements Closeable {
 				return res;
 			}
 		};
+	}
+
+	private Dependency applyManagedDependencies(Dependency d, List<Dependency> managedDeps) {
+		Artifact art = d.getArtifact();
+		if (art.getVersion().isEmpty()) {
+			Optional<Artifact> ma = managedDeps	.stream()
+												.map(Dependency::getArtifact)
+												.filter(a -> a.getGroupId().equals(art.getGroupId())
+														&& a.getArtifactId().equals(art.getArtifactId()))
+												.findFirst();
+			if (ma.isPresent()) {
+				return new Dependency(ma.get(), d.getScope(), d.getOptional(), d.getExclusions());
+			}
+		}
+		return d;
 	}
 
 	private List<Dependency> getManagedDependencies(Dependency dependency) {
--- a/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
@@ -182,16 +182,18 @@ public class ArtifactResolver implements Closeable {
 															.collect(Collectors.groupingBy(Dependency::getScope));
 
 			List<Dependency> deps = scopeDeps.get(JavaScopes.COMPILE);
-			List<Dependency> managedDeps = null;
+			List<Dependency> managedDeps = deps	.stream()
+												.flatMap(d -> getManagedDependencies(d).stream())
+												.collect(Collectors.toList());
+
 			if (scopeDeps.containsKey("import")) {
 				// If there are any @pom artifacts we'll apply their
-				// managed dependencies to the given dependencies
+				// managed dependencies to the given dependencies BEFORE ordinary deps
 				List<Dependency> boms = scopeDeps.get("import");
 				List<Dependency> mdeps = boms	.stream()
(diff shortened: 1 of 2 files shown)


cstamas:
Oh, "I see what have you done" :smile: returning "magic" as JBang is not Maven and it allows use of GA it seems...

quintesse:
> Oh, "I see what have you done" 😄 returning "magic" as JBang is not Maven and it allows use of GA it seems...

Could you explain what "we have done"? 😅 
What does "allow use of GA" mean and why does Maven not allow it?

cstamas:
Expressing something like this is impossible in Resolver (or at least, ModelBuilder "fills in the gaps" if needed, but that is Maven not Resolver alone):
https://github.com/jbangdev/jbang/blob/main/src/test/java/dev/jbang/cli/TestEdit.java#L121-L123

quintesse:
Is there a way you think we _should_ have done it to be able to work with BOMs?

cstamas:
> Is there a way you think we _should_ have done it to be able to work with BOMs?

Unsure what is the right answer here :smile:
But I dislike that `@pom` means BOM. As I understand the syntax is `@type` (ie. `@fatjar`), and while POM _is type_ "pom", what happens if you want to use POM that is not BOM what is grouping dependencies (a common pre-BOM pattern)?

quintesse:
> > Is there a way you think we _should_ have done it to be able to work with BOMs?
> 
> Unsure what is the right answer here 😄 But I dislike that `@pom` means BOM. As I understand the syntax is `@type` (ie. `@fatjar`), and while POM _is type_ "pom", what happens if you want to use POM that is not BOM, but instead is grouping dependencies (a common pre-BOM pattern)?

Yeah, @maxandersen would be able to explain this better, but IIRC it was done because we needed to treat BOMs differently and we thought "pom" as a type wasn't actually used by Maven so we could use it ourselves. Only years later we found out that was incorrect.

We're still discussing if this is something we can "correct" (JBang is not 1.0 yet so we could make a breaking change).

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
```diff
@@ -182,16 +182,19 @@ public List<ArtifactInfo> resolve(List<String> depIds) {
...
 			List<Dependency> deps = scopeDeps.get(JavaScopes.COMPILE);
-			List<Dependency> managedDeps = null;
+			List<Dependency> managedDeps = deps	.stream()
+												.flatMap(d -> getManagedDependencies(d).stream())
+												.collect(Collectors.toList());
+
 			if (scopeDeps.containsKey("import")) {
 				// If there are any @pom artifacts we'll apply their
-				// managed dependencies to the given dependencies
+				// managed dependencies to the given dependencies BEFORE ordinary deps
```
cstamas, line 191: comment should be "... and ADD THEM BEFORE ordinary dep depMgt sections"

src/main/java/dev/jbang/dependencies/ArtifactResolver.java
```diff
@@ -182,16 +182,19 @@ public List<ArtifactInfo> resolve(List<String> depIds) {
...
 				List<Dependency> boms = scopeDeps.get("import");
 				List<Dependency> mdeps = boms	.stream()
 												.flatMap(d -> getManagedDependencies(d).stream())
 												.collect(Collectors.toList());
 				deps = deps.stream().map(d -> applyManagedDependencies(d, mdeps)).collect(Collectors.toList());
-				managedDeps = mdeps;
+				managedDeps.addAll(0, mdeps);
```
maxandersen, line 197: if we do this do we even need to set dependencies?

maxandersen: oh we do because managed dependences aren't resolved, just used for guiding proper resolution.

maxandersen: actually - why does it fix things to add all deps to managed dependencies?

cstamas: No, we don't do that. What we do is "pull all dependency dependencyManagement section (if any)." Dependencies are still on level 1, it is only their management section pulled up.

maxandersen:
okey - this fix almost looks too simple to be good but I can't find fault with it.

It will cause some changes to users dependency resolution but I think to the better.

What it does is that JBang now will honor any dependencies management section (which is what Maven4 will do to) meaning that when you do `jbang agav:1.3.2` will result the same as if import scoped agav:1.3.2 and all its dependencies managed dependencies first and then resolve.

Where as before it would not honor the managed deps of transitive dependencies.

This makes things much more consistent and remove/reduce need for having to use bom-pom's as one can now just rely on the included dependency owns managed dependencies.

PR #1862, opened 2024-11-13
fix: fixes issue with too long command lines on windows

### LINKED ISSUE #1861: jbang fails with `CreateProcess error=206, The filename or extension is too long` error on Windows
jbang fails while installing an application (jbang app install yaks@citrusframework/yaks) due to `CreateProcess error=206, The filename or extension is too long` on Windows.

**To Reproduce**
1. run `jbang app install yaks@citrusframework/yaks`
2. See error

**Expected behavior**
Yaks installed successfully.

**JBang version**
Paste output of `jbang version --verbose`

There is a workaround (depicted here: https://www.howtogeek.com/266621/how-to-make-windows-10-accept-file-paths-over-260-characters/), but usually Windows dev machines are restricted and can't apply the workaround. 

Other projects, such as Spring Boot and CXF, replaced the long command line with and argument file.

quintesse:
We were already doing this for the `java` command, but not for the `javac` and `native-image` commands becuase they were run using the `ProcessBuilder` API and it turns out it has a limit of 32K characters on Windows.

Fixes #1861

### DIFF
--- a/src/main/java/dev/jbang/source/buildsteps/CompileBuildStep.java
+++ b/src/main/java/dev/jbang/source/buildsteps/CompileBuildStep.java
@@ -130,7 +130,11 @@ public abstract class CompileBuildStep implements Builder<Project> {
 	}
 
 	protected void runCompiler(List<String> optionList) throws IOException {
-		runCompiler(CommandBuffer.of(optionList).asProcessBuilder().inheritIO());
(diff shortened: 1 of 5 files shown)


quintesse:
No tests for, because they would take more work than the fix itself 😅

maxandersen:
YOLO!

PR #1874, opened 2024-11-27
refactor: Separated JDK management code into its own module

### LINKED ISSUE #1857: Refactor JDK management code into separate module
This module should be its own artifact that can be published and used by 3rd parties.

quintesse:
Fixes #1857


src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -6,22 +6,101 @@
...
+import dev.jbang.jvm.jdkproviders.*;
 
 public class JavaUtil {
 
 	private static Integer javaVersion;
 
+	private static JdkManager jdkManager = null;
+
+	@Nonnull
+	public static JdkManager jdkManager() {
```
maxandersen, line 29: we seem to call this in many places - but how is going to take into affect jbang provided config of i.e. providers?

quintesse: Right now the idea is that the config gets used exactly in that file you're referencing. So the builder will use the config to initialize the jdk manager, including the providers. The jdk manager itself is then completely agnostic where its configuration comes from.

maxandersen: ok so in jbang cli we would not call jdkmanger() but have our own instance configured from jbang config?

quintesse: Well, this `JavaUtil` class _is_ in the jbang CLI, so this code here is exactly where that instance using jbang config is created.

quintesse: NB: If you look a bit lower in the `createProvider()` method you see it references `config`, which is the jbang config. The missing part I mentioned in our chat is exactly that: tying together the config with the instance creation :-)

maxandersen:
small thing: how come the package is called `.jvm` when everything else is named JDK+something ?

build.gradle
```diff
@@ -18,20 +18,85 @@ plugins {
 	id 'maven-publish'
 }
 
-javadoc {
-	options.encoding = 'UTF-8'
-	//remove this to see all the missing tags/parameters.
-	options.addStringOption('Xdoclint:none', '-quiet')
-}
+allprojects {
```
maxandersen, line 21: no matter what I think the changes to enable additional modules in this repo should stay. enable us to split up or add things where it makes the most sense.

src/main/java/dev/jbang/util/JBangFormatter.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Date;
+import java.util.logging.Formatter;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+/**
+ * This JUL Formatter is used to format JUL log messages in such a way that they
+ * look similar to JBang log messages.
+ */
+public class JBangFormatter extends Formatter {
```
maxandersen, line 14: hmm - originally I explicitly kept "ux output" separate from "logging output" ...is there a reason why we want to redirect all output to JUL ?

src/main/resources/logging.properties
```diff
@@ -0,0 +1,4 @@
+handlers=java.util.logging.ConsoleHandler
```
maxandersen, line 1: why do we need this now? (not against it - just want to understand the rationale)

build.gradle
quintesse: Any changes I made to the gradle build file are there only to get things to work, which was already difficult enough. Very likely I made mistakes or did it in an improper way.

src/main/java/dev/jbang/util/JBangFormatter.java
quintesse: Well, you might not want all of the logging output to be confused with UX output, I can understand that.

But there are two things that come to mind:

1. our output mechanism sure does look like logging (having levels like warn, inifo, error, debug, etc) 
2. to make the new module really independent from JBang I needed a way to generate output that didn't use JBang's system and I didn't want to introduce some new interface

So that's why I decided to switch the module's output to use plain logging but then the output would look different from before, so that's why I decided to capture the logging output and make it look like JBang output.

Now, I thought it would be nice to do the same for _anything_ that could be outputting logging info, but that's just something I decided in the moment and we could easily change that to only work for the module and nothing else.

src/main/resources/logging.properties
quintesse: It's just to make any logging appear as if it comes directly from JBang (with our normal "[jbang]" prefix for every line), which was needed to make the logging output from the new module appear exactly the same as it did before.

quintesse:
> small thing: how come the package is called `.jvm` when everything else is named JDK+something ?

Historical reasons, because that was the tool is called that I want to use this module for. But it can of course be anything we want.

quintesse:
Yay, all checks green! I've marked the PR as final now that devkitman has been published and all the tests pass.

build.gradle
```diff
@@ -25,6 +25,7 @@ javadoc {
 }
 
 repositories {
+	mavenLocal()
```
quintesse, line 28: Max, I added this so I could use locally installed SNAPSHOT versions of devkitman. We can leave this in, right?

build.gradle
```diff
@@ -90,10 +91,12 @@ sourceSets {
 sourceSets.main.compileClasspath += sourceSets.java9.output.classesDirs;
 
 dependencies {
-	implementation 'com.offbytwo:docopt:0.6.0.20150202'
+	implementation 'dev.jbang:devkitman:0.1.2'
+
+//	implementation 'com.offbytwo:docopt:0.6.0.20150202'
```
quintesse, line 96: This probably shouldn't be here, but it seemed unused to I commented it out and it slipped into the PR. If you want I can take it out.

build.gradle
```diff
@@ -238,6 +241,7 @@ compileJava9Java {
 shadowJar {
 	minimize() {
 		//exclude(dependency('org.slf4j:slf4j-api:.*'))
+		exclude(dependency('dev.jbang:devkitman:.*'))
```
quintesse, line 244: devkitman uses services so this is needed or classes will be missing.

build.gradle
maxandersen: hmm - I would say if it must be here it should be after mavenCentral() so  the build isn't injected with any local bits you have.

build.gradle
maxandersen: lol - thats a ancient leftover from very early jbang...never needed. lets just nuke it.

src/main/resources/logging.properties
maxandersen: ok but:
a) what makes it end up in stderr not stdout?
b) doesn't this meddling with jul logging make it harder to embed?

src/test/java/dev/jbang/BaseTest.java
```diff
@@ -61,6 +65,19 @@ void initEnv(@TempDir Path tempPath) throws IOException {
 
 	@BeforeAll
 	static void init() throws URISyntaxException, IOException {
+		try {
+			// The default ConsoleHandler for logging doesn't like us changing
```
maxandersen, line 69: yeah so all this is why i refrained from using JUL for jbang output. why is this in the test class? shouldn't it be in main code path too?

src/main/resources/logging.properties
quintesse: a) what do you mean? all JBang messages go to stderr, right? (at least any of the `xxxMsg()` methods, which is what the JUL logging replaces here)

b) also not sure what you mean here. Using JUL logging is pretty normal, isn't it? You could use it just fine even with logging completely disabled. There's no essential output being generated using JUL.

src/test/java/dev/jbang/BaseTest.java
quintesse: But in this module JUL is only used for logging, there is no essential output being generated with it. It's a library, it shouldn't do any essential output by itself, it only returns information which the caller can then decide to output or not. The logging is really only used for logging.
Thing is we do have tests that specifically check for certain log lines, we shouldn't really be doing that but I didn't want to change the tests so I added this code to make sure that the output between monolithic JBang and modular JBang would remain 99.9% the same.

maxandersen (approved):

Issue #1776, opened 2024-03-22
option for turning on/off integrations


maxandersen:
as discussed in #1774 there can be situatons where the classpath contains integrations you dont want to run.

we should at a minimum have a global off switch, maybe `--no-integrations` where `--integrations` are default ?

possibly consider  `--integrations=none` and `--integrations=all` is default and open up for allowing list of '.'  names to turn on/off specific integrations?

jmini:
Couldn't it be something that is controlled in the comments of the file itself? Similar to other keywords like: `//DEPS`, `//REPOS`, `//JAVA`.

I would prefer having it in the source than having to type it on the command line every time I execute the script.

quintesse:
@maxandersen perhaps this should be part of the build steps feature I had been working on some time ago? There were several ways to select which steps to run, including from the sleeve file. We could then great the quarkus integration as a kind of legacy build step.

maxandersen:
How would that look?

jmini:
@maxandersen @quintesse we would be very interested by this feature, since we currently can't execute a client library that is built with quarkus but that has no dependency on quarkus. What can we do to push this forward?

maxandersen:
submit a PR with a flag, maybe `--integrations` that take an optional list of names that will match on substring. Allowing you to do:

`--no-integrations` that runs nothing.
`--integrations` that runs as is today
`--integrations=quarkus` only run things matching 'quarkus'
`--no-integrations=quarkus` run everything but things matching `quarkus`

....if wanna keep it simple then just start with `--integrations/--no-integrations` (`negatable=true` in picocli)

maxandersen:
realized its pretty easy to do.  draft at #1875 to try.

jmini:
This works perfectly. See: https://github.com/orgs/jbangdev/discussions/1774#discussioncomment-11761212

Thank you very much.

PR #1875, opened 2024-12-02
feat: add --no-integrations option

maxandersen:
realized #1776 didn't require much for the first MVP.

adds a `--no-integrations` option to disable execution of integrations.

### DIFF
--- a/itests/noints.java
+++ b/itests/noints.java
@@ -1,0 +1,8 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//NOINTEGRATIONS
+
+public class noints {
+    public static void main(String... args) {
+        System.out.println("No integrations here");
+    }
+}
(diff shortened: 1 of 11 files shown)


maxandersen:
opening as draft as not sure this is how it should look. 

@jmini give it a try for your executable and tell me if works for you or things missing.

src/main/java/dev/jbang/util/Util.java
```diff
@@ -77,7 +77,7 @@ public class Util {
 			Pattern.MULTILINE);
 
 	public static final Pattern mainClassMethod = Pattern.compile(
-			"(?<=\\n|\\A)(?:public\\s)\\s*(class)\\s*([^\\n\\s]*)");
```
quintesse, line 80: This is an unrelated change, isn't it?

quintesse:
@maxandersen I updated the PR with the "actual" code. (The `--(no-)integrations` option changes compile output, like for example `--native`, and should therefore be part of the `Project` options. I changed the PR to reflect that.

(I _am_ starting to dislike having to copy all options to the `Alias` class though)

src/main/java/dev/jbang/util/Util.java
maxandersen: yeah thatwas definitely not intended.

quintesse:
Hmmm weird, the error is still there...

quintesse:
Ok, problem fixed and I added support for a `//NOINTEGRATIONS` tag as well.

PR #1877, opened 2024-12-03
ensure -parameters is set for javac

maxandersen:
I've [had enough ](https://github.com/mariofusco/site-summarizer/pull/2) with java not by default adding `-parameters` 

since java 8 `-parameters` been supported and if you use any Quarkus, Spring, Micronaut or any other annotation driven framework or library you need `-parameters` to have functional code.

Only downside I can think of is that the bytecode gets slightly bigger and pre-Java 8 code that somehow assumes names starts with "argN" breaks - that is minuscule impact  IMO.

But it is what kept me from adding it until now.

Now I'm realizing that it actually is worse to NOT include it due to so many libraries requiring it and having to tell everyone to include `//JAVAC_OPTIONS -parameters` to fix possible obscure errors is not worth worth.

we do the same for `-g` (debug symbols) and thus I implemented it exactly the same here - if input is a javasource file enable it. kotlin et.al. not affected (kotlinc already defaults to this).

Also multiple -parameters is ignored by javac so users already having it are not affected.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -173,3 +173,8 @@ jobs:
           echo "::add-matcher::.github/karatewindows-matcher.json" 
           .\itests.cmd
           echo "::remove-matcher owner=karate-windows::"
+      - name: Write out Unit Test report annotation for forked repo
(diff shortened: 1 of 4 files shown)


quintesse:
Great idea!

PR #1878, opened 2024-12-03
Update Kotlin to 2.1.0

yostane:
Updated Kotlin to the latest stable version 2.1.0

https://kotlinlang.org/docs/releases.html#release-details


maxandersen:
I get the wish to use latest; but should we always update latest here ?

for Java we stay on Java 11 for default, and will allow to use //JAVA - and if users need specific kotlin they use //KOTLIN

I'm not sure how stable/nonstable kotlin but wont moving up aggressively risk break users?

yostane:
Good question 😊.

In terms of stability, Kotlin 2.1.0 is a stable version. But I agree that it's generally safer to wait a bit after a new release before changing the default version.

I think that we can close this PR since we also have the option of choosing a Kotlin version.

yostane:
To be merged later

PR #1879, opened 2024-12-04
Fixes for --preview option and using build.jbang

quintesse:
FIrst of all JBang will now correctly handle trying to run `jbang --preview somefile.java` and secondly the message when unsuccessfully trying to run a folder (eg `jbang --preview .`) will now mention `build.jbang` (before it would only mention `main.java`).
And finally running `jbang --preview .` will now actually work if the folder contains a `build.jbang` file.

### DIFF
--- a/src/main/java/dev/jbang/Main.java
+++ b/src/main/java/dev/jbang/Main.java
@@ -2,6 +2,7 @@ package dev.jbang;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import dev.jbang.cli.JBang;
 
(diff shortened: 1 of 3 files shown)


quintesse:
Btw @maxandersen , isn't it time to remove this from "preview" and promote this to a full feature?

maxandersen:
+1 on the improvements.

about full feature my issue is that the syntax isn't really doing more than just pass in bunch of directives.

quintesse:
For me it's the fact that you can remove the lines from the .java files and into a file that's completely JBang specific. Very useful for those situations where you really don't want to put weird non-Java tags in your source files, but you _do_ want the ease-of-use of JBang.

Issue #1886, opened 2024-12-17
Basic Auth in the URL is not supported


bartoszpop:
**Is your feature request related to a problem? Please describe.**
Currently, Basic Auth for catalog-ref is only supported by setting the JBANG_AUTH_BASIC_USERNAME and JBANG_AUTH_BASIC_PASSWORD env variables. It is not scalable as there might be multiple catalogs hosted in different locations.

**Describe the solution you'd like**
The solution is to access Basic Auth in the url, e.g. http://username:password@example.com.

PR #1887, opened 2024-12-18
feat: support Basic Auth in the catalog-ref URL (#1886)

bartoszpop:
This pull request adds support for Basic Auth credentials in the catalog-ref url. They will take precedence over JBANG_AUTH_BASIC_USERNAME/JBANG_AUTH_BASIC_PASSWORD env variables.

#1886

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -1123,11 +1123,20 @@ public class Util {
 
 	private static void addAuthHeaderIfNeeded(URLConnection urlConnection) {
 		String auth = null;
-		if (urlConnection.getURL().getHost().endsWith("github.com") && System.getenv().containsKey("GITHUB_TOKEN")) {
+		URL url = urlConnection.getURL();
+		if (url.getHost().endsWith("github.com") && System.getenv().containsKey("GITHUB_TOKEN")) {
 			auth = "token " + System.getenv("GITHUB_TOKEN");
 		} else {
-			String username = System.getenv(JBANG_AUTH_BASIC_USERNAME);
-			String password = System.getenv(JBANG_AUTH_BASIC_PASSWORD);
+			String username;
+			String password;
+			if (url.getUserInfo() != null) {
+				String[] credentials = url.getUserInfo().split(":", 2);
+				username = credentials[0];
+				password = credentials.length > 1 ? credentials[1] : "";
+			} else {
+				username = System.getenv(JBANG_AUTH_BASIC_USERNAME);
+				password = System.getenv(JBANG_AUTH_BASIC_PASSWORD);
+			}
 			if (username != null && password != null) {
 				String id = username + ":" + password;
 				String encodedId = Base64.getEncoder().encodeToString(id.getBytes(StandardCharsets.UTF_8));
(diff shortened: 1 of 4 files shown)


src/main/java/dev/jbang/util/Util.java
```diff
@@ -99,6 +101,16 @@ public class Util {
 	private static Boolean downloadSources;
 	private static Instant startTime = Instant.now();
 
+	private static final StringSubstitutor urlSubstitutor = createUrlSubstitutor();
```
maxandersen, line 104: we have `PropertiesValueResolver` to do property replacement. don't think there is a need to use different one here?

src/test/java/dev/jbang/util/TestUtil.java
```diff
@@ -158,4 +160,13 @@ void testDispostionFilename() {
 		// assertThat(Util.getDispositionFilename("inline;
 		// filename*=iso-fake-1''dummy"), equalTo(""));
 	}
+
+	@Test
+	void testUrlBasicAuth() throws IOException {
+		URLConnection connection = new NoOpUrlConnection(new URL("https://SomeUser:${properties:src\\test\\resources\\url.properties::password}@example.com"));
```
maxandersen, line 166: i'm not following what that syntax is meant to do? 

loading a relative file and grap a value from it?

src/main/java/dev/jbang/util/Util.java
bartoszpop: Done

src/test/java/dev/jbang/util/TestUtil.java
bartoszpop: Yes, this syntax is defined in [org.apache.commons.text.StringSubstitutor](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html). Anyway, as suggested in the first comment, I replaced it with PropertiesValueResolver.

bartoszpop:
@maxandersen comments addressed, pls review

bartoszpop:
@maxandersen any changes to have it merged?

maxandersen:
so i like the idea i'm just reluctant to add it as it applies only to github but github isn't the only one where this would be useful.

bartoszpop:
@maxandersen it applies to any URL, e.g. we use jbang-catalog.json to refer to another externally hosted jbang-catalog.json:
```
{
  "catalogs": {
    "example": {
	"catalog-ref": "https://example.com/jbang-catalog.json"
    }
  }
}
```
which does support Basic Auth, but currently we can't use the inline syntax:
```
{
  "catalogs": {
    "example": {
	"catalog-ref": "https://user:pass@example.com/jbang-catalog.json"
    }
  }
}
```
and we need to use JBANG_AUTH_BASIC_USERNAME and JBANG_AUTH_BASIC_PASSWORD instead.

maxandersen:
what I meant was that it should not just be for catalog-ref urls but any other url we refer to, maven repos, github and other urls but I think it actually already are doing that - i just got mislead by the the commit/issue message talking only about catalogs.

quintesse:
Needs a `./gradlew spotlessApply`

bartoszpop:
@quintesse @maxandersen rebased and reformatted

maxandersen:
build errors my bad of a bad manual merge. fixed.

maxandersen (approved):

PR #1891, opened 2024-12-29
Remove spurious line endings in Wrapper's Powershell script

vorburger:
This avoid a half second of friction when adding the wrapper to projects with like to trim whitespace at the end of lines, using e.g. tools like [`.editorconfig`](https://editorconfig.org/) with trim_trailing_whitespace = true` enforced e.g. with https://pre-commit.com.


quintesse (approved):

quintesse:
Thanks!

PR #1892, opened 2024-12-29
Fix a few ShellCheck SC2086 & SC2155 in Wrapper's Linux script

vorburger:
See https://www.shellcheck.net/wiki/SC2086 & https://www.shellcheck.net/wiki/SC2155.

### DIFF
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -17,7 +17,7 @@ script_dir() {
     [[ $script != /* ]] && script=$dir/$script
   done
   dir=$( cd -P "$( dirname "$script" )" >/dev/null 2>&1 && pwd )
-  echo $dir
+  echo "$dir"
 }
 
 download() {
@@ -141,7 +141,7 @@ else
     else
       jburl="https://github.com/jbangdev/jbang/releases/download/v$JBANG_DOWNLOAD_VERSION/jbang.tar";
     fi
-    download $jburl "$TDIR/urls/jbang.tar"
+    download "$jburl" "$TDIR/urls/jbang.tar"
     if [ $retval -ne 0 ]; then echo "Error downloading JBang from $jburl" 1>&2; exit $retval; fi
     echo "Installing JBang..." 1>&2
     rm -rf "$TDIR/urls/jbang"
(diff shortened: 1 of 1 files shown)


vorburger:
Oups, hang on, at least one of the change proposed (automagically) here actually appears to break the wrapper... do not merge!

vorburger:
> at least one of the change [originally] proposed (automagically) here actually appears to break the wrapper

Found it - that `${JBANG_JAVA_OPTIONS}` should not be quoted, so I've undone that, and it works again.

It would be good to have a `# shellcheck disable=SC2086` there, to avoid future confusion - so I've added that.

Please do carefully review to make sure that this doesn't break anything else (but I think as-is now it's fine).

PS: I'm not sure (didn't check) if CI caught this - is there integration test coverage for the wrapper?

src/main/scripts/jbang
```diff
@@ -219,8 +219,10 @@ fi
 
 ## run it using command substitution to have just the user process once jbang is done
 export JBANG_RUNTIME_SHELL=bash
-export JBANG_STDIN_NOTTY=$([ -t 0 ] && echo "false" || echo "true")
+JBANG_STDIN_NOTTY=$([ -t 0 ] && echo "false" || echo "true")
```
quintesse, line 222: why did this line get split (and the other not)?

vorburger: See https://www.shellcheck.net/wiki/SC2155 ... so it's because this line uses `$( ... )` and the other two do not.

maxandersen:
Last time I tried shell check it was overly cautious but happy to see if we can improve :) this does change quotes so even though we do have some decent level of intégration tests in itest i'm not confident it tests for these. 

It would be great to actually have some test showing the issue shell check is worried about over just fixing it do reduce lint warnings.

vorburger:
> Last time I tried shell check it was overly cautious but happy to see if we can improve :) this does change quotes so even though we do have some decent level of intégration tests in itest i'm not confident it tests for these.

FWIW I **did** actually test these changes myself... ;-) and it seems to still work.

> It would be great to actually have some test showing the issue shell check is worried about over just fixing it do reduce lint warnings.

That would be cool (but I wouldn't want to be the one contributing it). To me, this is more like fixing a warning from a compiler or build tool - you typically perhaps wouldn't write a test?

maxandersen:
YOLO :)

PR #1898, opened 2025-01-07
docs: Add "Disable integrations" section

jmini:
Documents https://github.com/jbangdev/jbang/pull/1875

See https://github.com/jbangdev/jbang/issues/1776, or https://github.com/orgs/jbangdev/discussions/1774#discussioncomment-11761212


quintesse:
Thanks @jmini ! Really appreciate your eye for detail :-)

PR #1901, opened 2025-01-10
fix: if no deps, avoid npe; also make sure POMs refd exist

cstamas:
If there are no deps (or only imports, which does not make sense), do not belly up. Also, make sure that all the refd POMs (of deps and imports) are present, as by def Maven uses "forgiving" policy.

### DIFF
--- a/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
+++ b/src/main/java/dev/jbang/dependencies/ArtifactResolver.java
@@ -346,17 +346,26 @@ public class ArtifactResolver implements Closeable {
 	private ArtifactDescriptorResult resolveDescriptor(RepositorySystemSession session, Artifact artifact) {
 		try {
 			if (artifact.getVersion().trim().isEmpty()) {
-				return new ArtifactDescriptorResult(new ArtifactDescriptorRequest(artifact, context.remoteRepositories(), ""));
+				return new ArtifactDescriptorResult(
+						new ArtifactDescriptorRequest(artifact, context.remoteRepositories(), ""));
 			}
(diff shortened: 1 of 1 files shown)


quintesse:
The strict resolving seems to break some tests, right?

cstamas:
Even if now it works, this is _wrong_. The deps with ranges should be first "materialized" as exact versions, as nothing guarantees that you load POM of v1 and then resolve v2 (this is very bent, still, no guarantee for it).

maxandersen:
sorry for being daft here - whats the wrong part? afaik we didn't do anything different with version ranges than normal maven...but you write as if we did - so curius to grok it before merging?

cstamas:
no problem per se, feel free to merge. It is only me tinkering... 
For me the wrong part is that ranges are resolved twice (which in a session will yield same result UNLESS jbang equiv of mvn -U (`--fresh`?) is used AND a deploy happened between the two invocations -- which is a very small windows, but is not zero).

All in all, is fine to merge, but needs more. Will add PR later.

quintesse:
@cstamas it seems these changes have caused issues with some of our users. Could you perhaps take a look at this issue and see what's going on? https://github.com/enola-dev/enola/issues/1040

cstamas:
I don't think so. In fact, i think these changes made some issues now apparent, that were hidden before... We will see...

cstamas:
So I was right :smile: 
JBang discovered an issue https://github.com/bazel-contrib/rules_jvm_external/issues/1324

quintesse:
Awesome! :-)

I don't understand though how it worked in JBang version 0.122.0? Does that have to do with the "lax resolver" PR you opened? If so I would love if you could explain the "how" to me, because it's a mystery to me 😅

PR #1906, opened 2025-01-14
docs: Update Adoptium URL in javaversions.adoc

yostane:
Replace https://adoptopenjdk.net/releases.html with https://adoptium.net/temurin/releases, in javaversions.adoc, to be up to date with AdoptOpenJDK moving to the Eclipse foundation. In fact, when we open https://adoptopenjdk.net/releases.html, we get this message: 

> 24th July 2021: AdoptOpenJDK is moving to the Eclipse Foundation and rebranding.
> Our July 2021 and future releases will come from Adoptium.net


maxandersen:
thanks!

PR #1907, opened 2025-01-14
Add `jbang jdk list --available` in javaversions.adoc

yostane:
I propose to add this command to the documentation so that the user can check the available versions without leaving this terminal.


maxandersen (approved):

maxandersen:
makes sense! thanks!

@all-contributors please add @yostane  for docs

Issue #1910, opened 2025-01-23
Attributes defined in jbang-catalog.json should but appended to the end of the command.


bartoszpop:
**Describe the bug**
Attributes defined in jbang-catalog.json should but appended to the end of the command.

**To Reproduce**
The below `jbang-catalog.json` results with
1. Create the following jbang-catalog.json
{
	"aliases": {
		"camel": {
                        "script-ref": "org.apache.camel:camel-jbang-main:4.8.3",
			"main": "main.CamelJBang",
			"arguments": ["--quarkus-version=3.16.0"]
		}
	}
}

2. run the command `jbang camel@apache/camel run * --runtime=quarkus`. It will produce jbang9236616212127187886.args

```
-classpath
"(...)"
main.CamelJBang
--quarkus-version=3.17.0 <--- this should go at the end
run
.camel-jbang
.camel-jbang-run
somefile.camel.yaml
jbang-catalog.json
--runtime=quarkus

```

**JBang version**
[jbang] [0:664] jbang version 0.121.0

quintesse:
Well... I understand that for your use-case you want it like that, but I think this is actually worse (besides the possibility that we could break existing code). The current implementation is based on how aliases work in shells (bash, zsg, etc), where user supplied arguments are also appended to the end. Why? So the user can override a parameter supplied by the alias, if alias params are put at the end they would always have precedence and the user would be much more limited.

A possible work-around for this would be to allow the Camel CLI to get its quarkus version from a system property, eg:

```json
{
   "aliases": {
      "camel": {
         "script-ref": "org.apache.camel:camel-jbang-main:4.8.3",
         "main": "main.CamelJBang",
         "arguments": ["--quarkus-version=3.16.0"],
         "properties": {
            "quarkus.version": "3.17.0"
         }
      }
   }
}
```

So you'd get something like:

```
-classpath
"(...)"
-Dquarkus.version=3.17.0
main.CamelJBang
run
.camel-jbang
.camel-jbang-run
somefile.camel.yaml
jbang-catalog.json
--runtime=quarkus
```

quintesse:
Oh, I just thought of another work-around.... except that it doesn't work, which _does_ seem a bug. This is what I thought you could do:

Create 2 aliases, with one referring to the other, eg:

```json
{
   "aliases": {
      "camel": {
         "script-ref": "camel-run", <-- not sure if this works, or that you need "camel-run@apache"
         "arguments": [],
         "properties": {
[...]

PR #1912, opened 2025-01-23
fix: Append alias arguments from jbang-catalog.json to the end of the command (#1910)

bartoszpop:
Append alias arguments from jbang-catalog.json to the end of the command (#1910)


bartoszpop:
Closing the pr as per discussion in #1910

PR #1913, opened 2025-01-23
refactor: merge alias attributes

quintesse:
When one alias refers to another we now merge the attributes of the two aliases (when dealing with lists and maps, in the case of simple values the "parent" alias overrides the "child" alias).

See #1910


maxandersen:
I think this might be right but not fully sure ... it does seem to be breaking behavior but maybe what we had before was worse.

got some examples of what we solve and what we break?

quintesse:
It's difficult, I'm not sure how many examples we have "in the wild" of aliases referring to other aliases. And then they'd both need to have arguments (or other options) defined. Only in those cases the behaviour has been changed.

Thing is, I just tried it and although the PR does what it says it does, it unfortunately doesn't actually fix the problem mentioned in #1910 :-(

Issue #1911, opened 2025-01-23
Jython support in JBang


wfouche:
See https://github.com/orgs/jbangdev/discussions/1909 for a  proposal to add Jython support to JBang.

wfouche:
Might just use Groovy instead.

wfouche:
Link to GraalPy documentation  on Jython compatibility

* https://www.graalvm.org/python/docs/#migrating-jython-scripts

wfouche:
Note: Java, Kotlin and Groovy all have compilers that can compile source code to class files.  Jython used to have a working jythonc compiler, but this compiler is not supported anymore and does not work with Jython 2.7.4 (latest stable version).  It is unclear whether Jython 3.0 will have a jythonc compiler.

maxandersen:
@wfouche doesn't even need a compiler but some kind of boot strap generation to package things and make it runnable would be relevant I reckon?

wfouche:
@maxandersen, here's an initial  attempt at creating a prototype solution to bootstrap a Jython program with JBang. The prototype is written in Python, but can easily be translated to Java.

$ python3   jython-cli.py   test.py

`test.py` (Jython program augmented with JBang-like annotations)
```python
from __future__ import print_function

##DEPS io.leego:banana:2.1.0
##JYTHON 2.7.4

import io.leego.banana.BananaUtils as BananaUtils
import io.leego.banana.Font as Font

text0 = "Jython 2.7"
text1 = BananaUtils.bananaify(text0, Font.STANDARD)

print(text1)
```

![Image](https://github.com/user-attachments/assets/17e4d321-9734-4904-9ed7-01a4fd6f9c5d)

`jython-cli.py`:
```python
from __future__ import print_function
import os
import sys

text = """

import org.python.util.jython;

public class __CLASSNAME__ {

    public static void main(String... args) {
        jython.main(args);
    }
}

"""

def main():
    scriptFilename = sys.argv[1]
    javaClassname = "_" + os.path.basename(scriptFilename)[:-3]
    javaFilename = javaClassname + ".java"
    deps = []
    version = "2.7.4"
    with open(scriptFilename) as f:
        lines = f.readlines()
        tag1 = "##DEPS"
        tag2 = "##JYTHON"
        for line in lines:
            if len(line) > len(tag1):
                if line[:len(tag1)] == tag1:
                    list = line.split()
                    dep = list[1]
                    deps.append(dep)
            if len(line) > len(tag2):
[...]

PR #1927, opened 2025-02-10
docs: show how to run Jython scripts

### LINKED ISSUE #1926: Document how a Jython script can be run using JBang
Jython scripts occasionally require JAR files from Maven Central. Unlike Groovy, Jython lacks a built-in mechanism to resolve these dependencies. However, you can bootstrap a Jython script along with its Maven dependencies using JBang. This technique ensures that your Jython scripts can seamlessly access the necessary dependencies from Maven Central.

Create fiie Jython.java and add all required `DEPS` to it.
```

//DEPS org.python:jython-standalone:2.7.4
//DEPS io.leego:banana:2.1.0

import org.python.util.jython;

public class Jython {

        jython.main(args);
    }

}
```

test.py
```
from __future__ import print_function

import io.leego.banana.BananaUtils as BananaUtils
import io.leego.banana.Font as Font

text0 = "Jython 2.7"
text1 = BananaUtils.bananaify(text0, Font.STANDARD)
print(text1)
```

Running test.py
```
$ jbang run Jython.java test.py
```

Output

Image

wfouche:
Closes #1926

### DIFF
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -375,7 +375,7 @@ public class Jython {
 }
 ----
 
-=== Step 2 - Create Jython script (test.py)
+=== Step 2 - create Jython script (test.py)
 
 .test.py
 [source,python]
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -290,7 +290,6 @@ You have 1 arguments! First is YOLO!%
 Jython scripts occasionally require JAR files from Maven Central.
 Unlike Groovy, Jython lacks a built-in mechanism to resolve external dependencies.
 However, one can bootstrap a Jython script along with its Maven dependencies using JBang.
-The technique illustrated below ensure that Jython can seamlessly access the Maven dependencies.
 
 === Step 1 - create file Jython.java
 
--- a/docs/modules/ROOT/pages/usage.adoc
+++ b/docs/modules/ROOT/pages/usage.adoc
@@ -368,7 +368,7 @@ import org.python.util.jython;
 
 public class Jython {
 
-    public static void main(String... args) {
+    public static void main(String[] args) {
         jython.main(args);
     }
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/usage.adoc
```diff
@@ -285,6 +285,63 @@ Hello, World!
 You have 1 arguments! First is YOLO!%
 ----
 
+== Running Jython (.py) (EXPERIMENTAL)
```
maxandersen, line 288: I grok the incentive to have it here but this is not really jbang support for jython - its example of using the jython library.

wfouche: Totally agree. I'll change the heading and update the pull request

`== Running Jython scripts`

wfouche:
@maxandersen , patiently waiting for this pull request to be looked at again.

albertocavalcante (approved):
LGTM

maxandersen:
Sorry if I wasn't clear - I was not merging this as-is as this is NOT jython support in jbang - it is just showing how to use the jython library like any other library. Also this is just python2 which we really shouldn't promote for people to use.

Furthermore it is not in any way handling caching of the build etc. like jbang is thus I'm not liking that it is listed in same places as the grooy, kotlin etc. which actually does these things.

Remind me again why we can't add proper .py support using jython or graalpy?

quintesse:
> Remind me again why we can't add proper .py support using jython or graalpy?

I don't think there is a _technical_ issue per se. Something can be made that handles .py files.
But "proper" support? I don't know, the whole issue with Python 2 vs 3 and two different possible engines (jython vs graalpy), both with their pros and cons. It seems to me that we'd run the risk of having a majority of people (that want to use Python with JBang) unhappy vs happy.
If we choose to do this than probably jython is easier to implement than graalpy I guess? (Given that jython support can be made to work like we do for kotlin and groovy)

wfouche:
@maxandersen , this pull request can be closed. I've shared this information on the Jython mailing list.

We still have open issue #1911 which is about real Jython support in JBang.

PR #1932, opened 2025-02-25
ci: fix report workflow

quintesse:


quintesse:
Well this either doesn't work, or the action isn't using the fixed workflow file unless it's merged into main. Don't know if you want to test that theory @maxandersen , otherwise we can just close this PR unmerged.

maxandersen:
missed this - fixed in main now

PR #1935, opened 2025-03-04
chore: fix justfile for windows

quintesse:
The actions for opening a web browser with the results from the tests didn't work on Windows. Now it does.

### DIFF
--- a/.justfile
+++ b/.justfile
@@ -1,4 +1,4 @@
-open := if os() == "macos" { "open" } else { "xdg-open" }
+open := if os() == "macos" { "open" } else if os() == "windows" { "start" } else { "xdg-open" }
 
 #@default:
 #    just --choose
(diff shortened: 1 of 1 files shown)


.justfile
```diff
@@ -12,7 +12,7 @@ test:
     ./gradlew test
 
 preitest := if path_exists('build/install/jbang/bin') != 'true' {
-  './gradlew installDist -x test'
+    './gradlew spotlessApply installDist -x test'
```
maxandersen, line 15: why need to run spotlessapply again here?

quintesse: Not sure what you mean by "again". If you've not run the "build:" and want to run the "itests:" it has to do a build, right? It just seemed more consistent to have the same command as the "build:" action.

maxandersen (approved):

PR #1936, opened 2025-03-06
feat: export maven/gradle project

### LINKED ISSUE #1044: jbang export maven/gradle project

tadayosi:
My initial attempt to fix #1044. I know there are still rough edges but at least the following basic usages work. Let me send it for your review so that I can check I'm on the right track.

Add two subcommands to 'jbang export':
* jbang export maven
* jbang export gradle
 
Those generate a template maven/gradle project the user can start developing the script further.

Fix #1044

Usage examples:
```console
jbang export maven hello.java # generates hello/ project with org.example.project:hello:999-SNAPSHOT
jbang export gradle -g dev.jbang.hello -a hello -v 1.0.0 -O hello2 hello.java # generates hello2/ with dev.jbang.hello:hello:1.0.0
```

### DIFF
--- a/docs/modules/ROOT/pages/exporting.adoc
+++ b/docs/modules/ROOT/pages/exporting.adoc
@@ -19,6 +19,8 @@ Note, the local generated jar will have classpath references that are machine de
 jar use `jbang export portable <script>` and the dependent jars will be put in `lib` directory and
 generated jar will have relative references to the jars in the `lib` folder.
 
+You can also use `jbang export gradle|maven <script>` to export the set of the script, its additional sources and resources, along with their dependencies, to a Gradle or Maven project, allowing you to proceed with full-scale development in your favorite IDE.
+
 == Exporting to Maven Repository
 
 If your application or script need to be used from another java project it can be beneficial to publish your jar into a maven repository.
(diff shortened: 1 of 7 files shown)


tadayosi:
@maxandersen @quintesse Please review this.

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -502,3 +498,234 @@ private Path getJlinkOutputPath() {
...
+
+		createProjectForExport(ctx, projectDir);
+
+		return EXIT_OK;
+	}
+
+	private void createProjectForExport(BuildContext ctx, Path projectDir) throws IOException {
+		Project prj = ctx.getProject();
+		String projectName = projectDir.getFileName().toString();
+		Path tmpProjectDir = Files.createTempDirectory(projectName);
```
quintesse, line 561: This line, plus the `Files.move()` on line 590 won't work on Windows if the directory pointed to by `projectDir` is on a different drive than `%TEMP%`.

So I'd just skip the temp folder and immediately copy/create any files directly into the target folder.

quintesse (changes requested):
Looks good! Fix the one thing I mentioned and I think this should be ready to merge.

src/main/java/dev/jbang/cli/Export.java
tadayosi: Thanks! I will change it accordingly.

tadayosi:
@quintesse Applied your feedback. Thanks!

quintesse (dismissed):
Code-wise it all looks good to me! Perhaps you could add a couple of tests?

quintesse:
@tadayosi please be advised I updated your PR. I rebased it on top of main to have it include some recent fixes to the CI process. Without that the CI build would fail.

tadayosi:
@quintesse Added tests for basic scenarios.

tadayosi:
@quintesse Could you tell me if there's anything else that blocks it from merging?  Thank you!

quintesse (dismissed):
LGTM

quintesse:
> @quintesse Could you tell me if there's anything else that blocks it from merging? Thank you!

No, everything looks good. Now we just wait for @maxandersen to merge it. (But he's been very busy so it might take a bit of time)

maxandersen:
looking at this - looks great thus far but really should have something in docs on this. wanna add that in this pr ?

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -502,3 +498,239 @@ private Path getJlinkOutputPath() {
...
+	}
+
+	@Override
+	void renderBuildFile(BuildContext ctx, Path projectDir, String fullClassName) throws IOException {
+		Project prj = ctx.getProject();
+		ResourceRef templateRef = ResourceRef.forResource("classpath:/export-pom.qute.xml");
+		Path destination = projectDir.resolve("pom.xml");
+
+		Map<String, String> properties = new HashMap<>();
+		if (prj.getJavaVersion() != null) {
```
maxandersen, line 693: default jbang java is 8...should we explicilty set it?

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -502,3 +498,239 @@ private Path getJlinkOutputPath() {
...
+			properties.put("maven.compiler.source", prj.getJavaVersion());
```
maxandersen, line 694: this is not handling case of `//JAVA 17+` it shuold remove the +.

maxandersen:
this could probably be handled in updates but just listing here for awareness:

ignores //PREVIEW, breaks on kotlin/other languages, quarkus integration not handled (so `jbang -t qli test.java; jbang export maven test.java` doesnt work), and probably more :)

but definitely a win.

lets at least fix the java version and some basic docs about it.

src/main/java/dev/jbang/cli/Export.java
tadayosi: While it totally makes sense for building JBang itself, I'd guess Java 8 is a bit too restrictive for normal JBang users. I'd imagine most users just expect the default local Java version to be used unless they explicitly set `//JAVA`. I think this can be achieve just by using `JavaUtil.javaVersion`. Let me update the code accordingly and make it always set a Java version.

src/main/java/dev/jbang/cli/Export.java
tadayosi: Thanks for pointing it out. I'll modify it to use `JavaUtil.javaVersion` which should handle `+` cases.

tadayosi: Done

tadayosi:
@maxandersen Thanks for your review! Fixed the java version issue and added basic docs for the new subcommand to `exporting.adoc`. If I'm not mistaken, the CLI Reference docs are something auto-generated later, is that right?  Please review this again.

src/main/java/dev/jbang/cli/Export.java
tadayosi: Ah, thanks for your explanation. So, I agree that we should only set the version if the script explicitly set `//JAVA`. As Tako said, it seems to be really an edge case where the user fails with it.

I'll go ahead and change the PR again so that it strictly follows `//JAVA` and when absent it won't set the Java versions.

tadayosi:
@maxandersen @quintesse Fixed the java version issue again by applying Tako's suggestion. Now I believe using `JavaUtil.minRequestedVersion()` is the right way. I hope this finally gets it right.

docs/modules/ROOT/pages/exporting.adoc
```diff
@@ -45,3 +47,110 @@ You should only need to change `hello.java` to match your application/script.
...
+}
+----
+
+=== To a Gradle project
+
+You can export the script to a Gradle project with the GAV `dev.jbang:myapp:1.0.0-SNAPSHOT` as follows:
+
+[source,bash]
+----
+jbang export gradle --group dev.jbang --artifact myapp --version 1.0.0-SNAPSHOT hello.java
```
maxandersen, line 85:
```suggestion
jbang export gradle --group org.acme --artifact myapp --version 1.0.0-SNAPSHOT hello.java
```

lets not confuse use of the jbang namespace.

tadayosi: Done

src/main/java/dev/jbang/cli/Export.java
maxandersen: i'm -0 on it but lets just see - we can change it later if problematic :)

maxandersen:
@all-contributors please add @tadayosi for code

tadayosi:
@maxandersen @quintesse Thanks!

maxandersen:
just realized this is moving default package names into package names based on GAV - that isn't right. working on a patch.

PR #1943, opened 2025-03-10
fix: Add auth token to githubusercontent requests (#1942)

### LINKED ISSUE #1942: Authorization issue accessing a URL from a private github repository
```
> jbang version
0.124.0
```

If I try to run a JBang script from a private GitHub repo as described in [urls-from-trusted-sources](https://www.jbang.dev/documentation/guide/latest/usage.html#urls-from-trusted-sources) I'm getting a error.

```
> jbang https://github.com/teggr/a-private-repo/blob/main/formatjson/formatjson.java 

[jbang] [ERROR] Could not download https://github.com/teggr/a-private-repo/blob/main/formatjson/formatjson.java
```

If I turn on the verbose logging, I can see that it is unable to access the raw content containing the script.

```
[jbang] [81:278] [ERROR] Could not download https://github.com/teggr/a-private-repo/blob/main/formatjson/formatjson.java
dev.jbang.cli.ExitException: Could not download https://github.com/teggr/a-private-repo/blob/main/formatjson/formatjson.java
        at ...
Caused by: java.io.FileNotFoundException: No file to download at https://raw.githubusercontent.com/teggr/a-private-repo/main/formatjson/formatjson.java. Server replied HTTP code: 404
```

I have my `GITHUB_TOKEN` defined and available in my environment.

I can access this file using a manual curl command. 

```
> curl -v -H "Authorization: token $TOKEN" "https://raw.githubusercontent.com/teggr/a-private-repo/main/formatjson/formatjson.java" 
```

This looks to be related to the code in [Utils::addAuthHeaderIfNeeded](https://github.com/jbangdev/jbang/blob/62526aeb0ddc6ce9ff80063b8892d6cc0bad5c3d/src/main/java/dev/jbang/util/Util.java#L1124).

```
String auth = null;
		if (urlConnection.getURL().getHost().endsWith("github.com") && System.getenv().containsKey("GITHUB_TOKEN")) {
			auth = "token " + System.getenv("GITHUB_TOKEN");
		} else {
```
If I run JBang with debugging, I can see that the value of urlConnection is the raw `https://raw.githubusercontent.com/t` rather than the original `https://github.com/` value so the authentication token is not being application to this url. Which will lead to the `404` response from github and the failure of the script to run.

A suggested fix would be to add the `githubusercontent.com` host to a set of allowed hosts where we add the github token.

teggr:
Added support for `githubusercontent.com` as a location where we load the GITHUB_TOKEN to allow access to private repos

Closes #1942


quintesse (dismissed):
LGTM

quintesse:
@teggr almost. Can you update your PR after having run `./gradlew applySpotless`? Thanks!

teggr:
OK @quintesse . patch applied. thanks - ran against java 11

teggr:
merged in master for the PR, applied the spotless plugin

bumping this for re-review @quintesse or @maxandersen if you find a moment :)

maxandersen:
failure was unrelated github actions issue thus merging as all builds passed.

PR #1949, opened 2025-03-15
fix: export maven/gradle should not touch default packages

maxandersen:
#1936  add export maven/gradle but it modifies default package source files. 

It shouldn't; or rather - I see no point in adding package names unless the user actually need them as it potentially break how things are run/built.

and its perfectly valid java code in a maven or gradle project to have no package.

Also added a few info messages to help users understand its not perfect :)


tadayosi (approved):
Thanks. The original behaviour was just because it was how the Camel CLI works. I totally agree with what you said for JBang.

PR #1953, opened 2025-03-16
fix: avoid downloading java 21 for no good reason

maxandersen:
#1044  added test with `//JAVA 21+` which adds about 20-30 seconds to the runtime of TestExport as it triggers autodownload of java 21 on at least 2 unittests methods.

made it use 11+ instead and things won't be as bad unless you run with 8 :)


maxandersen:
the test error is some blusky remote service specific issue. thus merged anyway.

Issue #1955, opened 2025-03-18
Docs - CLI Reference outdated


tadayosi:
It appears that the CLI Reference hasn't been updated for some time:
https://www.jbang.dev/documentation/guide/latest/cli/jbang.html

For instance, `jbang-export` only lists three subcommands, while the latest at least has six (`native`, `fatjar` and `jlink` are missing).

I tried `./gradlew clidoc` locally but it didn't work for some reason:
```
$ ./gradlew clidoc

> Task :clidoc FAILED
Generating cli man pages
[jbang] Building jar for genadoc.java...
error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider org.eclipse.sisu.space.SisuIndexAPT6 not found
1 error
[jbang] [ERROR] Error during compile
[jbang] Run with --verbose for more details. The --verbose must be placed before the jbang command. I.e. jbang --verbose run [...]

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':clidoc'.
> Process 'command 'jbang'' finished with non-zero exit value 1
```

maxandersen:
hmm - interesting. here i thought I years ago had it autoupdating....checking.

maxandersen:
ok so clidoc is easy fix (disable annotation processing - though proper fix is separately needed as annotation processors shouldn't fail here).

PR #1956, opened 2025-03-18
re-enable cli doc build

maxandersen:
- fix: cli doc generation broken
- fix: add back cli doc in build

#1955 revealed cli-doc was broken - so this fixes it and re-add clidoc building into the build so this shouldn't happen again.


maxandersen:
for now disabling clibuild as it itself is dependent on jbang.

PR #1957, opened 2025-03-19
fix: allow env as alias for java-env

maxandersen:
tired of forgetting its java-env and not just env.

keeps things backwards compatible.


maxandersen:
by accident also includings simple fix to justfile ;)

PR #1967, opened 2025-03-23
refactor: toassertj

maxandersen:
assertj is just nicer IMO - better autocomplete and errormessages are nice

Use this link to re-run the recipe: https://app.moderne.io/builder/Rq6dt8U1O?organizationId=SkJhbmc%3D


maxandersen:
this was an attempt in using Moderne openrewrite to move to assertj - seems it did not catch all ;)

quintesse:
What I find surprising is that it added dependencies but none were removed?

maxandersen:
Hamcrest is default in junit so maybe just not something the recipe does?

PR #1968, opened 2025-03-23
fix: move to use junit+allure over karate

maxandersen:
I love Karate and I'll always be a fan of it but
as time passed I realized it is a big dependency to have
with all the javascript/nashorn compabilitie for something
that for us is just calling out to a shell and does string comparisons.

This starts removing need for using Karate and instead uses plain
JUnit with assertj as gradle integration tests + using Allure to collect intermediate step results.

That last part was/is a key feature of karate where we can see all the steps
before and after a reported failure.

To run the build and see report try out `just itestreport`

end result is to remove karate/itests possibly even the asserta ones (they only run on non-windows anyways) and have it all handled in gradle run and standard junit execution and reporting.


maxandersen:
this is related to #1967  but it is not required we move everything to assertj - can do it iteratively.

maxandersen:
goal of this would eventually remove need for the various workarounds/managing of karate tests.

maxandersen:
added fix to ensure actual built jbang is on path so it will succeed in github action

quintesse:
That would be nice, because running the karate tests always messes up my system somewhat. Hopefully when running from JUnit we can set up a proper test environment.

maxandersen:
Hmm - karate was supposed to run isolated - done the same setup now at least :)

quintesse:
For me it always created a couple of files and modified one. I almost committed those changes by mistake a number of times.

maxandersen:
@quintesse weird thing - I rebased with main and now test on windows fails with it cannot delete tempdirs in only one specific test TestJdk - which I haven't touched in this code.

any idea if things changed somehow with respect to streams/files not being closed?

quintesse:
Well you did make a bunch of changes to the test setup in Gradle for example. So perhaps the tests are being run in a different way? Because the test code in main doesn't have this error so it must be some combination with your code that causes this.

I'd try to see what path exactly is causing the error. And who is trying to delete it at what point in the code.

maxandersen:
if you spot something you believe could cause changes to the main test run i'm all ears. all seem innocent to me (at least for now ;)

quintesse:
The Gradle changes are a big shout-out to me. Changes related to tests -> suddenly test start failing. How? No idea, Gradle is like black magic to me :-)

quintesse:
Ah yes, I've seen this before, it's a bug in jUnit :-)

quintesse:
See this issue I opened for it 2 months ago: https://github.com/junit-team/junit5/issues/4299

quintesse:
It seems the fix was released in version 5.12, I've tried and changed the deps in Gradle to 5.12.2 and the problem goes away.

Again one of those fine Windows special cases where everything needs to work differently just for the fun of it. ;-)

quintesse:
Also see this code in devkitman: https://github.com/jbangdev/jbang-devkitman/blob/main/src/test/java/dev/jbang/devkitman/BaseTest.java#L69-L73

```java
	// TODO Remove once https://github.com/junit-team/junit5/issues/4299 is released
	@AfterEach
	protected void cleanupEnv() throws IOException {
		FileUtils.deletePath(config.installPath);
	}
```

maxandersen:
Finally Green! gdman \r\n lines in windows :)

quintesse:
Curious, I'm getting a couple of errors:

1. `RunIT/shouldPassStdIn` - uses `cat` which doesn't exist on Windows
2. `JshIT/shouldForceJsh` - "Expected error output to be empty but was <WARNING: Using incubator modules: jdk.incubator.foreign, jdk.incubator.vector". Probably depends on the Java version used to run the tests
3. `JshIT/shouldForceJshSources` - same
4. `ExportIT/testBasicExportNoClasspath` - "Expected exit code to be <0> but was <2>", I'm guessing it's because the first line of the test tries to run `rm`, which of course doesn't exist on Windows

How did 1 and 4 not fail during the GitHub actions build?

PR #1969, opened 2025-03-23
chore(deps): sdkman, update java to 11.0.26-tem

wfouche:
Set Gradle=8.9 and Java=11.0.26

### DIFF
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -2,4 +2,7 @@
 # Add key=value pairs of SDKs to use below
 
 # Java
-java=11.0.25-tem
+java=11.0.26-tem
+
+# Gradle
+gradle=8.9
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -3,9 +3,3 @@
 
 # Java
 java=11.0.26-tem
-
-# Gradle
-gradle=8.9
-
-# Maven
-maven=3.9.9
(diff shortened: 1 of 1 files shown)


maxandersen (approved):

wfouche:
@maxandersen ,  added Maven to .sdkmanrc since it is required to be able to test / use "export maven" functionality on a clean dev system.

wfouche:
@maxandersen , please review and merge (if you're approve).

.sdkmanrc
```diff
@@ -2,4 +2,10 @@
 # Add key=value pairs of SDKs to use below
 
 # Java
-java=11.0.25-tem
+java=11.0.26-tem
+
+# Gradle
+gradle=8.9
+
+# Maven
```
maxandersen, line 10: why set mvn here? jbang itself doesn't require/use it?

quintesse: And really there's no need for Gradle either because we're using the Gradle wrapper (which is recommended over using your own)

maxandersen: ah its for easier testing of jbang export gradle/mvn

...but that actually is a good point- maybe the exports should add mvn/gradle wrapper instead?

wfouche: > .but that actually is a good point- maybe the exports should add mvn/gradle wrapper instead?

Agree.  That would solve the problem. The wrappers are missing at the moment.

wfouche: In anticipation of wrappers being generated in the future for Maven and Gradle, I have removed the entries for them from .sdkmanrc

PR #1970, opened 2025-03-23
fix: export gradle, set jvmArgs and compilerArgs in build.gradle

### LINKED ISSUE #1961: build.gradle: should define applicationDefaultJvmArgs
**Describe the bug**
To ensure the same runtime behaviour as if the application is run directly using JBang, when using Gradle the same JVM options originally defined via setting JBANG_JAVA_OPTIONS should be carried over to the Gradle runtime environment in the build.gradle file created by `jbang export gradle App.java`.

**Expected behavior**
Property `applicationDefaultJvmArgs` should contain values sourced from JBANG_JAVA_OPTIONS, otherwise it should be empty.

If `JBANG_JAVA_OPTIONS` is not defined, then applicationDefaultJvmArgs is empty.
```
application {
    .....
    applicationDefaultJvmArgs = []
}
```

If `JBANG_JAVA_OPTIONS` is  defined, then populate applicationDefaultJvmArgs accordingly. For example if `JBANG_JAVA_OPTIONS` has value `-server -Xmx2g -XX:+UseZGC -XX:+ZGenerational`
```
application {
    .....
    applicationDefaultJvmArgs = ['-server', '-Xmx2g', '-XX:+UseZGC', '-XX:+ZGenerational']
}
```

**JBang version**
0.125.0

wfouche:
Closes #1961


wfouche:
```
//RUNTIME_OPTIONS -server -Xms2g -Xmx2g -XX:+UseZGC -XX:+ZGenerational
```

is translated to

```
applicationDefaultJvmArgs = ['-server', '-Xms2g', '-Xmx2g', '-XX:+UseZGC', '-XX:+ZGenerational']
```

wfouche:
`//PREVIEW`

is translated to 

`applicationDefaultJvmArgs = ['--enable-preview']`

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -660,6 +660,16 @@ void renderBuildFile(BuildContext ctx, Path projectDir, String fullClassName) th
 		boolean isGroovy = ctx.getProject().getMainSource() instanceof GroovySource;
 		boolean isKotlin = ctx.getProject().getMainSource() instanceof KotlinSource;
 		String kotlinVersion = isKotlin ? ((KotlinSource) ctx.getProject().getMainSource()).getKotlinVersion() : "";
+		StringBuilder jvmArgs = new StringBuilder();
+		if (ctx.getProject().enablePreview()) {
```
maxandersen, line 664: if you add enable-preview support you'll need to enable on the compile side too. otherwise wont do much.

wfouche:
ci-build failed, please restart the failed jobs.  The build passes all tests on my dev machine.

quintesse:
Jobs restarted (you see why we need https://github.com/jbangdev/jbang/pull/1941 @maxandersen :-) )

Btw, I assume that if we have support for `//RUNTIME_OPTIONS` -> `applicationDefaultJvmArgs ` we also need support (maybe even more important than the runtime options) for `//COMPILE_OPTIONS` -> `compilerArgs`?

src/main/resources/export-build.qute.gradle
```diff
@@ -43,9 +43,12 @@ dependencies {
 application {
 	mainClass = '{fullClassName}'
 	applicationDefaultJvmArgs = [{jvmArgs}]
+{#if enablePreview eq 'true'}
+	applicationDefaultJvmArgs += ['--enable-preview']
+{/if}
 }
 
-{#if enableJavaPreview eq 'true'}
+{#if enablePreview eq 'true' and language eq 'java'}
```
maxandersen: I *think* you also need to set specific release/source flag  - you can see how we do it in jbang itself at

https://github.com/maxandersen/jbang/blob/main/src/main/java/dev/jbang/source/buildsteps/CompileBuildStep.java#L74

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -660,6 +660,20 @@ void renderBuildFile(BuildContext ctx, Path projectDir, String fullClassName) th
...
 		String kotlinVersion = isKotlin ? ((KotlinSource) ctx.getProject().getMainSource()).getKotlinVersion() : "";
+		StringBuilder jvmArgs = new StringBuilder();
+		for (String arg : ctx.getProject().getRuntimeOptions()) {
+			if (jvmArgs.length() > 0) {
+				jvmArgs.append(", ");
+			}
+			jvmArgs.append(String.format("'%s'", arg));
+		}
+		StringBuilder compilerArgs = new StringBuilder();
+		for (String arg : ctx.getProject().getMainSourceSet().getCompileOptions()) {
```
quintesse, line 671: Looks good in general, but perhaps you could introduce a method for this to avoid the code duplication?

src/main/resources/export-build.qute.gradle
```diff
@@ -42,4 +42,19 @@ dependencies {
...
+	applicationDefaultJvmArgs += ['--enable-preview']
+{/if}
 }
+
+{#if language eq 'java'}
+compileJava {
+{#if enablePreview eq 'true'}
+	options.compilerArgs += ['--enable-preview']
+{/if}
+{#if compilerArgs ne ""}
```
quintesse, line 56: Compiler args are not only for Java, they can also be for Kotlin or Groovy

quintesse (changes requested):
Just a couple of additional changes and it will be perfect.

src/main/resources/export-build.qute.gradle
quintesse: @maxandersen Just as an aside, we should probably use `--release` instead of `-source` for JDKs >= 9. I'll open an issue for that

wfouche: //PREVIEW now adds  the following additional compiler options for Java.

```
options.compilerArgs += ['--release', '{javaVersion}', '--enable-preview', '-Xlint:preview']
```

I see that others recommend adding -Xlink:preview to provide a  reminder to the user of the  specific preview features that are in use.

```
App/src/main/java/App.java:15: warning: [preview] unnamed classes are a preview feature and may be removed in a future release.
void main(String[] args) {
^
1 warning
```

src/main/java/dev/jbang/cli/Export.java
wfouche: Created method gradleArgs(List<String> options)

src/main/resources/export-build.qute.gradle
wfouche: Functionality implemented and tested with the following language specific compiler options:

Groovy
* //COMPILE_OPTIONS --tolerance=5

Kotlin
* //COMPILE_OPTIONS -progressive

quintesse (dismissed):
LGTM, thanks @wfouche !

wfouche:
Implemented a minor code cleanup change, to use consistently use prj instead of prj and ctx.getProject()

wfouche:
Fixed generating build.gradle correctly when setting //PREVIEW but not setting the Java version with //JAVA as well.

wfouche:
Done losts of testing with Java, Groovy and Kotlin projects, they all converted flawlessly to Gradle projects that build and run without issues. I think I'll stop here.

wfouche (commented):
Done lots of testing with Java, Groovy and Kotlin projects, they all converted flawlessly to Gradle projects that build and run without issues. I think I'll stop here.

wfouche:
I'm creating additional tests to test the new export-gradle functionality. Will submit the changes when properly tested.

quintesse (commented):
Does last commit mean that you _have_ to use the `@JvmStatic` annotation to make things work?

quintesse:
Ok, I just found an issue with the project export that's also present in the currently merged code.

For example you're using `classpath_log.java` in your tests and your code assumes that the class name is the same as the file name, but that's actually not the case! It's `classpath_example`. But that's okay, it's not recommended to do that perhaps, but as you can see it compiles just fine.

Thing is that I didn't look closely enough last time when I okayed your PR because the full class name is already available in the Project! You can simply do `String fullClassname = prj.getMainClass()`. It will use the name from the compiled code, so if languages like Kotlin do any "name mangling" (like adding "Kt") it should return the correct name.

wfouche:
>  You can simply do String fullClassname = prj.getMainClass()

I tried this change but then assertThat on line 295 of TestExport.java fails

```
assertThat(build, containsString("mainClass = 'classpath_log_bom'"));
```

So I had to settle for:

```
data("fullClassName", isKotlin ? prj.getMainClass() : fullClassName)
```

wfouche:
Thanks @quintesse and @maxandersen for the guidance you provided.

quintesse:
> > You can simply do String fullClassname = prj.getMainClass()
> 
> I tried this change but then assertThat on line 295 of TestExport.java fails

Well that has to be investigated, because if that value is incorrect than we have a bigger problem :-)

wfouche:
> @wfouche I pushed some fixes to your branch. The fullClassname you were using isn't needed, the tests were wrong, not the code :-)

Fantastic! Thanks.

wfouche:
@maxandersen and @quintesse , I think this it. I've got nothing else to push to this branch.

Please review the changes.

maxandersen:
> Definitely not recommended, as Java expects the filename to be the same as the classname if the file contains only one class. I think the Java test class name should be changed to classpath_log. I tested this change, and all the tests succeeded.

just to educate/inform - it is incorrect that java expects the filename to match class - that is only the case for public classes. And there is no public class here. So its perfectly valid java.

maxandersen:
looks good!

PR #1983, opened 2025-03-28
refactor: Now using `devkitman` module

### LINKED ISSUE #1857: Refactor JDK management code into separate module

### LINKED ISSUE #1950: jbang/devkitman breaks jdk lookup logic
something is not right.
I'm running latest jbang test suite and a request for java 21+ turns into java 25 requests because my ~/.jbang/ jdk caches a 25 version.

blocking next jbang release.

quintesse:
Second try.

Fixes #1857
Should also fix #1950


maxandersen:
is this supposed to still be draft?

quintesse:
Well, I had left it in draft to take a good look at the problem you found, but I can't find anything.
So perhaps you can take it for a spin and report if you find anything weird?

quintesse:
Tests failed, but they are the typical ones that have to do with our tests hitting rate limits. So have to wait a bit and re-run them.

quintesse:
Ok, tests passed.

maxandersen:
rebased

maxandersen (approved):
lgtm!

quintesse:
Love this nitpick:

> [nitpick] The constant name DEFAULT_JDK paired with the value 'currentjdk' might be confusing; consider aligning the naming convention so that the name and value clearly reflect its purpose.
public static final String DEFAULT_JDK = "currentjdk";

Because it's totally right :-)
And I would definitely change it if I could, but we already have the command `jbang jdk default` (which is what I decided to name everything in the code) and the symbolic link that's been `currentjdk` already for ages. No way to change any of those two without breaking backward compatibility unfortunately. Still, cool it catches things like that.

PR #1991, opened 2025-04-07
fix: export maven/gradle, install runtime wrappers

### LINKED ISSUE #1987: Export maven/gradle should create Maven/Gradle wrappers
pom.xml or build.gradle files are currently generated by the export command, but wrappers for maven or gradle are not generated.

(1) Display a message to the user saying please initialize the exported project by running command:

* mvn wrapper:wrapper -Dmaven=3.9.9, or
* gradle wrapper --gradle-version 8.9

(2)  Run the wrapper commands automatically from JBang (assuming Maven and Gradle are pre-installed).

(3) Generate  init_wrapper.{sh,cmd) files that can be used to generate the wrapper files.

init_wrapper.{sh,cmd} will either contain "mvn wrapper:wrapper" or "gradle wrapper"

Which solution is preferred?

wfouche:
Closes #1987


wfouche:
This code has been tested to work with Maven and Gradle on Linux and Windows.

quintesse:
Looks like a good start, but why did you remove the JAR from the resources?

wfouche:
> Looks like a good start, but why did you remove the JAR from the resources?

Encountered a weird issue. File gradle-wrapper.jar (binary file) was not copied to the JBang JAR file, so it could not be read at runtime.  Does thiis make any sense?

So I reverted to downloading the file at runtime. But would prefer to have it in the JBang JAR if possible.

quintesse:
Aha, I guess it must be something in the Gradle assembly process that filters out files that it thinks should not be part of the output.

quintesse:
Just as a test, try renaming the file to give it a different extension from ".jar" and see if that works.

wfouche:
> Just as a test, try renaming the file to give it a different extension from ".jar" and see if that works.

Renamed it to -jar instead of .jar; and it works fine at runtime.

Will update the pull request shortly.

quintesse:
Ok, good to know, still I'd like to find out how we can get it to work with its original name.
Why? Because if we keep the original name updating the wrappers to neewer versions is much easier: we can just go into the resources/dist/gradle or maven folder and run the required gradle/maven command to update the files and commit.

wfouche:
> Ok, good to know, still I'd like to find out how we can get it to work with its original name.

It might be added to the classpath if it is stored as a .jar file, and then it will be loaded and the classed it contains made available to the JVM. Maybe not what we want.

I think it would be fine to keep it as a -jar file, and also simple enough to rename the file when updating it in the future.

wfouche:
Adding scrupt `update-version.sh` to the dist/gradle folder to be used in the future to update the Gradle wrapper files.

```
#!/bin/bash

export GRADLE_VERSION=8.9

mv gradle/wrapper/gradle-wrapper-jar gradle/wrapper/gradle-wrapper.jar

touch build.gradle
touch settings.gradle

./gradlew wrapper --gradle-version $GRADLE_VERSION

mv gradle/wrapper/gradle-wrapper.jar gradle/wrapper/gradle-wrapper-jar

rm -f -r .gradle build.gradle settings.gradle

tree -a

git status
```

wfouche:
Created an `update-version.sh` script for Maven as well.

quintesse:
> Adding scrupt `update-version.sh` to the dist/gradle folder to be used in the future to update the Gradle wrapper files.

Good enough 👍

wfouche:
Done with coding and testing; ready to be reviewed.

wfouche:
@quintesse , please restart the build. (smile)

src/main/java/dev/jbang/cli/Export.java
```diff
@@ -619,6 +626,34 @@ String getJavaVersion(Project prj, boolean minorVersionFor8) {
...
+		dstDir.mkdirs();
+		InputStream ifs = this.getClass().getResourceAsStream(srcPath);
+		Files.copy(ifs, Paths.get(dstPath), StandardCopyOption.REPLACE_EXISTING);
+
+		boolean isWindows = System.getProperty("os.name").toLowerCase().contains("windows");
+
+		if (isWindows)
+			return;
+
+		if (execFlag) {
```
quintesse, line 643: Ok, everything is perfect for merging, but .... I wonder if you could do me a favor and perform a small refactor: there is already some code in JBang to set a file as executable (See https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/cli/App.java#L192). Could you perhaps extract that, move it to Util.java (just below the isExecutable() method for example) and make Export use that one as well?

wfouche: That's a good idea. I'll implemented the requested changes.

quintesse (approved):

quintesse:
Thanks for your effort @wfouche !

PR #1992, opened 2025-04-08
chore: minor export cleanup/improvements

quintesse:
There was a bit too many instanceof tests for my taste so I made the `Source` return its `Type` (that was an existing type but it wasn't being exposed) and extended the `Type` to include the "maven source folder" name (which is the name of the folder that sources should be in when using the maven project structure, eg: "src/main/**java**")

I also made sure that the code will exit with an error when trying to export an unsupported language (especially useful now that people are looking to support other languages like Python/Jython)

### DIFF
--- a/src/main/java/dev/jbang/cli/Export.java
+++ b/src/main/java/dev/jbang/cli/Export.java
@@ -22,7 +22,6 @@ import dev.jbang.catalog.CatalogUtil;
 import dev.jbang.dependencies.*;
 import dev.jbang.source.*;
 import dev.jbang.source.resolvers.AliasResourceResolver;
-import dev.jbang.source.sources.GroovySource;
 import dev.jbang.source.sources.KotlinSource;
 import dev.jbang.util.JarUtil;
 import dev.jbang.util.JavaUtil;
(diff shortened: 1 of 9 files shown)


wfouche:
@quintesse,  I like the refactoring you've done. Using EnumSet will further simplify the code.

```
  Source.Type srcType = prj.getMainSource().getType();
  if (srcType != Source.Type.java && srcType != Source.Type.groovy && srcType != Source.Type.kotlin) {
	  throw new ExitException(EXIT_INVALID_INPUT, "Unsupported source type: " + srcType.name());
  }
```

could be replaced by

```
EnumSet<Source.Type> SupportedSourceTypes = EnumSet.of(Source.Type.java, Source.Type.groovy, Source.Type.kotlin);
....
Source.Type srcType = prj.getMainSource().getType();
if (!SupportedSourceTypes.contains(srcType)) {
    throw new ExitException(EXIT_INVALID_INPUT, "Unsupported source type: " + srcType.name());
}
```

maxandersen (dismissed):

maxandersen:
looks good. merge when you all happy :)

p.s. i just for fun tried requesting copilot to see whta it would do

quintesse:
> Using EnumSet will further simplify the code

I was actually looking how to do that yesterday because I couldn't remember what the exact name of that helper class was, but got distracted. Thanks!

PR #2011, opened 2025-05-08
test: make sure itests leave project directory clean

quintesse:
Some itests leave create files in the project folder and leave them around.
This then complicates committing any changes because you have to be careful that you don't accidentally commit those files as well.
These changes avoid that by making a temporary copy of the entire itests folder and run the tests from there.

Also fixed one test that I guess depended on the order that the tests were run in because there doesn't exist a `helloworld.jar` in the itests folder.


maxandersen:
good fix - the hellojar.jar is funny as it was helloworld.jar in old tests so apparently always been an issue :)

PR #2013, opened 2025-05-12
feat: handle documentation references both in source references and in alias catalog

### LINKED ISSUE #174: Allow for Script documentation
Some scripts are getting pretty complicated, more so than perhaps is easily handled by any `--help` options. So it would be nice to have a command like:

```
jbang docs alias@catalog
```

Possible options are to look for `*.html`, `*.md` or `*.txt` files with the same name as the script and open them up somehow.
We could also add a `docs-url` to the entries in a Catalog file and open that in a browser.

andsel:
- Closes #174 

## Feature description
Implements a new `DOCS` directive that permit insert a link or file reference that contains the documentation for the script. The documentation reference can be also assigned to an alias of a script and in that case takes precedence over the directive present in the script.
The documentation reference can be opened with a newly added info subcommand, name docs, like: 
```
jbang info docs <script|alias>
```

## Describe changes at code level
The main change is to collect and resolve the reference present in the new `DOCS `directive in `TagReader.getDocs`.
During JBang project creation or update the docs reference is collected and stored, to be later read in the `ScriptInfo` helper class.
On the `cli.Alias` command a new `--docs` parameter is added to be accept a manually provided doc reference and store in the catalog. This catalog information is used in the `ScriptInfo`'s initialization, after a resolution to a ResourceRef, to eventually overwrite the directive coming form the script.

## How to test
Use a script like following for the testing part:
```java
//DESCRIPTION This is a description test
//DOCS this_exists.txt
//DOCS ./this_exists.txt
//DOCS /tmp/this_exists.txt
//DOCS http://www.jbang.dev/documentation/guide/latest/index.html
//DOCS does-not-exist.txt

public class docsexample {

        out.println("Hello " + (args.length>0?args[0]:"World"));
    }
}
```

Build a new JBang distribution and and uncompress in tmp folder:
```
> ./gradlew distTar
> cp build/distributions/jbang-0.126.1.12.tar /tmp
> cd /tmp
> tar xf jbang-0.126.1.12.tar
```

In the same `tmp` folder create a new file `this_exists.txt` with some content.

### Browse doc from script
```
/tmp/jbang-0.126.1.12/bin/jbang info docs docsexample.java
```
Opens a browser or an editor.

### Browse doc from alias
Add an alias overriding the file's contained directive and check it opensù
```
/tmp/jbang-0.126.1.12/bin/jbang alias add --docs https://www.jbang.dev/documentation/guide/latest/javaversions.html docsexample.java
/tmp/jbang-0.126.1.12/bin/jbang alias list --format=json
/tmp/jbang-0.126.1.12/bin/jbang info docs docsexample
```

Check it opens the web browser to https://www.jbang.dev/documentation/guide/latest/javaversions.html .

Remove the alias
```
/tmp/jbang-0.126.1.12/bin/jbang alias remove docsexample
/tmp/jbang-0.126.1.12/bin/jbang alias list --format=json
```

Add an alias without a docs ref forcing
```
/tmp/jbang-0.126.1.12/bin/jbang alias add docsexample.java
/tmp/jbang-0.126.1.12/bin/jbang info docs docsexample
```

Verify if opens the editor or the browser pointing to the resource defined in the first `DOCS` directive
## Dev tasks
- [x] complete the implementation of display the doc reference
- [x] cover with unit test
- [x] implement the catalog retrieval, giving it precedence over source file directive
- [ ] check if integration tests are feasible
- [ ] update the documentation

## Summary by Sourcery

Implement support for documentation references via a new DOCS directive and alias option, and expose them through an `info docs` command to list and optionally open linked documentation.

New Features:
- Introduce a DOCS directive to annotate scripts with documentation file or URL references
- Add an `info docs` subcommand to display and open documentation references for scripts and aliases
- Enable specifying documentation references for aliases using a new `--docs` option

Enhancements:
- Extend the project and alias catalog to collect and resolve DocRef instances
- Prioritize alias-provided documentation references over inline script directives

Tests:
- Add unit tests for file and URL documentation resolution
- Add integration tests to verify `info docs` output and browser-opening behavior


src/main/java/dev/jbang/cli/Info.java
```diff
@@ -308,3 +317,42 @@ public Integer doCall() throws IOException {
 		return EXIT_OK;
 	}
 }
+
+@CommandLine.Command(name = "docs", description = "Open the documentation file into the default browser.")
```
quintesse, line 321: Change "into" -> "in"

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -308,3 +317,42 @@ public Integer doCall() throws IOException {
...
+		ScriptInfo info = getInfo(false);
+		System.out.println("Command invoked: " + info.docs);
+
+		URI uri = validateDocsReferenceAndTransformToUri(info);
+		Desktop.getDesktop().browse(uri);
+		return EXIT_OK;
+	}
+
+	private static URI validateDocsReferenceAndTransformToUri(ScriptInfo info) {
+		if (info.docs.startsWith("http")) {
```
quintesse, line 335: Once the type of info.docs is `ResourceRef` none if this is necessary because that type handles all of the URL vs absolute vs relative paths etc.

src/main/java/dev/jbang/source/Project.java
```diff
@@ -37,6 +37,7 @@ public class Project {
 	private final Map<String, String> manifestAttributes = new LinkedHashMap<>();
 	private String javaVersion;
 	private String description;
+	private String docs;
```
quintesse, line 40: The type of this shouldn't be a `String` but a `ResourceRef`

src/main/java/dev/jbang/source/Project.java
```diff
@@ -180,6 +181,17 @@ public Project setDescription(String description) {
 		return this;
 	}
 
+	@Nonnull
+	public Optional<String> getDocs() {
```
quintesse, line 185: Same here, use `ResourceRef`

src/main/java/dev/jbang/source/TagReader.java
```diff
@@ -153,6 +154,17 @@ protected boolean isDescriptionDeclare(String line) {
 		return line.startsWith(DESCRIPTION_COMMENT_PREFIX);
 	}
 
+	public Optional<String> getDocs() {
```
quintesse, line 157: Don't return a `String` but a `ResourceRef`. You can look at methods like `collectFiles()` on how to do that. (Although in this case the method would be much simpler because //DOCS doesn't need to support wildcards. But it does need a resourceRef and a siblingResolver like we do for //SOURCES and //FILES). Don't hesitate to ask me questions if you get stuck!

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -94,6 +97,8 @@ static class ScriptInfo {
 		String description;
 		String gav;
 		String module;
+		String docs;
```
quintesse, line 100: The type of this shouldn't be a `String` but a `ResourceRef`. And once that is done you don't need `rootFile`.

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -179,6 +184,10 @@ private void init(BuildContext ctx) {
 			}
 			gav = prj.getGav().orElse(null);
 			description = prj.getDescription().orElse(null);
+			docs = prj.getDocs().orElse(null);
+			if (prj.getMainSource() != null) {
```
quintesse, line 188: This `if` can be removed once we switch to `ResourceRef` (NB: any relative paths are relative to `Project.getResourceRef()`, not to the source files).

quintesse (changes requested):
A good first step! But there's one important change that has to be made, which is to change the type of docs from `String` to `ResourceRef`. I've left comments on the relevant sections. It makes the code slightly more complicated but it makes sure the existing machinery in JBang takes care of locating the files.

src/main/java/dev/jbang/source/TagReader.java
andsel: Hi @quintesse thank's a lot for your first feedback and suggestion. It's not clear to me why the `getDoc` method should use a `SiblingResourceResolver`.
I think of this possible implementation:
```java
public Optional<ResourceRef> getDocs(ResourceResolver resolver) {
	return getTags()
		.filter(this::isDocsDeclare)
		.map(s -> s.substring(DOCS_COMMENT_PREFIX.length()))
		.map(resolver::resolve)
		.findFirst();
}
```

Invoking from `ProjectBuilder.createJbangProject` as
```java
prj.setDocs(tagReader.getDocs(ResourceResolver.forResources()).orElse(null));
```

So why should it use a `SiblingResourceResolver` ?

quintesse: > So why should it use a `SiblingResourceResolver` ?

Because a user can do `jbang info docs http://some/repo/or/url/code.java` and if the code.java file has a line like `//DOCS mydocs.html` then the html file has to be retrieved from `http://some/repo/or/url/mydocs.html` and that's what the resource refs are for.

quintesse: Btw if you look at how `TagReader.collectFiles()` is called you'll see it simply wraps `ResourceResolver.forResources()` in a `mew SiblingResolver()`, eg:

```java
ss.addResources(tagReader.collectFiles(resourceRef,
	new SiblingResourceResolver(resourceRef, ResourceResolver.forResources())));
```

So, it's not much more complicated.

quintesse: Oh, I see there might also be somewhat of a miscommunication: I'm not saying the *type* of the argument should be a `SiblingResourceResolver`, in fact it should be a plain `ResourceResolver`.

Explanation: the way the resolver will be used is always to resolve siblings (when using relative paths the docs are "siblings" of the main source file, as are source and resource files). You can see this in the implementation of `collectSources()` and `collectFiles()`, the parameter is `ResourceResolver siblingResolver`, the "sibling" is only in the name.

quintesse: Sure, although I wouldn't worry about that too much right now. You have to think that possibly the documents aren't files that can be shown directly in the browser. Think of .md and .adoc files that have to be transformed first, which means having them locally available is most likely required anyway.

Still I could imagine the use for something like a `LazyResourceRef` that would delay the resolving part to the moment that somebody actually wants to access the contents of the resource. Then as long as the code that displays the docs never asks for the contents (because it simply passes the url to a browser)  it would never get downloaded. But that could be a future improvement, so again, don't worry about it now.

andsel: Fixed by dcee2c8f3913144b4027efc8e94d84bd8e2d1ce1

src/main/java/dev/jbang/cli/Info.java
andsel: Fixed by dcee2c8f3913144b4027efc8e94d84bd8e2d1ce1

src/main/java/dev/jbang/source/TagReader.java
```diff
@@ -153,6 +155,27 @@ protected boolean isDescriptionDeclare(String line) {
...
 
+	public Optional<ResourceRef> getDocs(ResourceResolver resolver) {
+		return getTags()
+			.filter(this::isDocsDeclare)
+			.map(s -> s.substring(DOCS_COMMENT_PREFIX.length()))
+			.map(TagReader::createResourceReference)
+			.findFirst();
+	}
+
+	private static ResourceRef createResourceReference(String s) {
```
quintesse, line 166: No, this isn't how it's supposed to be done, you _have_ to use the resolver. Like I said before, don't worry about things being downloaded, that's something we can fix later.

quintesse (changes requested):
Much better, there's still a couple of minor things and one major _must fix_ item.

src/main/java/dev/jbang/catalog/Alias.java
```diff
@@ -58,6 +58,8 @@ public class Alias extends CatalogItem {
 	public final Map<String, String> manifestOptions;
 	@SerializedName(value = "java-agents")
 	public final List<JavaAgent> javaAgents;
+	@SerializedName(value = "docs-ref")
```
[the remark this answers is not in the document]
andsel: solved by ca23d2880668d0958bc9ac40636fbf837e6fff78

src/main/java/dev/jbang/catalog/Alias.java
```diff
@@ -118,6 +120,7 @@ public Alias(String scriptRef,
 			Boolean enableSystemAssertions,
 			Map<String, String> manifestOptions,
 			List<JavaAgent> javaAgents,
+			String docsRef,
```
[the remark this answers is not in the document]
andsel: solved by ca23d2880668d0958bc9ac40636fbf837e6fff78

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -563,6 +564,7 @@ private Project updateProject(Source src, Project prj, ResourceResolver resolver
 			ss.addNativeOptions(src.getNativeOptions());
 			prj.addRepositories(src.tagReader.collectRepositories());
 			prj.addRuntimeOptions(src.getRuntimeOptions());
+			prj.setDocs(src.tagReader.getDocs(new SiblingResourceResolver(srcRef, ResourceResolver.forResources())).orElse(null));
```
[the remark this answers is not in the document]
andsel: solved by ca23d2880668d0958bc9ac40636fbf837e6fff78

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -321,6 +321,7 @@ private Project createJbangProject(ResourceRef resourceRef) {
 		TagReader tagReader = new TagReader.JbangProject(contents,
 				it -> PropertiesValueResolver.replaceProperties(it, getContextProperties()));
 		prj.setDescription(tagReader.getDescription().orElse(null));
+		prj.setDocs(tagReader.getDocs(new SiblingResourceResolver(resourceRef, ResourceResolver.forResources())).orElse(null));
```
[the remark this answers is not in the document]
andsel: solved by ca23d2880668d0958bc9ac40636fbf837e6fff78

andsel:
@quintesse thank's a lot for the addition, so it has to always download the url and then let it open in the browser.

I'll proceed by aligning the documentation.

quintesse:
> so it has to always download the url and then let it open in the browser.

No, that's just a temporary implementation detail that can be fixed later. It's of no consequence to the docs feature.

andsel:
@quintesse which is the best place to document this feature?
- should I add a new section (like `caching.adoc`) to describe what's this documentation feature it?
- -or- add paragraph in the "Aliases & Catalog" ? I propose this because the feature also touches the aliases creation, when a an alias can be created with a specific link to the documentation for it, which overrides the tag present in the script.

quintesse:
Btw, I'd be willing to merge this PR without the docs and leave that to be done in #2032

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -182,10 +185,26 @@ private void init(BuildContext ctx) {
 			}
 			gav = prj.getGav().orElse(null);
 			description = prj.getDescription().orElse(null);
-			docs = prj.getDocs().orElse(null);
+			initDocsReference(prj);
+
 			module = prj.getModuleName().orElse(null);
 		}
 
+		private void initDocsReference(Project prj) {
```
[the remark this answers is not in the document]
quintesse: Hmm wait a sec, what is it exactly what you are trying to accomplish here?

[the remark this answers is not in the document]
quintesse: > We can move this

Just do what I said in my last message and it will do exactly what you propose

quintesse (dismissed):
LGTM

andsel:
@quintesse thanks a lot for your all your guidance on this! :-D

quintesse:
> @quintesse thanks a lot for your all your guidance on this! :-D

And thank you for your implementation and in moving this issue forward! :-)

maxandersen:
trying this out and having a few issues:

- failing on file with no jbang/docs in them; it should help/guide the user not just fail.

```
just jbang info docs examples/readme.adoc
[jbang] [ERROR] Cannot invoke "dev.jbang.source.ResourceRef.isURL()" because "info.docs" is null

- fail on file with multiple //DOCS and first one is missing

This just fails on the file not being there - but there are multiple others that does exist...can't / shouldn't we show it?

As mentioned on #174 "I would treat it as first (or last dependent on pov) one wins and everything else is secondary." - in case of first one not being there I would print/show the other options.

- auto-open on urls

I'm not sure its super safe to open browser or default editor on a file/resource - could be tricky...I would suggest we give a chance to cancel it. Either asking expliclitly or do the "wait 5 seconds" to open as we do for trust checking.

Could still have a `jbang info docs --open test.java` to just open directly 

- ignoring multiple //DOCS
I would print all the docs found, not just open first one.

- not a blocker, but should the list of docs found not be listed in `jbang info tools` as a new key info?

**Comment by quintesse** [comment]
> This just fails on the file not being there - but there are multiple others that does exist...can't / shouldn't we show it?

Dunno, that what likely make the code a whole lot more complex for something that shouldn't happen (a file shouldn't have multiple //DOCS). It would be a bit like a file having both a `//JAVA 170+` and a `//JAVA 17+` and then saying "why can't it just pick the one that is correct?". Yes, we could, but should we?

> in case of first one not being there I would print/show the other options.

~But that would require doing a resolution at a point that you don't even know if the user is going to want the information. (You don't know if they'll be running `jbang info` or `jbang run`). Again, related to the example above, if we'd try the different ??JAVA options we'd be doing possibly several failing remote request only to find out that we want Jdk 17 which was installed already.~

> I would print all the docs found, not just open first one.

Ok, this would mean that we simply allow multiple //DOCS tags, just like //SOURCES and //FILES and that "docs" would be an array of values. That could work. And if we make them all Lazy we can delay resolution to the last possible moment.

Still, if you do that I would find it pretty weird that we'd only allow you to open a single one using `--open`.

Also, if the docs reference is for example `./docs.html` the output of `info docs` (without --open) would be pretty useless. (And it would be replicated verbatim by the `info tools` command, once it's updated, as you mentioned)

I don't immediately have a good solution, I just want to point out that what we have now works well IMO and is easy to understand. These new suggestions seems to make things less well defined and less useful.

**Comment by andsel** [comment]
These are my thoughts on the multi-docs theme:
 I think that printing all the docs reference found could be useful, in particular I could imagine a case where there is a `//DOCS` in the source file and another from the alias catalog. In such case it's selected the one coming from the catalog and would help the user to understand why wasn't selected the one provided in the source file.

> - fail on file with multiple //DOCS and first one is missing
While the definition of missing is obvious on a local file (the file doesn't exist) when it's a URL we have 2 choices:
1. assure that the URL return an HTTP  2xx code (following eventual 302 and 301 codes)
2. consider the URL as always a valid reference.

In case 1 we have to navigate the URL, and could be not a stable result, in case the server or network are unreachable.
In case 2. (as it is today) we let the user to face the eventual error.


> but should the list of docs found not be listed in `jbang info tools` as a new key info?

Also if we decide to don't follow the multi-docs path, the information for the single doc is useful in the output of that command.


> - auto-open on urls

> I'm not sure its super safe to open browser or default editor on a file/resource - could be tricky...I would suggest we give a chance to cancel it. Either asking expliclitly or do the "wait 5 seconds" to open as we do for trust checking.

> failing on file with no jbang/docs in them; it should help/guide the user not just fail.

This is a bug fixed with commit 94939af6361e797be6555173f6375552887b521f

> Could still have a `jbang info docs --open test.java` to just open directly 

I agree with the fact to request the user the acknowledgment to open an url.

**Comment by andsel** [comment]
@quintesse I think the commit 7984ce528c6cefa1d970753dbabb332ce08a0309 broke the opening the doc reference defined in an alias.

So for example if I have a script which declare a DOCS directive like:
```
//DOCS http://www.jbang.dev/documentation/guide/latest/index.html
```
and then with the following command I define an override of the doc reference in an alias pointing to the same script, like in 
```
jbang alias add --docs https://www.jbang.dev/documentation/guide/latest/javaversions.html itests/docsrun.java
```

When I open the docs reference for the alias with:
```
jabang info docs docsrun
```

I would expect it open the `https://www.jbang.dev/documentation/guide/latest/javaversions.html`  and not the one from the script directive `http://www.jbang.dev/documentation/guide/latest/index.html`.

As I can understand the problem comes from this line:
https://github.com/jbangdev/jbang/pull/2013/commits/7984ce528c6cefa1d970753dbabb332ce08a0309#diff-0ed5e949a3d2d5b76e97b0a03e821cf4cb9d71645790f03c4eb28699f5f8e62cR674

`docs` is always `null`, while it should check `alias.docs`. 
Please could you confirm that?

**Comment by quintesse** [comment]
> `docs` is always `null`, while it should check `alias.docs`.

Indeed, the if [in line 674](https://github.com/jbangdev/jbang/commit/7984ce528c6cefa1d970753dbabb332ce08a0309#diff-0ed5e949a3d2d5b76e97b0a03e821cf4cb9d71645790f03c4eb28699f5f8e62cR674) is inverted, it should check for == null, just like the rest (and as you state right now that will always be true, still I'd like the if to remain in case the docs ever is not null in the future).

**Comment by andsel** [comment]
@quintesse 
> still I'd like the if to remain in case the docs ever is not null in the future

Do you mean that when `updateFromAlias` it's always expected that `doc` field is null? The only case I could imagine where it's non null, is when `updateFromAlias` is invoked twice.

In this case, should it raise a sort of programmatic error? (which is an error that the user can't do anything and we can't add any valuable call to action for him).

**Comment by quintesse** [comment]
> Do you mean

No, I mean that I want that if-statement to remain there, even if, for now, it's result will always be true. Just humor me :-)

**Comment by andsel** [comment]
> No, I mean that I want that if-statement to remain there, even if, for now, it's result will always be true. Just humor me :-)

@quintesse done with 95b7a9a14a3919ae76ca15a92bea36d737d14788

**Review by quintesse — DISMISSED** [review]
LGTM

**Comment by quintesse** [comment]
Now we have to finish the discussion about one doc, multiple docs, auto-open or not etc.

To me multiple docs don't make sense when you also have a way to open them, but you're only going to open one.

To me printing out a bunch of URLs or file paths also don't make sense. What is the use of that? To do anything useful with that you'd have to copy & paste them into a browser or reader.

If you run `info docs` you _know_ you're going to open a browser window, it's not a command that you can run by accident. (Except perhaps the very first time when you're new to JBang)

So my vote is for keeping things like they are now, it makes the most sense to me.

Edit: NB, I want to clarify: it's not that I don't see why people might like wanting to add multiple doc links, it's just that I don't see a nice way to make that feature really useful and make sense

**Comment by maxandersen** [comment]
My terminal when printing filenames/urls lets me click on them directly and/or pass it to other tools to process.

That + I can't really know upfront which docs are relevant for me for a set of combined scripts makes it for me makes it
it relevant to at least be able to show them.
 
One could argue that the automatic open is the challenge here?

if it was not automatically opened and `jbang info docs` did this:

```
jbang info doc.java
this_do_not_exists.txt
https://xam.dk/notthere
this_exists.txt
/tmp/this_exists.txt
http://www.jbang.dev/documentation/guide/latest/index.html
does-not-exist.txt
```

I as a user could pick the one relevant to me, click on it and it opens.

and if not auto open you do:

```
jbang info doc.java --open
Opening first out of 7 doc: this_do_not_exists.txt
To see all run jbang info doc.java
```

or if we flip it back to current:

```
jbang info doc.java
Opening first out of 7 doc: this_do_not_exists.txt
To see all run jbang info doc.java --no-open
```

(--no-open is also used for jbang edit)

and 

```
jbang info doc.java --no-open
this_do_not_exists.txt
https://xam.dk/notthere
this_exists.txt
/tmp/this_exists.txt
http://www.jbang.dev/documentation/guide/latest/index.html
does-not-exist.txt
```

...also, remember how we deal with maven repos - they are key/value pairs but key is optional.

we could allow:

//DOCS javadoc=https://javadoc.io/whereever
//DOCS website=https://....
//DOCS main=https://...

and then it would be possible to do `jbang info doc.java javadoc` ...and on `jbang info doc.java` we would show `main` if avaiable otherwise first in list.

for me there are plenty of reasosn why multiple docs are relevant and why having info also in `jbang info tools` makes sense.

File: src/main/java/dev/jbang/cli/Info.java [hunk 14]
```diff
@@ -308,3 +308,37 @@ public Integer doCall() throws IOException {
...
+		}
+		return validateDocsReferenceAndTransformToUri(info);
+	}
+
+	private static Optional<URI> validateDocsReferenceAndTransformToUri(ScriptInfo info) {
+		if (info.docs.isURL()) {
+			return Optional.of(URI.create(info.docs.getOriginalResource()));
+		}
+		if (!info.docs.exists()) {
+			throw new ExitException(EXIT_INVALID_INPUT, "Invalid documentation file path: " + info.docs.getFile());
```
maxandersen: this should not fail. info needs to succeed even if the doc path does not exist. otherwise tools using this on incomplete files would fail.

itests/docstest1.java
```diff
@@ -0,0 +1,10 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//DOCS readme.md
+
+import static java.lang.System.*;
+
+public class docstest1 {
+    public static void main(String... args) {
+        out.println("Docs test 1"));
```
maxandersen: all the tests has a ")" too much on this line resulting in compile/build errors if trying to use them. intentional?

src/main/java/dev/jbang/cli/Info.java
andsel: Good point, I'll replace with a warning log more helpful for the user.

itests/docstest1.java
andsel: I think it's a typo

andsel:
Regarding the multi vs single doc question. I thought that original intention of the feature was to drive the user to sort of landing page for the script/project, where it could find the description of the tool plus other pointers. In this spirit having multiple DOCS is an error, that could be possible, and have to work with that. 

@maxandersen in case you want to provide multiple doc links to the user, then the script developer can provide a single page with multiple link, a sort of TOC. We can inform the user that we are opening/displaying the first doc reference of a list, in case there a many, but just to let him know why we selected that URL/file in case he was editing a doc directive and not seeing what he expected.

quintesse:
Re:

```
jbang info doc.java
this_do_not_exists.txt
https://xam.dk/notthere
this_exists.txt
/tmp/this_exists.txt
http://www.jbang.dev/documentation/guide/latest/index.html
does-not-exist.txt
```

@maxandersen I can understand that we call all technically do this, but my question, with capital letters en question marks on purpose is, WHY???

Who on earth is ever going to use any of that?
We're solving an issue from 5 years ago of something that would "nice to have" and we're turning it into a full-fledged documentation feature challenging the library of congress... for what? For scripts that are 10 lines long?

Edit: to me this is about having the feature where you can run `info docs <script>` and docs pop up. People can add a dozen extra links as comment in the code if they want, as we can do with any code file. But the //DOCS line is specifically for JBang and specifically to do one simple thing.

src/main/java/dev/jbang/cli/Info.java
quintesse: Why should it not fail? Isn't that exactly how tools would detect that we haven't been able to get the docs or open the browser?

If what you want is to get a reference to the doc, for example to display it yourself in some way (instead of using the default which is to open a browser) we could instead add `--print` option which would not open the browser but print out the reference on the console. (Which is then meant to be piped to something else or read by some tool)

But in all these cases having a failing error code to be able to detect there is no docs link is a good thing, right?
(eg. you could do something like `mdreader $(jbang info docs myscript.java)` and it would properly fail to run, instead of opening up the reader and probably getting an error or just an empty window)

maxandersen: No. Info tools is there to give tools info and should not fail to give all the remaining correct info. 

If we fail here a typo in docs or simply a website down would fail the IDE ability to resolve dependencies. So no should not fail.

maxandersen:
It matches all the other interactions we have. I got it implemented with very few chances and fixing a few bugs in meantime. I'll push it later tonight or tomorrow morning so you can see.

src/main/java/dev/jbang/cli/Info.java
quintesse: But this is "info docs", there is no "other information", it just opens a browser.

There's also no dependencies to resolve. "info docs" is a completely separate command.

Btw, the docs reference _should_ be made part of the "info tools" output (I think I mentioned it before at some point) and in that case of course it shouldn't fail (but it also won't be de-referenced because we'd just be putting the actual reference, so no chance of getting errors)

maxandersen:
@andsel good work - I pushed commits to do the named docs via key/value as discussed above.

Now it does this:

```
This is a description test
main:
  file:///Users/manderse/code/jbangdev/jbang/docsexample.java
  https://xam.dk/notthere
  http://www.jbang.dev/documentation/guide/latest/index.html
  does-not-exist.txt (not found)
javadoc:
  /tmp/this_exists.txt (not found)
```

it uses ansi colors to make it look pretty :)

atm it defaults to not open...but I can be convinced to flip it so one would need to use `--no-open` but let me know what you all think.

quintesse:
Great! Just a quick question: is it important to show the "not found" message? It could possibly save a lot of requests that are not really necessary anyway, right? (We don't need the eocs to be locally available in the cache)

quintesse:
Also, don't forget to add the docs to the `info tools` output, like we discussed

src/main/java/dev/jbang/source/DocRef.java
```diff
@@ -0,0 +1,69 @@
...
+		String repoid = null;
+
+		if (split.length == 1) {
+			reporef = split[0];
+			repoid = reporef.toLowerCase();
+		} else if (split.length == 2) {
+			repoid = split[0];
+			reporef = split[1];
+		} else {
+			// more than one =, fail for now. todo support escaping
```
quintesse: Why not use `split("=", 2)`? That way any "="-s become part of the reopref (which for URLs is _very_ common!)

src/main/java/dev/jbang/source/TagReader.java
```diff
@@ -103,6 +103,23 @@ protected Stream<String> extractRepositories(String line) {
 		return Arrays.stream(line.split(" // ")[0].split("[ ;,]+")).skip(1).map(String::trim);
 	}
 
+	protected Stream<String> extractDocs(String line) {
+		return Arrays.stream(line.split(" // ")[0].split("[ ;,]+")).skip(1).map(String::trim);
```
quintesse: Are you sure you want to allow multiple references per //DOCS line?? `;` and `,` are perfectly valid URL characters, so then we _must_ introduce some kind of escaping mechanism making things more complicated

src/main/java/dev/jbang/source/resolvers/SiblingResourceResolver.java
```diff
@@ -55,7 +55,9 @@ public ResourceRef resolve(String resource, boolean trusted) {
 			}
 			ResourceRef result = resolver.resolve(sr, true);
 			if (result == null) {
-				throw new ResourceNotFoundException(resource, "Could not find " + resource);
+				// resolve returns null throw new ResourceNotFoundException(resource, "Could not
```
quintesse: Why did you do this?

src/main/java/dev/jbang/source/LazyResourceRef.java
```diff
@@ -39,6 +40,18 @@ public Path getFile() {
 		return resolvedRef.getFile();
 	}
 
+	@Nullable
+	// @Override
+	public Optional<Path> getFileMaybe() {
```
quintesse: This doesn't seem to be used?

maxandersen:
> Great! Just a quick question: is it important to show the "not found" message? It could possibly save a lot of requests that are not really necessary anyway, right? (We don't need the eocs to be locally available in the cache)

I need to resolve it to give the path to open. So I know the state anyway.

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -304,36 +324,111 @@ class Jar extends BaseInfoCommand {
...
+	Map<String, List<DocRef>> getDocsMap(ScriptInfo info) {
+
+		Map<String, List<DocRef>> docsMap = new LinkedHashMap<>();
+
+		if (info.docs != null) {
+			for (DocRef doc : info.docs) {
+				docsMap.compute(doc.getId().equals(doc.getRef().getOriginalResource()) ? "main" : doc.getId(),
+						(k, v) -> {
+							if (v == null) {
+								return new ArrayList<>();
```
quintesse: This can't be right, can it?

src/main/java/dev/jbang/source/LazyResourceRef.java
```diff
@@ -34,7 +34,7 @@ public Path getFile() {
 		if (resolvedRef == null) {
 			resolvedRef = resolver.resolve(getOriginalResource());
 			if (resolvedRef == null) {
-				throw new LazyResolveException();
+				return null;
```
quintesse: Weren't you going to leave the exception in place?

quintesse:
> > Great! Just a quick question: is it important to show the "not found" message? It could possibly save a lot of requests that are not really necessary anyway, right? (We don't need the eocs to be locally available in the cache)
> 
> I need to resolve it to give the path to open. So I know the state anyway.

No, for URLs you don't need the path. So for URLs you're now actually downloading the HTML page.
That's why @andsel first checked if the ref was a URL and only if it was not he would get the Path.
(The whole point in creating Lazy resolvers was so we didn't have to resolve and download a bunch of unneeded files :-) )

maxandersen:
And thats what the code does atm , right?

src/main/java/dev/jbang/cli/Info.java
maxandersen: why not? simplest way I know to create a multimap ?

src/main/java/dev/jbang/source/DocRef.java
maxandersen: makes sense.

src/main/java/dev/jbang/source/LazyResourceRef.java
maxandersen: missed that one.

src/main/java/dev/jbang/source/TagReader.java
maxandersen: i guess :)

src/main/java/dev/jbang/cli/Info.java
maxandersen: doh - see it now. missing the add. why the tests passed I don't know :) fixing.

andsel:
> @andsel good work - 

@maxandersen thanks, I would have been more effective to the project, but at the end opened the discussion on this. Hope to be more helpful in another PR.

maxandersen:
> thanks, I would have been more effective to the project, but at the end opened the discussion on this. Hope to be more helpful in another PR.

you have been more helpful - we needed someone to go do the start :)

i've pushed update that should fix the test failures.

maxandersen:
@sourcery-ai review

maxandersen:
@andsel sorry for noise - just for fun enabled sourcery.ai - didn't realize it would @-mention people ;)

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -253,9 +311,9 @@ public Integer doCall() throws IOException {
 				Object v = f.get(info);
 				if (v != null) {
 					if (v instanceof String || v instanceof Number) {
-						System.out.println(v);
+						out.println(v);
```
andsel, line 295: I think that use the static import for System.out and then using the `out` here is little bit confusing that this out is that one from System package instead of a field.

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -304,7 +362,68 @@ class Jar extends BaseInfoCommand {
...
+
+	@Parameters(arity = "0..*", description = "Documentation file/links to open")
+	List<String> docsToOpen;
+
+	@Override
+	public Integer doCall() throws IOException {
+
+		ScriptInfo info = getInfo(false);
+
+		URI toOpen[] = new URI[1];
```
andsel: There is a practical reason for this instead of:
```suggestion
		URI[] toOpen = new URI[1];
```

?

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -304,7 +362,68 @@ class Jar extends BaseInfoCommand {
...
+			docs.forEach(doc -> {
+
+				String uripart = doc.toString();
+				String suffix = "";
+
+				if (toOpen[0] == null && "main".equals(id)) {
+					toOpen[0] = doc;
+				}
+
+				out.println(String.format("  %s%s", uripart, suffix));
```
andsel: Here `suffix` is always empty string, so maybe redundant?

src/main/java/dev/jbang/cli/Info.java
andsel: Do we really need an array of 1 URI instead of of just an URI which could be empty? In case we could used `Optional<URI>`, which has a plus on readability of the intention, but has a minus on object allocation, however not impactful because just one allocation per iteration and we aren't looping on big collection.

andsel (commented):
Tinkering a little bit with the new implementation and left some suggestions on code 😅

quintesse:
@maxandersen I've updated the PR, can you take a look if it's what you wanted?

src/main/java/dev/jbang/cli/Info.java
```diff
@@ -304,7 +343,73 @@ class Jar extends BaseInfoCommand {
...
+					} catch (IOException e) {
+						Util.infoMsg("Documentation file to open not found: " + toOpen[0]);
+					}
+				} else {
+					Util.infoMsg("No documentation file to open found");
+				}
+			}
+		} else {
+			Util.infoMsg("Use --open to open the documentation file in the default browser.");
+		}
```
andsel:
```suggestion
       if (!open) {
            Util.infoMsg("Use --open to open the documentation file in the default browser.");
	    return EXIT_OK;
        }
        if (GraphicsEnvironment.isHeadless()) {
            Util.infoMsg("Cannot open documentation file in browser in headless mode");
	    return EXIT_OK;
        }

        if (toOpen[0] == null) {
            Util.infoMsg("No documentation file to open found");
	    return EXIT_OK;
        }
        try {
            Desktop.getDesktop().browse(getDocsUri(toOpen[0]));
        } catch (IOException e) {
            Util.infoMsg("Documentation file to open not found: " + toOpen[0]);
        }
```

@quintesse do you think a more assertive style in the ifs would help readability?

quintesse: Sure, it avoids excessive nesting in a place where nesting doesn't really do anything useful.

andsel: Fixed by 9ee8eb06490d398f8272136c583a96044830f70a

maxandersen (approved):
LGTM

maxandersen:
looks good - took a bit as I wanted to ensure intellij and vscode was not failing on these. they work fine hence merging.

Issue #2043, opened 2025-06-05
Repos parameter during app install command should be considered even if the //REPOS comment is defined


mcarlett:
When you want to use some maven artifacts not available on the repositories defined in the `//REPOS` comment, it should be useful to honor the [option](https://www.jbang.dev/documentation/guide/latest/cli/jbang-app-install.html#_options) `--repos` to add additional repositories (as stated in the description)

So a possible solution would be considering `--repos` argument as additional repository even if there are already defined repositories

example:

the MyScript.java file contains 

`//REPOS central=https://repo1.maven.org/maven2,mycustom.repo1=https://maven.repository.example.org/repo1
`

the install command is

`jbang app install  --repos=https://maven.repository.example.org/repo2 MyScript.java`

the install process will use all the 3 repositories

maxandersen:
That would not be consistent. imagine you start by doing `jbang app install --repos internalrepo myapp.java` - all is fine, now myapp.java gets updated to use some bad repo - your app.java will fail/resolve differently.

Thats why we apply same logic as maven /gradle does that the top level repo def wins over any "lower" repo definitiions.

Thus if you need to use multiple repos you have to list them all at app install or jbang run or jbang alias.

maxandersen:
before closing this - what is you *actual* use case where you want this?

maxandersen:
fyi, you can use `jbang info tools --select=repositories MyScript.java` to get the repo urls to reuse it in another call.

mcarlett:
This feature would be useful once you have to build something that is not in the public repos, so the scenario is:
- create jbang script that uses internal dependencies (I mean, deployed in an internal Maven repository)
- test the jbang script, installing it using the internal repository from the command line (to avoid testing the script and then updating it before releasing)
- once all the dependencies are resolvable from the public repositories (or at least the one configured in the jbang script) it is no longer necessary to use `--repos` option

quintesse:
Ok, but couldn't you simply do:

- Create script without //REPO so you don't have to change it once the deps are public
- Run `jbang app install --repos central,https://maven.repository.example.org/repo2 MyScript.java`

And use that during development?

Or is there a reason you want to have that custom repo set in the source file? And in your example you're adding a second custom repo on top of that. We're just trying to figure out if that's a common situation or not.

mcarlett:
> * Run `jbang app install --repos central,https://maven.repository.example.org/repo2 MyScript.java`

[...]

PR #2044, opened 2025-06-05
fix: pass `app install` build options to pre-build process

mcarlett:
Because "app install" pre-builds the app that it installs, it's not
enough to only copy the build options to the installed app startup
script, but they also need to be passed to the pre-build process and
that was missing.

Fixes https://github.com/jbangdev/jbang/issues/2043

also:

feat: added --no-build option to app install


maxandersen:
thanks for the PR but as outlined in https://github.com/jbangdev/jbang/issues/2043#issuecomment-2943707008 this would make the system of launches inconsistent.

If you can argue otherwise do feel free to (re)open issue with rationales for it.

quintesse:
(and of course we get failing tests again because of rate limiting grrr)

PR #2048, opened 2025-06-06
Fix: `--source-type` is now honoured for stdin and for existing non-`.java` files

### LINKED ISSUE #1975: source-type not honored
copy this to clipboard:

```
//DEPS dev.jbang:devkitman:0.1.2
//JAVA 21+
//PREVIEW

import dev.jbang.devkitman.*;

void main(String[] args) {
	var jdkManager = JdkManager.create();
	var jdk = jdkManager.getOrInstallJdk(args.length > 0 ? args[0] : "11+");
	System.out.println("JDK: " + jdk);
	System.out.println("Installed: " + jdkManager.listInstalledJdks());
	System.out.println("Providers: " + JdkProviders.instance().allNames());
}
```

run it with `pbpaste | jbang --source-type -` and it still try to run this as .jsh even though its forced to be java.

lutzleonhardt:
Closes #1975

### 🐞 Problem
| Scenario | Expected | Current behaviour |
|----------|----------|-------------------|
| `pbpaste \| jbang --source-type=java -` | Treated as Java source | Piped script cached as `.jsh` and executed via JShell |
| `jbang --source-type=java myscript.jsh` | Compiled as Java (or rejected) | File copied as `myscript.jsh.java` → `javac` “bad file name” error |

### 🔧 What changed
* **LiteralScriptResourceResolver**  
  * Accepts the `forceType` parameter and always prefers it over auto-detection when caching stdin.  
  * Keeps the existing _“prepend underscore if the name starts with a digit”_ safeguard.
* **RenamingScriptResourceResolver**  
  * When `--source-type` is supplied, we strip the original extension before adding the forced one  
    (`myscript.jsh` ➜ `myscript.java` instead of `myscript.jsh.java`).

### ✅ Tests
* Added `TestSourceType` with two cases:
  1. `stdinHonoursForceType()` – verifies that piping to `-` is handled as Java when requested.
  2. `forceJavaOnJshFile()` – verifies that forcing Java on a `.jsh` file succeeds and does **not** generate a `.jsh.jar`.

* Manual test
image

### ♻ Regression risk
Localised to two resolvers; no logic elsewhere changed. Existing behaviour for regular `.java`, `.jsh`, remote, and gav resources stays untouched.

---

### 🙌 Credits
Most of the root-cause analysis was assisted by [Brokk AI](https://brokk.ai).  
If you’re curious, I recorded the full debugging session:  
https://youtu.be/t_7MqowT638?si=QVySbqoy0mqLcljz  
(*24 min video – optional to watch, of course*).

---

### 📋 Checklist
- [x] Code builds and all existing tests pass  
- [x] New unit-tests added  
- [x] `--enable-preview` workflow verified manually  
- [ ] Release notes updated (if required)


src/main/java/dev/jbang/source/resolvers/RenamingScriptResourceResolver.java
```diff
@@ -87,7 +87,14 @@ public ResourceRef resolve(String resource) {
 						original = original.substring(original.indexOf("\n"));
 					}
 
-					String name = probe.getName() + (forceType != null ? "." + forceType.extension : "");
+					String name;
+					if (forceType != null) {
+						// Remove existing extension and add the forced extension
+						String baseName = probe.getName().replaceFirst("\\.[^.]+$", "");
```
quintesse: Please use `Util.sourceBase()` here

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
```diff
@@ -55,16 +61,42 @@ public String description() {
...
+				basename = getMainClass(scriptText).orElse(basename);
+			}
+		}
+		
+		// Ensure basename is a valid Java identifier for .java files to avoid
+		// "Unsupported class file major version" errors when using preview features.
+		// The preview feature (JEP 445) derives implicit class names from filenames,
+		// and Java class names cannot start with digits. SHA-256 hashes can start
+		// with digits (0-9), so we prefix with underscore to make them valid.
+		if (".java".equals(suffix) && !Character.isJavaIdentifierStart(basename.charAt(0))) {
```
quintesse: Interesting, isn't that something we need to do in the RenamingScriptResourceResolver as well?

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
```diff
@@ -26,6 +26,12 @@
  * reference to that file.
  */
 public class LiteralScriptResourceResolver implements ResourceResolver {
+	private final dev.jbang.source.Source.Type forceType;
```
quintesse: Please introduce an import for this type. We only use this syntax when absolutely necessary.

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
quintesse: And don't we perhaps need to filter out other characters that aren't legal in Java ids as well?
If so we better introduce a method for this in Util.

quintesse (changes requested):
Looks really good, just a couple of small changes and questions.

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
```diff
@@ -55,16 +61,42 @@ public String description() {
 	}
 
 	public static ResourceRef stringToResourceRef(String resource, String scriptText) throws IOException {
+		return stringToResourceRef(resource, scriptText, null);
```
quintesse: There are actually 3 other places in the code that call this function directly that might benefit from the new forceType parameter. One is in `Run` where we have `scriptMixin.forceType` and another is in `MarkdownSource` where we could pass either `null` or `Type.markdown`, I think either will work. And in that case we can remove this method as no longer needed.

src/test/java/dev/jbang/cli/TestSourceType.java
```diff
@@ -0,0 +1,49 @@
+package dev.jbang.cli;
```
quintesse: All our `jbang run` tests live in `TestRun`, these tests should be moved there.

We know this is not the best organization of tests but it's what we use for now and we should stick to it.

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
lutzleonhardt: You are completely right!
The problem is more dominant in the `LitzeralScriptResourceResolver`, because the temp file name is created from a hash function:

image

But your suggestion is more error prone. I will add a general utility function for that

`public static String toJavaIdentifier(String baseName) { … }`

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
lutzleonhardt: Thx - Done.

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
lutzleonhardt: Thx for the hint - Done.

src/main/java/dev/jbang/source/resolvers/RenamingScriptResourceResolver.java
lutzleonhardt: Done

src/test/java/dev/jbang/cli/TestSourceType.java
lutzleonhardt: OK, thx for the guidance - Done

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
lutzleonhardt: It is working as expected, here you see the wrong filename is fixed and can be used as valid implicit classname:

image

lutzleonhardt: The same command with the current/official jbang version:

image

lutzleonhardt: There you see the `bad file name`

lutzleonhardt:
@quintesse Thanks for your feedback. I have implemented your suggestions so far. Let me know if something is wrong or missing -thx.

src/main/java/dev/jbang/util/Util.java
```diff
@@ -2001,4 +2002,60 @@ public static String substituteRemote(String arg) {
...
+	/**
+	 * Converts an arbitrary string into a valid Java identifier that can be used as
+	 * a class name. This is particularly important for JEP-445 implicit classes
+	 * where the compiler derives class names from file names.
+	 *
+	 * @param baseName The input string to convert
+	 * @return A valid Java identifier, never null or empty
+	 */
+	public static String toJavaIdentifier(String baseName) {
+		if (baseName == null || baseName.isEmpty()) {
```
quintesse: This part I find a bit weird and unnecessary. Just mark the parameter as `@Nonnull`. If somebody passes an empty string something is probably wrong anyway. Better not to hide that by returning forced correct values.

src/main/java/dev/jbang/util/Util.java
```diff
@@ -2001,4 +2002,60 @@ public static String substituteRemote(String arg) {
...
+			return "__";
+		}
+
+		StringBuilder sb = new StringBuilder();
```
quintesse: Pff this is so much annoying AI-generated crap, sorry :-)
We're not using this function to generate files that users will see, so it's okay to create names that are not "nice".
We can change all this to a simple:

```java
return "_" + basename.replaceAll("[^\\p{L}\\p{N}_]", "_");
```

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -2643,4 +2646,53 @@ void testReadingNoAddExportsOnJava8() throws IOException {
...
+			System.setIn(originalIn);
+		}
+	}
+
+	@Test
+	void testForceJavaSourceTypeOnJshFile() throws Exception {
+		// create a .jsh file with unique name
+		Path p = cwdDir.resolve("sourceTypeTest.jsh");
+		try {
+			dev.jbang.util.Util.writeString(p,
```
quintesse: Import please

src/main/java/dev/jbang/source/resolvers/LiteralScriptResourceResolver.java
```diff
@@ -54,17 +61,39 @@ public String description() {
...
+			}
+		} else {
+			suffix = ".jsh";
+			if (hasMainMethod(scriptText)) {
+				suffix = ".java";
+				basename = getMainClass(scriptText).orElse(basename);
+			}
+		}
+
+		// Ensure basename is a valid Java identifier for .java files to avoid
```
quintesse: We can remove these comments, it's explained on the `toJavaIdentifier()` method. No need to repeat it.

src/main/java/dev/jbang/source/resolvers/RenamingScriptResourceResolver.java
```diff
@@ -87,7 +87,20 @@ public ResourceRef resolve(String resource) {
 						original = original.substring(original.indexOf("\n"));
 					}
 
-					String name = probe.getName() + (forceType != null ? "." + forceType.extension : "");
+					String name;
+					if (forceType != null) {
+						// Remove existing extension and add the forced extension
+						String baseName = Util.sourceBase(probe.getName());
+						// Ensure basename is a valid Java identifier for .java files to avoid
```
quintesse: Same here

src/test/java/dev/jbang/util/TestUtil.java
```diff
@@ -174,4 +174,53 @@ void testUrlBasicAuth() throws IOException {
 		assertThat(connection.getRequestProperty("Authorization"),
 				equalTo("Basic Sm9obkRvZTpWZXJ5U3Ryb25nUGFzc3dvcmQx"));
 	}
+
+	@Test
+	void testToJavaIdentifier() {
+		// Valid names stay the same
```
quintesse: This generated stuff is at least _somewhat_ useful, but boy do AI models love to hear themselves talk. There are so many tests here doing the same thing. It's as if they're paid by the character ;-)

quintesse (changes requested):
Still some minor changes to be made, but almost there!

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -2643,4 +2646,53 @@ void testReadingNoAddExportsOnJava8() throws IOException {
 				not(containsString(
 						"--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED")));
 	}
+
+	@Test
+	void testStdinHonoursForcedSourceType() throws Exception {
+		String SOURCE_TYPE_STDIN_CODE = "        ///usr/bin/env jbang \"$0\" \"$@\" ; exit $?\n" +
+				"        //JAVA 21+\n" +
```
quintesse: We can't dictate Java versions in our tests (at least not in the ones that get built or run). The Java version is dictated by the environment.

So this test should simply test if the forcing of source type works, nothing more (I'd just use the same jsh -> java test like in `testForceJavaSourceTypeOnJshFile`).

If you want to test the fact that we don't generate invalid filenames it's better to create a different test that skips actual compilation. (Search for tests that use `JavaCompileBuildStep` if you're feeling brave)

src/main/java/dev/jbang/util/Util.java
lutzleonhardt: Wouldn't say it's crap, it's a general Java identifier sanitization and satisfy the unit test cases (with the smallest amount of modifications). 
But for the case of sanitize filenames it's overengineered, I agree.
If the filenames doesn't matter, we can use this short method... or just use a "_" + guid anyway for the basename?
I am wondering why they decided to use the filename as classname wihtout sanitize it...

src/main/java/dev/jbang/util/Util.java
lutzleonhardt: Yes, and as you pointed out the function is not use-case oriented. 
Done

src/main/java/dev/jbang/util/Util.java
lutzleonhardt: Done :)

src/test/java/dev/jbang/cli/TestRun.java
lutzleonhardt: Ups, you mentioned this before.
Done

src/test/java/dev/jbang/util/TestUtil.java
lutzleonhardt: Haha, this was a good one ;)

src/test/java/dev/jbang/cli/TestRun.java
lutzleonhardt: Agree, let the test only focus on one thing

lutzleonhardt:
Let's iterate on it until it's satisfying :) Thanks for you input, let me know when something is missing

lutzleonhardt:
@quintesse Anything missing here from my side? Thanks

quintesse:
@lutzleonhardt if you have the time you can take a look at why the IT tests are failing. If not I'll look into it as soon as aI have some time.

maxandersen:
just testing this and at first glance it solves the reported bug (good) but it also introduce a regression and fail at handling this:

`jbang --code "$(cat itests/helloworld.java)" jbangtest`

it does not save as a helloworld.java file but instead _helloworld.java which worked previously. Thats what the failed itests is about.

src/main/java/dev/jbang/util/Util.java
```diff
@@ -2001,4 +2001,16 @@ public static String substituteRemote(String arg) {
...
+	/**
+	 * Converts an arbitrary string into a valid Java identifier that can be used as
+	 * a class name. This is particularly important for JEP-445 implicit classes
+	 * where the compiler derives class names from file names.
+	 *
+	 * @param baseName The input string to convert
+	 * @return A valid Java identifier, never null or empty
+	 */
+	public static String toJavaIdentifier(@Nonnull String baseName) {
+		return "_" + baseName.replaceAll("[^\\p{L}\\p{N}_]", "_");
```
maxandersen, line 2020: why is it always adding "_" ?

maxandersen: shouldn't it avoid touching it if it is an already valid name?

maxandersen:
I've added code to detect identifers so it wont add _ unless really need to. wdyt?

@quintesse wiremock generated new files from foojay. expected i guess?

src/main/java/dev/jbang/util/Util.java
```diff
@@ -2011,6 +2011,40 @@ public static <K, V> Entry<K, V> entry(K k, V v) {
 	 * @return A valid Java identifier, never null or empty
 	 */
 	public static String toJavaIdentifier(@Nonnull String baseName) {
+		if(isValidJavaIdentifier(baseName)) {
```
maxandersen: this is what I suggest added so we dont unnecessarily add _'s to names that are valid.

If we dont it breaks when our code elsewhere detect public class helloworld should be named helloworld.

src/main/java/dev/jbang/util/Util.java
lutzleonhardt: Yes, you are right. I think it was like that on the initial pull request. Then we decided to apply the naming fix in general

src/main/java/dev/jbang/util/Util.java
lutzleonhardt: Yes, I agree, we should not touch the name when it's still valid.
lgtm

maxandersen:
the pr was blocked/failing due to bad formatting. pushed update with that done.

maxandersen (approved):
lgtm

PR #2051, opened 2025-06-07
chore(deps): updated the versions of gradle and maven installed by jbang export gradle/maven

wfouche:


wfouche:
This functionality was tested with a JBang Java 21 application that was exported to Gradle and Maven, and both tested to build and run.

maxandersen:
thanks.

 I'm wondeirng if there would be value in these versions be possible to customize on the export? or even fetch "latest" as an option? 

for now merging this :)

maxandersen (approved):
lgtm

wfouche:
> I'm wondeirng if there would be value in these versions be possible to customize on the export? or even fetch "latest" as an option?

@maxandersen ,

I think this can be accomplished as follows (provided --version is allowed):

* `jbang export gradle --version=8.9  org/example/App/java`
* `jbang export maven --version=3.9.8  org/example/App/java`

The current implementation would need to be updated to parse the version number, and execute one of the following two commands once the project has been exported in order to update the local version of gradle or maven:

* For Gradle, command  `./gradlew --no-daemon wrapper --gradle-version  <version>` will update the local version to the requested version.

* And for Maven command `./mvnw -N wrapper:wrapper -Dmaven=<version>` will do the same.

PR #2053, opened 2025-06-10
refactor: made ResourceRefs lazy by default

quintesse:
ResourceRefs can now obtain their files or input streams lazily if so desired. This makes LazyResourceRef and LazyResourceResolver obsolete, so they have been removed.


quintesse:
As I self-merged the previous commit that introduced LazyResourceRefs I decided I could also self-merge the commit that removes it again :-)

Now this does directly affect the implementation of existing code paths so we should make sure everything really keeps working.

PR #2064, opened 2025-06-16
Clarification on //DEPS

gunnarmorling:
This just cost me five min and some grey hair ;)

### DIFF
--- a/docs/modules/ROOT/pages/dependencies.adoc
+++ b/docs/modules/ROOT/pages/dependencies.adoc
@@ -43,7 +43,7 @@ class classpath_example {
 	}
 }
 ----
-<.> `//DEPS` must be placed at the start of line and can be one or more space separated dependencies.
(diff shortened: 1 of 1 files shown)


quintesse:
While I do understand your frustration, the thing is that there should be no space for *any* of the directives, not just `//DEPS`. So how do we make that clear? (Hopefully without having to repeat that sentence for each and every directive)

maxandersen:
fwiw its quite intentional for two reasons:

a) faster to search/match on `^//DEPS ` 
b) the space lets you comment out the directive. if space was not meaningfull you wouldn't be able to comment it out.

maxandersen:
worth putting it here as the first/direct place its listed - we probably should go add a table of directives where we could also place info like this.

maxandersen:
@all-contributors please add @gunnarmorling  for docs

PR #2071, opened 2025-06-21
chore: enable toolchain for gradle using java 11

maxandersen:
avoids me having to remember to set java version before build :) 

@quintesse seeing any issue using java 11 as default as it can compile to 8 ?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,9 +45,13 @@ repositories {
 }
 
 java {
+	toolchain {
+		languageVersion = JavaLanguageVersion.of(11)
+	}
 	withJavadocJar()
 	withSourcesJar()
 }
+
 
 // See https://docs.gradle.org/current/dsl/org.gradle.api.reporting.ReportingExtension.html
 // Extension is provided via Gradle's `reporting-base` plugin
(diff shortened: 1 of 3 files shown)


wfouche:
FYI, I've been building JBang using Java 11 (via .sdkmanrc file) for a while, and have not encountered any issues.

*buildDist.sh*
```bash
#!/bin/bash
source "$HOME/.sdkman/bin/sdkman-init.sh"
sdk env

./gradlew clean
./gradlew assemble
./gradlew installDist

ls -l build/install/jbang/bin

./build/install/jbang/bin/jbang version
```

quintesse:
@maxandersen no issue whatsoever, in fact I think it's a very good idea.
The only thing I would suggest is to not use the Gradle tool chain stuff, in my experience it's very limiting because it means that you _must_ have a Java 11 installed, which kinda sucks IMO. In devkitman I did it the way you once suggested and it works great: https://github.com/jbangdev/jbang-devkitman/blob/main/build.gradle#L11-L21 , because it means you can use _any_ JDK that's 11 or newer.

Edit: if I'm wrong and the toolchain can be used to select any JDK that's of the selected version _or newer_, then that would be great and preferable of course. I just wasn't able to find this option in the docs or online.

maxandersen (commented):
adding this as I for build dont really think it should be option to build with newer unless you really want

quintesse:
> adding this as I for build dont really think it should be option to build with newer unless you really want

For me it's the other way around: why are you making my dev experience worse by forcing me to install and have active a JDK 11, just for compiling JBang? :-)

maxandersen:
well you dont have to install anything ...it happens automatically..

quintesse:
Hmm it doesn't for me, it says I have to use a JDK 11 and fails.

maxandersen:
ah - because I hadn't included settings.gradle that has the foojay resolver.

https://docs.gradle.org/current/userguide/toolchain_plugins.html#toolchain_plugins
https://github.com/gradle/foojay-toolchains

I was considering if we should add a jbang devkitman resolver but thought lets start with this first :)

maxandersen:
going to merge this - although ultimately we should find way to have it use latest foojay installer but that will need we make grade build use java 17+ but then use toolchain 11...which should work but lets do that in isolation.

PR #2073, opened 2025-06-23
chore(deps): sdkman, update java to 11.0.27-tem

wfouche:
`sdk env` will now activate Java 11.0.27


wfouche:
@quintesse , please review (and merge).

quintesse:
The change is ok, but I sure hope sdkman will soon implement https://github.com/sdkman/sdkman-cli/issues/1154 , because having to update for each minor version is a bit of a nuisance.

wfouche:
>The change is ok

Thank you, so can this pull request please be merged?

> but I sure hope sdkman will soon implement https://github.com/sdkman/sdkman-cli/issues/1154 , because having to update for each minor version is a bit of a nuisance.

I think there is a benefit to use a specific version of the Java toolset. All software has (unknown) bugs, so it is best to be bug-compatible. Doesn't matter who builds the software, it will be compiled in the same way. If we are happy to use very specific versions of libraries, why not the compiler (Java) itself as well?

maxandersen:
if goal is to set specific version we should use gradle tool chains - sdkman is not what we use in builds.

wfouche:
@maxandersen , I'm looking forward to the next release of JBang that will contain this fix. Hopefully sometime before the end of July.  :-)

Windows, at the moment, is my primary development platform and I use JBang on it frequently.

PR #2074, opened 2025-06-23
feat: use java.util.Scanner for keyboard input

### LINKED ISSUE #2070: Util.askInput - keyboard input is not echoed to the console on Windows
**Describe the bug**

Image

Keyboard input is not echoed to the console on Windows, but it is on Linux and I assume MacOS. A new JBang user that I introduced to JBang found this very confusing.

**Expected behavior**
On Windows when prompted for a response, if you type '3' then '3' should appear on the screen, just like in Linux.

**JBang version**
0.126.3

**Additional context**
The ConsoleInput class could be re-implemented using JLine. Keyboard read timeouts are supported by JLine
* https://github.com/jline/jline3
* Cross-platform support: Works on Windows, macOS, Linux, and other Unix-like systems

wfouche:
Fixes #2070


wfouche:
@maxandersen , @quintesse - please review this pull when you have time. The new readLine() implementation works great on Windows and all the other platforms I tested it on. I still need to test the code on macOS, but should be able to do so in the next day or so.

Update:
* Tested the code changes on macOS, and it worked as expected, without any issues.

quintesse:
See my comment here https://github.com/jbangdev/jbang/issues/2070#issuecomment-3001715837

quintesse (dismissed):
LGTM

src/main/java/dev/jbang/util/ConsoleInputReadTask.java
```diff
@@ -14,19 +13,7 @@ public ConsoleInputReadTask(InputStream in) {
...
-				while (!br.ready()) {
-					Thread.sleep(200);
-				}
-				input = br.readLine();
-			} catch (InterruptedException e) {
-				return null;
-			}
-		} while ("".equals(input));
-		return input;
+		Scanner sc = new Scanner(in);
```
maxandersen: im trying to recall why we did the old - and as far as i can grok it had to do with curl install approach not doing the right thing - we'll need to test that still works.

wfouche: Changed the updates to conform to the original implementation. Now using Scanner in non-blocking mode.

quintesse: Perhaps we should add some comments to this class so we remember next time why it exists and is implemented in this way 😉

wfouche:
The `readTask()` function was redundantly invoked during retries and successive readLine() calls. A task is now instantiated only once and re-used across task invocations.

src/main/java/dev/jbang/util/ConsoleInputReadTask.java
wfouche: Added comments and additional documentation.

wfouche:
Ready to be reviewed again.

quintesse:
> The `readTask()` function was redundantly invoked during retries and successive readLine() calls. A task is now instantiated only once and re-used across task invocations.

I'm not entirely sure this is correct. @maxandersen , do you remember why you implemented retries?

I suppose, given the fact that catching an IOException only results in a warning message and then it retries, it's because the /dev/tty might be busy somehow? (I don't think it's actually used because retries always seems to have the value 1) Because that would mean we'd need to retry opening the console, ergo you can't cache the readTask() result.

But then again if there's a timeout it will also loop if tries > 1, which doesn't seem correct either. So, not sure if I'm misunderstanding that code, or if because of historical reasons it's more complicated than it needs to be. (Or if it needs to be complicated it probably needs to be fixed :-) )

wfouche:
@quintesse , I've reverted the changes to instantiate the reader task only once. We can get back to this in a future round of updates.  For now keeping this pull request focussed on just resolving reading a line of text properly from the keyboard in Windows.

quintesse (approved):
LGTM

wfouche:
@maxandersen , do you need additionals tests to be run before this pull request can be accepted?

I did test the changes on MacOS, Windows, and Linux. Worked perfectly fine on all of them. This is a very simple and targeted change and should not "upset the apple cart".

maxandersen:
And this doesn't block so the timeout is honored and will not block cli runs forever ?

wfouche:
> And this doesn't block so the timeout is honored and will not block cli runs forever ?

Yes, it is completely non-blocking and timeouts work (as expected) if you do not enter a valid line of text (followed by ENTER) in the allotted time. I carefully tested the timeout functionality on Windows, macOS and Linux.

I have also documented the ConsoleInputReadTask class. The file now contains more documentation than code.

wfouche:
@maxandersen , every time I run `jbang edit file.java` on Windows (which is often) I encounter this keyboard issue. Looking forward to this pull request being accepted.

If you have any other concerns that should be addressed, please let me know.

maxandersen (approved):

wfouche:
@maxandersen , I'm looking forward to the next release of JBang that will contain this fix. Hopefully sometime before the end of July. :-)

Windows, at the moment, is my primary development platform and I use JBang on it frequently.

maxandersen:
I'll get a release pushed out this weekend :) 

Really would be nice we had a earlyaccess release that gets built from main to easy install :)

maxandersen:
been trying to release but hitting issue with publishng to new maven central publish api using jreleaser.

https://github.com/jreleaser/jreleaser/issues?q=is%3Aissue%20state%3Aopen%20author%3Amaxandersen

will have to look later to figure out if can find workaround/fix.

wfouche:
> been trying to release but hitting issue with publishng to new maven central publish api using jreleaser.

I've encountered issues with jreleaser before, and have decided never to use the latest version and wait a while before upgrading. It is a very complex problem domain in which jreleaser operates, and I am happy to be a stable release or two behind.

wfouche:
Thanks for the release @maxandersen , what a marathon!

Updated to 127.15 on my Windows machine, and happy to be able to interact properly with JBang via the keyboard.

maxandersen:
> > been trying to release but hitting issue with publishng to new maven central publish api using jreleaser.
> 
> 
> 
> I've encountered issues with jreleaser before, and have decided never to use the latest version and wait a while before upgrading. It is a very complex problem domain in which jreleaser operates, and I am happy to be a stable release or two behind.

Me too but needed to update to get the new maven central releaser api support.

PR #2076, opened 2025-06-23
feat: have jbang init know if nakedmains are ok to generate

maxandersen:
quick proof of concept...you can now pass --java version to init  `jbang init --java 25 hello.java` and if java version is >=25 it will default to naked mains.

```


void main(String... args) {
    out.println("Hello World");
}
```

this also add other build settings to init which we could turn into have the right values set. like compier flags etc.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -130,7 +130,7 @@ dependencies {
 	implementation 'org.apache.commons:commons-text:1.11.0'
 	implementation 'org.apache.commons:commons-compress:1.27.1'
 	implementation 'info.picocli:picocli:4.7.5'
-	implementation 'io.quarkus.qute:qute-core:1.12.2.Final'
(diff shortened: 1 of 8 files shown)


quintesse:
This is pretty cool, but it seems _very_ specific for naked mains, couldn't we simply pass the java version as a value and do the `if` in the template file? That seems like a good example for people to come up with their own extensions instead of adding a feature that's only useful for a single thing.

maxandersen:
i couldnt get qute to behave right ;/

maxandersen:
qute has some issue with int comparison - so for now making explicit "compactSourceFiles" property as thats the official name for this feature in java.

maxandersen:
also added that it generates //JAVA <version> when requested and when we know its needed (i.e. compact source MUST have //JAVA 25+ to work unless you always run with java 25.

Further made the compact make use of java.lang.IO static imports.

Issue #2069, opened 2025-06-19
jbang info tools MUST not fail on missing resources


maxandersen:
not sure if old or new issue but if you have:

```
//FILES donotexist.properties

public class test {}
```

jbang info tools fails with something like:

```
[jbang] [ERROR] Could not find 'donotexist.properties' when resolving 'donotexist.properties' in test.java via Chain of [Trusted remote resource, Classpath resolver, File Resource resolver]
[jbang] Run with --verbose for more details. The --verbose must be placed before the jbang command. I.e. jbang --verbose run [...]
```

This breaks IDE's usage as now they don't get information about the dependencies etc.

We must still return a result, even if only partial.

related to https://github.com/jbangdev/jbang-idea/issues/129

fbricon:
jbang-vscode manages to [parse](https://github.com/jbangdev/jbang-eclipse/blob/main/dev.jbang.eclipse.core/src/main/java/dev/jbang/eclipse/core/internal/process/JBangInfoExecution.java#L151-L163) such [errors](https://github.com/jbangdev/jbang-eclipse/blob/main/dev.jbang.eclipse.core/src/main/java/dev/jbang/eclipse/core/internal/process/JBangInfoExecution.java#L38-L39) already. Hope your fix doesn't break that

<img width="885" alt="Image" src="https://github.com/user-attachments/assets/e7df9fb7-d076-4959-87ef-bc5627c74208" />

maxandersen:
ah - so its not new behaviour :)

It would still continue to work - but we would probably put dedicated error "markers" in the json output instead.
like, an "errors" section in the json instead of dumping out raw text :)

quintesse:
Now that we made all resource refs lazy for the //DOCS support we can simply test if the sources/files exist and adjust the json output accordingly. I would suggest to simply not set the `backingResource` element in such cases, that would flag that the resource could not be obtained.

Edit: before publishing that version we'd of course have to make sure that the IDE plugins are updated first.

PR #2077, opened 2025-06-26
fix: no exceptions on missing sibling (re)sources

quintesse:
We used to throw exceptions when sibling sources or resources were
missing or if an error occurred while resolving them. But this makes it
difficult to use the `jbang info` command on a source that has missing
siblings. So now we no longer throw exceptions, but use the special
`UnresolvableResourceRef` to indicate that the resource could not be
resolved. This allows the `jbang info` command to still work and show
information about the source, even if some of its siblings are missing.

Fixes https://github.com/jbangdev/jbang/issues/2069

### DIFF
--- a/itests/abstagsnix.java
+++ b/itests/abstagsnix.java
@@ -1,0 +1,11 @@
+
+//SOURCES Two.java
+///SOURCES /nested/*.java
+
+//FILES /res/resource.properties renamed.properties=/res/resource.properties
+//FILES META-INF/application.properties=res/resource.properties
+
+public class abstagsnix {
+    public static void main(String... args) {
+    }
+}
--- a/itests/abstagswin.java
+++ b/itests/abstagswin.java
@@ -1,0 +1,11 @@
+
+//SOURCES Two.java
+///SOURCES C:/nested/*.java
+
+//FILES C:/res/resource.properties renamed.properties=C:/res/resource.properties
+//FILES META-INF/application.properties=res/resource.properties
+
+public class abstagswin {
+    public static void main(String... args) {
+    }
+}
--- a/itests/badtags.java
+++ b/itests/badtags.java
@@ -1,0 +1,9 @@
+
+//SOURCES _missing_.java
+
+//FILES _missing_.properties renamed.properties=_missing_.properties
+
+public class badtags {
+    public static void main(String... args) {
+    }
+}
--- a/src/main/java/dev/jbang/cli/Info.java
+++ b/src/main/java/dev/jbang/cli/Info.java
@@ -395,16 +395,16 @@ class Docs extends BaseInfoCommand {
 			});
 		});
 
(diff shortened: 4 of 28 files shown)


quintesse:
Hi @fbricon , could you take a look how this PR affects the VSCode plugin?

The main differences are:

- No more exceptions while simply parsing the project, so `info tools` should at least always return a result (unless there are serious problems, but not missing (re)source files)
- The JSON output will be the same as the current code when scripts have no problems. But for scripts with missing (re)sources this PR jbang info tools _will_ return JSON, instead of displaying an error. The JSON will look something like:

```json
  "files": [
    {
      "originalResource": "resourcethatdoesnotexist.properties",
      "target": "renamed.properties",
      "error": "not resolvable from itests/brokenresource.java via Chain of [Trusted remote resource, Classpath resolver, File Resource resolver]"
    }
  ],
```

PS: The error text is highly technical and probably not of much use to users, but the existence of the property can be used to determine that the (re)source could not be retrieved for whatever reason.

fbricon:
Since we expected those errors to be parsed from the standard out/err output, we don't get the proper diagnostic location anymore, so I'll have to make some changes in https://github.com/jbangdev/jbang-eclipse/blob/main/dev.jbang.eclipse.core/src/main/java/dev/jbang/eclipse/core/internal/process/JBangInfoExecution.java#L158-L168

Screenshot 2025-06-27 at 13 54 03

fbricon:
I realize your PR doesn't cover errors like 

> //SOURCE missing.java

which I was trying. And when I try:

> //FILES missing.java, 

vscode doesn't show any errors.

I definitely need to debug the backend side.

quintesse:
> I realize your PR doesn't cover errors like
> 
> > //SOURCE missing.java
> 
> which I was trying. And when I try:

I will look into that.

> 
> > //FILES missing.java,
> 
> vscode doesn't show any errors.
> 
> I definitely need to debug the backend side.

I would imagine that the plugin will need changes. The idea is that there won't be any errors thrown anymore just for getting information. It also means it will tell you when multiple files are missing, instead of failing at the first missing file it encounters.

quintesse:
> I realize your PR doesn't cover errors like
> 
> > //SOURCE missing.java
> 
> which I was trying. And when I try:

This has now been fixed. Could you take a look when you  have time @fbricon ?

maxandersen:
we really should get this ina. yes ide wont report missing stuff until updated to *also* cover this much more sane way to report missing resources.

maxandersen:
this was a merged already. missed to mark this done.

maxandersen:
whoops - nope not merged yet.

maxandersen:
@fbricon could we get vscode updated to support this? I can deal with intellij support.

maxandersen (dismissed):

quintesse:
Rebased PR on main

PR #2082, opened 2025-07-02
chore(deps): default Kotlin version set to 2.1.21

wfouche:


maxandersen:
it really feels wrong we keep changing the default...

wfouche:
The thinking here is that users that start writing new Kotlin or Groovy scripts (today) should be using the latest stable versions of Kotlin or Groovy. This is all we are doing here.

Kotlin 2.2.0 is the latest version, but 2.1.x is the stable branch and 2.1.21 the latest stable release. Previously we support Kotlin 2.0.x as the stable release branch.

We should not changes the default Kotlin and Groovy versions often, it might only happen once a year or not at all.

wfouche:
@maxandersen , the merge went the wrong way it seems.

maxandersen:
> @maxandersen , the merge went the wrong way it seems.

wdym? seems correct to me? moving to 2.1.21?

wfouche:
> wdym? seems correct to me? moving to 2.1.21?

I was expecting to see the commits merged to https://github.com/jbangdev/jbang/commits/main/

maxandersen:
it was just this PR being rebased with main.

PR #2083, opened 2025-07-02
chore(deps): default Groovy version set to 4.0.27

wfouche:


wfouche:
The thinking here is that users that start writing new Kotlin or Groovy scripts (today) should be using the latest stable versions of Kotlin or Groovy. This is all we are doing here.

Groovy 4.0.27 is the latest stable release of Groovy. Previously we supported Groovy 4.0.14 as the stable release of Groovy for JBang. A lot of bugs have been squashed since the 4.0.14 release.

We should not change the default Kotlin and Groovy versions often, it might only happen once a year or not at all.

wfouche:
@maxandersen , the merge went the wrong way it seems.

quintesse:
> same as kotlin - feels wrong. would be so much nicer default was 4+ than fixed version.

Groovy is downloaded from Maven while Kotlin comes from GitHub releases, so it should be possible to get a list of versions and select the latest. It's a bit more work of course to make that happen, but shouldn't be extremely hard.

wfouche:
> It's a bit more work of course to make that happen, but shouldn't be extremely hard.

Consider starting to use Milestones for the JBang project again, because then issues can be prioritized for the next release and attached to a Milestone. Milestones will also give everybody more insight into what functionality (most likely) will be in the next release. At the moment we have lots of open issues, some will possibly never be receive any attention.  "We need a sheep herder!" (smile)

quintesse:
Milestones require planning, planning requires time, time which a small team doing work in their free time often doesn't have, and when they have it they prefer to spend it on fixing/creating instead of planning :-)
Even when we have Milestones things often come along which get put in there at a moment's notice and other things that were in there have to be dropped because of lack of time or interest. So it becomes more of a administrative burden without any advantage for the people doing the work. So unless Max or I feel like spending the time on managing milestones, because we see a clear advantage for _us_, I don't see it happening.

wfouche:
Just asking that you think carefully about the pros and cons of using Milestones. Pros are that you can advertise low-hanging fruit type issues where you need help with that others could help fix by the next release. Issues can also be created for sections of the documentation to be updated. Milestones gives the project the ability to visibly plan future releases. Might also give it more momentum in resolving issues.

Post the 1.0 release (whenever that happens) I think using Milestones will be essential to the continued success of the project.

PR #2088, opened 2025-07-03
fix: add -t as short hand for --source-type

maxandersen:
while testing the fix for #1975 I got tired of typing out `--source-type java` all the time so thought `-t` would be good to have as option.

not merging it directly as `-t` could probably be used for lot of other things ... so if other suggestions on ways we could avoid having to always fully type `--source-type` i'm all ears....otherwise lets merge it as #1975 actually is quite useful now that it actually works ;)

### DIFF
--- a/src/main/java/dev/jbang/cli/ScriptMixin.java
+++ b/src/main/java/dev/jbang/cli/ScriptMixin.java
@@ -17,7 +17,7 @@ public class ScriptMixin {
 			"--files" }, converter = CommaSeparatedConverter.class, description = "Add additional files.")
 	List<String> resources;
 
-	@CommandLine.Option(names = { "-t",
+	@CommandLine.Option(names = { "-T",
 			"--source-type" }, description = "Force input to be interpreted as the given type. Can be: java, jshell, groovy, kotlin, or markdown")
 	Source.Type forceType;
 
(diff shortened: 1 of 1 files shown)


quintesse:
We can always use `-T` if you think it's useful but not that important. That way `-t` can be used for a future option that we think is important.

src/main/java/dev/jbang/cli/ScriptMixin.java
```diff
@@ -17,7 +17,7 @@ public class ScriptMixin {
 			"--files" }, converter = CommaSeparatedConverter.class, description = "Add additional files.")
 	List<String> resources;
 
-	@CommandLine.Option(names = {
+	@CommandLine.Option(names = { "-t",
```
maxandersen:
```suggestion
	@CommandLine.Option(names = { "-T",
```

maxandersen:
> We can always use `-T` if you think it's useful but not that important. That way `-t` can be used for a future option that we think is important.

good idea!

wfouche:
> By default, in Java picocli, options -T and -t are considered different options.

@maxandersen 

https://github.com/jbangdev/jbang/commit/45fc206621d9b75b01837da6b07f61e06b027533

The commit message says `-t` but the actual  implementation uses `-T`.  Best to edit the commit message and change it to `-T`.

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

PR #2113, opened 2025-07-14
fix: `JBANG_JDK_VENDOR` can again be used

quintesse:
The latest version of devkitman supports setting the "distro" of the JDK to install, which allows us to use the `JBANG_JDK_VENDOR` environment variable again to select the JDK vendor. Which was a feature that was missing since the switch to devkitman. (In devkitman the term "distro" was chosen instead of "vendor".)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,7 +125,7 @@ sourceSets {
 sourceSets.main.compileClasspath += sourceSets.java9.output.classesDirs;
 
 dependencies {
-	implementation 'dev.jbang:devkitman:0.1.2'
+	implementation 'dev.jbang:devkitman:0.2.0'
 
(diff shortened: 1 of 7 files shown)


maxandersen:
Should this be JBANG_JAVA_VENDOR ? ( Can still have the old as fallback but I'm inclined to have just one clean set of java-* settings)

Wdyt?

quintesse:
I don't mind, it's simple enough to do.

maxandersen:
+1 on making it JBANG_JAVA_VENDOR as that could translate well to `--java-vendor` and heck, even `//JAVA_VENDOR`

quintesse:
I hadn't made the change yet :-)
I'll open an issue for the JBANG_JAVA_VENDOR  (#2124)

PR #2114, opened 2025-07-15
chore: verbose mode now way less spammy with https

quintesse:
The security loggers for remote connections were generating way too many log messages. This fixes that.

### DIFF
--- a/src/main/resources/logging.properties
+++ b/src/main/resources/logging.properties
@@ -1,4 +1,5 @@
 handlers=java.util.logging.ConsoleHandler
 .level=INFO
+jdk.event.security.level=WARNING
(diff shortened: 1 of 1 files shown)


quintesse:
PS: the other way to fix this would be to have `--vrebose` only switch the JBang loggers from `INFO` to `ALL`, but I actually like the fact that other loggers start showing more information as well. But the security loggers were way too much. Of course we can't go adding more and more exceptions, so if this happens again in the future we might have to do what I said in the first sentence.

maxandersen:
We have the option to have verbose take optional value to grant finger control. 

Because yeah https traffic was too much :)

PR #2121, opened 2025-07-24
refactor: Migrate Hamcrest assertions to AssertJ

maxandersen:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.java.testing.hamcrest.MigrateHamcrestToAssertJ?organizationId=QUxML09wZW4gU291cmNlL0pCYW5n

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -138,6 +138,7 @@ dependencies {
 	implementation 'org.codejive:java-properties:0.0.7'
 
 	implementation "org.slf4j:slf4j-nop:1.7.30"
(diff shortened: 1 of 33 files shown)


maxandersen:
re-ran to see if it gotten better

maxandersen:
this is still borked. rejecting.

quintesse:
Borked in what way? (just curious)

maxandersen:
it doesn't compile (bad version) but even when fixed still lots of ham crest not ported as It doesn't seem to handle much of the matchers we use

PR #2122, opened 2025-07-26
Make App parent command to exentd from BaseCommand to expose --help option

andsel:

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -31,10 +31,10 @@ import dev.jbang.util.Util;
 
 import picocli.CommandLine;
 
-@CommandLine.Command(name = "app", description = "Manage scripts installed on the user's PATH as commands.", subcommands = {
+@CommandLine.Command(name = "app", mixinStandardHelpOptions=true, description = "Manage scripts installed on the user's PATH as commands.", subcommands = {
 		AppInstall.class, AppList.class,
 		AppUninstall.class, AppSetup.class })
-public class App {
+public class App extends BaseCommand {
 
 	public static void deleteCommandFiles(String name) {
 		try (Stream<Path> files = Files.list(Settings.getConfigBinDir())) {
@@ -44,6 +44,12 @@ public class App {
 		} catch (IOException e) {
 			// Ignore
 		}
+	}
+
+	@Override
+	public Integer doCall() throws IOException {
+		warn("composite command, specify one of the available subcommands [install, list, uninstall, setup]");
+		return EXIT_INVALID_INPUT;
 	}
 }
 


maxandersen:
you need to run formatting - `just format` or `./gradlew spotlessApply`

quintesse:
@andsel so actually to _real_ change that needs to be made is this:

```
@CommandLine.Command(name = "app", description = "Manage scripts installed on the user's PATH as commands.", subcommands = {
   AppInstall.class, AppList.class, AppUninstall.class, AppSetup.class },

--> mixinStandardHelpOptions = true)
```

If you add that attribute to the annotation it will just work.

Extra points if you fix it for the other commands as well :-)

Edit: Ok, i see now that you used that option before and then removed it again. But the correct option (AFAICS) is the one I just described earlier: only adding the flag should be enough.

andsel:
The problem with `mixinStandardHelpOptions` is that i prints a standard message 
```
Show this help message and exit.
```

Instead of something more customized on JBang:
```
Display help/info. Use 'jbang <command> -h' for
                            detailed usage.
```

The flag `mixinStandardHelpOptions` also adds a `--version` which doesn't print anything.


### With  mixinStandardHelpOptions

Screenshot 2025-07-29 at 09 19 30


### Without mixinStandardHelpOptions extending BaseCommand

Screenshot 2025-07-29 at 09 19 05


Hinheriting from `BaseCommand` we can also print a helpfull message when the user issue just the app command without any subcommand:

``` 
jbang add
```
Screenshot 2025-07-29 at 09 37 11

quintesse:
Hmm the --version is indeed somewhat unfortunate. About the help being more useful without any arguments, that's debatable, I find the original "missing subcommand" and then a full list of options and command with explanations more useful than a single line saying that I should choose one of the commands x, y or z but without any information about what each command does.

Btw, in your example "Without mixinStandardHelpOptions extending ..." it also shows a --version, does that one work?

maxandersen:
we don't use mixing standard help elsewhere afaik. Exactly because it's help is not great.

quintesse:
Ok, I just did a test with an alternative solution:

1. Create a HelpMixin class with the contents of `boolean helpRequested` from the BaseCommand class
2. Use that mixin in BaseCommand and in all toplevel commands that need help (App, Info, etc etc)
3. There are a couple of tests the need to be fixed from `run.helpRequested` to `run.helpMixin.helpRequested`

This way we get a --help option that works, no --version that doesn't work and a full help message when you don't give any arguments

andsel:
Hi @quintesse, do you have a PR that contains this change?

quintesse:
No, I just tested it quickly locally and threw away the changes :-)

andsel:
Oks thanks! I'll move forward with your idea in this PR.

andsel:
Hi @quintesse I've added a couple of commits to recreate what you described and added a couple of test just for the `app` top command.
Let me know what do you think, so that I can replicate this in all the other top commands that has subcommands.

quintesse:
Looks good! Want to do this for the other commands that lack help?

andsel:
yes, I'll do in this PR if it's ok for you.

quintesse:
Yes, perfect 👍

quintesse:
LGTM, if you think you're ready with this PR you can make it final (ready for review).

andsel:
I've found that also `cache` (clear) `jdk` (list, install, uninstall, home, java-env, default) `trust` (add, remove) `wrapper` (install) have sub-commands, so I have to validate also those.

quintesse:
Indeed. Hopefully it's as simple as adding the help mixin to the class.

andsel:
Thanks @quintesse , now it's ready to get reviewed :-)

quintesse (approved):
Nice job @andsel !

PR #2129, opened 2025-07-30
include version.txt in distributions

### LINKED ISSUE #1047: Binary distributions do not include version.txt
Apparently the build is set to copy `tmp/version.txt` into the generated distribution

https://github.com/jbangdev/jbang/blob/3adf1b50133ccd652048aeed2b6173e93940a77c/build.gradle#L116-L130

However this file is not created at the appropriate time and it's missing from the final distribution archives

```
$ tar tvf ~/Downloads/jbang-0.81.2.tar 
drwxr-xr-x  0 0      0           0 Jan  2  1970 jbang-0.81.2/
drwxr-xr-x  0 0      0           0 Jan  2  1970 jbang-0.81.2/bin/
-rw-r--r--  0 0      0     6216055 Jan  2  1970 jbang-0.81.2/bin/jbang.jar
-rwxr-xr-x  0 0      0        6324 Jan  2  1970 jbang-0.81.2/bin/jbang
-rw-r--r--  0 0      0        5385 Jan  2  1970 jbang-0.81.2/bin/jbang.cmd
-rw-r--r--  0 0      0        7173 Jan  2  1970 jbang-0.81.2/bin/jbang.ps1
drwxr-xr-x  0 0      0           0 Jan  2  1970 jbang-0.81.2/tmp/
```

```
$ unzip -l ~/Downloads/jbang-0.81.2.zip 
Archive:  /Users/aalmiray/Downloads/jbang-0.81.2.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  02-01-1980 00:00   jbang-0.81.2/
        0  02-01-1980 00:00   jbang-0.81.2/bin/
  6216055  02-01-1980 00:00   jbang-0.81.2/bin/jbang.jar
     6324  02-01-1980 00:00   jbang-0.81.2/bin/jbang
     5385  02-01-1980 00:00   jbang-0.81.2/bin/jbang.cmd
     7173  02-01-1980 00:00   jbang-0.81.2/bin/jbang.ps1
        0  02-01-1980 00:00   jbang-0.81.2/tmp/
---------                     -------
  6234937                     7 files
```

If this file is to be included then the `versionTxt()` task must be set as a dependency or `distZip`, `distTar`, `latestDistZip`, and `latestDistTar`.

maxandersen:
this fixes #1047 but I'm actually not even sure we want it in there at all.

@quintesse this would put a /version.txt in all the distributions - do we want that? any overlap in some install mechanisms?
should it have a different name?

if we don't need it I'll make a PR that removes the unnecessary attempt on including the file - our GitHub Action builds does rely on the file but only on filesystem not in the built archives.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -505,6 +505,8 @@ task clidoc(type: Exec) {
 //didn't find better way yet.
 distZip.dependsOn(versionTxt)
 distTar.dependsOn(versionTxt)
+latestDistZip.dependsOn(versionTxt)
+latestDistTar.dependsOn(versionTxt)
 
 build.dependsOn(latestDistZip)
 build.dependsOn(latestDistTar)


quintesse:
Hmm I don't think we need it, no. And if we put it in there someone might start using and depending on it

maxandersen:
Yeah i kinda feel mutating manifest.mf to have trace of version would be better mechanism.

quintesse:
Well, if you think you'd might need it at some point, the advantage of the version.txt is that even shell scripts will be able to read it.

maxandersen:
closing this as we dont need/want it in distribution.

And made mf@jbangdev to help scripts to get manifest info :)

`jbang mf@jbangdev `jbang info jar dev.jbang:jbang-cli:RELEASE` Version`

maxandersen:
replaced with #2156

PR #2133, opened 2025-08-01
feat: scope deps

### LINKED ISSUE #2126: Scoped and managed dependencies
This issue is intended to be the parent of all the various issues related to the features needed for more fine grained dependency mechanisms than jbangs current bruteforce `//DEPS` which is basically one classpath for both build/compile/boot/runtime....which works in about 90% of cases.

And that latter is to be remembered - whatever we do here I do not think we should change the working default behaviour - its easy to understand and use. I will rather not have any finegrained controls than give up easy to use in 90% of cases.

That said - there are starting to be more cases where some finer control is needed, lets gather the ones we know of:

- [ ] define managed dependencies - in maven speak: dependency management pom import - to control what specific version of a library to be used. We have @pom today but it is problematic as it prevents normal pom dependency used in projects like graalvm truffle #1683 
- [ ] being able to exclude (transitive) dependencies - useful to leave out i.e. optional log4j dependencies included by some libraries #150 
- [ ] be able to mark dependencies as buildtime only (graalvm "compileOnly", maven "provided"); useful for things like annotationprocessors and usecases like jreleaser extension where the runtime classpath really shouldn't include jreleaser itself (see [PR discussion](https://github.com/jreleaser/jreleaser/pull/1951#issuecomment-3129965955))
- [ ] be able to define module deps(//MDEPS) as well as claspath deps (//DEPS) #559
- [ ] consider how boot-module-path and agent-module-path fits in this #618 and #1662

Currently the above I'm leaning towards the following principles/conclusions (mainly here to validate if things stays consistent, but curious to hear feedback on it):

* scopes feels very much like "tags" on a dependency type - i.e. one way to express it could be asciidoc block property `//DEPS[build]` or `//DEPS{build}` to make parsing less ambigious as [] is valid in version ranges.
* test dependency / test scope is NOT currently in the cards as that would be more a separate project/script that uses existing jbang script. The tests has the other script as a dependency; the main project doesn't know about the tests...we might want to make that a thing later but then thats really a different kind of relationship - not a classpath defined dependency scope.
* exclusions in maven is AFAICS done only on transitive dependencies on a specific explicit dependency, need to use enforce rules for global exclusions. ["provided" does it partially ](https://stackoverflow.com/questions/4716310/is-there-a-way-to-exclude-a-maven-dependency-globally )[feature requests exist](https://issues.apache.org/jira/browse/MNG-1977) for global definable - gradle has global exclude options but [seem to vary over time ](https://stackoverflow.com/questions/21764128/how-do-i-exclude-all-instances-of-a-transitive-dependency-when-using-gradle) - i think a first step is to define global excludes...and that it will be useful being able to do group excludes .. like `//!DEPS org.jreleaser:*` but also be able to do it at buildtime...`//!DEPS[build] org.jreleaser`
* we'll need to find out how users/tools get the resolved classpaths/scopes.

## 🧪 Proposed Design Directions

To support these features incrementally and maintain JBang’s focus on simplicity, the following ideas are proposed for review and refinement:

### 🎯 Scoped Dependencies

Introduce optional **scopes** (tags) on dependencies to control when they are available in the lifecycle:

```java
//DEPS{compile} com.example:lib:1.0
//DEPS{runtime} com.example:lib:1.0
````

* **Shorthand** style (`{scope}`) or **key=value** style (`{scope=compile}`) allowed.
* Multiple scopes can be combined: `{compile,runtime}` (equivalent to today’s default).
* This allows fine-grained control over classpath assembly for compile and runtime phases without affecting current behavior (scopeless `//DEPS` would still default to both).

### 📦 Module Dependencies (`//MDEPS`)

Introduce `//MDEPS` for declaring **JPMS module dependencies** which are resolved **not** through Maven but used to build the JPMS `module-path` or `--add-modules` settings:

```java
//MDEPS java.sql
//MDEPS{static} jdk.crypto.ec
```

This supports modular Java applications, GraalVM agent injection, etc., aligning with long-standing requests.

### 🚫 Global Exclusions (`//!DEPS`)

Instead of requiring Maven-style per-dependency exclusions, JBang can introduce a simpler **global exclusion directive**:

```java
//!DEPS commons-logging:commons-logging
//!DEPS org.slf4j:*
```

* Excludes matching dependencies from **transitive resolution**.
* **Explicitly declared `//DEPS` still take precedence** (i.e. explicit includes override global excludes).
* Wildcards (`*`) allowed in artifactId or groupId for convenience.

We can also support **scope-specific exclusions**:

```java
//!DEPS{runtime} org.jreleaser:*
```

This example removes `org.jreleaser` dependencies **only from the runtime classpath**, while keeping them available for compile as needed.

### 🧩 Resolution Rules

A few behavioral principles underpin the above:

* `//!DEPS` exclusions apply **only to transitive dependencies** by default.
* Explicit `//DEPS` always **wins** over a global exclusion.
* Scopes define **when** a dependency is added to the classpath:

  * `compile` = available only during compile
  * `runtime` = available only at runtime
* Wildcard support in exclusions (`group:*` or `group:artifact*`) gives lightweight power without full dependency graph introspection.

---

## Open questions/concerns

- Is there a usecase for exclusions only defined on certain dependencies?
- if we only have global exclusions these would have to be added on all deps in maven export
- are there more properties to "add" to a dependency than scopes?
- should "agent" and "boot" be a scope ?
- how is managed deps expressed if @pom not allowed? `//DEPS[managed=true, scope=compile]`?
- is `//DEPS` equal to `//DEPS[scope=runtime,compile]` ? 
- anything missed here?
 
## 📌 Next Steps

Each of the checklist items in this parent issue can be implemented incrementally. The above proposals aim to balance **power and flexibility** without compromising JBang’s **zero-config default UX**.

Feedback on naming, syntax, parsing, and implementation trade-offs is very welcome!

maxandersen:
this PR will grow to have scope deps.

for now this is just implementing precursor for #2126 as a mechanism to store scopes and possible other attributes on dependency.

Basically can parse a string into a map of key->List<string> to capture any kind of options that we add to a MavenCoordinate as a DependencyAttributes.

With this in place doing additional filtering when resolving becomes somewhat trivial.

We can utilize this notion of attributes on potential anything

Currently implemented syntax is inspired by asciidoc: https://docs.asciidoctor.org/asciidoc/latest/attributes/positional-and-named-attributes/

I initially started with https://github.com/yupiik/tools-maven-plugin/blob/d1e8b97ebcae032684a0fafac426bcc25819089f/asciidoc-java/src/main/java/io/yupiik/asciidoc/parser/Parser.java#L1634 but it was too simple so with a bit a LLM help got something that supports things like:

`g:a:v` or `g:a:v{}` -> scope will be build, run
`g:a:v{build}` -> scope will be build
`g:a:v{build, boot} -> scope will be build , boot
`g:a:v{build, scope=boot} -> scope will be build , boot
`g:a:v{build, scope=boot,%transitive} -> scope will be build , boot and there will be a "transitive=[true]

...I even allow quotes...
`g:a:v{build, scope=boot,comment="This is probably over kill",%transitive} -> scope will be build , boot and there will be a "transitive=[true] and "comment"=["This is probably over kill"]

...will allow for things like:

`g:a:v{build,exclude=org.jreleaser:*}` which is pretty nice too imo :)

oh...and I converted to assertJ in the tests where I need to...hamcrest need to retire.

...will update this main description as things evolve and this eventually will Fix #2126


src/main/java/dev/jbang/dependencies/DependencyAttributes.java
```diff
@@ -0,0 +1,29 @@
...
+	// Java 8 Set.of() is sad :)
+	static Set<String> defaultScopes = Collections.unmodifiableSet(new HashSet<>(Arrays.asList("build", "run")));
+	Set<String> scopes;
+
+	public DependencyAttributes(Map<String, List<String>> attributes) {
+		this.attributes = attributes;
+	}
+
+	// lazy computed; dont store hashset unless truly needed
+	public boolean includeInScope(String scope) {
```
maxandersen, line 24: I don't like the word scope - mainly because these aren't actually maven scopes but rather something that gets translated into a scope.

`includeInConfig(String config)` maybe ? (gradle use configuration as name for different scenarios)

maxandersen: and it would be `defaultConfigs` if follow same route.

src/main/java/dev/jbang/dependencies/DependencyAttributes.java
```diff
@@ -0,0 +1,61 @@
...
+		return AttributeParser.toStringRep(attributes, "scope");
+	}
+
+	@Override
+	public String toString() {
+		return toStringFormat();
+	}
+
+	@Override
+	public boolean equals(Object o) {
```
maxandersen, line 54: so first push resulted in funky TestModules errors that was caused by equals method missing on Attributes - which is fair..but reveals a conondrum - how do we define equals for these? i've implemented a few tests (which currently fails) showing situations that we probably want to make equals...which we can by unifing attributes...but should we?

src/test/java/dev/jbang/source/TestModule.java
```diff
@@ -43,6 +43,15 @@ public class TestModule extends BaseTest {
 			"    }\n" +
 			"}\n";
 
+	String srcWithVersionRangeDep = "//MODULE testmodule\n" +
+			"//DEPS info.picocli:picocli:4.6+\n" +
```
maxandersen: added this test as realized the module support seem to rely on *exact* string matches rather than understanding of resolved dependencies - is that a known/expected limitation @quintesse ?

src/main/java/dev/jbang/dependencies/DependencyAttributes.java
maxandersen: just did an update where i split it out in a DependencyRequest even though I do thing the other properties makes sense to have on the "thing" that encodes the dependency - but it seems like we have places where we treat MavenCoordinate as absolutes which can be used as keys....that is NOT the case even with these new info because MavenCoordinates can have  version ranges defined.

src/test/java/dev/jbang/source/TestModule.java
maxandersen: looks like Module generation should only compare on the group + artifict + type ..ignore version range or any properties?

quintesse: I'm not sure what you're referring to @maxandersen . What exact string matches? What understanding of resolved dependencies? There's not really a relationship between modules and maven dependencies implemented (at least as far as I remember).

maxandersen: it's https://github.com/jbangdev/jbang/pull/2135

PR #2135, opened 2025-08-02
fix: code should not rely on MavenCoordinate equals/hashcode

maxandersen:
While doing #2133 i spotted that ModuleUtil relied on MavenCoordinate to be used as key in list - that wont do what you most likely want given a mavencoordinate is more like a "request for a dependency" so it can have version ranges in it and soon scope info. 

So I'm isolating this fix as I believe its a bug on its own right and to not make scope deps fix having too many things in it.

### DIFF
--- a/src/main/java/dev/jbang/dependencies/MavenCoordinate.java
+++ b/src/main/java/dev/jbang/dependencies/MavenCoordinate.java
@@ -1,6 +1,5 @@
 package dev.jbang.dependencies;
 
-import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


quintesse:
Sure, seems logical.

Not okaying yet because a) I don't understand the module tests you added/changed and b) the commented out code should be removed, right? (A comment saying we shouldn't have equals/hash + reason is probably a good idea , so we don't fall into this trap in the future)

maxandersen:
The module test is a dependency with version range. 

All the exisiting tests failed after I added attribute syntax support and first I thought i broke something but then I realized even if we implemented equals/hashxode to make the tests passes code using version ranges would still fail as resolved maven coordinate has no version ranges and thus wouldn't be equals. And contains would still fail. Hence why i made it explicitly used managed key (g:a:type) which is what module util code should have used originally anyway.

quintesse:
Ok, understood, LGTM (still, you'll remove the commented code, right? :-) )

quintesse (dismissed):

maxandersen:
@quintesse ok, moved the description to top and removed the commented out equals/hashcode implementations

quintesse (approved):

PR #2139, opened 2025-08-05
fix: renovate[bot] should ignore major version updates to wrappers for maven and gradle

### LINKED ISSUE #2138: Renovate[bot] should ignore files in src/main/resources/dist
Distribution files for Gradle and Maven are manually updated at a time of our choosing using scripts:

* maven/update-version.sh
* gradle/update-version.sh

renovate[bot] should ignore files in folder  `src/main/resources/dist` and beyond.

This issue can be resolved by adding config option `ignorePaths` to the renovate.json file.

```
{
    "ignorePaths": [
        "src/main/resources/dist/**/*"
     ]
}
```

wfouche:
Closes #2138

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -17,6 +17,16 @@
       "description": "Wiremock >3 requires Java 11",
       "matchPackageNames": ["com.github.tomakehurst:wiremock"],
       "allowedVersions": "<3.0.0"
+    },
+    {
+      "matchManagers": ["maven-wrapper"],
+      "matchUpdateTypes": ["major"],
+      "enabled": false
+    },
+    {
+      "matchManagers": ["gradle-wrapper"],
+      "matchUpdateTypes": ["major"],
+      "enabled": false
     }
   ]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -2,6 +2,9 @@
   "extends": [
     "config:best-practices"
   ],
+  "ignorePaths": [
+    "src/main/resources/dist/**/*"
+  ],
   "packageRules": [
     {
       "matchPackageNames": ["io.quarkus.qute:qute-core"],
--- a/renovate.json
+++ b/renovate.json
@@ -2,9 +2,6 @@
   "extends": [
     "config:best-practices"
   ],
-  "ignorePaths": [
-    "src/main/resources/dist/**/*"
-  ],
   "packageRules": [
     {
       "matchPackageNames": ["io.quarkus.qute:qute-core"],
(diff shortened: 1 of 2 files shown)


maxandersen:
Why dont we remove the custom update scripts and use renovate bot hooks instead ?

wfouche:
> Why dont we remove the custom update scripts and use renovate bot hooks instead ?

I don't think using Renovatebot in the JBang project adds any value. The number of Jars and Plugins to update are not that many, and at the moment they are not being updated regularly and consistently.  So maybe Renovatebot is providing a false sense of security. Gson, Jsoup, Maveniverse, just to name a few,  have more recent versions available.

It is also not the best policy to just go with the latest version of a given artifact, for example Gradle 9.0 is available but from a stability perspective I am more inclinded to use Gradle 8.14.3.

My preference is to manually decide what version to upgrade to, and do this regularly (preferable as the first step after every release of JBang so that we have en entire release cycle to test the updated dependencies)

If it was my decision to make, I would remove Renovatebot from the JBang project and simple use the Gradle Versions plugin (https://github.com/ben-manes/gradle-versions-plugin)

```
$ ./gradlew dependencyUpdates

The following dependencies have later milestone versions:
 - com.google.code.gson:gson [2.10.1 -> 2.13.1]
     https://github.com/google/gson
 - eu.maveniverse.maven.mima:context [2.4.20 -> 2.4.30]
     https://www.maveniverse.eu/mima
 - eu.maveniverse.maven.mima.runtime:standalone-static [2.4.20 -> 2.4.30]
     https://www.maveniverse.eu/mima
 - info.picocli:picocli [4.7.5 -> 4.7.7]
     https://picocli.info
 - org.codehaus.plexus:plexus-java [1.2.0 -> 1.3.0]
     https://codehaus-plexus.github.io/plexus-pom/plexus-languages/plexus-java/
 - org.jsoup:jsoup [1.17.1 -> 1.21.1]
     https://jsoup.org/
  - ......
```

maxandersen:
Just configure renovate to stay away from 9.x updates. 

Nice sideeffect is that it documents why/how updates are limited.

maxandersen:
Updating deps in bulk when I remember to run it is imo worse than renovate bot that at least provide context, opens a PR and I can reject with a reason.

wfouche:
@maxandersen , I spend some time today to learn more about how renovatebot works.

> Just configure renovate to stay away from 9.x updates.

Rather configure it to not perform major upgrades of the maven and gradle wrappers. Major upgrade should be done manually and can be tested to see if we should use the new major version of either Gradle or Maven.

I agree that Renovatebot is a good idea, and will update the pull request accordingly.

The individual upgrade scripts for Maven and Gradle in resources/dist should remain. They'll be used in the future to test major upgrades from 3.9 to 4.0, or 4.0 to 5.0 (Maven) or 8.x to 9.0, or 9.0 to 10.0 (Gradle).

maxandersen:
thans @wfouche

Issue #2147, opened 2025-08-06
Performance of JBang 128.7 vs JBang 126.3 on Win11.WSL2(Ubuntu)


wfouche:
JBang 128.7 is slow to get going on WSL2(Ubuntu).

```
$ time jbang version
0.128.7

real    0m6.452s
user    0m1.142s
sys     0m0.729s
```

JBang 126.3 is quick as expected.

```
$ time jbang version
0.126.3

real    0m0.326s
user    0m0.576s
sys     0m0.141s
``` 

What could be causing the six second delay?

wfouche:
Strace files for JBang 0.126.3 vs 0.128.7

[jbang_0_126_3.txt.gz](https://github.com/user-attachments/files/21622744/jbang_0_126_3.txt.gz)

[jbang_0_128_7.txt.gz](https://github.com/user-attachments/files/21622745/jbang_0_128_7.txt.gz)

quintesse:
Hi @wfouche , given the fact that it doesn't do this on my machine, would you perhaps be willing to do some debugging?

This would involve building the code with some changes. I'm especially interested in finding out if this was caused by commit  964753279435c2fd4d6c3454a63cd96f69f07bbc

We'd have to revert that commit by doing `git revert 964753279435c2fd4d6c3454a63cd96f69f07bbc` and then building the code and seeing if that improved things.

maxandersen:
Is this same as #2137 ?

wfouche:
@quintesse ,  with commit https://github.com/jbangdev/jbang/commit/964753279435c2fd4d6c3454a63cd96f69f07bbc removed JBang starts quickly and performance is as expected.

wfouche:
> Is this same as [#2137](https://github.com/jbangdev/jbang/issues/2137) ?

They seem to be related. The delay is just much longer on SuSE than what I see on WSL2(Ubuntu).

WSL2(Ubuntu) automatically mounts the Windows Drive C: as /mnt/c and some folders from /mnt/c are in $PATH. Accessing /mnt/c/... is slower than accessing the local native ext4 filesystem.

On the SuSe system a few directories in $PATH might be on an NFS or other share.

But why is the plugin functionality causing long delays for command "$ time jbang version" ?

quintesse:
[...]

Issue #2137, opened 2025-08-04
JBang CLI hangs on SUSE Linux Enterprise Server15 SP6


bvahdat:
After installation using the curl command: [`curl -Ls https://sh.jbang.dev | bash -s - app setup`](https://www.jbang.dev/download/) on `SUSE Linux Enterprise Server15 SP6`, running the CLI just hangs without any error:

```
$> jbang -version
```

The installed version is:

```
$> cat ./.jbang/cache/urls/jbang/version.txt
0.128.7
```

The installed JDK seems to be properly in place:

```
$> ./.jbang/currentjdk/bin/java -version
openjdk version "17.0.16" 2025-07-15
OpenJDK Runtime Environment Temurin-17.0.16+8 (build 17.0.16+8)
OpenJDK 64-Bit Server VM Temurin-17.0.16+8 (build 17.0.16+8, mixed mode, sharing)
```

quintesse:
What happens if you run `jbang --verbose version`? (Note the lack of -- for version, it's intentional!)

bvahdat:
Same behaviour, it hangs.

bvahdat:
Also the installation log looks O.K. without any errors / warning:

```
$> curl -Ls https://sh.jbang.dev | bash -s - app setup
Downloading JBang ...
Installing JBang...
Downloading JDK 17. Be patient, this can take several minutes...
Installing JDK 17...
[jbang] Default JDK set to 17 (17.0.16+8 [dynamic], current, /home/a89600396/.jbang/cache/jdks/17, [Ga, Jdk]))
```

bvahdat:
> Same behaviour, it hangs.

And the same as for this one:

```
$> jbang app install camel@apache/camel
```

bvahdat:
> Same behaviour, it hangs.

Sorry @quintesse, wrong answer of me above. I had to wait long enough to see the output, as it takes more than 2 minutes:

```
$> time jbang --verbose version
[jbang] [135:967] jbang version 0.128.7
Cache: /home/a89600396/.jbang/cache
Config: /home/a89600396/.jbang
Repository: /home/a89600396/.m2/repository
Java: /home/a89600396/.jbang/cache/jdks/17 [17.0.16]
OS: linux
Arch: x64
Shell: bash
0.128.7

real	2m16.213s
user	0m2.252s
sys	0m0.172s
````

bvahdat:
> > Same behaviour, it hangs.
> 
> And the same as for this one:
[...]

PR #2149, opened 2025-08-07
fix: this fixes part of the slowness on startup

quintesse:
The setup of help information for PicoCli is now done lazily which improves JBang startup time in general but specifically on systems that for some reason are slow when scanning the PATH for plugins. We also improved direct plugin lookup by not using full file scanning.

See #2147 and #2137

### DIFF
--- a/src/main/java/dev/jbang/cli/JBang.java
+++ b/src/main/java/dev/jbang/cli/JBang.java
@@ -10,7 +10,10 @@ import static picocli.CommandLine.Model.UsageMessageSpec.SECTION_KEY_COMMAND_LIS
 import static picocli.CommandLine.Option;
 import static picocli.CommandLine.ScopeType;
 
+import java.io.File;
+import java.io.IOException;
 import java.io.PrintWriter;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -24,7 +27,9 @@ import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.Future;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import dev.jbang.Configuration;
 import dev.jbang.util.Util;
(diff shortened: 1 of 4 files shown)


src/main/java/dev/jbang/util/Util.java
```diff
@@ -1695,16 +1696,16 @@ public static Path searchPath(String cmd, String paths, Predicate<Path> pathFilt
 
 	private static Stream<Path> executables(Path base) {
 		if (isWindows()) {
-			return Stream.of(Paths.get(base.toString() + ".exe"),
-					Paths.get(base.toString() + ".bat"),
-					Paths.get(base.toString() + ".cmd"),
-					Paths.get(base.toString() + ".ps1"));
+			return Stream.of(Paths.get(base + ".exe"),
```
maxandersen, line 1701: Will this work if in a bash env on windows that we don't check base ?

src/main/java/dev/jbang/cli/JBang.java
```diff
@@ -265,52 +270,37 @@ public static CommandLine getCommandLine(PrintWriter localout, PrintWriter local
...
-		if (!cmds.isEmpty()) {
-			sections.put("External", new ArrayList<>(cmds.keySet()));
-		}
-		CommandGroupRenderer renderer = new CommandGroupRenderer(sections, cmds);
-		return renderer;
+		return new CommandGroupRenderer();
 	}
 
-	private static Map<String, String> findExternalCommands() {
-		Map<String, String> result = new TreeMap<>();
```
maxandersen, line 283: Yeah my bad if i missed this. We already had this bad experience doing Quarkus cli and removed dynamic help listing and only include for commands that been "installed"

src/main/java/dev/jbang/util/Util.java
quintesse: This code has been here for ages, it has not changed. I only removed the unneeded `.toString()`.

But on WIndows things like bash will still run .exe and .bats. Given the fact that we only use `searchPath()` for finding executables and not bash scripts we're okay not including the extension-less name on Windows. Also, in the cases we'd like to find a command that might be a bash script, eg. `jbang` itself, on Windows it is still preferable to run the .cmd or .ps1 version (because the bash script uses wget or curl which might either not be available on Windows or work slightly differently). So unless we have a situation where we positively want to find a bash script on Windows this code will mostly do the right thing.

PR #2151, opened 2025-08-08
refactor: made findNearestWith() and plugin lookup code more flexible

quintesse:


maxandersen:
what does this actually do? just clean up or?

quintesse:
yes, no new features, no change in behaviour. It's just an internal API change to make it possible to use the code in more circumstances. I wrote it for the plugins, but now that it's not needed anymore I didn't want to lose those improvements.

Issue #2152, opened 2025-08-09
allow resolve properties on aliasrefs


maxandersen:
in //DEPS we allow use of property replacement to support things like autodetected properties like ${os.detected.*}  and user specified config ${quarkus.version}.

But we don't apply those when running as a scriptref.

Making 

`jbang https://github.com/HanSolo/jarkanoid/releases/download/17.0.17/jarkanoid-${os.detected.jfxname}-17.0.17.jar` not work....when it would be nice it could.

see https://github.com/HanSolo/jarkanoid/releases/download/17.0.17/jarkanoid-${os.detected.jfxname}-17.0.17.jar for the standard naming patterns 

note, its not just https, `jbang org.openjfx:javafx-graphics:11.0.2:${os.detected.jfxname}` fails too ...which it shouldn't imo given `//DEPS` works.

writing this out i'm thinking more and more that on scriptref's we should apply property replacement before we pass it further down...though we don't want to loose the original value either - which is where i'm currently stomped.

maxandersen:
@quintesse  i'm thinking we could probably put this into resourceResolvers - add a call to `PropertiesValueResolver.replaceProperties(item, ...)` in select places.

I think we didn't do that in start as a) resourceresolvers weren't around then b) resourceresolvers don't have access to the current build context so can't access relevant properties.

But maybe we should enable them to have that access? wdyt?

quintesse:
FOr me the question is more: why do you want to keep the original value around? That's normally not how property replacement works. You do it early so lower pieces of code don't need to deal with it, right?

And then there'd be no need to pass things (even deeper) down.

PR #2153, opened 2025-08-09
fix: resource references with property resolving

maxandersen:
minimal draft fix for #2152 

this feels almost too easy...but would this work ?

### DIFF
--- a/src/main/java/dev/jbang/source/resolvers/CombinedResourceResolver.java
+++ b/src/main/java/dev/jbang/source/resolvers/CombinedResourceResolver.java
@@ -1,14 +1,18 @@
 package dev.jbang.source.resolvers;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Properties;
 import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 
+import dev.jbang.dependencies.Detector;
 import dev.jbang.source.ResourceRef;
 import dev.jbang.source.ResourceResolver;
+import dev.jbang.util.PropertiesValueResolver;
 
 /**
  * A <code>ResourceResolver</code> that, when given a resource string will
@@ -19,14 +23,19 @@ import dev.jbang.source.ResourceResolver;
 public class CombinedResourceResolver implements ResourceResolver {
 	private final List<ResourceResolver> resolvers;
 
+	private Properties contextProperties = new Properties();
+
 	public CombinedResourceResolver(ResourceResolver... resolvers) {
+		new Detector().detect(contextProperties, Collections.emptyList());
+
 		this.resolvers = Arrays.asList(resolvers);
 	}
 
 	@Override
 	public ResourceRef resolve(String resource, boolean trusted) {
+		final String resolvedResource = PropertiesValueResolver.replaceProperties(resource, contextProperties);
 		return resolvers.stream()
-			.map(r -> r.resolve(resource, trusted))
+			.map(r -> r.resolve(resolvedResource, trusted))
 			.filter(Objects::nonNull)
 			.findFirst()
 			.orElse(null);
--- a/src/test/java/dev/jbang/cli/TestRun.java
+++ b/src/test/java/dev/jbang/cli/TestRun.java
@@ -2102,6 +2102,25 @@ public class TestRun extends BaseTest {
 	}
 
 	@Test
+	void testMagicPropertyViaRun() throws IOException {
+
+		String res = "https://github.com/HanSolo/jarkanoid/releases/download/17.0.17/jarkanoid-${os.detected.jfxname}-17.0.17.jar";
+		// todo fix so --deps can use system properties
+		CommandLine.ParseResult pr = JBang.getCommandLine()
+			.parseArgs("run", res);
+
+		Run run = (Run) pr.subcommand().commandSpec().userObject();
+
+		ProjectBuilder pb = run.createProjectBuilderForRun();
+
+		Project prj = pb.build(res);
+		String line = run.updateGeneratorForRun(CmdGenerator.builder(prj)).build().generate();
+
+		assertThat(line, containsString(" --module-path "));
+
+	}
+
+	@Test
 	void testScriptCliReposAndDeps(@TempDir File output) throws IOException {
 		String base = "" +
 				"public class test {\n" +


maxandersen:
might make a PropertyResourceResolver(ResourceResolver) to separate concerns.

maxandersen:
Things to check before approving: that we don't cache based on originalresourceref keys.

quintesse:
That's too easy, yes. There's no guarantee, nor reason, to assume that the `CombinedResourceResolver` is always the "first" (or "top") resolver to be called. Right now it is, but that's not required. So if that changes for some reason, perhaps even just for a single specific use-case in the future, this wouldn't work and we'd have to add property replacers to other resolvers as well.

So given the fact that you do nothing more than replace properties on the input I don't see why this can't be done _before_ that input gets passed to the resolver. It just seems to me we're mixing responsibilities here.

Edit: Personally I'd just look to all the places where we call the resolver and see if we can't put the property replacement there.

If that's somehow too much work I'd suggest a "fake" resolver which does nothing more than replace the properties before passing on the work to a wrapper resolver. ANd then make sure that we always return that one as the "top" resolver.

maxandersen:
> Edit: Personally I'd just look to all the places where we call the resolver and see if we can't put the property replacement there.

that made me go look and i located resolveChecked which is the "root" of all the CombinedResource resolvers I could find.

I updated it to do the property replacement - and since its in the project builder it can get and use the buildContext which has user specifcied propertiers.

So far thats nice; but I see two issues:

1) to my surprise we actually put system properties into this build context - not sure we should allow that ..or at least clean out a few keys like user.home...but not sure
2) it is also used for source ref resolution...meaning `//SOURCE  ${user.home}/somewhere` suddenly becomes accessible.

...writing this out I think i'm going to update this to take a parameter where the property replacement will be enabled and only do that for the "root" build/alias call.

maxandersen:
note: just realizing prperty replacement as is implemented now also support env.<key> for environment variable lookup so definitly dont want it available *everywhere*

maxandersen:
well heck - we already support property replacement in all Tags (incl. files) I just realize. look for "replaceProperties"

we avoid/reduce the impact on //FILES by files not allowing absolute paths only relative ones.

maxandersen:
so yeah "//FILES myfile=../../../.bash_profile" works ...but so does Path.of("../../../.bash_profile") in a main method of any java code ...so yeah; not any worse or better risk than what running anything from online apps does.

Hence adding verification/checksum checks is the more important part.

maxandersen:
@quintesse assuming tests passes I think this is good to go.

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -258,12 +258,41 @@ private List<MavenRepo> allToMavenRepo(List<String> repos) {
...
+	 * data.
+	 * 
+	 * i.e. jbang run someref${os.detected.os}.jar is ok but //FILES with
+	 * ${user.home} is definitely not.
+	 * 
+	 * @param resolver
+	 * @param resource
+	 * @return
+	 */
+	private ResourceRef resolveChecked(ResourceResolver resolver, String orginal_resource,
```
quintesse, line 284: This is already merged, but `original_resource`? Wtf? Did you use AI to generate this? 😁

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -258,12 +258,41 @@ private List<MavenRepo> allToMavenRepo(List<String> repos) {
...
+	 * and handles retry on caching.
+	 * 
+	 * Note: only enable property replacement if you know the resource is not
+	 * something that can be embedded/included directly in the project from users
+	 * data.
+	 * 
+	 * i.e. jbang run someref${os.detected.os}.jar is ok but //FILES with
+	 * ${user.home} is definitely not.
+	 * 
+	 * @param resolver
```
quintesse, line 280: Either put proper docs or remove the `@params` :-)

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -258,12 +258,41 @@ private List<MavenRepo> allToMavenRepo(List<String> repos) {
...
+	 * @return
+	 */
+	private ResourceRef resolveChecked(ResourceResolver resolver, String orginal_resource,
+			boolean propertyReplacement) {
+		final String resource = propertyReplacement
+				? PropertiesValueResolver.replaceProperties(orginal_resource, getContextProperties())
+				: orginal_resource;
+
+		if (!resource.equals(orginal_resource)) {
+			Util.verboseMsg("Resolving resource ref: " + orginal_resource + " -> " + resource);
```
quintesse, line 291: Logging this is good, but I think the wording could be improved. In this case if there are no properties to be replaced you won't get a "resolving ..." message, which is wrong because we definitely are resolving. So I'd say that we'd either have to _always_ log something or change the wording to something like "Applying properties" instead of "Resolving".

src/main/java/dev/jbang/source/ProjectBuilder.java
maxandersen: This is what was already there afaik. Unless you identify as AI it was human :)

quintesse: Don't gaslight me, this code is new, the diff shows you just added it :-)

I would never ever write a (Java) variable name with an underscore, the very idea gives me shivers 😁

maxandersen: The only answer must be that I'm an AI then :)

src/main/java/dev/jbang/source/ProjectBuilder.java
maxandersen: but something gets logged everytime - its just only when applying properties we put the "-> <property resolved value>" 

I did that to avoid it to say 99% of the time "Resolving resource ref: abc -> abc"

but if you can think of a better one do open pr.

quintesse: Well, that's why I said to simply change the text. Because it's not resolving, it's "applying properties to resource ref" or  something like that. Just to avoid any confusion that that line suggests that resolving has been done/performed.

maxandersen: But resolve xyz is what we wrote before applying any properties. So are you saying the old one was wrong?

quintesse: Maybe I'm just misunderstanding things. Doesn't matter. Not important,

PR #2174, opened 2025-08-12
fix: have way to force/encourage jbang to use ui not console/tty

### LINKED ISSUE #2150: explict control over notty ui?
in @jbangdev/jbang-launch i'm struggling that when i run jbang from a java process it is started in a way that jbang shell script always thinks it has a tty but it actually does not.

so i would like to explore how we can handle this better.

Should we make a JBANG_PREFER_GUI or similar that jbang-launch could set?

@quintesse wdyt?

maxandersen:
fixes #2150 

@quintesse wdyt?

### DIFF
--- a/src/main/java/dev/jbang/util/ConsoleInput.java
+++ b/src/main/java/dev/jbang/util/ConsoleInput.java
@@ -1,5 +1,6 @@
 package dev.jbang.util;
 
+import java.awt.GraphicsEnvironment;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


quintesse:
I don't know... isn't this too "dumb"? You're now assuming that a Gui will always be available, so if somebody sets that to "true" when they are using ssh or docker then things will break because there will be no fallback to the console, right?

maxandersen:
@quintesse fair point. I added a "if !headless" so it will only honor prefer_gui if gui is actually avaialble. wdyt?

PR #2180, opened 2025-08-14
docs: describe how to update dependencies

### LINKED ISSUE #2148: Document how to update //DEPS
Create a section called `Updating //DEPS` in dependencies.adoc describing how the `deps@jbangdev` command can be used to  find newer versions of dependencies.

$ jbang run deps@jbangdev asciidoctorj@asciidoctor

Image

wfouche:
Closes #2148


docs/modules/ROOT/pages/dependencies.adoc
```diff
@@ -78,6 +78,26 @@ is the equivalent of the following Maven dependency declaration:
 ```
 ====
 
+== Updating `//DEPS`
+
+The dependencies in JBang programs must be updated manually by editing the Java, Groovy, or Kotlin file. Use alias `deps@jbangdev` to check if never versions of dependencies are available.
```
maxandersen: Not *must* but *can* be updated manually. Since ie. Renovatebot actually can do it automatically. 

Also this is not updating but more reporting on what updates are available, right ?

wfouche: I've updated the text, and it is not more accurate.

maxandersen (approved):

PR #2183, opened 2025-08-15
fix: replace os detector with nisse

maxandersen:
kmotd is incompatible with newer maven including maveniverse so need to decoupe from it.

first attempt here is using Nisse from @cstamas but might go for what @aalmiray done in jreleaeser and simply copied in the needed classes - but this stuff really should be a independent library...so nisse it is for now :)

### DIFF
--- a/src/main/java/dev/jbang/dependencies/Detector.java
+++ b/src/main/java/dev/jbang/dependencies/Detector.java
@@ -1,38 +1,52 @@
 package dev.jbang.dependencies;
 
+import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
-public class Detector extends kr.motd.maven.os.Detector {
+import dev.jbang.util.Util;
+
+import eu.maveniverse.maven.nisse.core.NisseConfiguration;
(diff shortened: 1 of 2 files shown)


src/main/java/dev/jbang/dependencies/Detector.java
```diff
@@ -1,38 +1,52 @@
...
+				}
+			} else if (os.equals("windows")) {
+				os = "win";
 			}
-		} else if (os.equals("windows")) {
-			os = "win";
+			properties.setProperty("os.detected.jfxname", os);
+
+		} catch (IOException e) {
+			Util.warnMsg("Failed to detect OS properties, continuing without them.", e);
```
maxandersen, line 71: not a fan I have to do this - but also can't spot why .build() of config should fail...so maybe not a big issue.

raised question at https://github.com/maveniverse/nisse/issues/95

cstamas:
Please hop to https://github.com/maveniverse/nisse/releases/tag/release-0.6.0 where all issues were fixed 

Also see PR https://github.com/jbangdev/jbang/pull/2193

Issue #2189, opened 2025-08-23
sh.jbang.dev script does not add ~/.jbang/bin to Fish Shell's PATH


vorburger:
For me, on a fresh environment, with [my dotfiles](https://github.com/vorburger/vorburger-dotfiles-bin-etc), following https://www.jbang.dev/download/ for https://docs.enola.dev/use/#jbang, in a 🐟  https://fishshell.com :

```shell
$ curl -Ls https://sh.jbang.dev | bash -s - app setup
...

$ jbang app install enola@enola-dev
[jbang] Resolving dependencies...
[jbang]    dev.enola:enola:0.0.1-SNAPSHOT
[jbang] Dependencies resolved
[jbang] Command installed: enola

$ enola version
fish: Unknown command: enola
``` 

This has nothing to do with earlier Fish support related issues such as #732 or #875.

The root cause for this problem is that the sh.jbang.dev script adds this to `~/.bashrc` :

```
# Add JBang to environment
alias j!=jbang
export PATH="$HOME/.jbang/bin:$PATH"
```

but the Fish shell doesn't read `~/.bashrc`, of course.

@maxandersen would you like a PR with a proposed contribution with a fix for this problem?

Where is the source code of the bash install script that's served from https://sh.jbang.dev ?

@edewit FYI

maxandersen:
https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/cli/App.java#L439-L440

vorburger:
Doing (something like) https://github.com/vorburger/vorburger-dotfiles-bin-etc/commit/02bd6943fdd68fc65bdddfa2620ec764f22bff2f fixes (works around) it locally, for me.

> https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/cli/App.java#L439-L440

ACK, that would need something like the aforementioned... although probably better as an `~/.config/fish/conf.d/jbang.fish` ("overlay") than by touching the (shared) `~/.config/fish/config.fish`.

I'll see if I can have a go at that.

PR #2190, opened 2025-08-23
clean: Minor changes in class App

vorburger:
Just some clean-up before making actual functional for #2189.

@maxandersen


vorburger:
FTR: This is a subset of #2191 so if that gets merged then this one can be closed.

PR #2191, opened 2025-08-23
fix: Install adds ~/.jbang/bin to PATH via .config/fish/conf.d/jbang.fish

### LINKED ISSUE #2189: sh.jbang.dev script does not add ~/.jbang/bin to Fish Shell's PATH

vorburger:
Should fix #2189.

But I've written this "blindly", and not actual run it.

@maxandersen during development, how do I make it "do the same thing that `curl -Ls https://sh.jbang.dev | bash -s - app setup` would do", to test? What `./gradlew ...` do I need to run to get a `jbang` somewhere? And then launch it as ... what, `???/jbang app setup`, presumably?

Let's get #2190 merged first, so that I can rebase this and it only has the actual fix.

Draft, for now; until I've actually been able to test it.

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -8,7 +8,11 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -76,7 +80,7 @@ class AppInstall extends BaseBuildCommand {
 		// dependencyInfoMixin.validate();
 		boolean installed = false;
 		try {
-			if (scriptMixin.scriptOrFile.equals("jbang")) {
+			if ("jbang".equals(scriptMixin.scriptOrFile)) {
 				if (name != null && !"jbang".equals(name)) {
 					throw new IllegalArgumentException(
 							"It's not possible to install jbang with a different name");
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -444,14 +444,14 @@ class AppSetup extends BaseCommand {
 				// Update shell startup scripts
 				if (Util.isMac()) {
 					Path bashFile = getHome().resolve(".bash_profile");
-					changed = changeScript(binDir, jdkHome, bashFile) || changed;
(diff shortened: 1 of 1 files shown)


vorburger:
> @maxandersen during development, how do I make it "do the same thing that `curl -Ls https://sh.jbang.dev | bash -s - app setup` would do", to test? What `./gradlew ...` do I need to run to get a `jbang` somewhere? And then launch it as ... what, `???/jbang app setup`, presumably?

It's not via `./bootstrap.sh`, because that's currently broken; see #2192.

maxandersen:
I'm renovating a barn atm - maybe @quintesse who is master of shell integration have a minute?

maxandersen:
But basically just run jbang the script without curl. 

https://github.com/jbangdev/jbang/blob/main/src/main/scripts/jbang is what gets run.

maxandersen (dismissed):
I added a missing logging and the j! abbreviation for consistency with the other shells.

and thanks - this PR made me checkout fish again and realized it now works with jbang /// shell trick AND have much saner defaults and less setup required compared to bash/zsh so changed my default shell to fish now :)

vorburger:
> I added a missing logging and the j! abbreviation for consistency with the other shells.

Thank You! I don't have to do anything anymore, but just wait, right? Or do I need to "accept" anything? Just asking to avoid "PR deadlock".

> and thanks - this PR made me checkout fish again and realized it now works with jbang /// shell trick AND have much saner defaults and less setup required compared to bash/zsh so changed my default shell to fish now :)

Awesome! 👍🏽 😎 Welcome (back) to Fishing 🎣 with The 🐠 🐟 Great Fish Shel!

maxandersen:
Yes all good. I'll merge when I can verify the test passing

maxandersen:
@all-contributors add @vorburger for code

PR #2193, opened 2025-08-23
feat: use nisse 0.6 for os detection

cstamas:
Update PR #2183 to Nisse 0.6.0 and simplify things: we use only OS source, so no need for "compatibility", just apply naming strategy we need (prefix) and we are done.


maxandersen:
awesome @cstamas - the fixes in nisse makes this as simple as I hoped.  merging.

maxandersen (approved):

PR #2195, opened 2025-08-25
feat: fish completion

maxandersen:
Screenshot 2025-08-25 at 16 32 52
Screenshot 2025-08-25 at 16 32 03

PR #2196, opened 2025-08-25
fish completion using copy of picocli pr

maxandersen:
- **feat: fish completion**
- **feat: fish completion (forked while waiting for picocli)**

a version of #2195 which copies the picocli code into jbang - once picocli releases we can use it ...or keep this if they dont

### DIFF
--- a/src/main/java/dev/jbang/cli/Completion.java
+++ b/src/main/java/dev/jbang/cli/Completion.java
@@ -2,11 +2,24 @@ package dev.jbang.cli;
 
 import java.io.IOException;
 import java.io.PrintStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
 
 import picocli.AutoComplete;
 import picocli.CommandLine;
+import picocli.CommandLine.Model.CommandSpec;
+import picocli.CommandLine.Model.OptionSpec;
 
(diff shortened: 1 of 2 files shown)


maxandersen:
Screenshot 2025-08-25 at 16 32 03
Screenshot 2025-08-25 at 16 32 52

maxandersen:
changes no other code and i like fish now :)

src/main/java/dev/jbang/cli/Completion.java
```diff
@@ -27,4 +55,256 @@ public int completion() throws IOException {
 		out.flush();
 		return EXIT_OK;
 	}
+
+	// copied from picocli fish PR https://github.com/remkop/picocli/pull/2463
```
quintesse, line 59: So why not wait for that? It must be because you personally want to use it because otherwise I'd say : why waste the time :-)

quintesse (approved):

src/main/java/dev/jbang/cli/Completion.java
maxandersen: Because the pr for picocli been waiting for 2+ years and I'm now using fish because of its nice working out of box features :)

quintesse: Two years?? Wow. Ok yeah, good point :-)

PR #2197, opened 2025-08-25
fix: first catalog is now minimal

maxandersen:
got tired of java-agents, catalog and aliases was added unnecessarily when doing `jbang alias add -f . env@jbangdev`

this pr uses gson serializer to not serialize nulls/empties.


maxandersen:
Well they aren't nulls. They are empty. So since i override the serialie behavior i need to handle both cases. 

Ultimately better fix is probably to set them to null so we can represent them being present or not in the files and avoid unnecessary rewrites?

we would need to add more consistent null checks. 

Btw. We seems o create Json parser/writer lots of places - lots of duplicated init...

quintesse:
> Btw. We seems o create Json parser/writer lots of places - lots of duplicated init...

Yes, it didn't seem like a big deal, lots of example code seems to do the same so I assumed it was pretty cheap

maxandersen:
Less worried about cost - more that it's multiple calls duplicated multiple places.

maxandersen:
merging for now as adding null checks everywhere turned to be much bigger refactor.

if someone complains about empty list/sets being rewritten to not being specifed we can look into handle that case.

Issue #2198, opened 2025-09-05
network logging is annoying


maxandersen:
this output on `--verbose` is not useful imo:

```
[jbang] [2025-09-05 10:25:13] X509Certificate: Alg:SHA1withRSA, Serial:20:06:05:16:70:02, Subject:OU=certSIGN ROOT CA, O=certSIGN, C=RO, Issuer:OU=certSIGN ROOT CA, O=certSIGN, C=RO, Key type:RSA, Length:2048, Cert Id:4005006620, Valid from:04/07/2006, 19.20, Valid until:04/07/2031, 19.20
[jbang] [2025-09-05 10:25:13] X509Certificate: Alg:SHA256withRSA, Serial:11:00:34:b6:4e:c6:36:2d:36, Subject:OU=certSIGN ROOT CA G2, O=CERTSIGN SA, C=RO, Issuer:OU=certSIGN ROOT CA G2, O=CERTSIGN SA, C=RO, Key type:RSA, Length:4096, Cert Id:2181086866, Valid from:06/02/2017, 10.27, Valid until:06/02/2042, 10.27
[jbang] [2025-09-05 10:25:13] X509Certificate: Alg:SHA1withRSA, Serial:15:c8:bd:65:47:5c:af:b8:97:00:5e:e4:06:d2:bc:9d, Subject:OU=ePKI Root Certification Authority, O="Chunghwa Telecom Co., Ltd.", C=TW, Issuer:OU=ePKI Root Certification Authority, O="Chunghwa Telecom Co., Ltd.", C=TW, Key type:RSA, Length:4096, Cert Id:3632331159, Valid from:20/12/2004, 03.31, Valid until:20/12/2034, 03.31
[jbang] [2025-09-05 10:25:13] X509Certificate: Alg:SHA1withRSA, Serial:00, Subject:OU=Go Daddy Class 2 Certification Authority, O="The Go Daddy Group, Inc.", C=US, Issuer:OU=Go Daddy Class 2 Certification Authority, O="The Go Daddy Group, Inc.", C=US, Key type:RSA, Length:2048, Cert Id:4023522997, Valid from:29/06/2004, 19.06, Valid until:29/06/2034, 19.06
[...]

PR #2199, opened 2025-09-05
fix: only activate dev.jbang logging when verbose

maxandersen:
suggested fix to #2198 

dont activate logging for all of jdk (""), just "dev.jbang"

wdyt @quintesse ?


maxandersen:
merging this as current log levels are just too noisy. @quintesse if something missing due to devkitman we can patch release.

PR #2200, opened 2025-09-06
fix: improved WebServer example

wfouche:
The httpserver dep is not required, and it also uses an older version of httpserver than what is included in Java 8+


wfouche:
Testing with Java 8

```
jbang run --java 8 WebServer.java
[jbang] Building jar for WebServer.java...
Current Java Version: 1.8.0_462
Server started at http://localhost:8000
```

Testing with Java 21

```
jbang run --java 21 WebServer.java
[jbang] Building jar for WebServer.java...
Current Java Version: 21.0.8
Server started at http://localhost:8000
```

maxandersen:
Afair this doesn't build/work in ides without the external classes on the classpath?

In any case the point of the example is showing how with some dependency can do something useful. 

So removing this deps would make the example kinda not relevant

wfouche:
> Afair this doesn't build/work in ides without the external classes on the classpath?

It does work fine without explicitly defining the httpserver dep because this dependency has been superceeded by built-in JVM functionality, IDEs have no problem resolving the classnames.

image

wfouche:
> In any case the point of the example is showing how with some dependency can do something useful.

In that case it would be a better example if a well-maintained dependency is used instead. Will update the example to use Undertow.

wfouche:
image

readme.adoc
```diff
@@ -131,21 +131,24 @@ class hello implements Runnable {
...
-        HttpServer server = HttpServer.create(new InetSocketAddress(8000), 0);
-        server.createContext("/", exchange -> {
-            String response = "Hello from JBang!";
-            exchange.sendResponseHeaders(200, response.length());
-            exchange.getResponseBody().write(response.getBytes());
-            exchange.close();
-        });
+        Undertow server = Undertow.builder()
+            .addHttpListener(8000, "localhost")
+            .setHandler(new WebServer()).build();
```
maxandersen: could simplify it to not require separeate method declaration, inheritance and the new constructor.

```suggestion
            .setHandler(exchange -> exchange.getResponseSender().send("Hello from JBang!")).build();
```

wfouche:
The implementation using Javalin is very simple.

image

image

readme.adoc
```diff
@@ -131,23 +131,17 @@ class hello implements Runnable {
...
-            String response = "Hello from JBang!";
-            exchange.sendResponseHeaders(200, response.length());
-            exchange.getResponseBody().write(response.getBytes());
-            exchange.close();
-        });
-        server.start();
-        System.out.println("Server started at http://localhost:8000");
+    public static void main(String[] args) {
+        String m = "Hello from JBang!";
+        Javalin app = Javalin.create(/*config*/)
```
maxandersen: no point in assigning it to a variable if you dont want to call start separately.

PR #2205, opened 2025-09-11
fix: moved weather.java to examples folder

wfouche:
It is better to keep larger example programs as runnable source files in the documention tree rather than fully embedding them in [source] blocks in Asciidoc files.

Moved the weather.java example program to examples/weather.java and using include::example$weather.java[] to reference the source file from the Asciidoc documentation.

$ ./weather.java London


wfouche:
@maxandersen , please review.

quintesse:
I'm not sure this is correct, AFAIK all examples were moved to https://github.com/jbangdev/jbang-examples a long time ago. Not sure if it would be confusing to re-introduce an examples folder again. Then what examples should go here and which ones should go into the examples repo? But let's see what @maxandersen has to say about this.

wfouche:
This is an Antora best practice. It supports 4 folders for content: pages, partials, images and examples.

quintesse:
Ok, so perhaps we should move all that into the docs folder itself? Just seems messy to have all those toplevel folders just because a tool does it like that by default.

Edit: it would also make it obvious that the examples (and images?) are related to the docs.

wfouche:
> Ok, so perhaps we should move all that into the docs folder itself?

This will work, provided we publish all the examples in an appendix or other sub-section in the documentation. I like this idea, because then the examples are easier to find by JBang users and the maintainers of the documentation will ensure that the examples are all nicely formatted, that they are concise, and have up to date dependencies, etc.  The jbang-examples repo can then also be removed.

In the Readme.adoc file in the examples folder I'm already talking about how the source files should be formatted.

```bash
export GJF_JAR=com.google.googlejavaformat:google-java-format:1.28.0

jbang run $GJF_JAR -r weather.java
```

And just have a small but carefully curated set of example programs.

docs/modules/ROOT/examples/README.adoc
```diff
@@ -0,0 +1,16 @@
+= JBang Examples
+
+This folder contains example programs that are included in the JBang documentation. Keeping the examples as runnable source files outside of the Asciidoc documentation files ensures that the programs can be tested and updated without having to update any documentation files.
+
+== Formatting
+
+In order to make best use of the limited space available to present source code examples, we use the Google Java Formatting style which indents the code by two spaces.
+
+The Google Java Formatting style do not support JBang comments such as `//DEPS` and will format it as `// DEPS`. After formatting a program file, edit it manually to remove the space after each `//`.
```
maxandersen: why can't we just use the default formatting we apply on all of jbang source code?

maxandersen: to be clear - not going to be ok relying on google fmt tool until they fix their bad stance on forced // formatting :)

wfouche:
Got spotless working via Maven (pom.xml) and ignoring all JBang directives using `spotless:{off,on}`.

```java

// spotless:off
//DEPS info.picocli:picocli:4.7.7
//DEPS com.fasterxml.jackson.core:jackson-core:2.20.0
//DEPS com.fasterxml.jackson.core:jackson-databind:2.20.0
// spotless:on

import java.net.URI;
import java.net.http.HttpClient;
import java.net.http.HttpRequest;
import java.net.http.HttpResponse;
import picocli.CommandLine;
import picocli.CommandLine.Parameters;
```

But it still reformats the first line which we currently cannot do much about.

* https://github.com/google/google-java-format/issues/1217
* https://github.com/google/google-java-format/issues/1218

It is easy to fix though, and I've created a simple script to do just that.

docs/modules/ROOT/examples/README.adoc
wfouche: > to be clear - not going to be ok relying on google fmt tool until they fix their bad stance on forced // formatting :)

I got it working and `// spotless:{off,on}` are recognised.

maxandersen:
Huh? Spotless doesn't use Google format it uses eclipses formatter afaik?

wfouche:
> Huh? Spotless doesn't use Google format it uses eclipses formatter afaik?

The actual formatter that Spotless use to format Java source code is configurable. It can be Google Java Formatter, Eclipse's Formatter, or others. 

image

maxandersen:
Yes, I get that spotless in general can use other formatters.
What I don't get is why the wish of using google java formatter in this repo for the docs when the spotless config we use today is eclipse format based.

Are you suggesting we maintain two different java formatting styles?

Thats why I asked: "why can't we just use the default formatting we apply on all of jbang source code?"

wfouche:
> Thats why I asked: "why can't we just use the default formatting we apply on all of jbang source code?"

I was trying to find an easy to use source formatting solution that looks great in the documentation (uses the available space optimally) but one that we can document so that JBang users can also adopt the same strategy, and format JBang source file uniformingly. Google Java Formatter is easy to use but it breaks JBang directives. So I had to look at spotless again. I'll test the default formatting style and formatter used by the JBang project itself.

wfouche:
Now using `jbang-fmt` as formatter.

docs/modules/ROOT/examples/formatExamples.sh
```diff
@@ -1,10 +1,3 @@
...
 
-mvn spotless:apply
-
-find src -name "*.java" | while read -r file; do
-    echo "Processing file: $file"
-    jbang run FixJBangLine1.java $file
-done
-
-rm -f -r target
+jbang run jbang-fmt@jbangdev/jbang-fmt --style google --line-length 90 --java 21 --indent-with space --indent-size 2 src
```
maxandersen: why are you insisting on not using the style we use everywhere else ? :)

just use the style file in the project and you should be good, right?

wfouche: You're right. Slowly acclimatizing.  :-)

wfouche:
@maxandersen , the `default` style is now used.

maxandersen:
? Wdym

wfouche:
> ? Wdym

Typo. I corrected the sentence. "now"

docs/modules/ROOT/examples/formatExamples.cmd
```diff
@@ -0,0 +1 @@
+jbang run jbang-fmt@jbangdev/jbang-fmt --line-length 90 --java 21 --indent-with space --indent-size 2 src
```
maxandersen: instead of having 2 files for these just make a jbang-catalog file with a jbang-fmt alias in ...then you just need to run `jbang jbang-fmt` no matter what OIS you are on

maxandersen: btw. why not just include this source folder in the spotlessapply build? then its formatted automatically?

maxandersen: aah - indent-size 2 ...got it. we could consider changing our default too but thats a bigger change:)

wfouche:
> all good now, but i suggest you make that jbang-catalog instead of separate commands per os.

Done. That is a nice improvement.

docs/modules/ROOT/examples/README.adoc
```diff
@@ -0,0 +1,11 @@
+= JBang Examples
+
+This folder contains example programs that are included in the JBang documentation. Keeping the examples as runnable source files outside the Asciidoc documentation files ensures that the programs can be tested and updated without having to update any documentation files.
+
+Run the following command to apply formatting changes to the example Java files in the `src` folder.
+
+[source,shellscript]
+----
+jbang run jbang-fmt
```
maxandersen:
```suggestion
jbang jbang-fmt
```

run is implied

maxandersen (approved):
thanks! merging

PR #2206, opened 2025-09-12
feat: requestedJavaVersion honored by init templates

maxandersen:
alternative/replacement of #2143 

in #2143 it was suggested to force add "//JAVA 25+" when compact source enabled (automatically enabled when Java > 25) but I turned it down as it was resulting in inconsistent behavior.

Since then I myself been caught by having done `jbang init --java 25 hello.java` and then when running `./hello.java` it didn't work because my environments java was less than Java 25.

so this PR suggests introducing a `requestedJavaVersion` property which will be set to null if no specific java version requested; and other set to the raw string of the arg value in `--java <arg>`

Then use that to generate `//JAVA ${requestedJavaVersion}` if requested version is present, and in the case of compact source files will get `//JAVA 25+` if you did not request a specific java file...meaning no matter how you get a compact source file - it will always have at least `//JAVA 25+` in it OR what you requested using `--java <arg>`

In addition I updated the compact source to use `IO.println()` to be able to cut out the import line.

```
//JAVA 25+

void main(String... args) {
    IO.println("Hello World");
}
```

I also added ```{#if requestedJavaVersion}
//JAVA {requestedJavaVersion}
{/if}```

to the templates I could see it being meaningful.

wdyt @wfouche and @quintesse - I think this should make us all more happy than annoyed ;)

### DIFF
--- a/src/main/java/dev/jbang/cli/Init.java
+++ b/src/main/java/dev/jbang/cli/Init.java
@@ -89,6 +89,7 @@ public class Init extends BaseCommand {
 		int reqVersion = buildMixin.javaVersion != null ? JavaUtil.minRequestedVersion(buildMixin.javaVersion)
 				: JavaUtil.getCurrentMajorJavaVersion();
 
+		properties.put("requestedJavaVersion", buildMixin.javaVersion);
 		properties.put("javaVersion", reqVersion);
 		properties.put("compactSourceFiles", reqVersion >= 25);
 		// properties.put("magiccontent", "//no gpt response. make sure you ran with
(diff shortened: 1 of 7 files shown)


maxandersen:
btw. yes - this is up to the individual templates wether they want to honor this or not. but we already have templates setting specific java versions so i dont think there is much way around this.

wfouche:
I like this change, and it is future proof.  :-)

quintesse:
I think it's a good compromise. Merge it! :-)

quintesse (approved):

PR #2207, opened 2025-09-13
fix: make tests faster by doing less

maxandersen:
used the perf graph in allure report to identify the top candidates for tests being slow.

reduced 3-4 tests of 15-30s length due to excessive dependency fetching which was not really needed for what they tested.


quintesse:
Yay! 👍

Issue #2212, opened 2025-09-15
HTTP: 403 - Access Denied


reportmill:
It seems my website is starting to reject jbang requests. Here is an example:

`jbang https://reportmill.com/jbang/SnapCodeAll.java`

With --verbose I get:

```
[jbang] [0:173] Resolving resource ref: https://reportmill.com/jbang/SnapCodeAll.java
[jbang] [0:235] Requesting HTTP GET https://reportmill.com/jbang/SnapCodeAll.java
[jbang] [0:235] Headers {User-Agent=[JBang/0.130.0 (Mac OS X/15.5/aarch64) Java/17.0.12/Oracle Corporation]}
[jbang] [0:611] HTTP: 403 - Access Denied. You do not have permission to perform this action. Contact your administrator
[jbang] [0:618] [ERROR] Could not download https://reportmill.com/jbang/SnapCodeAll.java
dev.jbang.source.ResourceNotFoundException: Could not download https://reportmill.com/jbang/SnapCodeAll.java
	at dev.jbang.source.resolvers.RemoteResourceResolver.lambda$resolve$0(RemoteResourceResolver.java:49)
	at dev.jbang.source.resolvers.FileResourceResolver$FileResourceRef.getFile(FileResourceResolver.java:87)
	at dev.jbang.source.ProjectBuilder.resolveChecked(ProjectBuilder.java:285)
	at dev.jbang.source.ProjectBuilder.build(ProjectBuilder.java:260)
	at dev.jbang.cli.Run.doCall(Run.java:72)
    <more deleted>
```

My website is served up by Cloudflare. Curl or Chrome seem to fetch this url file fine. I wonder if Cloudflare doesn't like the user-agent for some reason.

reportmill:
As a sanity check, I decided to set up a GitHub jbang-catalog repo, and I'm having a similar issue:

```
[jbang] [0:413] Requesting HTTP GET https://raw.githubusercontent.com/reportmill/jbang-catalog/HEAD/jbang-catalog.json
[jbang] [0:414] Headers {User-Agent=[JBang/0.130.0 (Mac OS X/15.5/aarch64) Java/17.0.12/Oracle Corporation]}
[jbang] [0:483] No catalog found at https://github.com/reportmill/jbang-catalog/blob/HEAD/jbang-catalog.json
dev.jbang.source.ResourceNotFoundException: Could not download https://github.com/reportmill/jbang-catalog/blob/HEAD/jbang-catalog.json
[...]

PR #2213, opened 2025-09-16
fix: avoid java in user agent

maxandersen:

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -698,7 +698,17 @@ public class Util {
 					+ System.getProperty("java.vm.vendor", "<unknown>");
 			agent = "JBang/" + getJBangVersion() + " (" + System.getProperty("os.name") + "/"
 					+ System.getProperty("os.version") + "/" + System.getProperty("os.arch") + ") " + "Java/" + version;
+
+			// This is a hack to avoid being blocked by some websites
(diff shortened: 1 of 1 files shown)


maxandersen:
simple fix for #2212 - going with "avaj" as "j.a.v.a" is a mess to deal with in the queries/data processing I need to do when processing the agent string in data analytics.

src/main/java/dev/jbang/util/Util.java
```diff
@@ -705,6 +705,12 @@ static private String getAgentString() {
 			agent = "JBang/" + getJBangVersion() + " (" + System.getProperty("os.name") + "/"
 					+ System.getProperty("os.version") + "/" + System.getProperty("os.arch") + ") " + "Java/" + version;
 		}
+
+		// This is a hack to avoid being blocked by some websites
```
quintesse: Why add these lines? Why not simply edit line 706?
AFAIK the "agent" var is there only for caching purposes (not really sure why)

maxandersen: good catch - should be within the if.

I do it on the full string as the values fetch from vendor, version, arch etc. could in theory also contain the sequence "java"

maxandersen:
I pushed update - the corresponding update to jbang.dev workers was done in https://github.com/jbangdev/cloudflare-worker-ga4/commit/c206ae5faea09e61a39c92012e9f5176fe146f20

PR #2215, opened 2025-09-17
refactor: TagReader does much less data processing

quintesse:
Before we would parse all directives and convert them to internal objects as much as possible. Now we do as little processing as possible and mostly just return the raw data. This makes TagReader simpler and will allow us to extract it into its own library later.


src/main/java/dev/jbang/catalog/Alias.java
```diff
@@ -44,6 +44,8 @@ public class Alias extends CatalogItem {
 	public final Boolean nativeImage;
 	@SerializedName(value = "native-options")
 	public final List<String> nativeOptions;
+	@SerializedName(value = "source-type")
```
maxandersen, line 47: this is new feature,right? being able to set source type on aliases?

src/main/java/dev/jbang/resources/resolvers/AliasResourceResolver.java
```diff
@@ -1,4 +1,4 @@
-package dev.jbang.source.resolvers;
+package dev.jbang.resources.resolvers;
```
maxandersen, line 1: is the package rename really necessary?

just asking as without it this becomes a much smaller change.

src/main/java/dev/jbang/resources/resolvers/ClasspathResourceResolver.java
```diff
@@ -64,5 +69,14 @@ private static URL getResourceUrl(String resource) {
 		public boolean exists() {
 			return getResourceUrl(originalResource) != null;
 		}
+
+		@Override
+		public ResourceRef resolve(String resource, boolean trusted) {
+			if (!Util.isValidPath(resource)) {
+				return null;
+			}
+			String sibRef = "classpath:" + Paths.get(originalResource.substring(11)).resolveSibling(resource);
```
maxandersen, line 78: didn't realize we had classpath: syntax....so disappointing this fails:

```
jbang classpath:/pom.qute.xml
[jbang] [ERROR] Getting contents from resource not supported
```

or maybe not - can't really tell.

src/test/java/dev/jbang/source/TestSource.java
```diff
@@ -322,8 +322,8 @@ void testExtractGroovyOptions() {
 
 		assertEquals(Arrays.asList("--enable-preview"), s.getCompileOptions());
```
maxandersen, line 323: why the order flip? (no, probably fine - just curious as feel weird)

also ...why does it have to --enable-preview....?

src/main/java/dev/jbang/catalog/Alias.java
quintesse: Yes, although I would say it was more a _missing_ feature, because in general all command line options should be made available in aliases as much as possible. I guess this one just slipped through the cracks.

src/main/java/dev/jbang/resources/resolvers/AliasResourceResolver.java
quintesse: of course it's not _necessary_ but the code becomes a bit cleaner and easier to understand which classes belong together. There's lots of classes in "source" that are not directly related.  Right now with thee two new packages there's still "building" and "project" handling in there that at some point I'd like to split apart.

src/main/java/dev/jbang/resources/resolvers/ClasspathResourceResolver.java
quintesse: Never really tried that, it's something that's mostly used internally, but it _should_ work. Let me see if I can figure out why it doesn't.

src/test/java/dev/jbang/source/TestSource.java
quintesse: There's actually not a real reason. When refactoring the code I felt that the COMPILE/RUNTIME_OPTIONS were the "real" options and the JAVA[C]_OPTIONS were the legacy, old, options. So the "real" options should go first. But that was me assuming there was really no difference, especially because who would use both (real and old) options in the same file?
But if you think there could be an issue I'll change it back.

src/main/java/dev/jbang/resources/resolvers/ClasspathResourceResolver.java
quintesse: Ah yes, I now know why this doesn't work.

So first, you have to understand that the `classpath:/pom.qute.xml` part is a `ResourceRef`. Which support classpaths. But this is because resource refs are used throughout JBang for accessing resources that might come from anywhere. So for example classpath resources are often used for templates.

The other thing is that you can access the contents of resource refs either by file path or by input stream. THe internal code can decide which one it prefers. So for template handling it uses input streams. But in this case if you run `jbang classpath:` it has to go to the compiler, which means it needs the file version.

And this is the issue: because classpath resources were never used in their file form, because it means making a temporary file copy of something in a JAR on the file system, and why would you do that if you can use the input stream? So I never implemented that part for the `ClasspathResolver`.

Now, it would be pretty easy to implement, but I'm not sure there is an actual use-case? The only thing you could access with them would be internal JBang resources. And how would people know they even exist and what their names are?

We could also go the opposite way and remove the `ClasspathResolver` from the public lookup, which would cause `jbang classpath:/pom.qute.xml` to generate an error. And we'd only leave the resolver for internal lookups. (I'll have to double-check that won't complicate things, but I think it won't)

quintesse: PS: I just thought of something. In the `jbang init` command the `-t/--template` option takes an alias name. But this could easily be changed to a resource ref, which also supports aliases of course. In which case you could actually refer directly to template files, imagine `jbang init -t https://github.com/jbangdev/jbang-catalog/blob/main/templates/qmcp.java.qute`.

And in _that_ case `jbang init -t classpath:/pom.qute.xml ...` would work.

quintesse:
> whats missing from going out of draft?

nothing important that can't be done in a future PR

Because of all the changes I'd prefer to merge this ASAP.

Also because I want to use it to show a PoC implementation of scope parsing, which should now be much easier and much cleaner to do. Btw, what do you want to call the things that we put between `{ ... }` in a //DEPS or //COMPILE_OPTIONS line? Modifiers? Attributes? Properties? 

Edit: ready to merge

quintesse:
No wait, I want to take a look at a possible issue with integrations (I think there should be an issue but all IT tests work, so I'm not sure)

quintesse:
Ok, so that is now fixed. So I flipped the switch again to "ready".

quintesse:
@maxandersen I know you basically gave your OK already, but perhaps you could merge or at least approve this PR? It's pretty big so I don't want to "force merge" :-)

src/main/java/dev/jbang/resources/resolvers/ClasspathResourceResolver.java
maxandersen: that would actually be nice - skip need for alias/catalog definitions..though only work for simple things as name patterns not possible to define this way. worth exploring - but in separate pr/issue - this one is big enough as is :)

src/test/java/dev/jbang/source/TestSource.java
maxandersen: ok, so should be fine/ not have impact.

maxandersen:
i see the changes for handling quotes on various comand separated list butare there tests for them im missing?

quintesse:
> but are there tests for them im missing?

Actually my code changes made this error appear in existing Grapes tests, so the fact that we don't have an error anymore means it works :-)

But I've added a new test that explicitly tests the new quoting in //DEPS.

PR #2221, opened 2025-09-23
feat: added `jdk exec` command

quintesse:
Also introduced single letter aliases for several of the jdk subcommands.

### DIFF
--- a/src/main/java/dev/jbang/cli/Jdk.java
+++ b/src/main/java/dev/jbang/cli/Jdk.java
@@ -256,6 +256,10 @@ public class Jdk {
 			if (Util.getShell() == Util.Shell.bash) {
 				fullCmd = "env PATH=\"" + home + File.separator + "bin:$PATH\" JAVA_HOME='" + home + "' "
 						+ fullCmd;
+			} else if (Util.getShell() == Util.Shell.powershell) {
+				fullCmd = "{ $oldPath, $env:PATH, $oldHome, $env:JAVA_HOME=$env:PATH, \"" + home
+						+ "\\bin;$env:PATH\", $env:JAVA_HOME, '" + home + "' ; " + fullCmd
+						+ " ; $env:PATH, $env:JAVA_HOME=$oldPath, $oldHome }";
 			} else {
 				String path = home + "\\bin;" + System.getenv("PATH");
 				fullCmd = "set \"PATH=" + path + "\" && set \"JAVA_HOME=" + home + "\" && " + fullCmd;
--- a/src/main/scripts/jbang.ps1
+++ b/src/main/scripts/jbang.ps1
@@ -108,6 +108,7 @@ if (Test-Path "$jarPath.new") {
 
 # Find/get a JDK
 $JAVA_EXEC=""
+$oldJavaHome=$env:JAVA_HOME
 if (Test-Path env:JAVA_HOME) {
   # Determine if a (working) JDK is available in JAVA_HOME
   if (Test-Path "$env:JAVA_HOME\bin\javac.exe") {
@@ -156,9 +157,8 @@ if ($JAVA_EXEC -eq "") {
   }
 }
 
-$env:JBANG_RUNTIME_SHELL="powershell"
-$env:JBANG_STDIN_NOTTY=$MyInvocation.ExpectingInput
-$env:JBANG_LAUNCH_CMD = $PSCommandPath
+$oldShell, $oldNotty, $oldCmd=$env:JBANG_RUNTIME_SHELL, $env:JBANG_STDIN_NOTTY, $env:JBANG_LAUNCH_CMD
+$env:JBANG_RUNTIME_SHELL, $env:JBANG_STDIN_NOTTY, $env:JBANG_LAUNCH_CMD="powershell", $MyInvocation.ExpectingInput, $PSCommandPath
 $output = & "$JAVA_EXEC" $env:JBANG_JAVA_OPTIONS -classpath "$jarPath" dev.jbang.Main @args
 $err=$LASTEXITCODE
 
@@ -173,3 +173,5 @@ if ($err -eq 255) {
 } else {
   break
 }
+
+$env:JAVA_HOME, $env:JBANG_RUNTIME_SHELL, $env:JBANG_STDIN_NOTTY, $env:JBANG_LAUNCH_CMD=$oldJavaHome, $oldShell, $oldNotty, $oldCmd


src/main/java/dev/jbang/cli/Jdk.java
```diff
@@ -235,6 +236,37 @@ public Integer javaEnv(
...
+			jdk = jdkMan.getOrInstallJdk(versionOrId);
+		}
+		if (jdk.isInstalled()) {
+			Path home = ((dev.jbang.devkitman.Jdk.InstalledJdk) jdk).home();
+			String path = home + java.io.File.separator + "bin" + java.io.File.pathSeparator + System.getenv("PATH");
+			Path cmd = Util.searchPath(args.get(0), path);
+			if (cmd != null) {
+				args.set(0, cmd.toString());
+			}
+			ProcessBuilder pb = CommandBuffer.of(args).asProcessBuilder();
```
maxandersen: This doesn't seem right. 

Shouldn't jbang be printing out the command for the surrounding shell script to execute ?

maxandersen (commented):
What's the use case here ?

one off command execution of different java versions ?

maxandersen:
> What's the use case here ?
> 
> one off command execution of different java versions ?

from chat:

You can now for example do jbang jdk x -j 25 java -version and get your environment set to the proper Jdk without having to mess about with your environment. Very useful for one-shot commandsTako Schotanus: (in this case for example for printing which exact version of 25 you have installed) You can also do jbang jdk x -j 25 bash (or zsh or fish of course) and have more or less the same result as eval $(jbang jdk env 25)

^^^ put that in the docs :)

src/main/java/dev/jbang/cli/Jdk.java
quintesse: No, because you can't set the env vars and execute the command on Windows (at least not in a single line, which is the only thing we support on Windows). Even if we could we would be directly changing the user's env vars in their shell, affecting all future commands typed. (Windows does not have a way to set env vars only for the next command)

maxandersen: after seeing your rationale for the issue I grok it.

Thought initially it was just to run a java command ...not about running with env setup.

maxandersen:
something goes wrong with output handling.

`just jbang jdk x --java 23 bash`

now when typing in commands like `which java`, `ls` etc. no output is rendered until I exit bash.

quintesse:
> no output is rendered until I exit bash.

Ha! I hadn't noticed that :-)

quintesse:
I _know_ how to fix it, we do it elsewhere in the JBang code as well, but I hate it. The way to do it is the actively read all lines from the command's output stream and copy them to JBang's stdout. But I'm always thinking: why do I need to actively do anything? Yes, if I want to capture the output, but why can't I just fire-and-forget with inheritIO when I don't care about capturing?

quintesse:
Actually can't get this to work right. No idea what's going on.
Simple command, like java, work, but I can't get shells to work correctly.

maxandersen:
you'll need a separate thread to pump the stream.

thats what jash or ztexec does :)

quintesse:
I did that, 2 even, one for stdout and one for stderr, but it doesn't seem to work.

I can push that code if you want to take a look, perhaps you see what I'm doing wrong.

maxandersen:
hmm - then i reckon its due to https://github.com/jbangdev/jbang/blob/main/src/main/scripts/jbang#L226

that jbang execution is done in an expression.

That leads me back to my original comment that it would be more right if you generate a command which will be what the script executes.

That command could be jbang itself again.....but this time doing the actual process call.

its hacky...but I think that would work.

quintesse:
> generate a command which will be what the script executes.

sure, but like I said, I don't see how to do that on Windows.
- first of all technically, that I don't see how to set up the environment and run the command all in a single line, 
- or how to make it execute multiple lines (the latter is likely possible but the whole parsing of the JBang output is already pretty nasty), but then we'd be affecting the user's environment variables... which is exactly what I'm trying to avoid with this command

maxandersen:
I think i wasn't explaining my suggestion well enough - I'm not suggesting you will set the environemnt etc.

today jbang's main method simply prints the corresponding java command to run, when you do `jbang run xyz.java`, it does necesary builds and then prints out something akin to `<path to java> <classpath> ..etc.` which then gets executed by theshell script.

I'm saying have `jbang jdk exec --java 24 bash` have it generate a command which will in turn run the command.

either havea a `jbang secret exec` that does it or heck; generate a .java file which has the environment settings defined and you then build and execs. 

that should give you the effect you are after.

does that make sense?

maxandersen:
to be precise:

Today the code does this:

run jbang/jbang.cmd -> call java jbang that nested run process -> exit to script with a "dont run anything exit code"

what I suggest is:

run jbang/jbang.cmd -> call java jbang that generates "something" that can be run -> exit to script with "command line" to run -> "command line" gets executed.

quintesse:
I _guess_ that's a possibility... 🤔 

The "generating something" would only be necessary on Windows as on Linux/Mac we _can_ actually output a command that both sets the env vars and runs the user command. But on Windows we could generate a temp bat file or something.

The only downside I see is that we'd probably end up with a whole bunch of temp files we're not able to delete.

maxandersen:
we already have a cache that grows over time; you can avoid it by having a "special" exec command that just does what you do now...

ie.

`jbang jdk exec x y z` will result in running `java -m dev.jbang.cli.Main jbang.jar jdk exec --direct x y z`

quintesse:
I'm not sure how that special command would work? You mean that it would skip the eval part of the startup script and run "x y z" with the ProcessBuilder?

maxandersen:
more or less yes.

maxandersen:
@quintesse  pushed code that adds --direct flag and uses raw java to execute jbang.jar

I at first had it generate a java code that i used jbang command to execute but jbang shell wrapper is the culprit for the buffering of stderr so i went back to basics and just uses the selected java to execute jbang directly.
 
so now `jbang jdk x bash` works.

But There is though *one* problem.

`jbang jdk x bash -c "java -version"` does not work because of "lost quotes"

somewhere in the pipeline "java -version" gets split into separate arguments. Didn't figure out where exactly but thought I wuold share a better working version... and maybe you realize where the quoting is best done.

maxandersen:
btw. any reason jdk command aren't extending BaseCommand so it gets --config etc. ?

quintesse:
> pushed code that adds --direct flag and uses raw java to execute jbang.jar

Ok, I'll take a look. But I thought we had determined that we could just run the command directly? Simply as `PATH=xxx JAVA_HOME=xxx <user command>` on Linux/Mac and executing a temp batch file on Windows?

Because running JBang twice (and therefore the JVM) probably doesn't make things any faster 😅

quintesse:
> btw. any reason jdk command aren't extending BaseCommand so it gets --config etc. ?

Most likely because the `jdk` command doesn't need `--config`? (Or any of the other things from BaseCommand, but to be honest I have no idea, it's been a while :-) )

Edit: took a look at BaseCommand again and I don't really see any good reason why Jdk should _not_ extend it. So really no idea :-)

maxandersen:
-1 on making the execution different between the OS if we can avoid it - and we are talking running just jbang.jar so it's a tiny overhead imo.

quintesse:
I don't think it's so tiny, we just doubled the startup time (and JBang already isn't a speed monster). Also, it's not OS-specific, it's Shell-specific, if we're running in Bash on Windows we don't need a temp file.

Still, if we want to do it this way I'd short-circuit the whole JBang Main and just call into an alternative JdkExecMain or something. Because we know that large part of our slowness is Picocli for some reason. So avoiding that would most likely reduce the slowness to only JVM startup.

maxandersen:
that's fine. I just reused jdk exec --direct as it was already there and with this zero temp files are needed AND you can use --direct if speed of essence :)

quintesse:
> AND you can use --direct if speed of essence

This part I don't understand, I don't see how you, as a normal user, can use --direct "directly" to speed things up.

maxandersen:
Well with --direct it doesn't do the second call. 

So if you don't need stderr working it avoids the extra call.

quintesse:
Ok, but that's a hidden option users won't know about. So you'd have to make it non-hidden. And then try to explain what it does.

quintesse:
Btw, I just found out that .bat files can delete themselves. There's a bit of a trick to it to prevent spurious error messages, but it works: https://superuser.com/questions/1223388/can-i-make-a-batch-script-uninstaller-that-deletes-itself

maxandersen:
where are we with this one?

quintesse:
> where are we with this one?

I've not spent any time on this in the meantime

quintesse:
Ok, I think I got a simple solution that works. It depends on the fact that I was somewhat mistaken about how things work in Windows. My issues were that:

 1. we can't do something like `FOO=bar mycommand` like you can on Linux/Mac where the env var is temporary
 2. we'd need to return multiple lines to be simulate the same behaviour which we don't handle right now and would be difficult to do
 3. we'd be polluting the environment with vars that stick around after the command ends (see 1)

But it turns out that:

 1. still true, but
 2. we can do `set FOO=bar && mycommand` which is similar and still one line
 3. the env var sticks around after the command finishes yes, but... and this was the important realization, we're running from a `jbang.cmd` script that has `SETLOCAL` set which means that after the script finishes the environment is back to normal!

maxandersen:
Cool. And this works for .ps1 too?

quintesse:
> Cool. And this works for .ps1 too?

Hmm I thought it did, but it doesn't.

quintesse:
So I pushed a fix for PowerShell, which actually made me realize we pollute the environment in the ps1 script. And that's because PowerShell has no notion of `SETLOCAL` like .cmd scripts (and that is because PowerShell has its own system for local variables separate from environment variables while cmd/bat use the environment both for local as well as user/global variables.)

So I also updated the jbang.ps1 script to reset the environment variables to their previous values or delete them

src/main/scripts/jbang.ps1
```diff
@@ -173,3 +174,8 @@ if ($err -eq 255) {
 } else {
   break
 }
+
+$env:JAVA_HOME=$oldJavaHome
+$env:JBANG_RUNTIME_SHELL=$null
```
maxandersen: I know it's probably unlikely but if jbang calls jbang itself (which could happen nested) shouldn't these variables be restored ?

quintesse: But then the useful lifespan of these variables is already over. They are _really_ meant only for the process that gets started (which could be jbang)... 


hmmm looking at the jbang bash script we could probably change this:

```bash
export JBANG_RUNTIME_SHELL=bash
JBANG_STDIN_NOTTY=$([ -t 0 ] && echo "false" || echo "true")
export JBANG_STDIN_NOTTY
export JBANG_LAUNCH_CMD=$0
# shellcheck disable=SC2086
output=$(CLICOLOR_FORCE=1 "${JAVA_EXEC}" ${JBANG_JAVA_OPTIONS} -classpath "${jarPath}" dev.jbang.Main "$@")
```

to this:

```bash
NOTTY=$([ -t 0 ] && echo "false" || echo "true")
# shellcheck disable=SC2086
output=$(JBANG_RUNTIME_SHELL=bash JBANG_LAUNCH_CMD=$0 JBANG_STDIN_NOTTY=$NOTTY CLICOLOR_FORCE=1 "${JAVA_EXEC}" ${JBANG_JAVA_OPTIONS} -classpath "${jarPath}" dev.jbang.Main "$@")
```

Although in bash there's no way that the script can pollute the parent environment, so perhaps this is okay :-)

quintesse: What I understand so far about Powershell is that each shell has its own copy of the environment (that's why suggestions in forums mention that you could start a sub-powershell and have that one run your command to avoid polluting your local environment).

So the only case I can imagine where setting/unsetting environment variables could cause issues is if you'd run multiple jbang.ps1 scripts in parallel at the same time _in the same powershell instance_. You could possibly get a race condition where both scripts are setting the variables at the same time before they had a chance to run the java command. But that seems somewhat unrealistic.

quintesse: Btw, I wondered if, like *nix shells, Windows starts a different cmd/powershell for each script. And it turns out that for cmd it does, but for powershell it doesn't. A ps1 file gets run by the powershell itself it seems. (At least that's what I infer from watching the process monitor)

maxandersen: yes im not worrying about the unix scripts - just the powershell ...because either we need to set these to previous values or it does not matter to set them here..

quintesse: Well, it does matter because they leak out into your environment. So after running `jbang` in Powershell you suddenly have those extra variables in your environment that shouldn't be there. But sure, if you want I can restore them.

quintesse: Done

maxandersen (approved):

PR #2222, opened 2025-09-24
fix: update Java for command-line use to JDK 21

wfouche:
Java versions to use on the Command-line (shell) vs for Gradle builds are now JDK 21 vs JDK 11.

* build.gradle: add Adoptium vendor tag to Java 11 toolchain specification.
* .sdkmanrc: set Java version to 21.0.8-tem

image


maxandersen (approved):
lgtm

PR #2223, opened 2025-09-24
fix: parseJavaVersion simplified

wfouche:


quintesse (commented):
I must say I don't really see the simplification?
It's a couple of lines shorter, but it's also missing the null check.

For a real simplification you can probably do the work with a single regex, something like:

`(1\.)?(\d+).*`

Then you extract the second value and parse that into an int. (Still need to check for null in the beginning though)

wfouche:
> It's a couple of lines shorter, but it's also missing the null check.

The method has a `@NonNull` annotation. Is it simpler? Well the logic is improved and no exception handling and explicit null checking is needed anymore.

```java
@NonNull
public static int parseJavaVersion(String version) {
    String v = version;
    v = v.startsWith("1.") ? v.substring(2) : v;
    String n = v.replaceAll("(\\d+).*", "$1");
    return n.matches("\\d+") ? Integer.parseInt(n) : 0;
}
```

wfouche:
@quintesse , I had to close this PR because the commit message was wrong. Resubmitted as PR https://github.com/jbangdev/jbang/pull/2224

quintesse:
You know you can just change the commit message and re-push, right? :-)

wfouche:
> You know you can just change the commit message and re-push, right? :-)

OK, will do next time. :-)

PR #2224, opened 2025-09-24
fix: parseJavaVersion simplified

wfouche:
The logic is as follows:

* if version startsWith("1.") remove "1." from the start of the string.
* Java major version number is the int(numeric prefix of version).

```java
public static int parseJavaVersion(@NonNull String version) {
    String v = version;
    v = v.startsWith("1.") ? v.substring(2) : v;
    String n = v.replaceAll("(\\d+).*", "$1");
    return n.matches("\\d+") ? Integer.parseInt(n) : 0;
}
```

### DIFF
--- a/src/main/java/dev/jbang/util/JavaUtil.java
+++ b/src/main/java/dev/jbang/util/JavaUtil.java
@@ -126,21 +126,22 @@ public class JavaUtil {
 	// Match the first number in a Java version string, except if
 	// it starts with "1.", in that case, match the second number
 	public static int parseJavaVersion(String version) {
-		if (version == null)
-			return 0;
-		int len = version.length();
-		int idx = (len > 1 && version.charAt(0) == '1' && version.charAt(1) == '.') ? 2 : 0;
-		int n = 0;
-		for (int i = idx; i < len; i++) {
-			char c = version.charAt(i);
-			if (c >= '0' && c <= '9') {
-				n = n * 10 + (c - '0');
-			} else {
-				break;
-			}
-		}
-		return n;
-	}
+        if (version == null) {
+            return 0;
+        }
+        int len = version.length();
+        int idx = (len > 1 && version.charAt(0) == '1' && version.charAt(1) == '.') ? 2 : 0;
+        int n = 0;
+        for (int i = idx; i < len; i++) {
+            char c = version.charAt(i);
+            if (c >= '0' && c <= '9') {
+                n = n * 10 + (c - '0');
+            } else {
+                break;
+            }
+        }
+        return n;
+    }
 
 	public static boolean isOpenVersion(String version) {
 		return version.endsWith("+");


wfouche:
@quintesse , I think you will like this version better.  :-)

```java
	public static int parseJavaVersion(@NonNull String version) {
		int n = 0;
		int baseIdx = version.startsWith("1.") ? 2 : 0;
		for (int i = baseIdx; i < version.length(); i++) {
			char c = version.charAt(i);
			if (Character.isDigit(c)) {
				n = n * 10 + (c - '0');
			} else {
				break;
			}
		}
		return n;
	}
```

quintesse:
So I could imagine code like this:

```java
    // Match the first number in a Java version string, except if it starts with "1.",
    // in that case, match the second number.
    private static final Pattern JVERSION = Pattern.compile("^(1\\.)?(\\d+).*$");

	public static int parseJavaVersion(@Nullable String version) {
		if (version != null) {
            Matcher m = JVERSION.matcher(version.trim());
            if (m.matches()) {
                return Integer.parseInt(m.group(2));
            }
		}
		return 0;
	}
```

Which doesn't necessarily need a try-catch because the regex will guarantee we match only digits. Still, we are still vulnerable to overflow giving non-zero results. Although we could potentially fix that in the regex itself by limiting the number of digits we accept, eg:

```java
    private static final Pattern JVERSION = Pattern.compile("^(1\\.)?(\\d{1,3}).*$");
```

This would allow us to parse up to Java version 999 which will probably take a couple of years still :-)

Not 100% sure this is enough of an improvement though. But with the comment on the regex it's at least more understandable what's going on.

wfouche:
@quintesse , I've taken all your feedback into consideration:

```java
	// Match the first number in a Java version string, except if
	// it starts with "1.", in that case, match the second number
	private static final String JVERSION = "^(?:1\\.)?(\\d+).*";

	public static int parseJavaVersion(String version) {
		if (version == null)
			return 0;
		try {
			String num = version.replaceAll(JVERSION, "$1");
			return Integer.parseInt(num);
		} catch (NumberFormatException ignored) {
			return 0;
		}
	}
```

src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -123,13 +123,15 @@ public static String parseJavaOutput(String output) {
 		return null;
 	}
 
+	// Match the first number in a Java version string, except if
+	// it starts with "1.", in that case, match the second number
+	private static final String JVERSION = "^(?:1\\.)?(\\d+).*";
```
quintesse: This last commit is not really necessary IMO. The final static constant I introduced in my code was so the Pattern only needed to be compiled once. When using a String this advantage does not exist.

src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -123,17 +123,19 @@ public static String parseJavaOutput(String output) {
...
 	public static int parseJavaVersion(String version) {
-		if (version != null) {
-			try {
-				String[] nums = version.split("[-.+]");
-				String num = nums.length > 1 && nums[0].equals("1") ? nums[1] : nums[0];
-				return Integer.parseInt(num);
-			} catch (NumberFormatException ex) {
-				// Ignore
-			}
+		if (version == null)
```
quintesse: We don't use this syntax in the JBang code, we always use { } , even for single lines

maxandersen:
what issue/problem is this trying to fix?

src/main/java/dev/jbang/util/JavaUtil.java
quintesse: On the other hand, you could also use `Pattern` again and change the replace slightly like this:

```
String num = JVERSION.matcher(version.trim()).replaceAll("$1");
```

src/main/java/dev/jbang/util/JavaUtil.java
maxandersen: I tried looking at this, but I don’t really see the simplification here — to me it feels about equally hard/easy to follow. In general, I lean toward avoiding regular expressions when something more direct works just as well.

If this change is meant to handle a specific case, could you add a test to show that? Otherwise, if the goal was only simplification, I don’t think this achieves it.

maxandersen (changes requested):

wfouche:
I fixed a bug in a similar method in another project, and it sparked an interest in how JBang determines the JVM major version number.  It is hard to improve on the current implementation, especially if one considers raw performance (memory allocations and CPU cycles consimed).

This is a bit of an academic exercise, so no problem if this PR is declined :-)

I think the only real improvement over the current method,  is a C-like algorithm that do not allocate any memory and uses CPU cycles optimally. 

```java
	// Match the first number in a Java version string, except if
	// it starts with "1.", in that case, match the second number
	public static int parseJavaVersion(String version) {
        if (version == null) {
            return 0;
        }
        int len = version.length();
        int idx = (len > 1 && version.charAt(0) == '1' && version.charAt(1) == '.') ? 2 : 0;
        int n = 0;
        for (int i = idx; i < len; i++) {
            char c = version.charAt(i);
            if (c >= '0' && c <= '9') {
                n = n * 10 + (c - '0');
            } else {
                break;
            }
        }
        return n;
    }
```

This is an algorithm that GraalVM will compile to native code with great enthusiasm.

maxandersen:
if its shown this method is in a hotspot - sure - but for now i'll go for "if its not broken don't fix it" and decline this PR :)

quintesse:
I know the PR is already closed, but I like the discussion :-)

>  think the only real improvement over the current method

I mentioned this before, but I don't really like that this algorithm would happily parse "99999999999999999999999999999999999999999999999999999999" as if it's a valid number and return some random number (well, not random of course but I don't feel like figuring out what the actual number would be after having been around the overflow carousel a couple of times :-) )

Now is it realistic to expect that this would ever happen? No, of course not, but _if_ it happened it would probably cause some head scratching :-) "It can't find JDK 812? Where on earth does that number come from??"

wfouche:
> "99999999999999999999999999999999999999999999999999999999" as if it's a valid number 

Adding 

```java
if ( n > 999999) {
    return 0;
}
  ```
  
  resolves the overflow issue.
  
```java
//JAVA 25+

void main() {
    String version = System.getProperty("java.version");
    System.out.println("Java Version: " + parseJavaVersion("1.8.0_271"));
    System.out.println("Java Version: " + parseJavaVersion("17.0.271"));
    System.out.println("Java Version: " + parseJavaVersion("25-ea"));
    System.out.println("Java Version: " + parseJavaVersion("25.ea"));
    System.out.println("Java Version: " + parseJavaVersion("25"));
    System.out.println("Java Version: " + parseJavaVersion("aaa"));
    System.out.println("Java Version: " + parseJavaVersion("99999999999999999999999999999999999999999999999999999999"));
}

// Match the first number in a Java version string, except if
// it starts with "1.", in that case, match the second number
public static int parseJavaVersion(String version) {
    if (version == null) {
        return 0;
    }
    int len = version.length();
    int idx = (len > 1 && version.charAt(0) == '1' && version.charAt(1) == '.') ? 2 : 0;
    int n = 0;
    for (int i = idx; i < len; i++) {
        char c = version.charAt(i);
        if (c >= '0' && c <= '9') {
            n = n * 10 + (c - '0');
        } else {
            break;
        }
        if ( n > 999999) {
            return 0;
        }
    }
    return n;
}

public static int parseJavaVersion1(String version) {
    if (version != null) {
        try {
            String[] nums = version.split("[-.+]");
            String num = nums.length > 1 && nums[0].equals("1") ? nums[1] : nums[0];
            return Integer.parseInt(num);
        } catch (NumberFormatException ex) {
            // Ignore
        }
    }
    return 0;
}
```

wfouche:
What triggered me to submit a PR, was that the code looked so ugly with a null check and catching an exception for someting so simple.

To remove all the noise, one could have written the routine just as:

```java
	public static int parseJavaVersion(String version) {
        String[] nums = version.split("[-.+]");
        String num = nums.length > 1 && nums[0].equals("1") ? nums[1] : nums[0];
        return Integer.parseInt(num);
	}
```

quintesse:
In many of my personal projects I end up writing a helper function like:

```java
public static int parseInt(String num, int defaultValue) {
   try {
      return Integer.parseInt(num);
   } catch (NumberFormatException ex) {
      return defaultValue;
   }
}
```

Then you can often hide away some of the ugliness. But of course only in situations where there's an obvious "default value" that you can use.

And of course there's a reason there is a null check in the method... which is exactly because we don't want to repeat the same null check all over the place where the function gets called :-)

PR #2225, opened 2025-09-25
fix: fatjar conflicts of file vs directory

maxandersen:

PR #2228, opened 2025-09-25
fix: improve fatjar export

### LINKED ISSUE #2220: `fatjar`: Relaxed existance check on non-casesensitive filesystems
On Windows, `license` and `LICENSE` are the same file. Thus, `jpackage fatjar` does not work in some cases.

Example - on Windows

    jbang --verbose export fatjar https://github.com/JabRef/jabref/blob/main/.jbang/JabLsLauncher.java

```
[...]
[jbang] [25:798] Skipping duplicate file: META-INF/LICENSE
[...]
[jbang] [25:800] Unpacking artifact: xml-apis:xml-apis:jar:1.3.04=C:\Users\koppor\.m2\repository\xml-apis\xml-apis\1.3.04\xml-apis-1.3.04.jar
[...]
[jbang] [34:668] [ERROR] C:\Users\koppor\AppData\Local\Temp\fatjar4045727320793413749\license
java.nio.file.FileAlreadyExistsException: C:\Users\koppor\AppData\Local\Temp\fatjar4045727320793413749\license
```

quintesse:
We no longer throw an exception when we encounter both a file and a directory with the same name.

We now remove signature files from the fatjar to avoid errors at runtime.

Fixes #2220

### DIFF
--- a/src/main/java/dev/jbang/cli/Export.java
+++ b/src/main/java/dev/jbang/cli/Export.java
@@ -372,6 +372,17 @@ class ExportFatjar extends BaseExportCommand {
 					Util.verboseMsg("Unpacking artifact: " + dep);
 					UnpackUtil.unzip(dep.getFile(), tmpDir, false, null, ExportFatjar::handleExistingFile);
 				}
+				// Remove all signature files
+				Path metaInf = tmpDir.resolve("META-INF");
+				if (Files.exists(metaInf) && Files.isDirectory(metaInf)) {
+					try (DirectoryStream<Path> stream = Files.newDirectoryStream(metaInf, "*.{SF,DSA,RSA}")) {
+						for (Path entry : stream) {
+							Util.verboseMsg("Removing signature file: " + entry);
+							Files.delete(entry);
+						}
+					}
+				}
+				Util.verboseMsg("Creating jar: " + outputPath);
 				JarUtil.createJar(outputPath, tmpDir, null, prj.getMainClass(),
 						exportMixin.buildMixin.getProjectJdk(prj));
 			} finally {
--- a/src/test/java/dev/jbang/cli/TestExport.java
+++ b/src/test/java/dev/jbang/cli/TestExport.java
@@ -578,4 +578,23 @@ public class TestExport extends BaseTest {
 		assertThat(result.result, equalTo(BaseCommand.EXIT_OK));
 		assertThat(result.err, containsString("[WARN] Skipping conflicting duplicate file vs directory:"));
 	}
+
+	@Test
+	void testExportFatjarSignatures(@TempDir Path temp) throws Exception {
+		Util.writeString(temp.resolve("SIG"), "A dummy signature file");
+		String code1 = "" +
+				"//FILES META-INF/DUMMY.SF=SIG\n" +
+				"//FILES META-INF/DUMMY.DSA=SIG\n" +
+				"//FILES META-INF/DUMMY.RSA=SIG\n";
+		Util.writeString(temp.resolve("src.java"), code1);
+		String code2 = "//DEPS src.java\n";
+		Path src = temp.resolve("test.java");
+		Util.writeString(src, code2);
+		CaptureResult<Integer> result = checkedRun(null, "--verbose", "export", "fatjar", src.toString());
+		assertThat(result.result, equalTo(BaseCommand.EXIT_OK));
+		assertThat(result.err, containsString("Removing signature file:"));
+		assertThat(result.err, containsString("DUMMY.SF"));
+		assertThat(result.err, containsString("DUMMY.DSA"));
+		assertThat(result.err, containsString("DUMMY.RSA"));
+	}
 }


src/main/java/dev/jbang/cli/Export.java
```diff
@@ -372,6 +372,17 @@ int apply(BuildContext ctx) throws IOException {
 					Util.verboseMsg("Unpacking artifact: " + dep);
 					UnpackUtil.unzip(dep.getFile(), tmpDir, false, null, ExportFatjar::handleExistingFile);
 				}
+				// Remove all signature files
+				Path metaInf = tmpDir.resolve("META-INF");
+				if (Files.exists(metaInf) && Files.isDirectory(metaInf)) {
+					try (DirectoryStream<Path> stream = Files.newDirectoryStream(metaInf, "*.{SF,DSA,RSA}")) {
```
maxandersen, line 378: got example where this errors if not here?

src/test/java/dev/jbang/cli/TestExport.java
```diff
@@ -547,4 +548,49 @@ void testExportMavenProjectWithTags() throws Exception {
...
+		assertThat(result.result, equalTo(BaseCommand.EXIT_OK));
+		assertThat(result.err, containsString("[WARN] Skipping conflicting duplicate file vs directory:"));
+	}
+
+	@Test
+	void testExportFatjarFileDirConflict2(@TempDir Path temp) throws Exception {
+		Util.writeString(temp.resolve("LICENSE"), "A dummy license file");
+		Util.writeString(temp.resolve("src1.java"), "//FILES LICENSE");
+		Util.writeString(temp.resolve("src2.java"), "//FILES LICENSE/LICENSE=LICENSE");
+		String code = "//DEPS src2.java src1.java\n";
```
maxandersen:
```suggestion
		Util.writeString(temp.resolve("src2.java"), "//FILES LICENSE/LICENSE=LICENSE");
		Util.writeString(temp.resolve("src3.java"), "//FILES LICENSE/LICENSE/NESTED=LICENSE");
		Util.writeString(temp.resolve("src4.java"), "//FILES LICENSE/LICENSE/NESTED/NESTED=LICENSE");
		String code = "//DEPS src2.java src1.java src3.java src4.java\n";
```

checking it wont fail on nested dirs further down than broken parent.

src/main/java/dev/jbang/util/UnpackUtil.java
```diff
@@ -86,21 +87,25 @@ public static void unzip(Path zip, Path outputDir, boolean stripRootFolder, Path
...
+					if (zipEntry.isDirectory()) {
+						Files.createDirectories(entry);
+					} else if (zipEntry.isUnixSymlink()) {
+						Scanner s = new Scanner(zipFile.getInputStream(zipEntry)).useDelimiter("\\A");
+						String result = s.hasNext() ? s.next() : "";
+						Files.createSymbolicLink(entry, Paths.get(result));
 					} else {
-						defaultZipEntryCopy(zipFile, zipEntry, entry);
+						if (!Files.isDirectory(entry.getParent())) {
+							Files.createDirectories(entry.getParent());
```
maxandersen, line 99: wont this fail the parent exist but is a file?

maxandersen (commented):
yours is better than mine at #2225 because yours has test.

overall looks good but wondering if it still wont fail for multi level nested dirs.

ie..
license (gets created)
LICENSE/somefile.txt (fails create as parent overlaps, but is handled so just rint warning)
LICENSE/otherdir/otherfile.txt (this could fail as we just createDirecotries in case of parent is not a dir ...but that dir can't be created as its parent-parent is a file)

src/main/java/dev/jbang/cli/Export.java
quintesse: wdym? There's a test that checks that the files are properly removed, but there is not a test checking for the actual error. That would mean exporting a project with dependencies that have these files and then trying to run it. That seemed more complicated than the fix itself :-)

src/main/java/dev/jbang/util/UnpackUtil.java
quintesse: > wont this fail the parent exist but is a file?
> ... wondering if it still wont fail for multi level nested dirs.
> ... but that dir can't be created as its parent-parent is a file

That's why the PR lets the error occur and catches the `FileAlreadyExistsException` instead of trying to test preventively all possible situations where this could happen.

src/main/java/dev/jbang/cli/Export.java
maxandersen: i get that the test doesn't need to be real and can just check for nonexistence of the files.

I was more wondering if you bumped into example in real world which this removal fixes things.

src/main/java/dev/jbang/util/UnpackUtil.java
maxandersen: as you can see with the failure that is not what happens in this case :)

quintesse: Indeed, turns out that some errors are platform dependent, so on WIndows in this case it's actually the Win32 API that returns the file not found error.

It's so weird, the Java code actually checks (in a loop going up the hierarchy) if there is a parent that has the proper access rights, etc... but then they forget to check if it's actually a directory before handing things over to the Win32 API. So now I'm doing the same thing, but _with_ the proper check. It's a bit unfortunate, but I think it keeps the code simple enough.

src/main/java/dev/jbang/cli/Export.java
quintesse: Yes, the original #2220 that started all this. Once I fixed the file/directory conflict issue the next error was about the JAR not having a valid signature. And then the next error was about the Main Class... but that seems to have gone away? I can't reproduce it anymore (not sure if that is 🎉 or 🫤)

PR #2233, opened 2025-09-27
fix: better logging for jdk providers

maxandersen:
this provides some basic logging/help when using --jdk-providers.

i.e. --jdk-providers=donotexist silently use default but --jdk-providers=mise when no mise dirs (yet) fails...so needed some help figuring out what was doing wrong.

+ prints available providers so i dont have to go look in the code :)

### DIFF
--- a/src/main/java/dev/jbang/util/JavaUtil.java
+++ b/src/main/java/dev/jbang/util/JavaUtil.java
@@ -47,6 +47,8 @@ public class JavaUtil {
 				for (String providerName : names) {
 					if (PROVIDERS_ALL.contains(providerName)) {
 						providerNames.add(providerName);
+					} else {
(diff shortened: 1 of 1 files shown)


quintesse (approved):
In itself this PR is okay. The only comment I'd make is that you print warning both in the `provider()` method as well as the `build()` method. Doesn't this (possibly) show the warning twice? I mean, the `build()` method _must_ be called anyway.

maxandersen:
The current code just do one as it fails earlier. So in theory the second should never happen.

quintesse:
> The current code just do one as it fails earlier. So in theory the second should never happen.

I see. Because they already get filtered out in the `provider()` method so they never reach `build()`. Which means it's probably the warning in `build()` that's unnecessary because the only way to add providers is using the `providers()` methods. But you can leave it if you like.

PR #2237, opened 2025-10-01
docs: Fix JVM proxy args

xfl12345:
The syntax for JVM parameters only requires a single "-D".


quintesse:
You're absolutely right. Thanks!

PR #2245, opened 2025-10-10
feat: make it possible to run jitpack URLs directly

quintesse:
We can now run `jbang https://github.com/OWNER/REPO` directly by assuming we can use JitPack to build and run the repo.

### DIFF
--- a/src/main/java/dev/jbang/dependencies/JitPackUtil.java
+++ b/src/main/java/dev/jbang/dependencies/JitPackUtil.java
@@ -25,6 +25,10 @@ public class JitPackUtil {
 
 	private static final Pattern POSSIBLE_SHA1_PATTERN = Pattern.compile("^[0-9a-f]{40}$");
 
+	public static boolean possibleMatch(String ref) {
+		return !ensureGAV(ref).equals(ref);
+	}
+
 	public static String ensureGAV(String ref) {
 		try {
 			// If the reference is a URL we'll try to convert it to a proper GAV
--- a/src/main/java/dev/jbang/resources/resolvers/GavResourceResolver.java
+++ b/src/main/java/dev/jbang/resources/resolvers/GavResourceResolver.java
@@ -5,6 +5,7 @@ import java.util.function.Function;
 import org.jspecify.annotations.NonNull;
 
 import dev.jbang.dependencies.DependencyUtil;
+import dev.jbang.dependencies.JitPackUtil;
 import dev.jbang.dependencies.ModularClassPath;
 import dev.jbang.resources.ResourceRef;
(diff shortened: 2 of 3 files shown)


maxandersen:
Cool ..but do we need to rename the wildly used util in the same pr? :)

quintesse:
Because I got fed up with all that code being in Util. If I would have split it out then I'd have to make two PRs, with one waiting for the other :-)

And it's not a rename, it's extracting all the download code into its own file. The change is really quite simple.

quintesse:
I created the PR to be able to run this: https://github.com/Delawen/leyden-analyzer :-)

quintesse:
> the netutil just overshadows the pr ;)

Well, it's only 2 commits, one for the refactor and one for the feature. If you look at the separately it's pretty straightforward I think.

maxandersen:
cool - so I noticed that when I do ` jbang https://github.com/Delawen/leyden-analyzer` it fails, but I have to do ` jbang --repos jitpack https://github.com/Delawen/leyden-analyzer` to get it to work.

should do something like https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/dependencies/DependencyUtil.java#L78 where it will automagically add jitpack if not already present.

maxandersen:
also, once it waited on build I get this:

```
[jbang]    com.github.Delawen:leyden-analyzer:main-SNAPSHOT
[jbang] Dependencies resolved
[jbang] [ERROR] no main class deduced, specified nor found in a manifest
```

but i assume thats issue of leyden-analyzer not having main class specified in the jar.

quintesse:
> but i assume thats issue of leyden-analyzer not having main class specified in the jar.

Ah yes, It's because it's a Quarkus project so the standard JAR doesn't have a main because it's not runnable and with just the URL you can't tell it to use the ":runner" classifier. So the URL isn't the best project to test with.

quintesse:
> where it will automagically add jitpack if not already present.

Yeah, that's why it's still in draft, the place where it detects that we could treat the URL as a gav is deep down in the resolver and that's not a place where you can go add repos. So I might need to pull this code up and put it somewhere higher in the call stack.

maxandersen:
makes sense; but just be sure it doesn't get called before repos configured as we shouldn't override users possible order of repos.

quintesse:
The only repos could come from the --repos flag, right? So either the user already added it there or we have to add it, and once we add it we can just add it to the end, right?

quintesse:
After several tries, it turned out the change was actually a one-liner

maxandersen:
> After several tries, it turned out the change was actually a one-liner

so ...16 files changed for netutil rename 5 lines for the actual feature in the title - got it :)

PR #2248, opened 2025-10-10
feat: let build.jbang auto-select sources

quintesse:
When no sources are specified, we now look for common source locations.

### DIFF
--- a/src/main/java/dev/jbang/source/ProjectBuilder.java
+++ b/src/main/java/dev/jbang/source/ProjectBuilder.java
@@ -1,5 +1,7 @@
 package dev.jbang.source;
 
+import static dev.jbang.util.Util.MAIN_JAVA;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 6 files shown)


quintesse:
The code perhaps looks somewhat more complicated than necessary for this relatively minor feature, but that's because I've made a first setup of adding the ability to list children to ResourceRefs. For now this only works for file resources, but the idea is that we can implement this for remote resources as well (either limited to things like GitHub and GitLab where we can use their APIs or Git or perhaps by using WebDav-like techniques, we'll see)

The idea is to enable the use of `//SOURCES src/*.java` in remote scripts (but of course we need to be smart about it, we don't want to create a download request for each file in a folder with hundreds of files if someone tries to do this for a large project)

quintesse:
Merging because it's only used by a preview feature so shouldn't affect anything outside that.

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -366,7 +368,24 @@ private Project createJbangProject(ResourceRef resourceRef) {
...
 
 		boolean first = true;
-		List<Source> includedSources = allToSource(directives.sources(), resourceRef, sibRes2);
+		List<String> sources = directives.sources();
+		if (sources.isEmpty()) {
+			// if no sources are defined, we go look for a couple of possible options
+			if (resourceRef.resolve(MAIN_JAVA) != null) {
+				sources.add(MAIN_JAVA);
+			} else {
+				if (resourceRef.resolve("src/main/java") != null) {
```
maxandersen, line 377: Is it intentionally it will add all languages in bulk and not just one?

quintesse: Kotlin allows you to mix .java and .kt files, I don't know if Groovy allows the same, but I wouldn't be surprised. So I thought I'd just do it like this. It'll probably be a very uncommon situation anyway.

Issue #2243, opened 2025-10-09
find free debug port


maxandersen:
default java behaviour is that the debug port is fixed to what `address` property is set on the debug agent.

works great unless when you have a java app (like a jupyter kernel) that might get started/restarted multiple times and if debug turned on for the same port you get a conflict and app does not start.

It would be nice if one could do `--debug=address=5555?` and the `?` would be indicating to jbang to check if the port is free - if yes, start, if no - find free port and print out to err about chosen port and then start with that instead.

quintesse:
That seems like a _very_ specific feature and somewhat weird? Either you want it to be that port and things will probably fail anyway because in the end you didn't get the port you expected, or you actually look at the port number being printed... and therefore why would you care if it's 5555 or any other port?

maxandersen:
> That seems like a _very_ specific feature and somewhat weird? 

Its niche but not weird. It lets you run your app with debug and have it not fail just because the debug port was not available.

>Either you want it to be that port and things will probably fail anyway because in the end you didn't get the port you expected, 

No, i.e. when developing on jupyter notebook its jupyter that launches the kernel and can do so *multiple* times and I do not want it to fail just because debugger can not listen, so i use `--debug=5000?` if I want it to fail (which is a valid usecase also) I just keep using `--debug=5000`

> or you actually look at the port number being printed... and therefore why would you care if it's 5555 or any other port?

I can't see the port being printed when launched by some other app - so i rely on attach to process feature of IDE's - and they will list which port. Thus easier if its 5000 if it just worked - and only have to deal with finding right one if it is NOT 5000. 

Keeps it deterministic and simple in default case and only "complicated" when something causes conflict (i.e. multiple launches of kernel by jupyter)

PR #2250, opened 2025-10-11
dynamicdebug

maxandersen:
implements #2243 

`--debug=5000?` starts debug on 5000 so it will be consistent if no conflicts; and search only if conflict on the port.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -130,7 +130,7 @@ sourceSets {
 sourceSets.main.compileClasspath += sourceSets.java9.output.classesDirs;
 
 dependencies {
-	implementation 'dev.jbang:devkitman:0.2.0'
+	implementation 'dev.jbang:devkitman:0.3.0'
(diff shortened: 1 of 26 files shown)


maxandersen:
btw. this amaze balls to use so would love a review to spot if see issues with it.

I'm not too worried as it only gets activated if "?" present at the end of debug string.

I do think the current app install handling of debug string is not safe/consistent but that is an existing bug so leaving that for separate pr.

maxandersen:
@quintesse works/ok for you?

quintesse:
Sorry, haven't had time to look at it yet.

maxandersen:
I need this to stay sane debugging jbang kernel - if test passes merging in as it really shouldn't affect anyone unless they put a ? in the debug string...which is illegal normally ...

maxandersen:
merge in as no impact on other parts.

PR #2254, opened 2025-10-12
feat: print commands

maxandersen:
veles has a `-N` option that enables printing of commands - we have that with --verbose too but could be nice being able to just get the commands and not everything else.

this pr just adds the needed wrappers - but does not handle enabling it (its just always on)

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -953,7 +953,7 @@ public class Util {
 	}
 
 	public static Process run(ProcessBuilder pb) throws IOException {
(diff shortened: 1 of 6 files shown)


src/main/java/dev/jbang/util/Util.java
```diff
@@ -952,6 +952,11 @@ static class History {
 		String version;
 	}
 
+	public static Process run(ProcessBuilder pb) throws IOException {
```
maxandersen, line 955: should we add a "reason" argument to prefix with so its not just a raw print of commands?

Issue #2187, opened 2025-08-19
jbang search or jbang deps ?


maxandersen:
would be interesting to combine gavsearch@jbangdev and jpm search feature to allow easy lookup and edit of deps..

ideas:

`jbang dep add --deps org.hibernate:hibernate-core:4.7.8 myapp.java` would directly add that dependency to `myapp.java`
`jbang dep add --deps hibernate myapp.java` would search for '*hibernate*' artifacts (as 'hibernate' in it self in maven central and let you pick/choose artifacts to then get added/updated in myapp.java

even imagine `jbang dep add --deps hibernate,log4j myapp.java` would be two searches...one for hibernate and log4j would "constrain"/guide the search for log4j....could even parse/read myapp.java to know what is referring to.

maxandersen:
if anything this feels like a prime candidate for being a jbang plugin and not default included. wdyt @quintesse ?

maxandersen:
btw. remember jbang run has defaults it searches for when no arguments(i.e. main.java, jbang.build..) jbang dep/search could have the same.

maxandersen:
`jbang deps update` could incorporate #2148 deps@jbangdev

quintesse:
Yeah, definitely think this could be a plugin. The _only_ thing I can think if that could cause a bit of duplication of code is the part where you say (correctly): "remember jbang run has defaults it searches for when no arguments(i.e. main.java, jbang.build..) "
Because it means that lookup of default files would have to be duplicated in the plugin and might at some point get out of sync with jbang itself. Not sure if that's something we really have to worry about but I thought I'd at least mention it.

maxandersen:
These are reasons to extract those core logic as library. 

that said it also feels like something like deps mgmt is a core feature so let's see how it looks like :)

quintesse:
> These are reasons to extract those core logic as library.

Definitely

PR #2260, opened 2025-10-22
feat: jbang deps add <deps> <file>

maxandersen:
this is a POC for now of #2187 with intent of adding `jbang deps` subcommand that eventually lets users manage and search for dependencies.

- [ ] `jbang deps add g:a:v g:a2:v myapp.java`

```
jbang deps g:a:v g:a2:v g:a myapp.java
```

this is not cleaned up yet but thought would get feedback on its current behavior.

- does not validate existence of dependency (as it would make it slow but maybe we should at least have option for it? or just do a build after add?)
- allows adding g:a (without version) in case you have a bom
- adds after last found  //DEPS directive or last other directive and otherwise top of file
- supports .java and build.jbang
- does NOT support updating deps in included files; if you have that you wanna maintain it for your self anyway :) 

want to implement a `jbang deps search <depsquery> myapp.java` which when you found a match will delegate this to jbang deps add.

- [ ] `jbang deps search [--query query] [scriptref]`

variation/fork/copy of java-jpm search but using java 8+ apis.
still rough but works better than jbang gavsearch@jbangdev and similar to java-jpm but supports searching based on fullclass/classname queries (i.e. `fc:dev.jbang.Main` or `c:StatelessSession`

### DIFF
--- a/.github/actions/shared-build-setup/action.yml
+++ b/.github/actions/shared-build-setup/action.yml
@@ -28,12 +28,12 @@ runs:
         java-version: ${{ inputs.java-version }}
         distribution: temurin
     - name: setup-gradle
-      uses: gradle/actions/setup-gradle@06832c7b30a0129d7fb559bcc6e43d26f6374244 # v4
+      uses: gradle/actions/setup-gradle@748248ddd2a24f49513d8f472f81c3a07d4d50e1 # v4
       with:
           cache-overwrite-existing: true
           add-job-summary: ${{ inputs.add-job-summary }}
     - name: Setup tmate session
-      uses: mxschmitt/action-tmate@7b6a61a73bbb9793cb80ad69b8dd8ac19261834c # v3
+      uses: mxschmitt/action-tmate@c0afd6f790e3a5564914980036ebf83216678101 # v3
       if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled }}
       with:
         limit-access-to-actor: true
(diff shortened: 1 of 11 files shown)


maxandersen:
btw. for now the command require specific scriptref but should update this and other jbang to locate "nearest" scriptref from current dir.

maxandersen:
current progress: https://asciinema.org/a/Kp0r4AAg50KUkwoqvmpUkpzLl

maxandersen:
this is working decently imo, we can iterate and improve but better to get it out.

src/main/java/dev/jbang/cli/BaseCommand.java
```diff
@@ -30,7 +30,7 @@ public abstract class BaseCommand implements Callable<Integer> {
 	public static final int EXIT_INTERNAL_ERROR = 4;
 	public static final int EXIT_EXECUTE = 255;
 
-	private static final Logger logger = Logger.getLogger("org.jboss.shrinkwrap.resolver");
```
quintesse, line 33: Why this change? That logger was specifically there to shut up all the messages from the resolver. Or is that now not necessary anymore because of other changes you made? (The result should be that if you now wun with `--verbose` you could be getting a flood of resolver messages drowning out everything else)

maxandersen: We haven't used shrink-wrap for ages so shouldn't be needed anymore afaics?

quintesse: Aah, you're right, I'm probably confused with some other logger. Sorry

PR #2262, opened 2025-10-25
feat: POC of jbang native image compiled

maxandersen:
this is a quick'n'dirty native image build of jbang that is not using quarkus nor graalvm tasks. probably missing out on some optimizations because of that but this was the change with least needed changes elsewhere.

only tested on osx for now.

enabling nativeimage build in github action but i dont think we should have it enabled upfront as it will take too long for default builds.

only tested basic stuff so edge cases probably not hit yet - should get jbang script to pick up the native binary and run itest before releasing it :)

Things missing:

 - [x] update script to use the native binary when present and JBANG_USE_NATIVE=true
 - [ ] update scripts own dependency on jbang (jvm selection is done with jbang-cli)
 - [ ] figure out what is going on with non-latin characters on windows
 - [ ] our boot/init of jbang is reliant on scripts as we can't know if a binary will work on the platform we are running on
 - [ ] download of the binary by the scripts 
 - [ ] how updates will work (should update both jbang.jar and jbang-cli if available?)
 - [ ] should we consider supporting jbang-cli gets called directly? it would enable using jbang command lines in places like cursor mcp and other command lines that does not support invoking .ps1, .java files. i.e. if JBANG_* vars not set we do the exec method ourselves?

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -99,7 +99,7 @@ jobs:
       - name: upload-native-image
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
-          name: build-native-image-jbang-cli-${{ matrix.os }}
+          name: build-native-image-jbang.bin-${{ matrix.os }}
           path: build/native-image/*
           if-no-files-found: error
 
@@ -194,12 +194,12 @@ jobs:
           name: build-shared-build-${{ github.sha }}
           path: build
 
-      - name: download jbang-cli 
+      - name: download jbang.bin 
         uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
         with:
-          name: build-native-image-jbang-cli-${{ matrix.os }}
+          name: build-native-image-jbang.bin-${{ matrix.os }}
           path: build/native-image
-      - name: create install with jbang-cli
+      - name: create install with jbang.bin
         run: |
           ./gradlew --no-daemon installDist
       - name: integration-test-native-image
--- a/build.gradle
+++ b/build.gradle
@@ -78,7 +78,7 @@ publishing {
 	publications {
 		maven(MavenPublication) {
 			groupId = 'dev.jbang'
-			artifactId = 'jbang-cli'
+			artifactId = 'jbang.bin'
 
 			from components.java
 
@@ -243,9 +243,9 @@ def commonSpec = project.copySpec {
 		into 'bin'
 	}
 	// Conditionally include native image binary if it exists
-	if (project.file("${project.buildDir}/native-image/jbang-cli").exists()) {
+	if (project.file("${project.buildDir}/native-image/jbang.bin").exists()) {
 		from(project.buildDir.toPath().resolve('native-image')) {
-			include 'jbang-cli'
+			include 'jbang.bin'
 			into 'bin'
 		}
 	}
@@ -558,7 +558,7 @@ task nativeImage(type: Exec) {
 	group = "build"
 
 	def jarPath = shadowJar.archiveFile.get().asFile.path
-	def nativeExecName = "jbang-cli"
+	def nativeExecName = "jbang.bin"
 	def nativeImageName = "native-image"
 	def outputDir = layout.buildDirectory.dir("native-image").get().asFile
 	if (System.getProperty("os.name").toLowerCase().contains("windows")) {
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -237,7 +237,7 @@ class AppInstall extends BaseBuildCommand {
 			} else {
 				Path jar = Util.getJarLocation();
 				// TODO: this is duplicated in Wrapper.java - should be more shared.
-				if (!jar.toString().endsWith(".jar") && !jar.toString().endsWith("jbang-cli")) {
+				if (!jar.toString().endsWith(".jar") && !jar.toString().endsWith("jbang.bin")) {
 					throw new ExitException(EXIT_GENERIC_ERROR, "Could not determine jbang location from " + jar);
 				}
 				Path fromDir = jar.getParent();
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -216,18 +216,18 @@ if [[ -z "$JBANG_DIR" ]]; then JBDIR="$HOME/.jbang"; else JBDIR="$JBANG_DIR"; fi
 if [[ -z "$JBANG_CACHE_DIR" ]]; then TDIR="$JBDIR/cache"; else TDIR="$JBANG_CACHE_DIR"; fi
 if [[ -z "$JBANG_USE_NATIVE" ]]; then JBANG_USE_NATIVE="false"; fi
 
-## resolve jbang-cli binary or jar path from script location
+## resolve jbang.bin binary or jar path from script location
 binaryPath=""
 jarPath=""
 if [[ "$JBANG_USE_NATIVE" == "true" ]]; then
   # Look for native binary first if enabled
   if [[ "$os" == "windows" ]]; then
-    if [ -f "$abs_jbang_dir/jbang-cli.exe" ] && [ -x "$abs_jbang_dir/jbang-cli.exe" ]; then
-      binaryPath="$abs_jbang_dir/jbang-cli.exe"
+    if [ -f "$abs_jbang_dir/jbang.bin.exe" ] && [ -x "$abs_jbang_dir/jbang.bin.exe" ]; then
+      binaryPath="$abs_jbang_dir/jbang.bin.exe"
     fi
   else
-    if [ -f "$abs_jbang_dir/jbang-cli" ] && [ -x "$abs_jbang_dir/jbang-cli" ]; then
-      binaryPath="$abs_jbang_dir/jbang-cli"
+    if [ -f "$abs_jbang_dir/jbang.bin" ] && [ -x "$abs_jbang_dir/jbang.bin" ]; then
+      binaryPath="$abs_jbang_dir/jbang.bin"
     fi
   fi
 fi
--- a/misc/jbang_release_checker.java
+++ b/misc/jbang_release_checker.java
@@ -117,7 +117,7 @@ public class jbang_release_checker implements Callable<Integer> {
             var futures = List.of(
                 executor.submit(() -> checkGitHubReleases("jbangdev/jbang", targetVersion)),
                 executor.submit(() -> checkSDKMAN("jbang",targetVersion)),
-                executor.submit(() -> checkMavenCentral("dev.jbang", "jbang-cli", targetVersion)),
+                executor.submit(() -> checkMavenCentral("dev.jbang", "jbang.bin", targetVersion)),
                 executor.submit(() -> checkDockerHub("jbangdev", "jbang-action", targetVersion)),
                 executor.submit(() -> checkGHCR("jbangdev", "jbang-action", targetVersion)),
                 executor.submit(() -> checkHomebrew("jbang", targetVersion)),
--- a/src/main/java/dev/jbang/cli/Wrapper.java
+++ b/src/main/java/dev/jbang/cli/Wrapper.java
@@ -38,7 +38,7 @@ public class Wrapper {
 		}
 		try {
 			Path jar = Util.getJarLocation();
-			if (!jar.toString().endsWith(".jar") && !jar.toString().endsWith("jbang-cli")) {
+			if (!jar.toString().endsWith(".jar") && !jar.toString().endsWith("jbang.bin")) {
 				throw new ExitException(EXIT_GENERIC_ERROR, "Couldn't find JBang install location via " + jar);
 			}
 			Path parent = jar.getParent();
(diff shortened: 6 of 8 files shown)


quintesse:
Surprisingly few changes! 👍

maxandersen:
> Surprisingly few changes! 👍

also: haven't run any IT test beyond jbang init -t cli hello.java ;)

.github/workflows/ci-build.yml
```diff
@@ -65,6 +65,13 @@ jobs:
           java-version: 11
           distribution: temurin
           cache: gradle
+      - name: setup-graalvm
```
maxandersen, line 84: this shouldn't really be here - but just done to test having native image setup.

should probably move build of native-image to a separate step which comes after all the default building and testing.

src/it/java/dev/jbang/it/InitIT.java
```diff
@@ -30,7 +30,11 @@ public void testInitAndRun() throws IOException {
 			.succeeded();
 
 		String contents = new String(Files.readAllBytes(testFile));
-		assertThat(contents).contains("class test");
+
+		assertThat(contents).satisfiesAnyOf(
```
maxandersen, line 34: this is a consequence of default testing runs on java <25 so output would in those cases be with a class - but if you run the tests >=25 which is the case when native-image is built with java 25 then output will default to java 25...*even* if you dont have java 25 installed... not sure what we can do here... we technically would need to detect the java version of what java you have installed in case its not the same as the one one we are running with.

@quintesse ideas?

src/main/java/dev/jbang/cli/App.java
```diff
@@ -236,8 +236,9 @@ public static boolean installJBang(boolean force) throws IOException {
 				});
 			} else {
 				Path jar = Util.getJarLocation();
-				if (!jar.toString().endsWith(".jar")) {
-					throw new ExitException(EXIT_GENERIC_ERROR, "Could not determine jbang location");
+				// TODO: this is duplicated in Wrapper.java - should be more shared.
+				if (!jar.toString().endsWith(".jar") && !jar.toString().endsWith("jbang-cli")) {
```
maxandersen: with jbang-cli the binary used can have two different names.

I don't really grok why this test that is also in Wrapper.java cars it is a .jar and not jbang.jar specifically if anything...feels a bit wrong but for now just kept the same logic so its equally broken/working before/after native-image.

src/main/java/dev/jbang/resources/resolvers/ClasspathResourceResolver.java
```diff
@@ -47,12 +47,13 @@ protected ClasspathResourceRef(@NonNull String ref, ResourceResolver resolver) {
 		private static InputStream createStream(String resource) {
 			URL url = getResourceUrl(resource);
 			if (url == null) {
-				throw new ResourceNotFoundException(resource, "Resource not found on class path");
+				throw new ResourceNotFoundException(resource, "Resource " + resource + " not found on class path");
```
maxandersen, line 50: just added as without it was impossible/hard to figure out why resource lookup failed - with this we can know what is missing in the native-image config to be marked for inclusion.

src/main/scripts/jbang
```diff
@@ -54,6 +54,94 @@ javacInPath() {
   [[ -x "$(command -v javac)" ]] && ( [[ $os != "mac" ]] || /usr/libexec/java_home &> /dev/null )
```
maxandersen, line 54: these scripts was done by help of genai so we need to double check them before merging this in. potentially we could leave this change out for now but then manually wire it up.

general prompt was to add ability to use jbang-cli if available and if JBANG_USE_NATIVE is set to true, move code to reusable methods but otherwise keep changes minimal.

@quintesse ^

build.gradle
```diff
@@ -175,6 +176,8 @@ buildConfig {
 
 // to enable parallel test execution
 tasks.withType(Test).configureEach {
+	//environment "JAVA_HOME", "/Users/manderse/.local/share/mise/installs/java/graalvm-community-25.0.1"
```
maxandersen, line 179: these doesn't do anything atm but leaving in as reminder on how to tell gradle to run with specific java version for test.

needed as to run test with graalvm/java 25 when gradle itself does not support java 25 for running it.

build.gradle
```diff
@@ -239,6 +242,13 @@ def commonSpec = project.copySpec {
 		include 'jbang.ps1'
 		into 'bin'
 	}
+	// Conditionally include native image binary if it exists
+	if (project.file("${project.buildDir}/native-image/jbang-cli").exists()) {
```
maxandersen: im really not sure about this - but it makes it so that the dist folder will have jbang-cli inside it if and only if a jbang-cli is present...but that could be out of sync so if you want to be sure run a clean first.

src/it/java/dev/jbang/it/InitIT.java
quintesse: Well, why are we defaulting to 25 when using native? I'm not sure where that comes from, but that doesn't seem correct. There is probably code that assumes we're running with Java and that it can use the current java version to run stuff. We should use the same code that we normally use to detect what the "default" Java version is to use for running code. And if it's already using that code then we need to fix that code to detect it's running natively :-)

quintesse: Btw, the "proper" way to get the default JDK that should be used is to ask devkitman, using `JdkManager().getOrInstallJdk()`. Which, if you have a `Project` instance, can be done by calling `project.projectJdk()` (this will automatically take care of selecting the correct version if code has //JAVA or user specified --java). So if that code returns 25, then the issue is with devkitman not understanding native.

src/main/java/dev/jbang/cli/App.java
quintesse: I'm not 100% sure, but it might be because if you're running tests or if you somehow (?) run jbang using the published version in Maven the name isn't `jbang.jar` but `jbang-x.y.z.jar`? Edit: Or if you simply need to run a test and do `java -jar build/jbang-x.y.z.jar` I guess

src/it/java/dev/jbang/it/InitIT.java
maxandersen: i believe it comes from that if you haven't requested a specific java version you run with the same as jbang is using...but that is obviously not the case for native image ;)

maxandersen: the code is this in Iniit.java:

>int reqVersion = buildMixin.javaVersion != null ? JavaUtil.minRequestedVersion(buildMixin.javaVersion)
				: JavaUtil.getCurrentMajorJavaVersion();
				
there is no current project as its about to create it.

anything better?

src/main/java/dev/jbang/cli/App.java
maxandersen: yeah - I reckon that could be it. will keep as is and if we find better/different way we can fix that; for now it works as coded before :)

src/it/java/dev/jbang/it/InitIT.java
quintesse: > i believe it comes from that if you haven't requested a specific java version you run with the same as jbang is using

That's not the way it works, because you can't assume the current JDK is "usable". For example you might be running with a JRE which is unusable for JBang (because we need a compiler). So devkitman does several checks to ensure the current JDK is correct. One of the other tests is that it actually exists :-)

But I'll have to check if it still does so in this case. Perhaps the fact that the java.home is null throws it off somehow.

quintesse: > the code is this in Iniit.java:
> 
> > int reqVersion = buildMixin.javaVersion != null ? JavaUtil.minRequestedVersion(buildMixin.javaVersion)
> > : JavaUtil.getCurrentMajorJavaVersion();

Yeah, that is way to simplistic, this is what `getCurrentMajorJavaVersion()` does:

```
return parseJavaVersion(System.getProperty("java.version"));
```

That's way too simplistic. It's also a pretty old method that right now is only used for a couple of cases where we need to know what JDK we're running ourselves (for example to know if the JDK support modules).

> 
> there is no current project as its about to create it.
> 
> anything better?

Well I didn't say you _had_ to have a Project, I said _if_ you have a project :-)

The better way is the first thing I mentioned:, something like

```
        Jdk.@NonNull InstalledJdk jdk = JavaUtil.defaultJdkManager().getOrInstallJdk(buildMixin.javaVersion);
        int reqVersion = jdk.majorVersion();
```

src/main/java/dev/jbang/resources/resolvers/ClasspathResourceResolver.java
quintesse: Yeah, it's actually the toString() of ResourceNotFoundException that should be changed because the resource is already passed as the first argument. But that can be done some other time. For now this is good enough.

src/main/scripts/jbang
quintesse: Yeah, the move to reusable methods is nice but makes comparing what has changed a bit difficult :-)

quintesse: I also have the feeling that the these comments have in time gotten further and further away from the code they were meant to explain 🤣 

```
## https://stackoverflow.com/questions/1668649/how-to-keep-quotes-in-bash-arguments
## attempt to ensure each argument keeps its original quoting
```

src/main/scripts/jbang
```diff
@@ -126,13 +214,28 @@ fi
...
-if [ -f "$abs_jbang_dir/jbang.jar" ]; then
-  jarPath=$abs_jbang_dir/jbang.jar
-elif [ -f "$abs_jbang_dir/.jbang/jbang.jar" ]; then
-  jarPath=$abs_jbang_dir/.jbang/jbang.jar
-else
+## resolve jbang-cli binary or jar path from script location
+binaryPath=""
+jarPath=""
+if [[ "$JBANG_USE_NATIVE" == "true" ]]; then
+  # Look for native binary first if enabled
```
quintesse, line 223: I think we should be using OS checks here. Because if we have a distribution with multiple binaries it would choose the wrong one on Windows.

quintesse: And imagine you downloaded the Mac version on Linux by accident, or vice versa, you'd probably get a very confusing error. So I'd feel much better if from the start we're much more explicit about what we're trying to run, eg: `jbang.cli.mac`.

Ah yes, that's another important thing! Please, can we make the binary _not_ use a dash? Because it conflicts with the plugin system that looks for things in your PATH named `jbang-*`!

src/it/java/dev/jbang/it/InitIT.java
maxandersen: just to confirm - this will use locally available first, even java 8, right?

.github/workflows/ci-build.yml
maxandersen: build is now split up so general test/build not "held back" by native image build.

src/it/java/dev/jbang/it/InitIT.java
maxandersen: opened https://github.com/jbangdev/jbang/issues/2263 as this can be handled separately.

src/main/scripts/jbang
maxandersen: the AI didn't move them so i moved them same place as main have them - https://github.com/jbangdev/jbang/blob/main/src/main/scripts/jbang#L218

and yeah - not sure how it relates but that seems like organic comment regression :)

maxandersen: > Yeah, the move to reusable methods is nice but makes comparing what has changed a bit difficult :-)

well it was that or massive duplication to handle the two different ways to get to a jbang run command.

src/main/scripts/jbang
maxandersen: > I think we should be using OS checks here. Because if we have a distribution with multiple binaries it would choose the wrong one on Windows.

fair point. pushed fix for that

maxandersen: > And imagine you downloaded the Mac version on Linux by accident, or vice versa, you'd probably get a very confusing error. So I'd feel much better if from the start we're much more explicit about what we're trying to run, eg: `jbang.cli.mac`.

I get we should publish the binaries under .zip names that has os/arch in them and, im for having a os specific fallback but the actual default binary for me seems overkill. 

if you download linux jdk on mac you also get a weird error and vice versa
Do we really need it? If we need it - what naming pattern would we want to apply?

> Ah yes, that's another important thing! Please, can we make the binary _not_ use a dash? Because it conflicts with the plugin system that looks for things in your PATH named `jbang-*`!

fair point - `jbangcli` ? `jbang.bin` ? `jbangbin` ? `jbing` :) ?

maxandersen: I must say I was tempted to use `jbing` but my responsible adult devil on shoulder changed my mind - so I've made a commit that use `jbang.bin` for the binary - nice symmetry with `jbang.jar` and avoids the `jbang-<something>` issue.

src/it/java/dev/jbang/it/InitIT.java
quintesse: > just to confirm - this will use locally available first, even java 8, right?

It will use whatever JBang currently does. So it will use Java 8 if that's a) the version JBang itself is running with or b) it's what JAVA_HOME points to or c) it's the one that's currently first on your PATH or d) it's marked as the "default" JDK or e) it's the only JDK available :-)

PR #2264, opened 2025-10-28
fix: devkitman now passed default Java version

quintesse:
We weren't passing JBang's default Java version to devkitman, causing it to sue its own default version, which might be different. It also meant the user couldn't override this using JBANG_DEFAULT_JAVA_VERSION.

### DIFF
--- a/src/main/java/dev/jbang/util/JavaUtil.java
+++ b/src/main/java/dev/jbang/util/JavaUtil.java
@@ -29,7 +29,10 @@ public class JavaUtil {
 
 	@NonNull
 	public static JdkManager defaultJdkManager(List<String> names) {
-		return (new JdkManBuilder()).provider(names).build();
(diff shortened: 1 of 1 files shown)


maxandersen (approved):

maxandersen:
@quintesse what would be a good test for this? running without java in the path and see it picks up the default java version?

quintesse:
I think it only uses that value when installing new Jdks. AFAIR it's used when no Jdks are installed or if you're asking for let's say "9+" and you only have "8". It won't install "9" but instead it will install "defaultJavaVersion" as long as that version is equal to or larger than the one you requested. (So assuming the default is "17", asking for "24+" will install "24", but asking for "11+" will install "17")

So testing this might be somewhat complicated. Or not so much complicated but hard to test without incurring the cost of downloading a Jdk.

PR #2266, opened 2025-10-28
chore: add parent suite with os/arch/native or java in title + native image fixes

maxandersen:
small fix to have allure test suites have os/arch and native/not native in parent suites.

### DIFF
--- a/src/it/java/dev/jbang/it/CacheIT.java
+++ b/src/it/java/dev/jbang/it/CacheIT.java
@@ -2,11 +2,8 @@ package dev.jbang.it;
 
 import static dev.jbang.it.CommandResultAssert.assertThat;
 import static java.lang.System.lineSeparator;
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import org.junit.jupiter.api.Test;
-
-import dev.jbang.util.JavaUtil;
 
 public class CacheIT extends AbstractHelpBaseIT {
 
@@ -33,8 +30,6 @@ public class CacheIT extends AbstractHelpBaseIT {
 // * match err == "[jbang] Clearing cache for urls\n[jbang] Clearing cache for jars\n[jbang] Clearing cache for jdks\n[jbang] Clearing cache for kotlincs\n[jbang] Clearing cache for groovycs\n[jbang] Clearing cache for projects\n[jbang] Clearing cache for scripts\n[jbang] Clearing cache for stdins\n[jbang] Clearing cache for deps\n"
 	@Test
 	public void clearCacheAll() {
-		assumeTrue(!JavaUtil.inNativeImage()); // remove when https://github.com/jbangdev/jbang-devkitman/issues/87
-
 		assertThat(shell("jbang cache clear --all"))
 			.succeeded()
 			.errContains(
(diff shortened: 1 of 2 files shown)


maxandersen:
found way that does not require changing tests

itests/integrations/main.java
```diff
@@ -1,5 +1,5 @@
 ///usr/bin/env jbang "$0" "$@" ; exit $?
-
+//JAVA 13+
```
quintesse, line 2: Just curious, why this change?

quintesse: Oh, probably something to do with CDS?

src/main/java/dev/jbang/spi/IntegrationManager.java
```diff
@@ -244,14 +244,20 @@ protected IntegrationResult runIntegrationExternal(IntegrationInput input,
 
 		Path jbangJar = Util.getJarLocation();
 		args.add("-cp");
-		if (jbangJar.toString().endsWith(".jar")) {
-			args.add(jbangJar.toString());
+		if (JavaUtil.inNativeImage() && jbangJar.toString().endsWith(".bin")) {
```
quintesse: You probably have to take into account that on Windows it will be ".bin.exe"

src/main/java/dev/jbang/util/JavaUtil.java
```diff
@@ -22,6 +22,15 @@
 
 public class JavaUtil {
 
+	// best approximation we can get to check if we are in a native image without
+	// relying on graalvm specific properties
```
quintesse: is it me or does the comment state the opposite of what's implemented? To me it seems it _Does_ check a GraalVM property

src/main/scripts/jbang
```diff
@@ -224,10 +224,14 @@ if [[ "$JBANG_USE_NATIVE" == "true" ]]; then
   if [[ "$os" == "windows" ]]; then
     if [ -f "$abs_jbang_dir/jbang.bin.exe" ] && [ -x "$abs_jbang_dir/jbang.bin.exe" ]; then
       binaryPath="$abs_jbang_dir/jbang.bin.exe"
+    else
+      echo "WARNING: JBang native binary (jbang.bin.exe) not found or not executable in $abs_jbang_dir" 1>&2
```
quintesse: The part about "or not executable" doesn't make sense on Windows. There's no "executable" bit, if it's called ".exe" it's assumed to be executable. If it's not you'll get an error.

src/main/java/dev/jbang/util/JavaUtil.java
maxandersen: ah, should have said "classes" not "properties"

itests/integrations/main.java
maxandersen: yeah, im not sure why it didnt happen in normal testing but without it it complained about CDS not being available before java 13...

maxandersen:
> Left some comments, I think there's only one thing that probably should be fixed, the others are all minor things.

fixed them all - incl. the exe that was 1 of 2 errors in window nativ image test.

other is a NPE happening in natiemode only on w indows when running `jbang cache clear --all` - can you check if you can reproduce and get a `--verbose` output?

maxandersen:
thanks for the fix @quintesse - merging!

PR #2269, opened 2025-10-31
chore: bring back multi java test

maxandersen:
This took a tons of iterations/commeits/test runs to get right, but I think its there now: 

- shared-build-setup calculates short sha to get shorter artifact names
- all unit tests are run with Java 11 in github CI because only version the wiremock proxy hack works as of now but can be changed using -PtestJavaVersion for future testing
- `jbang.test.wiremock.enable`property added to to turn of wiremock, currenlty needed to run with something else than java 11 for tests
- added _JBANG_TEST_JAVA_VERSION to specify which default java version external launches in itests will use
- setup gtihub action for all LTS versions + OS's
- generate unique suite names to allow allure to combine all tests into on report without seeing other OS/java versions as retries
- generate combined report to see everything in one go.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -287,11 +287,14 @@ jobs:
     steps:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
       - id: shared-build
-        uses: ./.github/actions/shared-build-setup
+        uses: ./.github/actions/shared-build-setup       
       - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
         with:
           pattern: ${{ steps.shared-build.outputs.github-short-sha }}-*-allure-results-*
           path: build/allure-results
+          merge-multiple: true
+      - id: generate-report
+        run: ./gradlew --no-daemon allureReport --clean   
       - uses: ./.github/actions/shared-test-archiving
         with:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.adoc
```diff
@@ -42,6 +42,24 @@ BUILD SUCCESSFUL in 5s
 <.> This error happens as in the junit tests we test behaviors under different environment variables and to do that we need to modify internal state in java's `System` handling of properties. Java 9+ does not like that.
 <.> Already https://github.com/stefanbirkner/system-rules/issues/64[reported] but it can't be fixed as the jvm does not allow this; but fine for `jbang` as this is just for the test runs.
 
+=== Changing the Java version
+
+The build uses the `testJavaVersion` and `testJavaVendor` properties to set the Java version for the test runs.
+
+To change the Java version for the test runs, set the `testJavaVersion` and `testJavaVendor` properties in the `build.gradle` file.
```
quintesse, line 49: Extremely nitpicky, but when I read "set .... in the ... file" I assume I need to edit the file. Perhaps simply say something like "To change the Java version for the test runs, you can set the appropriate `testJavaVersion` and `testJavaVendor` properties.

quintesse (approved):
LGTM

maxandersen:
Fixed merged test reports. Merging so it can be used in the other awaiting prs.

PR #2321, opened 2025-11-18
fix: remote argument urls are now properly swizzled

### LINKED ISSUE #2285: Remote files (using `%https://....`) don't deal well with binaries
I tried using a remote reference to a PNG file, using  https://github.com/jbangdev/jbang/blob/main/images/jbang_icon.png to be exact, also tried with the direct permalink https://github.com/jbangdev/jbang/blob/86e29b0eddd106484c9e8eabd659151985f8cb93/images/jbang_icon.png but in both cases the downloaded file is a HTML file, not a binary.

quintesse:
Fixes #2285

### DIFF
--- a/itests/readimg.java
+++ b/itests/readimg.java
@@ -1,0 +1,10 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+
+import java.io.File;
+import javax.imageio.ImageIO;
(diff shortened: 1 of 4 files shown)


quintesse:
Not merging.

PR #2324, opened 2025-11-19
feat: JBangGPT -> JBang AI with multiple providers and no-preview

maxandersen:
realized i never really used jbanggpt because it was A) behind preview flag B) broken.

this fixes it, removes preview requirement + calls it JBang AI instead of GPT ...and introduce mechanism of picking a key from env and guess potential providers - incl. option of overriding it with `JBANG_AI_*` env.

should we add jbang config flags for them?

wdyt?


quintesse:
Yes, I think it would be a good idea to provide flags. It will also give more attention to this feature when people run `jbang init --help`

maxandersen:
implemented ai arggroup that uses environment/jbang options to look up values - @quintesse did i do that right ? 

also moved file check to before ai call so you dont wait several seconds to then get rejected.

`jbang init --ai-model=gpt-5.1 tetris.java implement tetris using whatever library that makes the most sense
, javafx, libgdx or something else as long as its java using jbang syntax in one file.`

gives you:

https://github.com/user-attachments/assets/bcfd573a-d4ac-45ba-9f61-32f23dc4cee3

https://gist.github.com/maxandersen/ad1eed2e9daf1be6285cbab9ce095511


i kinda wish we could come up with a way to detect ollama running locally but just a bit too many variants and models to make it worthwhile...user can always set the --ai-* config or env to make it work though.

maxandersen:
plan to merge this if tests passes as its isolated change and would like a release out

maxandersen:
i commented out providers i haven't had time to verify - just in case.
i moved the system prompt to a qute template + made it way more directed to increase likelihood of code actually working (but still not a gurantee - but still pretty damn good).

docs/modules/ROOT/pages/templates.adoc
```diff
@@ -79,4 +79,137 @@ Templates can refer to properties passed in using `-Dkey=value` on the `jbang in
...
+jbang init myapp.java "create a REST API client"
+----
+
+==== Configuration Properties
+
+You can set AI provider defaults using JBang's configuration system:
+
+[source,bash]
+----
+$ jbang config set default.ai.provider openai
```
quintesse, line 156: Everything looks good, except for this, the default values don't start with "default.". THe fact that in the code you have to do:

```
defaultValue = "${JBANG_AI_PROVIDER:-${default.ai.provider)}"
```

is just an internal implementation detail.

Also, special chars are removed from the names. In this case you're using a `.` in the name, which is not valid. So the actual name becomes: `aiprovider`. (Although the fqn can also be used, which would be `app.init.aiprovider`, but in this case that's not very interesting/useful)

src/main/java/dev/jbang/cli/AIOptions.java
```diff
@@ -0,0 +1,20 @@
+package dev.jbang.cli;
+
+import picocli.CommandLine;
+
+public class AIOptions {
+
+	@CommandLine.Option(names = {
+			"--ai-provider" }, description = "Provider to use (openai, openrouter, etc.) when using init with a prompt - if none specified will use the default provider from environment variables", defaultValue = "${JBANG_AI_PROVIDER:-${default.ai.provider)}")
```
quintesse, line 8: You can't do `default.ai.provider` here, you have to use the fully qualified name (without special chars like I said earlier), so it becomes `default.init.aiprovider`.

And finally I see you have typos in those lines, you're using a `)` instead of as `}`:

```
${JBANG_AI_PROVIDER:-${default.ai.provider)}
                                          ^
```

PR #2340, opened 2025-11-25
fix: updated graalvm information

wfouche:
Removed use of gu command

### DIFF
--- a/docs/modules/ROOT/pages/native-images.adoc
+++ b/docs/modules/ROOT/pages/native-images.adoc
@@ -189,10 +189,10 @@ Get a copy of the native binary:
 [source,bash]
 ----
 # Export native binary to current directory
-jbang export local --native myapp.java
+jbang export native myapp.java
 
 # Export to specific location
-jbang export local --native --output-dir=/usr/local/bin myapp.java
+jbang export native --output=/usr/local/bin/myapp myapp.java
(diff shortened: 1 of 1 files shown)


wfouche:
The following example still need to be updated.

```
# Compile to native image
jbang --native hello.java

# Run the native executable
./hello
```

No "hello" executable is created, and "jbang --native" compiles hello.java to a native image, stores it in the global JBang cache and then runs it.

Did the sematics of --native change at some point? When "export native" was added?

maxandersen:
Hmm. I think the example is wrong. 

It was always jbang --native hello.java and it would use the native binary if newer timestamp or rebuild.

wfouche:
@maxandersen , ready for review. Thanks.

This page is badly broken in its current un-edited state. These updates will allow others to at least build native image ./hello from hello.java and get started on the JBang native image journey.

wfouche:
>  @quintesse devkitman improvements enabling jbang itself to fetch a native-image capable jdk...

That cannot happen soon enough!

> I wonder if there aren't a page on graalvm.org that would be good to link to?

I added a link link to this Oracle GraalVM page for prerequisites and it simplifies the current page somewhat:

https://www.graalvm.org/latest/reference-manual/native-image/#prerequisites

PR #2342, opened 2025-11-26
fix: ai factory not handling defaults right

maxandersen:
not sure when it broke but the default mechanism for detecting ai providers was very broken. this fixes it so it does not keep attempting using open ai without a password.


maxandersen:
well thats surprising 15 errors in the yaml and github actions report it as passed...

PR #2349, opened 2025-12-05
fix: enable UTF-8 console output for jbang.{cmd,ps1,sh} on Windows

### LINKED ISSUE #2350: `jbang --help` console output on Windows
`jbang --help` is broken on Windows

Image

but works correctly on Linux and MacOS

Image

wfouche:
`jbang --help` on Windows does not display Unicode characters.

image

but now it does.

image

Fixes https://github.com/jbangdev/jbang/issues/2350


quintesse:
Although this definitely works the issue I have with this is that the encoding will persist after JBang exits. So we have basically changed the console's encoding affecting any commands we run afterwards.

Now in itself that might not be bad, in many (most?) cases this might actually improve things for the user. But it would be somewhat weird to see (some) apps behaving differently before JBang was run vs after JBang was run.

Of course we could try resetting the code page back again before exiting. Or point users to documentation on how to change codepages on the system level. (Although I definitely like being able to show correct output regardless of the user's system settings)

wfouche:
100% agree with what you are saying. However, I think this PR is better approach than asking users to globally enable 

image

because 

1. Most will never do this.

2. If it is set globally then the risk of  it breaking something is higher than localizing this setting to the specific Windows shell that JBang is running in. 

I think it is better to enable this setting as the PR proposes, than not to have it enabled at all.  It might break an ancient Windows command-line program that requires the default Windows code page. I don't think many of those are still in use by JBang command-line users.

wfouche:
Alternative CMD implementation.

```
REM Step 1 - save current code page
for /f "tokens=3" %%c in ('chcp') do set "ORIGINAL_CP=%%c"

REM Step 2 - set code page to UTF-8
chcp 65001 > NUL


REM Step n - restore original code page
chcp %ORIGINAL_CP% > nul
```

Something similar could be done for PowerShell. But I think this might be overkill.

CMD script code to only set the code page to 65001 if required:

```
setlocal
for /F "tokens=4" %%c in ('chcp') do (
    if "%%c" NEQ "65001" (
        chcp 65001 > nul
    )
)
endlocal
```

quintesse:
> Something similar could be done for PowerShell. But I think this might be overkill.

You might be right, I'll let @maxandersen decide

maxandersen:
I do think we should clean up - it's not good behaviour to modify users environment. We do the env cleanup for other variables; especially on windows. 

My main concerns are:

- does it affect the execution speed? I doubt it but windows can be surprising :) 
- impact on users execution. Should we run the users app with this on or off? I'm leaning towards keeping it on as UTF-8 is just easier :) but could imagine we would need to offer flag to NOT apply it - so should we do it just for jbang exec part?
- should we do similar on other OS for consistent behaviour ?

wfouche:
> does it affect the execution speed? I doubt it but windows can be surprising :)

Adds 7ms overhead on my 8-core dev machine

wfouche:
> impact on users execution. Should we run the users app with this on or off?

It should be on, we want to have nice things on Windows too!  :-)

```
set JBANG_APP_JAVA_OPTIONS=-Xmx1g
jbang run JvmRuntimeOpts.java
```
--- 🚀 JVM Runtime Options (VM Arguments) ---
Option 1: -Xmx1g

--- 📝 Application Arguments ---
No Application Arguments were passed to the main method.

--- ✅ Execution Complete ---

wfouche:
> should we do similar on other OS for consistent behaviour ?

Not needed on Linux or MacOS

wfouche:
> I do think we should clean up - it's not good behaviour to modify users environment.

OK, I'll fix it.

wfouche:
Fixed jbang.ps1 - it now restores the code page to its original value.

wfouche:
jbang,cmd recursively calling itself ...... oh no.

```
if "!binaryPath!"=="" if "!jarPath!"=="" (
  if not exist "%JBDIR%\bin\jbang.jar" (
    powershell -NoProfile -ExecutionPolicy Bypass -NonInteractive -Command "%~dp0jbang.ps1 version" > nul
    if !ERRORLEVEL! NEQ 0 ( exit /b %ERRORLEVEL% )
  )
  call "%JBDIR%\bin\jbang.cmd" %*
  exit /b %ERRORLEVEL%
)
```
I'm going to rename `jbang.cmd` to `_jbang.cmd` and then create a new `jbang.cmd` with the folloing code:

```
@echo off

rem Save current code page
for /f "tokens=2 delims=:" %%a in ('chcp') do set "_OriginalCP=%%a"
set "_OriginalCP=%_OriginalCP: =%"

rem Enable UTF-8 code page
chcp 65001 > nul

call _jbang.cmd %*

rem Restore original code page
chcp %_OriginalCP% > nul
```

 `exit /b <n>` to the rescue.

wfouche:
Only one test case fails on Windows:

```
	@Test
	public void shouldHandleSpecialCharacters() {
		assertThat(shell("jbang echo.java \" ~!@#$%^&*()-+\\:;\'`<>?/,.{}[]\"")).outIsExactly(
				"0: ~!@#$%^&*()-+\\:;'`<>?/,.{}[]" + lineSeparator());
	}
```

wfouche:
To get the special character test to pass, I reduced it to:

```
	@Test
	public void shouldHandleSpecialCharacters() {
		assertThat(shell("jbang echo.java \" ~@#$&*()-+\\:;'`<>?/,.{}[]\"")).outIsExactly(
				"0: ~@#$&*()-+\\:;'`<>?/,.{}[]" + lineSeparator());
	}
```

Removed special CMD shell characters % and  !  and ^

wfouche:
> Removed special CMD shell characters % and ! and ^

I did additional testing with what is for CMD is the standard way to escape these three characters.

1. ^!  is translated to !
2. %% is translated to %
3. ^^ should be translated to ^

The first two actually works fine in an input string to JBang.cmd. However ^^ does not translate to ^, but to ^^^^.

maxandersen:
why is unicode failing the escape code? are the special characters affected based on codepage? i know windows can be weird but that seems *really* weird. /cc @quintesse as he had the "most fun" with escape character handling.

wfouche:
It is fortunately not Unicode related. The string value of `%*` in `call jbangx.cmd %*` is modifed when passed to jbangx.cmd from jbang.cmd by CMD.exe. I confirmed this with this several echo debug statements in the CMD scripts.

wfouche:
Another approach would be to revert back to the previous implementation, and conditionally set the code page to 65001 using a new environment variable `JBANG_CMD_NO_UTF8` .  The code page will only be set to UTF-8 if JBANG_CMD_NO_UTF8 is not set to "1".

```
if not "%JBANG_CMD_NO_UTF8%" == "1" (
  chcp 65001 > nul
)
```

wfouche:
Adding line

```
chcp.com 65001 > /dev/null    # note the .COM extension
```

to ~/.bashrc for Git-Bash on Windows resolves the issue for Git-Bash as well.

wfouche:
@maxandersen , are you happy with the addition of the CMD specific environment  variable `JBANG_CMD_NO_UTF8` ?

maxandersen:
> It is fortunately not Unicode related. The string value of `%*` in `call jbangx.cmd %*` is modifed when passed to jbangx.cmd from jbang.cmd by CMD.exe. I confirmed this with this several echo debug statements in the CMD scripts.

not following - what is jbangx.cmd in that sentence? and how is that triggered in this PR ? or are you talking about something only happening in your own testing?

maxandersen:
also, googling around - this is only issue in cmd.exe ? windows terminal and other modern ones does not have this issue, correct?

maxandersen:
> if not "%JBANG_CMD_NO_UTF8%" == "1"

we used "true" in other places so lets stay consistent.

and shouldn't it be JBANG_WIN_NO_UTF8 ?

wfouche:
I'm done with coding and testing. It all works rather well with CMD, PowerShell and Git-Bash (if .bashrc is updated).

wfouche:
I'm done with coding and testing. It all works rather well with CMD, PowerShell and Git-Bash (without having to modify .bashrc).

src/main/scripts/jbang.ps1
```diff
@@ -203,5 +203,22 @@ if (-not $binaryPath) {
   }
 }
 
-# Execute jbang
-Invoke-JBang -binaryPath $binaryPath -jarPath $jarPath -javaExec $JAVA_EXEC -args $args
+# Save the current console encodings
```
quintesse: I find this part of the changes somewhat intrusive. A single line has become a whole block which is  of overall very little importance.

Besides I think an exception in setting the encoding will skip running JBang entirely!

So I'd like to see two methods being used: one that tries to set the encoding (and if it fails it simply does nothing and returns) and another that restores the encoding to the previous value (and again if it fails it just does nothing)

But... I also see that in the case of the bash and the CMD scripts we're not resetting the encoding to it's previous value, so why do so only for PS1? Either we do it for all or we do it for none (and I'd go for the latter honestly).

quintesse (changes requested):

src/main/scripts/jbang.ps1
maxandersen: > But... I also see that in the case of the bash and the CMD scripts we're not resetting the encoding to it's previous value, so why do so only for PS1? Either we do it for all or we do it for none (and I'd go for the latter honestly).

why do you not want to reset it back like we do for other variables too?

maxandersen:
> > It is fortunately not Unicode related. The string value of `%*` in `call jbangx.cmd %*` is modifed when passed to jbangx.cmd from jbang.cmd by CMD.exe. I confirmed this with this several echo debug statements in the CMD scripts.
> 
> not following - what is jbangx.cmd in that sentence? and how is that triggered in this PR ? or are you talking about something only happening in your own testing?

@wfouche still confused what this was about?

src/main/scripts/jbang.ps1
maxandersen: btw. I do agree that it seems weird having to add this much code and a try/catch for codepage setting? why not just set it - if it fails print warning or ignore and go on ?

src/main/scripts/jbang
```diff
@@ -142,6 +151,8 @@ execute_jbang() {
   fi
 }
 
+enable_console_utf8_if_needed_on_windows
```
maxandersen: where is the revert of the utf for shell?

src/main/scripts/jbang.cmd
```diff
@@ -1,4 +1,7 @@
 @echo off
+if not "%JBANG_WIN_NO_UTF8%" == "true" (
```
maxandersen: where is the revert of this?

src/main/scripts/jbang.ps1
maxandersen: > why do you not want to reset it back like we do for other variables too?

sorry - I misread your comment. yes we should be consistent with this or not do it at all.

wfouche: I wrote the PowerShell code first and then had difficulties in creating the same level of functionality with CMD. Will simplify the PowerShell implementation to be consistent with the other implementations.

src/main/scripts/jbang
wfouche: I changed PowerShell not to revert, to be consistent with SHell and Cmd.

src/main/scripts/jbang.cmd
wfouche: I changed PowerShell not to revert, to be consistent with SHell and Cmd.

wfouche:
> > > It is fortunately not Unicode related. The string value of `%*` in `call jbangx.cmd %*` is modifed when passed to jbangx.cmd from jbang.cmd by CMD.exe. I confirmed this with this several echo debug statements in the CMD scripts.
> > 
> > 
> > not following - what is jbangx.cmd in that sentence? and how is that triggered in this PR ? or are you talking about something only happening in your own testing?
> 
> @wfouche still confused what this was about?

I tried to add functionality to jbang.cmd to restore the code page at the end of the script, this was rather challenging so I decided that given that jbang.cmd calls itself somewhere in the code (which compliates the restore functionality given the weird langauge CMD script is), I would rename jbang.cmd to jbangx.cmd and then create a wrapper jbang.cmd script that saves the code page, change it to 65001, calls jbangx.cmd and when it returns restore the code page. This all worked perfectly fine except for special characters that were automatically processed by CMD Shell in the call from jbang.cmd to jbangx.cmd which I could not find a satisfactory fix for after trying really hard. All of those changes were reverted and the current implementation (three lines) created for CMD Shell. 

It is best to apply the KISS principle to the three implementations - Keep It Super Simple.

wfouche:
The PR now has a small snippet of documentation as well.

src/main/scripts/jbang.cmd
maxandersen: i dont follow - didnt we agree on reverting ?

im not following how calling jbang.cmd itself is an issue? both will record the previous codepage and return it to previous state?

wfouche: @maxandersen , the design for UTF8 support evolved organically, and we did not decide which way to go with it.

Ideally we want users to globally enable the Windows checkbox for "Beta: Use Unicode UTF-8 for worldwide language support", but few will do so and the majority will remain unaware of this option.

(1) That is why I thought the current PR provides a good middle-ground. We enable UTF8 when JBang is first run and leave it enabled for the lifetime of the shell instance. And if users don't like this behaviour they can disable the functionality by setting JBANG_WIN_NO_UTF8 to true.

(2) Alternatively we continue with the current PR functionality but restore the code page to its original value before JBang exits.

(3) But maybe you think this PR is not required, and we should just encourage JBang users on Windows to enable option  "Beta: Use Unicode UTF-8 for worldwide language support". If this becomes the default in future versions of Windows, then the current PR adds no value.

All that enabling option "Beta: Use Unicode UTF-8 for worldwide language support" does is to set three Windows Code Page related registry values to 65001.

```
set "KEY=HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Nls\CodePage"
set "VALUE=65001"

echo Setting ACP value to %VALUE%...
REG ADD "%KEY%" /v ACP /t REG_SZ /d "%VALUE%" /f

echo Setting MACCP value to %VALUE%...
REG ADD "%KEY%" /v MACCP /t REG_SZ /d "%VALUE%" /f

echo Setting OEMCP value to %VALUE%...
REG ADD "%KEY%" /v OEMCP /t REG_SZ /d "%VALUE%" /f
```

(4) We could also just document in the JBang user guide the steps required to enable UTF8 globally on Windows using the `intl.cpl` application, and modify `jbang --help` to print a warning message on Windows if UTF8 is not enabled.

```
    import java.io.BufferedReader;
    import java.io.InputStreamReader;

    public class CodePageChecker {
            try {
                Process process = Runtime.getRuntime().exec("chcp.com");
                BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()));
                String line;
                while ((line = reader.readLine()) != null) {
                    if (line.contains("Active code page: 65001")) {
                        System.out.println("Code page is 65001 (UTF-8).");
                        return; // Found the desired code page
                    }
                }
                System.out.println("Code page is not 65001 (UTF-8).");
            } catch (Exception e) {
                e.printStackTrace();
            }
        }
    }
 ```

Edit: Having extensively tested all the options discussed above, I now think that the best compromise is to just have JBang on Windows set the code page to 65001 (if it is set some other value) and restore it to the original value before exiting. This will give the best out-of-the-box experience to all Windows users of JBang.

**Comment by wfouche** [comment]
This is starting to feel like a Don Quixote kind of effort.

In jbang.sh because it execs itself under some circumstances, it will not reach the end of the script to restore the code page.

```bash
  if [ $err -eq 255 ]; then
    eval "exec $output"
    # should not reach here
    err=$?
  elif [ -n "$output" ]; then
    echo "$output"
  fi
  return $err
```

quintesse:
> This is starting to feel like a Don Quixote kind of effort.

Yeah, which is why I suggested to just forget about restoring the code page.
But of course that _could_ result in users getting confused if things start to work differently after the've run jbang.

Now, they can prevent that by setting the  `JBANG_WIN_NO_UTF8` variable, but then we're back again to the user having to do something to make things work. (And where do they set that variable permanently? Not all users know how to do that and perhaps do not _want_ to go through that trouble)

Of course if we turn it around and have a `JBANG_WIN_ENABLE_UTF8` then we're back to square one as well because who is going to set that instead of doing the "right" thing (= setting "Beta: Use Unicode UTF-8 for worldwide language support")?

So I'm conflicted. It would be nice to do this, but I fear that unless we can consistently reset the code page back to its original value we might be causing more issues than solving them. But @maxandersen is more of a risk taker than me, perhaps he feel differently and we simply forget about resetting the code page.

wfouche:
> Of course if we turn it around and have a JBANG_WIN_ENABLE_UTF8

I've come to the same conclusion, and the latest round of changes support this way of looking at the UTF-8 world on Windows.

If you opt-in then you're happy for UTF-8 to remain enabled in the current shell, once JBang exits and is no longer running.

I've found that enabling the "Beta: Use Unicode UTF-8 for worldwide language support" is intrusive and it causes Jython to print warnings, so it might break other programs as well.

Who will actually set JBANG_WIN_ENABLE_UTF8 to true? If we draw attention to this setting in `jbang --help` or when just `jbang` is run, we could inform users of JBang on Windows that UTF8 support can be enabled.

wfouche:
On Windows using `setx` from the CMD, PowerShell or Bash command-line one can create a new environment variable and assign a value to it. The value is persistent across reboots.

```
setx JBANG_WIN_ENABLE_UTF8 true
```

wfouche:
OK, this is the best approach yet.  :-)

JBang on Windows will now automatically enable UTF8 support if it is not enabled.

1. If env-var **JBANG_WIN_ENABLE_UTF8** is not defined:
*  Use setx to create the env-var in user space
* Use set (CMD), export (Bash) or ?? (PowerShell) to create the env-var in process space.
2. If env-var **JBANG_WIN_ENABLE_UTF8** == "true"
* chcp 65001

A user that do not want have UTF8 support to be enabled, simply run setx to set the env-var to `false` and then restart the shell session:
* setx **JBANG_WIN_ENABLE_UTF8** false
* Close shell, open a new Shell.

@quintesse, as a fellow Windows user, maybe you should be the first to have a look at this new approach. Support for CMD, PowerShell, and Git-Bash have been extensively tested.

src/main/scripts/jbang
```diff
@@ -9,6 +9,19 @@
 # The Java version to install when it's not installed on the system yet
 javaVersion=${JBANG_DEFAULT_JAVA_VERSION:-17}
 
+# Check if the script is running on Git Bash (MSYS environment)
+if [[ "$OSTYPE" == "msys" ]]; then
```
quintesse: I think you should move this down, below the place where we already do OS detection (eg around line 218) and check for `$os == windows` instead of using $OSTYPE.

quintesse:
Personally I'm still not really happy with something that just adds env vars to my system and on top of that asks _me_ to do things if I don't like it.

Are there any other known scripts that do this code page change trick (and then exit without changing it back)? If it's already "accepted behaviour" somehow then I'd be more willing to go along with it as well.

quintesse:
Btw, I'm sorry @wfouche that I keep coming up with obstacles, I know you're doing your best to find a good solution to this and that is very much appreciated!

wfouche:
@quintesse , console I/O for Java is stuck in legacy land on Windows because unlike Python3, Rust and others, it does not use the Windows Unicode API calls for Console I/O: WriteConsoleW and ReadeConsoleW.

JEP 400: UTF-8 by Default - https://openjdk.org/jeps/400

- Standardize on UTF-8 throughout the standard Java APIs, except for console I/O.

It is a common practice to set the Windows code page to 65001 for Java applications. And there are many ways to do this, some more intrusive than others.

The solution (with compromises) that I have implemented attempts to provide the best out-of-the-box JBang user experience on Windows. 

I hope that adding the JBANG_WIN_ENABLE_UTF8 environment variable automatically at runtime is not too big an issue. JBang could be modified to add this variable at installation time (if required). So adding it at runtime should not be a real problem.

quintesse:
> I hope that adding the JBANG_WIN_ENABLE_UTF8 environment variable automatically at runtime is not too big an issue

I just think that adding this variable 100% of the time when most likely 99.9% of our users will never touch it is not the correct way.

In that particular case, that someone who doesn't want JBang to do this, has to change a var that was added by us, it seems that the var is basically unnecessary in 99.9% of the cases. So if that 0.1% of the users have to got through the trouble of opening up the right control panel, and manipulating a variable anyway, I think we can ask them to _add_ the variable instead of changing it. That would mean not having to add any variables by default and it would only be used for people that don't want this behaviour.

(If we get lots of complaints we can always go back on this decision in a future update I guess)

wfouche:
>  I think we can ask them to add the variable instead of changing it.

That's a good idea.

src/main/scripts/jbang
wfouche: I would like to keep the code as it is. Hopefully we can remove this functionalty in the future. For that reason I would like to keep it separate from the rest of the code. This particular if statement (OSYTPE) also executes extremely quickly, and when it is not running in Git-Bash there will be no performance impact.

quintesse: It's not about speed or anything like that. It's about having a single way of detecting the OS. Yours for example doesn't take into account people using Cygwin. Now you could fix the test, but why do so when we already have a perfectly working OS detection? It's about not duplicating code (and thereby having 2 places for possible bugs). And I think it can just as easily be moved a bit lower down the file as where it is now :-)

wfouche: OK, I agree. I've moved the code and now it also supports cygwin, due to improved OS detection. Tested on cygwin as well.

src/main/scripts/jbang
```diff
@@ -197,6 +197,15 @@ case "$(uname -m)" in
     ;;
 esac
 
+## Enable UTF-8 code page on Windows unless explicitly disabled
+if [[ $os == windows ]]; then
+  if [[ -z "$JBANG_WIN_ENABLE_UTF8" ]]; then
```
quintesse: Wouldn't it be easier to have just one if-statement checking for "false"?

src/main/scripts/jbang.cmd
```diff
@@ -1,4 +1,11 @@
 @echo off
+if not defined JBANG_WIN_ENABLE_UTF8 (
```
quintesse: Same here

src/main/scripts/jbang.ps1
```diff
@@ -203,5 +203,11 @@ if (-not $binaryPath) {
   }
 }
 
+if ([System.Environment]::GetEnvironmentVariable('JBANG_WIN_ENABLE_UTF8') -eq $null) {
```
quintesse: And here

quintesse (changes requested):
I'm not sure if we should go back again to using `JBANG_WIN_DISABLE_UTF8` so we can keep `false` as the default value and have people set it to `true`.

wfouche:
I think it is better to use `JBANG_WIN_UTF8 ` which can be set to `true` or `false`.

The if statements have been simplified.

src/main/scripts/jbang.cmd
```diff
@@ -1,4 +1,10 @@
 @echo off
+if not defined JBANG_WIN_UTF8 goto :EnableUtf8
```
quintesse: I'd do something much simpler (didn't test):

```
@echo off
SETLOCAL ENABLEEXTENSIONS ENABLEDELAYEDEXPANSION

rem Set unicode codepage unless disabled
if "%JBANG_WIN_UTF8%" != "false" (chcp 65001 > nul)
```

quintesse: If you really really want to check for "empty/true" then even this seems nicer:

```
@echo off
SETLOCAL ENABLEEXTENSIONS ENABLEDELAYEDEXPANSION

rem Set unicode codepage unless disabled
if not defined JBANG_WIN_UTF8 (chcp 65001 > nul)
if "%JBANG_WIN_UTF8%" == "true" (chcp 65001 > nul)
```

wfouche:
Thanks @quintesse , the logic is now greatly simplified.

src/main/scripts/jbang
```diff
@@ -197,6 +197,13 @@ case "$(uname -m)" in
     ;;
 esac
 
+## Enable UTF-8 code page on Windows unless explicitly disabled
+if [[ "$os" == "windows" ]]; then
```
quintesse: Love it. Just one tiny change here: I'd collapse the ifs into a single one with &&.

quintesse (changes requested):
Just one last tiny change.

src/main/scripts/jbang
wfouche: Done.

quintesse (dismissed):
LGTM. Thanks for your patience @wfouche !

maxandersen:
sorry for being slow - i'm really not liking we mess with users codepages. If we can't reliably do that we should instead just document that users can call `chcp 65001` in ther shell/cmd init scripts - and we should just revert the codepage before calling exec - then jbang itself can do unicode and users apps/scripts are no wiser/more affected by this than they were before?

I saw mention of other java apps setting the codepage - i didnt find those; those i did find was calling things in a subprocess thus not affecting their outer environment. so i think we'll be a in "bad" company if we dont behave ;)

wfouche:
The Windows code page is only changed to UTF-8 in the current shell session and is not permanent. Without this PR, users would have to run JBang as

C:\>  chcp 65001 && jbang

on Windows to be compatible with JBang on MacOS and Linux where the console output is UTF-8 by default. The few users that do not want this behavious can set environment variable `JBANG_WIN_UTF8` to `false`.

Users new to JBang on Windows will think it is broken the first time they run it because UTF8 support is not enabled by default:

```
   JBang ??  Commonhaus Foundation, MIT License
? https://www.commonhaus.org/community
? https://jbang.dev
```

The incorrect output creates a bad impression when using JBang on Windows. Oh, it doesn't support Windows very well ....

There are documented workarounds for individual Shells to allow them to start with UTF-8 enabled:

* CMD:  %SystemRoot%\System32\cmd.exe /K "chcp 65001" 
* PowerShell:  powershell.exe -NoExit -Command "chcp 65001"
* Git-Bash .bashrc: chcp.com 65001 > nul

Users of JBang on Windows would prefer for JBang to automatically take care of UTF8 support on their behalf. Again, the few that don't want this behaviour (the exception) can set JBANG_WIN_UTF8 to false. 

This change will be a net positive for Windows users.

wfouche:
I'm in favour of releasing this PR, so that users on Windows can benefit from this change.

(1) JBang is not yet 1.0, so we are allowed to be a little more experimental with new functionality that is released.

(2) If this change is poorly received (and I don't think it will be) we can re-release JBang without this functionality, and users will still be able to disable the functionality by setting JBANG_WIN_UTF8 to false while waiting for the new release.

Sometimes we just need to confidently take a step forward, in order to be able to advance.

maxandersen:
> Users new to JBang on Windows will think it is broken the first time they run it because UTF8 support is not enabled by default:

yes, thats why I'm fine we enable it while jbang runs. That is ok.

> Users of JBang on Windows would prefer for JBang to automatically take care of UTF8 support on their behalf. Again, the few that don't want this behaviour (the exception) can set JBANG_WIN_UTF8 to false.

my issue is that with this jbang breaks users setup no matter if the user maybe explicitly set the code page to maybe utf-16 or some other codepage that is relevant for rest of the commands they run in their terminal, which might be running in a batch job, container or somewhere else where codepages settings are important. Its an inherent limitation of how windows works - and people relied on it for decades. 

I fully grok the ask to make jbang helpful - but its a general principle we follow that we help as much we can without breaking too hard.

ie. I could make jbang work without classes and main code but decide not to as it breaks users.

And if you insists its problematic jbang on windoes prints broken unicodes then I'll rather remove that printing than make a launcher that "pollutes/modifies" the users environment.

docs/modules/ROOT/pages/troubleshooting.adoc
```diff
@@ -294,10 +294,17 @@ jbang --verbose script.java
 **Problem**: Various Windows-specific problems
 
 **Solutions**:
+
 1. **Path length limits**: Use shorter paths, enable long path support
 2. **Antivirus interference**: Add JBang directories to exclusions
 3. **PowerShell execution policy**: `Set-ExecutionPolicy RemoteSigned`
 4. **Line endings**: Ensure scripts use LF, not CRLF
+5. **UTF-8 console support**: JBang will automatically enable UTF-8 console support by executing command  `chcp.com 65001`. To disable this functionality, create environment variable `JBANG_WIN_UTF8` and set it to `false` and restart the shell session. `JBANG_WIN_UTF8` can be set to `false` from the command-line using the `setx` command:
```
maxandersen, line 302: shouldn't we document where users actually should put that chcp.com 65001 to make it work for everything not just jbang?

wfouche:
The main motivation to add this functionality to JBang is less about JBang itself printing ?? characters on Windows, but nearly 100% about JBang Apps that work fine on Linux and MacOS and then fail to produce the expected output on Windows. Given that Java in JEP 400 has standardized on UTF-8, JBang *should* ensure a seamless UTF-8 experience across platforms including Windows. It does not, at the moment.

This program is a case in point.

```
import java.lang.management.ManagementFactory;
import java.util.List;

/**
 * JvmRuntimeOps.java
 * This program demonstrates how to retrieve and display:
 * 1. JVM Runtime Options (VM arguments).
 * 2. Application Arguments (arguments passed to the main method).
 */
public class JvmRuntimeOpts {

        // --- 1. Display JVM Runtime Options ---
        System.out.println("--- 🚀 JVM Runtime Options (VM Arguments) ---");

        // Use the ManagementFactory to get the RuntimeMXBean
        List<String> jvmArgs = ManagementFactory.getRuntimeMXBean().getInputArguments();

        if (jvmArgs.isEmpty()) {
            System.out.println("No explicit JVM Runtime Options found (default settings are in use).");
        } else {
            for (int i = 0; i < jvmArgs.size(); i++) {
                System.out.println("Option " + (i + 1) + ": " + jvmArgs.get(i));
            }
        }

        System.out.println("\n" + "--- 📝 Application Arguments ---");

        // --- 2. Display Application Arguments ---
        if (args.length == 0) {
            System.out.println("No Application Arguments were passed to the main method.");
        } else {
            for (int i = 0; i < args.length; i++) {
                System.out.println("Argument " + (i + 1) + ": " + args[i]);
            }
        }

        System.out.println("\n" + "--- ✅ Execution Complete ---");
    }
}
```

wfouche:
@maxandersen , the original code page is now restored when using CMD, PowerShell or Bash on (Cygwin or Git-Bash).

maxandersen:
> Given that Java in JEP 400 has standardized on UTF-8, JBang _should_ ensure a seamless UTF-8 experience across platforms including Windows. It does not, at the moment.

I get what you are after. And I applaud it. But unfortunately JEP 400 does NOT ensure seamless UTF-8 experience when running java - it ensures UTF-8 will mostly works IF you are running in a UTF-8 environment...subtle but important difference. Its still up to user to set things up properly.

src/main/scripts/jbang
```diff
@@ -133,11 +133,22 @@ execute_jbang() {
   fi
   
   if [ $err -eq 255 ]; then
-    eval "exec $output"
+    if [[ "$os" == "windows" ]] && [[ "$JBANG_WIN_UTF8" != "false" ]]; then
+      bash -c "$output"
```
maxandersen, line 137: why bash not exec here?

wfouche: We need the command to return to the curent shell , to be able to restore the code page setting at the end of the script.

eval "exec ..." does not return.

wfouche:
The alternative is to document what users should do to enable UTF-8 for each shell. 

*CMD* - add the following registry key:

`reg add "HKLM\Software\Microsoft\Command Processor" /v "Autorun" /t REG_SZ /d "@chcp 65001>nul" /f`

*Bash* (Git-Bash or Cygwin) - add the following line to `~/.bashrc`

`chcp.com 65001 > /dev/null`

*PowerShell* - add the following line to file $PROFILE

`$OutputEncoding = [Console]::InputEncoding = [Console]::OutputEncoding = [System.Text.UTF8Encoding]::new($false)`

This has the added benefit that programs run using `java` will also benefit from UTF-8 console support:

`java -jar JvmRuntimeOpts-fatjar.jar`

```
--- 🚀 JVM Runtime Options (VM Arguments) ---
No explicit JVM Runtime Options found (default settings are in use).

--- 📝 Application Arguments ---
No Application Arguments were passed to the main method.

--- ✅ Execution Complete ---
```

maxandersen:
ok - first, I'm really appreciative of your efforts here @wfouche its been very educational ! But starting to do different execution flows and jump through hoops that potentially breaks user flows is just not worth the hassle. 

Here is my suggestion:

1.  document the options you found - it would actually be nice to have that documented in one place rather than scattered all over the internet :)

2. (optionally, if you want) add JBANG_WIN_UTF8 which if true on windows we'll just do the right thing. 

That at least makes #1 have an option that is just set `JBANG_WIN_UTF8=true` and be done with it.

but I'm also fine just leave it at #1 as it is probably the best for users to just set those values or registry key anyways.

quintesse:
> as it is probably the best for users to just set those values or registry key anyways.

Not sure I agree here, the _actual_ thing they should do is set that beta feature in Windows:

image


Not messing about with registry keys because undoing those can be a nightmare.

quintesse:
Another option would be to use something like JNA and set the codepage from within JBang itself: https://github.com/java-native-access/jna/blob/master/contrib/platform/src/com/sun/jna/platform/win32/Wincon.java#L106

maxandersen:
> > as it is probably the best for users to just set those values or registry key anyways.
> 
> 
> 
> Not sure I agree here, the _actual_ thing they should do is set that beta feature in Windows:
> 
> 
> 
> image
> 
> 
> 
> 
> 
> Not messing about with registry keys because undoing those can be a nightmare.

Any indication setting those registry keys are anything different than toggling that registry key?

wfouche:
> Any indication setting those registry keys are anything different than toggling that registry key?

image

Enabling this option sets three registry values permanently to 65001:

- OEMCP - code page for console applications
- ACP - code page for GUI applications
- MACCP - the default Macintosh code page 

They are all under registry key

- HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Nls\CodePage

Running `chcp 65001`  temporarily changes the value of OEMCP that is active for the process and not the OEMCP registry value itself.

wfouche:
New PR created to document how to enable console UTF-8 support on Windows.

- https://github.com/jbangdev/jbang/pull/2356

wfouche:
> Another option would be to use something like JNA and set the codepage from within JBang itself

Currently evaluating this option.

wfouche:
This program uses JNA to save the current Windows code page, changes it to 65001, and restores the original code page again before exiting. System.out has to be reinitialized for Unicode output to work.

```java

//JAVA 25+

//DEPS net.java.dev.jna:jna:5.18.1
//DEPS net.java.dev.jna:jna-platform:5.18.1

import com.sun.jna.platform.win32.Kernel32;
import java.io.FileDescriptor;
import java.io.FileOutputStream;
import java.io.PrintStream;
import java.nio.charset.StandardCharsets;

void main(String... args) throws Exception {

    int CODEPAGE_UTF8 = 65001;

    if (!System.getProperty("os.name").toLowerCase().contains("windows")) {
        return;
    }

    // check if env var JBANG_WIN_UTF8 is true, otherwise return
    // Not yet implemented.
    
    // save current code page
    int currentOutputCP = Kernel32.INSTANCE.GetConsoleOutputCP();
    
    // set code page to 65001
    boolean setOutputOk = Kernel32.INSTANCE.SetConsoleOutputCP(CODEPAGE_UTF8);

    // remap stdout
    if (setOutputOk) {
        FileOutputStream fos = new FileOutputStream(FileDescriptor.out);
                
        PrintStream utf8Out = new PrintStream(
            fos, 
            true, 
            StandardCharsets.UTF_8.name()
        );
        
        // Replace the default System.out stream with the new UTF-8 stream
        System.setOut(utf8Out);
    }

    System.out.println("\nTesting UTF-8 output: \u2764");

    // Restore original code page
    Kernel32.INSTANCE.SetConsoleCP(currentInputCP);
    Kernel32.INSTANCE.SetConsoleOutputCP(currentOutputCP);

}
```

wfouche:
In the future if there ever is a need to move forward with this proposal, then the approach suggested by @quintesse to use JNA is the best way to implement the code page switching functionality.

maxandersen:
so im trying using utf-8 on a windows vm now and i'm for some reasons NOT seeing any effect of chaning the codepage.

Still gets ?? in the output.

maxandersen:
...and i can for the live of me not find that Beta setting anymore....im on Windows 11.

maxandersen:
> ...and i can for the live of me not find that Beta setting anymore....im on Windows 11.

the internet lied to me. Its not under Settings but under Control Panel...stupid.

maxandersen:
anyhow - just calling chcp did NOT work for me.

maxandersen:
ok im just dumb - or rather windows is stupid :)

chcp 65001 works everywhere but powershell, on powershell you need `[Console]::OutputEncoding = [System.Text.UTF8Encoding]::new($false)` ffs.

maxandersen:
that `jbang --utf8` idea is starting to grow on me :)

PR #2353, opened 2025-12-11
docs: list more alternatives

maxandersen:
realized there was so many popping up recently and others from past - so maybe just add them to the original docs - wdyt?

i also had a comparison matrix but wasnt looking great but if interest i can retry ;)


wfouche:
Might be worth mentioning https://github.com/elide-dev as well, although it is not a direct competitor.

maxandersen:
not spotting elide-dev be a script/app runner ? its more a compiler is it not?

maxandersen:
thought I had merged this so reopen, update with babashka and merge.

PR #2354, opened 2025-12-11
fix: use 17 as minium for alpine

maxandersen:
while making a homeassistant addon with jbang I noticed we are still downloading java 16 as default for alpine linux.

this try to fix that.

### DIFF
--- a/src/main/java/dev/jbang/util/Util.java
+++ b/src/main/java/dev/jbang/util/Util.java
@@ -955,7 +955,7 @@ public class Util {
 		if (!Files.exists(link)) {
 			// On Windows we use junction for directories because their
 			// creation doesn't require any special privileges.
-			if (getOS() == OS.windows && Files.isDirectory(target)) {
+			if (isWindows() && Files.isDirectory(target)) {
 				if (createJunction(link, target.toAbsolutePath())) {
 					return;
 				}
--- a/src/main/scripts/jbang
+++ b/src/main/scripts/jbang
@@ -152,7 +152,8 @@ case "$(uname -s)" in
   Linux*)
     os=linux
     if [ -f /etc/alpine-release ]; then
-      os=alpine-linux
+      os=alpine-linux 
+      libc_type=musl
       # alpine linux does not work below java 16 (or not available at time of writing)
       # Thus we set javaVersion to 17 if it's unset or less than 16
       if [[ -z "${javaVersion}" || "${javaVersion}" -lt 16 ]]; then


maxandersen:
hmm - i'm getting weird behaviour outside java 16...

```
ccdc483b96d1:/# /root/.jbang/cache/jdks/17/bin/javac
bash: /root/.jbang/cache/jdks/17/bin/javac: cannot execute: required file not found
ccdc483b96d1:/# /root/.jbang/cache/jdks/17/bin/java
```

javac is actually there..

so maybe we hardcoded that 16 for a good reason?

@quintesse you remember?

maxandersen:
running with verbose I get it is using this url:

https://api.foojay.io/disco/v3.0/packages?distro=temurin%2Caoj&javafx_bundled=false&libc_type=glibc&directly_downloadable=true&archive_type=tar.gz&operating_system=linux&package_type=jdk&version=25&release_status=ga&architecture=x64&latest=available

is that the expected one for alpine based linux on amd64?

have a suspicion its downloading something "unrunnable"

maxandersen:
ah it says libc_type=glibc ...that doesn't seem right on alpine that is musl based.

maxandersen:
and it should also be operating_system=alpine_linux.... i feel this worked before...maybe something missed in devkitman @quintesse ?

maxandersen:
in comparison:

https://api.foojay.io/disco/v3.0/directuris?distro=aoj&javafx_bundled=false&libc_type=glibc&archive_type=tar.gz&operating_system=alpine-linux&package_type=jdk&version=16&architecture=x64&latest=available

is what is downloaded explicitly by jbangsh script.

I also actually thought we used directuris in jbang itself but seems we dont?

maxandersen:
https://api.foojay.io/disco/v3.0/packages?distro=temurin%2Caoj&javafx_bundled=false&libc_type=musl&directly_downloadable=true&archive_type=tar.gz&operating_system=linux&package_type=jdk&version=25&release_status=ga&architecture=x64&latest=available

just changing libc_type=musl instead of glibc gets the right thing...wonder why jbang doesn't do that...i feel we had that before?

maxandersen:
yeah - the right url would be:

https://api.foojay.io/disco/v3.0/packages?distro=temurin%2Caoj&javafx_bundled=false&libc_type=musl&directly_downloadable=true&archive_type=tar.gz&operating_system=alpine_linux&package_type=jdk&version=25&release_status=ga&architecture=x64&latest=available

operating_system=alpine_linux and glib_c is musl

maxandersen:
yup, devkitman "forgot" about alpine_linux: https://github.com/jbangdev/jbang-devkitman/blob/main/src/main/java/dev/jbang/devkitman/util/OsUtils.java#L17

vs

https://github.com/jbangdev/jbang/blob/f3c68b2de321d21a4c312e283e4a0cbc9cdc38cb/src/main/java/dev/jbang/util/Util.java#L414

quintesse:
> just changing libc_type=musl instead of glibc gets the right thing...wonder why jbang doesn't do that...i feel we had that before?

I don't think I ever touched the startup scripts, so if they don't mention "musl" I don't think they ever did.

For the rest of the code, I don't see "musl" anywhere there either. Not in jbang nor in devkitman. Is it possible that it was added _just_ when I was extracting the Jdk management code into devkitman? I guess...

maxandersen:
@quintesse alpine was working after this 2021 pr: https://github.com/jbangdev/jbang/pull/839

and yes we dont detect musl anywhere - just alpine_linux ...but if we did - we wouldn't need to treat alpine_linux special.....but then again;not sure if we could even detect muslc outside detecting its on alpine_linux.

maxandersen:
anyhow...can we have a devkitman that works with alpine pretty please ? :)

quintesse:
> yup, devkitman "forgot" about alpine_linux

Ok, we'll have to add/re-add that! Weird though, the alpine part was already there in 2021, and I don't think devkitman is _that_ old. Btw, we should probably make JBang use devkitman's OSUtils so we don't get any de-syncs like this.

maxandersen:
if you want something to test on: https://gist.github.com/maxandersen/ff6b1fd0d9802457fa3ab28136d23c4d

quintesse:
I published a devkitman patch release: 0.3.2

maxandersen:
so it works but can't be tested on all platforms in github so ended up having to add funky assert/skip test that make the test only run on linux in github - but still runs when not in github IF docker/podman available. So it does risk being silent about possible breakage if you run tests in environment that does not have docker.

PR #2360, opened 2025-12-13
fix: enable UTF-8 console support in exported gradle projects

### LINKED ISSUE #2359: `jbang export gradle` and UTF-8
When experting a UTF-8 Java applications to a Gradle project, Unicode characters are not displayed correctly when though Windows is using code page 65001.

*Step 1 - run Java program with JBang*

```
C:\> jbang run JvmRuntimeOpts.java
--- 🚀 JVM Runtime Options (VM Arguments) ---
No explicit JVM Runtime Options found (default settings are in use).

--- 📝 Application Arguments ---
No Application Arguments were passed to the main method.

--- ✅ Execution Complete ---
```

*Step 2 - export to Gradle run using Gradle*

```
C:\> jbang export gradle JvmRuntimeOpts.java
C:\> cd  JvmRunTimeOpts
C:\> .\gradlew run
--- ? JVM Runtime Options (VM Arguments) ---
Option 1: -Dfile.encoding=UTF-8
Option 2: -Duser.country=ZA
Option 3: -Duser.language=en
Option 4: -Duser.variant

--- ? Application Arguments ---
No Application Arguments were passed to the main method.

--- ? Execution Complete ---
```

The fix is to also add Java runtime options:
- -Dstdout.encoding=UTF-8
- -Dstderr.encoding=UTF-8

wfouche:
Fixes #2359

### DIFF
--- a/src/main/resources/export-build.qute.gradle
+++ b/src/main/resources/export-build.qute.gradle
@@ -44,6 +44,7 @@ dependencies {
 application {
 	mainClass = '{fullClassName}'
 	applicationDefaultJvmArgs = [{jvmArgs}]
+	applicationDefaultJvmArgs += ['-Dstdout.encoding=UTF-8', '-Dstderr.encoding=UTF-8']
 {#if enablePreview eq 'true'}
 	applicationDefaultJvmArgs += ['--enable-preview']
 {/if}


maxandersen:
This doesn't seem right. You want all users to set those flags for all projects so when they run via Gradle in CMD.exe it will hide the issue of Gradle having disconnected terminals ? 

I'm not saying -1; just seems weird if there aren't a better way to handle it in gradle.

maxandersen:
closing this as it isnt something we can solve cleanly due to how windows works.

PR #2364, opened 2025-12-19
feat: -m supports pattern matching

maxandersen:
this is a draft of a idea that came to me bumping into projects with multiple mains.

i added printing of multiple mains in last release so user can copy/paste the name - this one uses that same code to let user get to interactively choose which main...but I didn't like it to happen automatically so came up with `-m <pattern>|classname`.

Basically if `-m` is not just a classname we treat it as "ask me".

this pr for now just implement `-m ?` so you get:

```
jbang -m ? someapp.jar
[jbang] No main class deduced, specified nor found in a manifest, but found these candidates:

(1) demo.Demo1
(2) demo.Demo2
(3) demo.Demo3
(4) other.Demo4
(5) other.Demo7
(6) my.appXY
(7) other.appX
(0) Cancel

[jbang] Type in your choice and hit enter. Will automatically select option (0) after 30 seconds.
```

could imagine we treat it as a regular expression and  you could do:

```
jbang -m other* someapp.jar
[jbang] No main class deduced, specified nor found in a manifest, but found these candidates:

(1) other.Demo4
(2) other.Demo7
(3) other.appX
(0) Cancel

[jbang] Type in your choice and hit enter. Will automatically select option (0) after 30 seconds.
```

and 

```
jbang -m other.app.? someapp.jar
[jbang] No main class deduced, specified nor found in a manifest, but found these candidates:

(1) other.appX
(0) Cancel

[jbang] Type in your choice and hit enter. Will automatically select option (0) after 30 seconds.
```

wdyt? too much or just enough to be useful (the code is trivial so feels like worth it to me)

should we do the asking by default or should we require it to be explicilty asked?

note that it uses our standard query mechanism so it will never block - at most delay 30 seconds.


maxandersen:
> LGTM

thanks - wdyt about wether we should do the asking by default or we should require pattern to be there?

i.e. should `jbang jarthathasoneormanymainclasses`  prompt and let user respond or should we print the list and inform they can use `-m *` ?

Actually, having -m * being needed is probably not good as * gets expanded. I guess `-m ?` and ? being treated as a special expression that gets converted to `*` regular expression would be simpler. I don't think ? will expand on any shell.

maxandersen:
note, `?` will expand if you have exactly one character file names....but that would be rare so treating as special trigger seems reasonable.

maxandersen:
I've updated this to only support glob patterns (?*) not full regex as its not really needed and characters like $ and others are actual valid java identifiers. 

Now it will default ask if no main specificed and will filter the list with the pattern - but for now I plan not running even if only single match as user have not explicitly chosen which main method to run.

PR #2370, opened 2025-12-30
fix: moved vscodium to end of editor list

### LINKED ISSUE #2369: JBang edit options
If a default editor is not configured for JBang, an options list is presented for the user to select the edit option to be executed.

Here's an example:

```
(1) Download and run VSCodium
(2) Use 'code'
(3) Use 'idea'
(0) Cancel
```

The option to download VSCodium should not be the first option to be presented to the user. Rather present the available editors that can be used first and then optionally allow the user to download and run VSCodim.

```
(1) Use 'code'
(2) Use 'idea'
(3) Download and run VSCodium
(0) Cancel
```

wfouche:
Fixes #2369

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -299,20 +299,19 @@ public class Edit extends BaseCommand {
 					"Do you want to";
 
 			List<String> options = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


maxandersen:
thanks for the PR, but -1 as discussed in #2369 as it makes less likely to work the first choice.

PR #2371, opened 2026-01-03
feat: add central-portal-snapshots maven repo shorthand/alias

maxandersen:
realized we are missing the snapshots repo for central portal uploads.

so now you can do `//REPOS central-portal-snapshots,central` to enable easy usage of published snapshots


maxandersen:
the failures are unrelated. some about alpine download which is separate issue.

Issue #1979, opened 2025-03-27
Support version pinning for aliases/scripts


quintesse:
The problem with running aliases and scripts from remote catalogs/sites is that the original author can update them and then we're stuck with several issues:

- do we update automatically (possibly breaking an existing script) or not (possibly forgoing critical bug/security fixes)?
- we can already somewhat tell JBang to update by using `--fresh` , but what if we don't want that and want to "pin" a specific version we already have?
- how can we notify the user that a newer version is available (if automatic updating is not available/enabled)?
- how can we support multiple versions and how do we make it possible to install a specific version?
- how can a user list the versions that are available?

Also see discussion here https://github.com/quarkiverse/quarkus-mcp-servers/issues/30

quintesse:
Having a section for versions was indeed something I thought of as well. Although I'm not sure why you chose to go with fatjars? Committing binaries alongside code is somewhat frowned upon and it would go against the idea that jbang always compiles the code on-the-fly (taking into account the Java version you're currently using etc).

So a very simple option could for example be:

```json
    "tulip-cli": {
      "script-ref": "scripts/tulip-cli.kt",
      "description": "Tulip CLI command",
      "versions":  {
        "2.1.4" : "scripts/tulip-cli-2.1.4.kt",
        "2.1.5":  "scripts/tulip-cli-2.1.5.kt",
        "2.1.6-testing": "scripts/tulip-cli-2.1.6.kt"
      }
    }
```

with an extended version if you want to provide more information, eg:

```json
    "tulip-cli": {
      "script-ref": "scripts/tulip-cli.kt",
      "description": "Tulip CLI command",
      "versions":  {
        "2.1.4" : "scripts/tulip-cli-2.1.4.kt",
        "2.1.5": {
           "scripts/tulip-cli-2.1.5.kt",
           "description": "New awesome features!"
        }
        "2.1.6-testing": {
           "scripts/tulip-cli-2.1.6.kt",
           "description": "Only for testing"
        }
      }
    }
```

But if your catalog is already on a versioning system like GitHub people will of course say "but why do I need to keep my versions manually when they are already versioned? That doesn't make any sense!". And they'd be right.

So even though this is an option, we need something more or something different.

quintesse:
So in case of having your catalog on Github/Gitlab/etc we could imagine something like this:

```json
    "tulip-cli": {
      "script-ref": "scripts/tulip-cli.kt",
      "description": "Tulip CLI command",
      "versions":  {
[...]

PR #2373, opened 2026-01-09
feat: initial support for alias versioning

wfouche:
Initial support for #1979 

```
$ jbang run camel:4.14.2@<username> version
[jbang]    org.apache.camel:camel-jbang-core:4.14.2
[jbang]    org.apache.camel.kamelets:camel-kamelets:4.14.2
[jbang]    org.apache.camel:camel-bom:4.14.2@pom
[jbang] Dependencies resolved
JBang version: 0.135.1
Camel JBang version: 4.14.2
```

*CamelJBang.java*

```java
//JAVA 17+
//REPOS central=https://repo1.maven.org/maven2,apache-snapshot=https://repository.apache.org/content/groups/snapshots/
//DEPS org.apache.camel:camel-bom:${jbang.app.version:4.16.0}@pom
//DEPS org.apache.camel:camel-jbang-core:${jbang.app.version:4.16.0}
//DEPS org.apache.camel.kamelets:camel-kamelets:${jbang.app.version:4.16.0}

package main;

import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;

/**
 * Main to run CamelJBang
 */
public class CamelJBang {

        CamelJBangMain.run(args);
    }

}
```

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -126,8 +126,8 @@ class AppInstall extends BaseBuildCommand {
 			Util.infoMsg("A script with name '" + name + "' already exists, use '--force' to install anyway.");
 			return false;
 		}
-		String scriptRef = scriptMixin.scriptOrFile;
 		ProjectBuilder pb = createProjectBuilder();
+		String scriptRef = pb.extractJBangAppVersion(scriptMixin.scriptOrFile);
 		Project prj = pb.build(scriptRef);
 		if (name == null) {
 			name = CatalogUtil.nameFromRef(scriptRef);
@@ -157,10 +157,20 @@ class AppInstall extends BaseBuildCommand {
 				|| Files.exists(binDir.resolve(name + ".ps1"));
 	}
 
-	private static void installScripts(String name, String scriptRef, List<String> runOpts, List<String> runArgs)
+	private static void installScripts(String name, String _scriptRef, List<String> runOpts, List<String> runArgs)
 			throws IOException {
 		Path binDir = Settings.getConfigBinDir();
 		binDir.toFile().mkdirs();
+		String version = System.getProperty("jbang.app.version");
+		String scriptRef = _scriptRef;
+		if (version != null) {
+			if (scriptRef.contains("@")) {
+				String[] parts = scriptRef.split("@", 2);
+				scriptRef = parts[0] + ":" + version + "@" + parts[1];
+			} else {
+				scriptRef = scriptRef + ":" + version;
+			}
+		}
 		if (Util.isWindows()) {
 			installCmdScript(binDir.resolve(name + ".cmd"), scriptRef, runOpts, runArgs);
 			installPSScript(binDir.resolve(name + ".ps1"), scriptRef, runOpts, runArgs);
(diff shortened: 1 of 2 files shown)


src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -263,8 +265,59 @@ public Project build(String resource) {
 		return build(resourceRef);
 	}
 
-	private ResourceRef resolveChecked(ResourceResolver resolver, String resource) {
+	public String extractJBangAppVersion(String _resource) {
```
maxandersen, line 268: this method does not feel right.

it is not actually extracting the version - it is detecting and putting it into a global system property and then returns it self. 

we shouldn't rely on global state for this - its very local for the exact resolution of versions for alias ref.

src/main/java/dev/jbang/cli/App.java
```diff
@@ -126,8 +126,8 @@ public boolean install() throws IOException {
 			Util.infoMsg("A script with name '" + name + "' already exists, use '--force' to install anyway.");
 			return false;
 		}
-		String scriptRef = scriptMixin.scriptOrFile;
 		ProjectBuilder pb = createProjectBuilder();
+		String scriptRef = pb.extractJBangAppVersion(scriptMixin.scriptOrFile);
```
maxandersen, line 130: it shouldn't have to extract a version here *always* should it ? normally should only be needed if alias actually have a version to set, right?

src/main/java/dev/jbang/cli/App.java
```diff
@@ -157,10 +157,20 @@ private static boolean existScripts(Path binDir, String name) {
 				|| Files.exists(binDir.resolve(name + ".ps1"));
 	}
 
-	private static void installScripts(String name, String scriptRef, List<String> runOpts, List<String> runArgs)
+	private static void installScripts(String name, String _scriptRef, List<String> runOpts, List<String> runArgs)
```
maxandersen, line 160: why need to put _ in the signature parameter?

src/main/java/dev/jbang/cli/App.java
```diff
@@ -157,10 +157,20 @@ private static boolean existScripts(Path binDir, String name) {
...
 			throws IOException {
 		Path binDir = Settings.getConfigBinDir();
 		binDir.toFile().mkdirs();
+		String version = System.getProperty("jbang.app.version");
+		String scriptRef = _scriptRef;
+		if (version != null) {
```
maxandersen, line 166: this logic seems in reverse of what I would expect.

I would assume if and only if a version is specified in the alias will we pass down version info to be resolved. No need to do this otherwise.

src/main/java/dev/jbang/source/ProjectBuilder.java
maxandersen: even if we had to use global state this method should be possible to test in isolation which we can't if reliant on global system property state.

wfouche:
This implementaton is largely a proof-of-concept.  Looking at the code again, I think one should rather enhance the ResourceRef class to support resource versioning. The current implementation removes the version information from the resource ref up-front and then adds it back where needed during back-end processing in order not to break current functionality. But it does not fit really well into the JBang processing model.

PR #2376, opened 2026-01-19
feat: relative repos. --repos=local=./build/repo

maxandersen:
with this you can do `//REPOS local=build/repo` and jbang will turn that into an absolute file url based on current directory.

Its mainly intended for case where you use `gradle publishAllPublicationsToBuildRepository` and get `/build/repo` with the built artifacts. 

with this I can put:

```jbang.properties
repos=central,gradle=build/repo
```

and when in root run scripts/aliases and have it resolve against local built repo.

it of course have the downside that artifacts will be copied to local maven.

I've for now marked it as experimental as I'm realizing we should have a way to resolve the path relative to where the repos ref is placed or at least be able to use property replacements to be more consistent and able to run from anywhere.

### DIFF
--- a/src/main/java/dev/jbang/cli/JBang.java
+++ b/src/main/java/dev/jbang/cli/JBang.java
@@ -89,6 +89,13 @@ public class JBang extends BaseCommand {
 		}
 	}
 
+	@CommandLine.Option(names = {
+			"-x",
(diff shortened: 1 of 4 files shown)


maxandersen:
missed failing test ...moving to draft

PR #2381, opened 2026-02-06
fix: add missing "/" before jbang-catalog.json

### LINKED ISSUE #2380: Error resolving hello@acme~experimental
`jbang hello@acme~experimental`  resolves to 

- `https://...../HEAD/experimentaljbang-catalog.json`

and it should resolve to

- `https://...../HEAD/experimental/jbang-catalog.json`

wfouche:
Fixes #2380

### DIFF
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -31,7 +31,12 @@ public class ImplicitCatalogRef {
 	}
 
 	private String repoUrl(String host, String infix) {
-		String fs = path.endsWith("/") ? "" : "/";
+		String fs = "";
+		if (!path.isEmpty()) {
+			if (!path.endsWith("/")) {
+				fs = "/";
+			}
+		}
 		return host + org + "/" + repo + infix + ref + "/" + path + fs + Catalog.JBANG_CATALOG_JSON;
 	}
 
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -31,7 +31,7 @@ public class ImplicitCatalogRef {
 	}
 
 	private String repoUrl(String host, String infix) {
-		String fs = !path.isEmpty() ? "/" : "";;
+		String fs = !path.isEmpty() ? "/" : "";
 		return host + org + "/" + repo + infix + ref + "/" + path + fs + Catalog.JBANG_CATALOG_JSON;
 	}
 
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -27,17 +27,11 @@ public class ImplicitCatalogRef {
 		this.org = org;
 		this.repo = repo;
 		this.ref = ref;
-		this.path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
+		this.path = path;
 	}
 
 	private String repoUrl(String host, String infix) {
-		String fs = "";
-		if (!path.isEmpty()) {
-			if (!path.endsWith("/")) {
-				fs = "/";
-			}
-		}
-		return host + org + "/" + repo + infix + ref + "/" + path + fs + Catalog.JBANG_CATALOG_JSON;
+		return host + org + "/" + repo + infix + ref + "/" + path + Catalog.JBANG_CATALOG_JSON;
 	}
 
 	/**
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -31,7 +31,13 @@ public class ImplicitCatalogRef {
 	}
 
 	private String repoUrl(String host, String infix) {
-		return host + org + "/" + repo + infix + ref + "/" + path + "/" + Catalog.JBANG_CATALOG_JSON;
+		String fs = "";
+		if (!path.isEmpty()) {
+			if (!path.endsWith("/")) {
+				fs = "/";
+			}
+		}
+		return host + org + "/" + repo + infix + ref + "/" + path + fs + Catalog.JBANG_CATALOG_JSON;
 	}
 
 	/**
--- a/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
+++ b/src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
@@ -27,7 +27,7 @@ public class ImplicitCatalogRef {
 		this.org = org;
 		this.repo = repo;
 		this.ref = ref;
-		this.path = path;
+		this.path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
 	}
 
 	private String repoUrl(String host, String infix) {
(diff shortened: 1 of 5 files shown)


maxandersen:
hmm - not sure what this is fixing?
got a test that fails we can add?

wfouche:
https://www.jbang.dev/documentation/jbang/latest/alias_catalogs.html

The advertised support for 

- jbang hello@acme~experimental
- hello found in acme/jbang-catalog/experimental/jbang-catalog.json

does not work without this PR.

The alias is resolved to acme/jbang-catalog/**experimentaljbang**-catalog.json with a missing '/' before jbang-catalog.json 

I'll gladly add a testcase, just tell me to which file I should add it and roughly how it should work.

maxandersen:
yeah was just looking - and it should have been in TestImplicitAlias but it does not seem to test implicit catalog much at all. so yeah, definitely should expand that one ;)

wfouche:
@quintesse , I reverted a change that you made to remove '/' if it was at the end of the path string. Maybe it was just debug code that you wrote and inadvertently submitted, or did it serve a real purpose? The test cases are not affected by this revert, but it does fix how "hello@acme~experimental" is resolved.

maxandersen:
@wfouche what pr/commit did the original change?

in any case, should add test for this in TestImplicitAlias so we dont loose it again :)

wfouche:
> @wfouche what pr/commit did the original change?

See changes made to file ImplicitCatalogRef.java by https://github.com/jbangdev/jbang/pull/1743/changes#diff-596ddf206078ca715256b4ea770cd920ab1e5c64908d43f912b47e81f720e180

- this.path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path

> in any case, should add test for this in TestImplicitAlias so we dont loose it again :)

Yes.

wfouche:
@maxandersen , 

- ImplicitCatalogRef.resolveImplicitCatalogUrl("jbangdev/jbang-examples").get() is resolved correctly, but
- ImplicitCatalogRef.resolveImplicitCatalogUrl("jbangdev~experimental").get() return a null value.

I don't have time to go too deep into this at the moment, but will create a new issue to for this to be fixed and a proper testcase to be created for jbangdev~jbang-examples.

wfouche:
Related issue #2383

maxandersen:
What would double ~ mean ?

wfouche:
I've updated the examples to eliminate any visual confusion it might have caused.  There is no double ~ , just a single ~ is being used.

wfouche:
@maxandersen , this PR is ready to be reviewed.

maxandersen:
@wfouche thanks - i'm seeing one change, no test to catch we dont break again - were you not adding a test to TestImplicitAlias?

wfouche:
@maxandersen , in the end I only reverted an earlier change that was made which broke the acme~experimental functionality. It would be great if you can merge the fix as is (and release a new version of JBang).

I spend time trying to create an additional test case, but the code is rather complex, hard to understand and the seemingly simple test case just did not work. After debugging for a while, I suspended the effort. But I did create issue https://github.com/jbangdev/jbang/issues/2383 for us to remember to follow up and create the required test case.

maxandersen:
What is the actually issue this final bugfix fixes? I'll add a test for it.

wfouche:
@maxandersen , the issue is easy to reproduce.

Just run `jbang --verbose hello@jbangdev~experimental`
....
[jbang] [4:106] Requesting HTTP GET https://gitlab.com/jbangdev/jbang-catalog/-/raw/HEAD/experimentaljbang-catalog.json

**experimentaljbang-catalog.json** is the issue, with the fix it is resolved to **experimental/jbang-catalog.json**

I also run the following app during testing.

```java
//DEPS dev.jbang:jbang.bin:0.136.0

//JAVA 21

import dev.jbang.catalog.Alias;
import dev.jbang.catalog.ImplicitCatalogRef;

public class hello {

        var s = "jbangdev";
        var z = ImplicitCatalogRef.resolveImplicitCatalogUrl(s).get();
        out.println(z.toString());

        s = "jbangdev/jbang-examples";
        z = ImplicitCatalogRef.resolveImplicitCatalogUrl(s).get();
        out.println(z.toString());

        s = "jbangdev~experimental";
        z = ImplicitCatalogRef.resolveImplicitCatalogUrl(s).get();
        out.println(z.toString());
    }
}
```

wfouche:
@maxandersen . I created a test case but it is not working 100%. Maybe you can spot the error and fix it.

I'm using the following hello app as part of the test:

- https://github.com/wfouche/jbang-catalog/tree/main/hello

It would be best to move it to the **jbangdev/jbang-catalog** repo, if it is to remain part of the test.

src/test/java/dev/jbang/cli/TestRun.java
```diff
@@ -390,6 +391,43 @@ void testJarViaHttps(@TempDir Path tdir) throws IOException {
 		}
 	}
 
+	@Test
```
quintesse: This seems like an insanely complex test for such a simple change :-)
I would suggest just adding a simple test to `TestImplicitAlias` where you test directly that the class returns what it is supposed to return.

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
```diff
@@ -27,7 +27,7 @@ private ImplicitCatalogRef(String org, String repo, String ref, String path) {
 		this.org = org;
 		this.repo = repo;
 		this.ref = ref;
-		this.path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
+		this.path = path;
```
quintesse: Let's first get a better test. But I'm not sure this is the correct fix.
The reason this line is here is to make sure that the path is always in a known format (in this case we known that it will never end in a slash) which makes it easier to build full URLs with it by simply concatenating strings.
So my preference would be in making changes to the code that creates URLs (like in the `repoUrl()` or `toString() methods).

quintesse (changes requested):

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
wfouche: Thanks @quintesse . I reverted the this.path update and fixed repoUrl().

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
```diff
@@ -31,7 +31,13 @@ private ImplicitCatalogRef(String org, String repo, String ref, String path) {
 	}
 
 	private String repoUrl(String host, String infix) {
-		return host + org + "/" + repo + infix + ref + "/" + path + Catalog.JBANG_CATALOG_JSON;
+		String fs = "";
+		if (!path.isEmpty()) {
```
quintesse: Well, we already know that path doesn't end with a "/", the constructor makes sure of that, so the second if isn't necessary :-)

And perhaps write it like:

```
String fs = !path.isEmpty() ? "/" : "";
```

quintesse (changes requested):

src/main/java/dev/jbang/catalog/ImplicitCatalogRef.java
wfouche: Fixed.

wfouche:
@maxandersen , finally we have a simple & working test case!  :-)

quintesse (commented):
LGTM 👍

quintesse (approved):
Stupid UI. It changed my Approve back to Comment just when I hit Submit.

quintesse:
Thanks @wfouche !

PR #2385, opened 2026-02-10
docs: missing --name flag in alias add

chabala:
Minor documentation correction, `--name` flag was missing from an alias add command example.


maxandersen:
@all-contributors add @chabala for docs

thank you!

PR #2388, opened 2026-02-14
scala language support

maxandersen:
Feedback welcome - I'm not a scala super user but thought why not try add it :)

scala-cli exists in scala world and is really good and heavily similar (possibly inspired by?) jbang but it doesn't supprot jbang catalog so I thought - why not make it possible in jbang. thus this is not trying to compete with scala-cli nor replace it - definitely use scala-cli where you can :)

For now this just add .scala support just like for .java and .kt, .groovy.

meaning, we dont do like scala-cli does and wrap .sc content in a magical main body nor do we at this time support: 
//DEPS com.lihaoyi::os-lib:0.9.1 but require user to be explicit about the package they need, i.e. //DEPS com.lihaoyi:os-lib_3:0.9.1.

I feel we should then rather call scala-cli directly and let it deal with that instead. 

Anyhow - feedback welcome :)

### DIFF
--- a/docs/modules/ROOT/pages/caching.adoc
+++ b/docs/modules/ROOT/pages/caching.adoc
@@ -16,4 +16,18 @@ In previous versions of `jbang`, Java 10+ direct launch of `.java` was used, but
 
 The caching goes to `~/.jbang/cache` by default, you can run `jbang cache clear` to remove all cache data from this folder.
 
+You can also clear individual cache categories:
+
+[source,bash]
+----
+# Clear compiled jars only
+jbang cache clear --jar
+
+# Clear dependency metadata cache
+jbang cache clear --deps
+
+# Clear downloaded Scala compiler cache
+jbang cache clear --scalac
+----
+
 The default cache location can be overwritten by the environment variable `JBANG_CACHE_DIR`. If `JBANG_DIR` environment variable is set, the `cache` folder will be placed there.
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -50,7 +50,7 @@ JBang is a tool that makes it easy to write and run Java scripts without the tra
 
 | **Zero Setup** | Run Java files directly - no project setup needed
 | **Dependency Management** | Declare dependencies with `//DEPS` - automatic resolution
-| **Multiple Languages** | Support for `.java`, `.jsh`, `.kt`, `.groovy`, and `.md` files
+| **Multiple Languages** | Support for `.java`, `.jsh`, `.kt`, `.groovy`, `.scala`, and `.md` files
 | **Native Images** | Generate native binaries with GraalVM
 | **IDE Integration** | Edit with full IDE support using `jbang edit`
 | **Templates** | Quick start with built-in templates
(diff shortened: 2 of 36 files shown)


chabala:
When I was looking at what it would take for me to implement this PR, I started by searching for 'kotlin' in the codebase, as an example of another experimentally supported JVM language. I'm not sure that I found all of these files or their Kotlin-specific analogues. Just some thoughts:
 * there's a `Type` enum in `Source`. Maybe some code could be made simpler by iterating over the types, e.g.

```java
		} else if (ext.equals("groovy")) {
			return new GroovySource(resourceRef, replaceProperties);
		} else if (ext.equals("scala")) {
			return new ScalaSource(resourceRef, replaceProperties);
		} else if (ext.equals("jsh")) {
			return new JshSource(resourceRef, replaceProperties);
```

if every `Type` has an associated Type-`Source` class, then `Type.findByExt(ext).getSource()`, either passing the parameters, or maybe just letting `getSource()` return a method reference to the constructor. This may be applicable to the language specific parts of `Cache` and `ProjectBuilder` as well.

Then fewer things need to change to support a new language, less chance of missing something or getting out of sync.

* there's a lot of duplicated documentation. Could more of it be generated from a common source? Maybe it already is and it's just being tracked as separate files.

wfouche:
> public static final String DEFAULT_SCALA_VERSION = "3.8.1";

Scala 3 versions from 3.3 LTS through Scala 3.7 support JDK 8 and above.

Scala 3 versions starting with 3.8 have a new minimum JDK version of 17.

The DEFAULT_SCALA_VERSION should be "3.7.4" to allow JBang to use JDK8.

wfouche:
> I feel we should then rather call scala-cli directly and let it deal with that instead.

I agree. Keep is simple for now, and primarily aim to support Scala programs using Java JARs (Maven Central)

wfouche:
I'm using `scala-cli` at the moment but looking will change my Scala example app to JBang once this functionality is released.

*App.scala using scala-cli*

```scala
//> using jvm 21
//> using dep io.github.wfouche.tulip:tulip-runtime:2.2.3
//> using javaOpt -Xmx2g -XX:+UseZGC -XX:+ZGenerational
//> using javaOpt -XX:+IgnoreUnrecognizedVMOptions
//> using javaOpt --enable-native-access=ALL-UNNAMED
//> using javaOpt --sun-misc-unsafe-memory-access=allow
//> using repositories m2local

package io.tulip

import io.github.wfouche.tulip.api.TulipApi

object App {
  def main(args: Array[String]): Unit = {
    TulipApi.generateReport(TulipApi.runTulip("benchmark_config.json"))
  }
}
```

maxandersen:
> Scala 3 versions starting with 3.8 have a new minimum JDK version of 17.
> 
> The DEFAULT_SCALA_VERSION should be "3.7.4" to allow JBang to use JDK8.

not sure I follow why ? our default java version is 17+ ...why should scala be set to what needs jdk8? the right jdk shuold be selected/downloaded should it not?

wfouche:
>  our default java version is 17+

Agreed. Ignore my comment. Nobody is going to use JDK8 with Scala anymore, but if they do need to then would have to specify the appropriate version of Scala that still support JDK8.

chabala:
Makes sense, I pin JBANG_DEFAULT_JAVA_VERSION=8 to minimize JDK installs, I can do the same for DEFAULT_SCALA_VERSION (or should it be JBANG_DEFAULT_SCALA_VERSION to match the other env variables?)

maxandersen:
Jbang_ prefix should be used for consistency.

PR #2389, opened 2026-02-15
fix(deps): upgrade Gradle plugins

wfouche:


wfouche:
> Dependency requires at least JVM runtime version 17. This build uses a Java 11 JVM.

PR #2431, opened 2026-03-23
feat: add WAR file support with full JAR parity

### LINKED ISSUE #2428: Executable WAR files
Some .war files are configured so that they are "executable" using command "java -jar":

```
$ curl -O https://download.structurizr.com/structurizr-2026.03.06.war
$ java -jar structurizr-2026.03.06.war version
09:32:17.237 [main] INFO com.structurizr.command.VersionCommand -- structurizr: 2026.03.06
09:32:17.252 [main] INFO com.structurizr.command.VersionCommand -- structurizr-*: 6.1.0
09:32:17.252 [main] INFO com.structurizr.command.VersionCommand -- Java: 21.0.10/Microsoft 
09:32:17.252 [main] INFO com.structurizr.command.VersionCommand -- OS: Windows 11 10.0 (amd64)
```

JBang is not able to run executable .war files.

Recommendation: JBang should be updated to attempt to run .war files as well. 

This change will allow the structurizr application to be run as:

- jbang run https://download.structurizr.com/structurizr-2026.03.06.war  version

Also see https://github.com/structurizr/structurizr/discussions/60

maxandersen:
This introduce notion "executableArchive" instead of "jars".  That is why so many files affected - just rename of isJar to isExecutableArchive.

details below - but basically .jar and .war now treated equally.

## Summary

Enables jbang to execute WAR files with the same capabilities as JAR files:
- WAR files work in all contexts: local files, remote URLs, Maven coordinates, aliases, and catalogs
- Extracts Main-Class from manifest and runs with war on classpath (not `java -jar`) to allow additional dependencies
- Full backwards compatibility - deprecated `isJar()` but kept for compatibility

## Implementation

- Added `isExecutableArchive()` and `hasExecutableExtension()` to detect both .jar and .war files
- Updated all call sites across 8 source files to use new methods
- Fixed resource resolver to prevent renaming .war files to .war.java

## Testing

- Added WarTestFixtures helper to eliminate test duplication  
- 21 new tests covering core functionality and edge cases
- Integration tests for local/remote war execution
- All existing tests continue passing ✅

## Documentation

- Added WAR examples to running.adoc, dependencies.adoc, jbang-run.adoc
- Documented identical treatment of JAR and WAR files

## Example Usage

```bash
# Local WAR file
jbang myapp.war

# Remote WAR file  
jbang https://download.structurizr.com/structurizr-2026.03.06.war version

# With additional dependencies
jbang --deps com.google.gson:gson:2.8.9 myapp.war

# In aliases
jbang alias add structurizr https://download.structurizr.com/structurizr-2026.03.06.war
```

Resolves #2428

### DIFF
--- a/src/main/java/dev/jbang/source/ProjectBuilder.java
+++ b/src/main/java/dev/jbang/source/ProjectBuilder.java
@@ -305,7 +305,7 @@ public class ProjectBuilder {
 		}
 
 		Project prj;
-		if (resourceRef.getFile().getFileName().toString().endsWith(".jar")) {
+		if (Project.hasExecutableExtension(resourceRef.getFile())) {
 			prj = createJarProject(resourceRef);
 		} else if (Util.isPreview()
 				&& resourceRef.getFile().getFileName().toString().equals(Project.BuildFile.jbang.fileName)) {
--- a/docs/modules/ROOT/pages/dependencies.adoc
+++ b/docs/modules/ROOT/pages/dependencies.adoc
@@ -116,6 +116,17 @@ jbang eu.maveniverse.maven.plugins:toolbox:0.1.9:cli@fatjar
 
 Without `@fatjar` it would resolve dependencies which would be redundant.
 
+== Dependencies with WAR Files
+
+You can add dependencies to WAR files just like JAR files:
+
+[source, bash]
+----
+$ jbang --deps org.postgresql:postgresql:42.3.1 app.war
+----
+
+WAR files support all dependency-related features including `//DEPS` directives, BOM POMs, and custom repositories.
+
 == Managed dependencies ("BOM POM"'s) [Experimental]
 
 When using libraries and frameworks it can get tedious to manage and update multiple versions.
--- a/docs/modules/ROOT/pages/running.adoc
+++ b/docs/modules/ROOT/pages/running.adoc
@@ -12,6 +12,25 @@ ifdef::env-github[]
 :warning-caption: :warning:
 endif::[]
 
+== Running WAR Files
+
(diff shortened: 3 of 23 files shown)


quintesse:
It will work (the tests show that) but the change has been made too generally. Most of the `isJar()` tests are to test if it's a compilation result of **our** compilation. The output of that is _always_ a JAR, never a WAR. Now, if it's a JAR it will obviously pass the `isExecutableArchive()` test, but so would a .war, even if that should never be possible. So it works, I'll accept, but this  is not really how it should have been fixed.

quintesse (approved):
Acceptable.

maxandersen:
@quintesse can you point to which ones you think was not done right ? no reason to accept something bad but if you already spotted them lets fix them.

src/main/java/dev/jbang/source/ProjectBuilder.java
```diff
@@ -305,7 +305,7 @@ public Project build(ResourceRef resourceRef) {
 		}
 
 		Project prj;
-		if (resourceRef.getFile().getFileName().toString().endsWith(".jar")) {
+		if (Project.hasExecutableExtension(resourceRef.getFile())) {
```
maxandersen, line 308: @quintesse i assume you mean this one that should use "old" comparison?

quintesse: Well no, that one in particular is actually okay. Anyway, I'm really not sure if it's worth the trouble trying to do it in a different way. It works and it doesn't do anything obnoxious so let's just merge it. I can always take a look at it in the future if I think there's an obviously better way.

PR #2433, opened 2026-03-25
feat: implement version pinning for JBang aliases

maxandersen:
Implements `alias:version@catalog` syntax to pin versions at invocation time without modifying catalog files.
                                         
   Try and fix with the syntax suggested in #1979

  **Examples:**
   ```bash
   jbang picocli:4.7.0                    # GAV version replacement
   jbang tool:v1.0@jbangdev               # Git URL branch replacement
   jbang app:2.0@myorg/repo/main          # Catalog ref replacement
   ```

   ## Implementation

   **Version replacement cascade:**
   1. Property replacement (`${jbang.app.version:default}`) - takes precedence

As in, if alias has property replacement we will NOT do automagic replacement because 
can't know user really want the replacement .

   3. Maven GAV replacement (`group:artifact:version` → `group:artifact:newversion`)
   4. Git URL replacement (GitHub/GitLab/Bitbucket branch/tag substitution)
   5. Catalog reference replacement (`alias@org/repo/ref` → `alias@org/repo/version`)
   6. **Hard error** if version specified but no pattern matches.

Did this with the thinking that if version replacement not possible its better we fail - but of course if jbang.app.version is included we 'll have to just trust the catalog writer.

For example:
   ```json
   {
     "aliases": {
       "quarkus": {
         "script-ref": "io.quarkus:quarkus-cli-${jbang.app.version:3.0.0}:3.7.0"
       }
     }
   }
   ```
   Running `jbang quarkus:3.5.0` resolves to `io.quarkus:quarkus-cli-3.5.0:3.7.0`

   ## Feedback Needed

   1. **Hard error vs warning**: Currently throws `ExitException` when version cannot be applied to unknown URL patterns. Is this the right behavior (least surprise) or should it warn and continue?

   2. **GAV detection heuristic**: Uses colon counting (2+ colons = GAV) and dot detection in first segment (groupId pattern). Edge cases where this might fail?

   3. **Catalog reference version pinning**: For registered catalog names (e.g., `tool:1.0@mycatalog` where `mycatalog` is a named catalog), currently errors since we can't modify the catalog URL. Should this work
   differently?

 
   ## Documentation

   Added complete "Version Pinning" section to `docs/modules/ROOT/pages/alias_catalogs.adoc` with syntax, examples, and all replacement strategies.

### DIFF
--- a/src/main/java/dev/jbang/catalog/Alias.java
+++ b/src/main/java/dev/jbang/catalog/Alias.java
@@ -179,7 +179,7 @@ public class Alias extends CatalogItem {
 	public static Alias get(String aliasName) {
 		HashSet<String> names = new HashSet<>();
 		Alias alias = new Alias();
-		Alias result = merge(alias, aliasName, Alias::getLocal, names);
+		Alias result = merge(alias, aliasName, Alias::getLocal, names, null);
 		return result.scriptRef != null ? result : null;
 	}
 
@@ -194,25 +194,37 @@ public class Alias extends CatalogItem {
 	public static Alias get(Catalog catalog, String aliasName) {
 		HashSet<String> names = new HashSet<>();
 		Alias alias = new Alias();
-		Alias result = merge(alias, aliasName, catalog.aliases::get, names);
+		Alias result = merge(alias, aliasName, catalog.aliases::get, names, null);
+		return result.scriptRef != null ? result : null;
+	}
+
+	public static Alias get(Catalog catalog, String aliasName, String requestedVersion) {
+		HashSet<String> names = new HashSet<>();
+		Alias alias = new Alias();
+		Alias result = merge(alias, aliasName, catalog.aliases::get, names, requestedVersion);
 		return result.scriptRef != null ? result : null;
 	}
 
 	private static Alias merge(Alias a1, String name, Function<String, Alias> findUnqualifiedAlias,
-			HashSet<String> names) {
-		// if this is a proper possible GAV, i.e.
-		// io.quarkiverse.mcp:artifact:1.0.0.Beta5@fatjar
-		// don't try interpret it.
+			HashSet<String> names, String requestedVersion) {
+		if (name.startsWith(":") || name.startsWith("@")) {
+			throw new RuntimeException("Invalid alias name '" + name + "'");
+		}
+
 		if (DependencyUtil.looksLikeAPossibleGav(name)) {
 			return a1;
 		}
+
 		if (names.contains(name)) {
 			throw new RuntimeException("Encountered alias loop on '" + name + "'");
 		}
+
 		String[] parts = name.split("@", 2);
-		if (parts[0].isEmpty()) {
+		if (parts[0].isEmpty() || parts[0].startsWith(":")) {
 			throw new RuntimeException("Invalid alias name '" + name + "'");
 		}
+
+		// EXISTING: Look up alias (unchanged)
 		Alias a2;
 		if (parts.length == 1) {
 			a2 = findUnqualifiedAlias.apply(name);
@@ -222,9 +234,29 @@ public class Alias extends CatalogItem {
 			}
 			a2 = fromCatalog(parts[1], parts[0]);
 		}
+
+		// EXISTING: Merge if found
 		if (a2 != null) {
 			names.add(name);
-			a2 = merge(a2, a2.scriptRef, findUnqualifiedAlias, names);
+
+			// If a version was requested for the alias invocation, it must apply to the
+			// final script-ref (GAV/URL/catalog-ref), not to an alias reference. We
+			// detect plain alias references by checking that the scriptRef looks like a
+			// valid alias name and does not contain URL/GAV/catalog special characters.
+			if (requestedVersion != null && Catalog.isValidName(a2.scriptRef)
+					&& !a2.scriptRef.contains(":")
+					&& !a2.scriptRef.contains("/")
+					&& !a2.scriptRef.contains("\\")
+					&& !a2.scriptRef.contains("@")) {
+				throw new ExitException(EXIT_INVALID_INPUT,
+						"Cannot apply version '" + requestedVersion + "' to alias reference '" + a2.scriptRef
+								+ "'. Version can only be applied to the final target. "
+								+ "Use the target alias directly: " + a2.scriptRef + ":" + requestedVersion);
+			}
+
+			a2 = merge(a2, a2.scriptRef, findUnqualifiedAlias, names, requestedVersion);
+
+			// EXISTING: Property merging (all unchanged from original code)
 			String desc = a1.description != null ? a1.description : a2.description;
 			List<String> args = a1.arguments != null && !a1.arguments.isEmpty() ? a1.arguments : a2.arguments;
 			List<String> jopts = a1.runtimeOptions != null && !a1.runtimeOptions.isEmpty() ? a1.runtimeOptions
@@ -263,9 +295,12 @@ public class Alias extends CatalogItem {
 			List<JavaAgent> jags = a1.javaAgents != null && !a1.javaAgents.isEmpty() ? a1.javaAgents : a2.javaAgents;
 			List<String> docs = a1.docs != null && !a1.docs.isEmpty() ? a1.docs : a2.docs;
 			Catalog catalog = a2.catalog != null ? a2.catalog : a1.catalog;
+
+			// EXISTING: Create merged alias
 			return new Alias(a2.scriptRef, desc, args, jopts, srcs, ress, deps, repos, cpaths, props, javaVersion,
 					mainClass, moduleName, copts, nimg, nopts, forceType, ints, jfr, debug, cds, inter, ep, ea, esa,
-					mopts, jags, docs, catalog);
+					mopts,
+					jags, docs, catalog);
 		} else {
 			return a1;
 		}
--- a/src/main/java/dev/jbang/catalog/Alias.java
+++ b/src/main/java/dev/jbang/catalog/Alias.java
@@ -3,10 +3,12 @@ package dev.jbang.catalog;
 import static dev.jbang.ExitException.EXIT_INVALID_INPUT;
 
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Properties;
 import java.util.function.Function;
 
 import org.jspecify.annotations.NonNull;
@@ -16,6 +18,8 @@ import com.google.gson.annotations.SerializedName;
 
 import dev.jbang.ExitException;
 import dev.jbang.dependencies.DependencyUtil;
+import dev.jbang.source.ProjectBuilder;
+import dev.jbang.util.PropertiesValueResolver;
 import dev.jbang.util.Util;
 
 public class Alias extends CatalogItem {
@@ -224,7 +228,8 @@ public class Alias extends CatalogItem {
 			throw new RuntimeException("Invalid alias name '" + name + "'");
 		}
 
-		// EXISTING: Look up alias (unchanged)
+		// Look up alias, expanding any ${jbang.app.version:default} in catalog
+		// references
 		Alias a2;
 		if (parts.length == 1) {
 			a2 = findUnqualifiedAlias.apply(name);
@@ -232,7 +237,9 @@ public class Alias extends CatalogItem {
 			if (parts[1].isEmpty()) {
 				throw new RuntimeException("Invalid alias name '" + name + "'");
 			}
-			a2 = fromCatalog(parts[1], parts[0]);
+			// Expand ${jbang.app.version:default} in catalog reference before fetching
+			String catalogRef = expandVersionProperty(parts[1], requestedVersion);
+			a2 = fromCatalog(catalogRef, parts[0]);
 		}
 
 		// EXISTING: Merge if found
@@ -323,6 +330,32 @@ public class Alias extends CatalogItem {
 	static Catalog findNearestCatalogWithAlias(Path dir, String aliasName) {
 		return Catalog.findNearestCatalogWith(dir, true, true,
 				catalog -> catalog.aliases.containsKey(aliasName) ? catalog : null);
+	}
+
+	/**
+	 * Expand {@code ${jbang.app.version:default}} in a string.
+	 * <p>
+	 * If {@code requestedVersion} is non-null, it's used as the value for
+	 * {@code jbang.app.version}. Otherwise, the default value after the colon is
+	 * used.
+	 *
+	 * @param value            the string potentially containing property references
+	 * @param requestedVersion the version to use, or null to use defaults
+	 * @return the string with properties expanded
+	 */
+	private static String expandVersionProperty(String value, String requestedVersion) {
+		if (value == null || !value.contains("${jbang.app.version")) {
+			return value;
+		}
+		Properties props = ProjectBuilder.getContextProperties(Collections.emptyMap());
+		if (requestedVersion != null) {
+			props.setProperty("jbang.app.version", requestedVersion);
+		}
+		String expanded = PropertiesValueResolver.replaceProperties(value, props);
+		if (!expanded.equals(value)) {
+			Util.verboseMsg("Expanded catalog reference: " + value + " → " + expanded);
+		}
+		return expanded;
 	}
 
 	/**
--- a/docs/modules/ROOT/pages/alias_catalogs.adoc
+++ b/docs/modules/ROOT/pages/alias_catalogs.adoc
@@ -98,6 +98,166 @@ The full format is `<alias>@<user/org>(/repository)(/branch)(~path)` or `<alias>
 
 |====
 
+== Version Pinning
+
+Version pinning allows you to specify a version when running an alias, enabling you to use different versions of the same tool without modifying the catalog definition. This is particularly useful for testing, compatibility checks, or using specific versions of tools.
+
+=== Syntax
+
+The version pinning syntax is: `alias:version@catalog`
+
+Examples:
+
+[source,bash]
+----
+jbang mytool:1.5.0@mycatalog      # Pin to version 1.5.0
+jbang tool:v2.0.0@jbangdev        # Pin to v2.0.0 from jbangdev catalog
+jbang app:1.2.3                   # Pin to 1.2.3 from local catalog
+----
+
+=== How It Works
+
+When you specify a version, JBang applies different replacement strategies depending on the alias's script reference:
+
+==== Maven GAV Coordinates
+
+For aliases that reference Maven artifacts, the version replaces the version component in the GAV coordinates:
+
+[source,json]
+----
+{
+  "aliases": {
+    "picocli": {
+      "script-ref": "info.picocli:picocli-codegen:4.6.0"
+    }
+  }
+}
+----
+
+Running `jbang picocli:4.7.0` resolves to `info.picocli:picocli-codegen:4.7.0`
+
+This works with:
+
+[cols="1,1,1",options="header"]
+|===
+|Alias `script-ref`
+|Invocation
+|Resolved `script-ref`
+
+|`group:artifact:version`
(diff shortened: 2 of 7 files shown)


src/main/java/dev/jbang/catalog/Alias.java
```diff
@@ -16,11 +18,18 @@
...
 import dev.jbang.util.Util;
 
 public class Alias extends CatalogItem {
 	@SerializedName(value = "script-ref", alternate = { "scriptRef" })
 	public final String scriptRef;
+	/**
+	 * Version requested by user via alias:version syntax. Transient because it's
+	 * not stored in catalog files - only used at runtime for version resolution.
+	 */
+	public transient String requestedVersion;
```
maxandersen: this is not needed - just calculate on need to know basis.

wfouche:
@maxandersen , the logic that is required is described in the following Python/Java like pseudocode:

```python3
import org.apache.commons.text.StringSubstitutor

String scriptRef = "camel:4.17.0@apache/camel"

# alias object can be null.
# version is an empty string or a valid version string ("1.0.2")
(Alias alias, String version) = getAlias(scriptRef)

if not alias:
    return

# Set version to the "default-version" property
# if version is empty. "default-version" might
# not be defined in which case it is also set by
# default to the empty string.
if version.length() == 0:
    version = alias.defaultVersion()

if version.length() > 0:
    # replace all instances of "${version}" in string values 
    # inside the alias record with the value of version.toString()

    # Convert the alias object to its JSON representation
    jsonText0 = alias.toJson()

    # Create a version HashMap
    versionMap = {"version": version}

    # Replace all instances of "${version}" with version.toString()
    sub = StringSubstitutor(versionMap)
    jsonText1 = sub.replace(jsonText0)

    # Create a new alias object from the version expanded JSON text
    alias = Alias.fromJson(jsonText1)

# Continue using alias object
```

Using the following `jbang-catalog.json` as a reference. Property `default-version` might not be present, in which case its value would be the empty string when retrieved from the Alias object.

```jsonc
{
  "aliases": {
    "camel": {
      "default-version": "4.18.0",
      "script-ref": "dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java",
      "properties": {
        "camel.jbang.version": "${version}",
        "camel-kamelets.version": "${version}"
      }
    }
  }
}
```

maxandersen:
@wfouche I don't know or follow at all what that code is trying to do.
default-version nor ${version} is not something required in any of the conversations we had.

Please show explicit cases that you believe is failing with this PR?

like:

Given the following catalog:
<example>

These aliases should resolve as this:

a:1.0 -> ...

wfouche:
@maxandersen , after our meeting today, I have the missing context that I needed to understand the reasoning behind the approach you follow in the PR.  I'll test this and see how well it works, and provide feedback.

wfouche:
Using the following jbang-catalog.json file

```json
{
  "aliases" : {
    "junit" : {
      "script-ref" : "org.junit.platform:junit-platform-console-standalone:6.0.3",
      "description" : "Launch the JUnit Platform from the console"
    },
    "cli" : {
      "script-ref" : "org.junit.platform:junit-platform-console-standalone:6.0.3",
      "description" : "Launch the JUnit Platform from the console"
    }
  }
}
```

I ran commands:

```
$ ./build/install/jbang/bin/jbang run cli --version
JUnit Platform Console Launcher 6.0.3
JVM: 21.0.10 (Eclipse Adoptium OpenJDK 64-Bit Server VM 21.0.10+7-LTS)
OS: Linux 6.6.87.2-microsoft-standard-WSL2 amd64

$ ./build/install/jbang/bin/jbang run cli:6.0.2 --version
JUnit Platform Console Launcher 6.0.2
JVM: 21.0.10 (Eclipse Adoptium OpenJDK 64-Bit Server VM 21.0.10+7-LTS)
OS: Linux 6.6.87.2-microsoft-standard-WSL2 amd64
```

Versioning of GAVs works! However, this feels a bit unreal to me.  :-)

I'll test a script next.

wfouche:
I've created three scripts:

```
$ jbang run hello/hello.java
Hello World
$ jbang run hello/1.0.0/hello.java
Hello World, 1.0.0
$ jbang run hello/1.0.1/hello.java
Hello World, 1.0.1
```

*jbang-catalog.json*

```json
{
  "aliases" : {
    "h": {
      "script-ref": "./hello/${jbang.app.version:}/hello.java"
    }
  }
}
```

*Testing versioning functionality*

```
$ ../build/install/jbang/bin/jbang run h
Hello World

$ ../build/install/jbang/bin/jbang run h:1.0.0
Hello World, 1.0.0
```

@maxandersen , it works very well. I like it. All that I found that is missing, is that JBang do not pass property `jbang.app.version` on to the script using -Djbang.app.version=1.0.0 , we need this for the solution to be complete.

Updated test program:

```java

import java.lang.System;

public class hello {
        System.out.println("Hello World, 1.0.0");
        String version = System.getProperty("jbang.app.version");
        if (version != null) {
            System.out.println("JBang App Version: " + version);
        } else {
            System.out.println("Property 'jbang.app.version' is not defined.");
        }
    }
}
```

I also tested with `script-ref` set to "./hello/${jbang.app.version:1.0.0}/hello.java"

```
$ ../build/install/jbang/bin/jbang run h
Hello World, 1.0.0

$ ../build/install/jbang/bin/jbang run h:1.0.1
Hello World, 1.0.1
```

The following alias could not be resolved:

```
    "camel": {
      "script-ref": "camel@apps/camel/${jbang.app.version:4.16.0}/jbang-catalog.json"
    }
```

```
$ ../build/install/jbang/bin/jbang run camel
[jbang] [ERROR] Unknown catalog 'apps/camel/${jbang.app.version:4.16.0}/jbang-catalog.json'
```

wfouche:
In summary, there are two issues to resolve:

- `jbang.app.version` should be set in the app JVM environment.
- support versioned script-refs pointing to `jbang-catalog.json` files

With these two issues resolved, I think we'll have a great and useful versioning solution.

maxandersen:
> * `jbang.app.version` should be set in the app JVM environment.

Why/what is the usecase for the app to know what jbang.app.version was used on the command line? That is not really the business of jbang to tell an app what version it is ? There is *zero* guarantees that version relates to the application version.
 
 I get it could be nice to have but it is not something to rely and thus not a fan of exposing it. 

> * support versioned script-refs pointing to `jbang-catalog.json` files

I assume you mean this error:

```
../build/install/jbang/bin/jbang run camel
[jbang] [ERROR] Unknown catalog 'apps/camel/${jbang.app.version:4.16.0}/jbang-catalog.json'
```

can you give me a repo/branch with what you have setup because this direct pointing to a jbang-catalog.json seems irrelevant...I assume its more about that the jbang.app.version is not replaced in case of no alias version specified?

wfouche:
> jbang.app.version should be set in the app JVM environment.

I've changed my mind. Don't set the property in the app JVM.

wfouche:
> can you give me a repo/branch with what you have setup 

https://github.com/wfouche/jbang/tree/dev/version-tests/tests

```json
{
    "hw": {
      "script-ref": "hello@./apps/hello/1.0.0/jbang-catalog.json"
    },
    "hw2": {
      "script-ref": "hello@./apps/hello/${jbang.app.version:1.0.0}/jbang-catalog.json"
    }
}
```

*Issue 1*

`hw` is resolved and works, `hw2` is not resolved.

wfouche:
*Issue 2*

```json
    "camel": {
      "script-ref": "./CamelJBang.java"
    }
```

`jbang run camel version` works, `jbang run camel:4.17.0 version` fails with error message:

- [jbang] [ERROR] Cannot apply version '4.17.0' to script-ref './CamelJBang.java'. No recognizable version pattern found. Supported patterns: Maven GAV, GitHub/GitLab/Bitbucket URLs, or use ${jbang.app.version:default} in the alias definition.
- It is perfectly valid to have a script-ref for a versioned alias to be just "./CamelJBang.java".

maxandersen:
'It is perfectly valid to have a script-ref for a versioned alias to be just "./CamelJBang.java".'

I dont see how. That is not a versioned alias. So no way to apply version Alias to it.

wfouche:
> I dont see how. That is not a versioned alias. So no way to apply version Alias to it.

The "how" is that we just use the version number specified for the alias in jbang.app.version to find the correct //DEPS to load.

```java

//JAVA 21+
//REPOS central=https://repo1.maven.org/maven2,apache-snapshot=https://repository.apache.org/content/groups/snapshots/

//DEPS org.apache.camel:camel-bom:${jbang.app.version:4.18.0}@pom
//DEPS org.apache.camel:camel-jbang-core:${jbang.app.version:4.18.0}
//DEPS org.apache.camel.kamelets:camel-kamelets:${jbang.app.version:4.18.0}

package main;

import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;

public class CamelJBang {

        CamelJBangMain.run(args);
    }

}
```

This does not work, but it should:
- $JBANG run camel:4.17.0 version

This works (simulating what we expect from alias versioning):
- $JBANG run -Djbang.app.version=4.17.0 camel version

maxandersen:
Ok that is not apps jvm environment. That's when the script it build into a jar. 

You'll need to pass the value via properties as you showed before. The app version is not automatically "inherited"

wfouche:
I hope we can change this. We don't necessarily want a separate Java file for each version of an app.

Below is how JBang support for Camel is defined today. How do you propose that this be changed so that

- jbang app install camel:4.17.0@apache/camel

will work with a release of JBang that support alias versioning?

https://camel.apache.org/manual/camel-jbang.html

```json
{
	"catalogs": {},
	"aliases": {
	  "camel": {
		"script-ref": "dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java",
		"description": "Run Apache Camel routes easily"
	  }
	}
}
```

```java

//JAVA 21+
//REPOS central=https://repo1.maven.org/maven2,apache-snapshot=https://repository.apache.org/content/groups/snapshots/
//DEPS org.apache.camel:camel-bom:${camel.jbang.version:4.18.0}@pom
//DEPS org.apache.camel:camel-jbang-core:${camel.jbang.version:4.18.0}
//DEPS org.apache.camel.kamelets:camel-kamelets:${camel-kamelets.version:4.18.0}

package main;

import org.apache.camel.dsl.jbang.core.commands.CamelJBangMain;

/**
 * Main to run CamelJBang
 */
public class CamelJBang {

        CamelJBangMain.run(args);
    }
}
```

maxandersen:
This is the chicken-egg thing I talked about - the alias itself is not versioned its the "thing" the alias points to that needs to be versioned; and when you point to a file relatively in a main branch...you are implicitly saying "give me whatever is the latest".

So you need to pick it you want whatever is latest or a specific named version.

So your alias could point to: `camel@apache/camel/camel-4.18.1` and use the versioned name of the jbang catalog 
or use `org.apache.camel:camel-jbang-core:4.18.0` + extra reps and use the versioned maven GAV you are after.

 Do note that in the first one the version number is actually "camel-4.18.1" which might not be what you are after but that is the version name used.
 
 You can do `camel@apache/camel/${jbang.app.version:main}` but then you'll need to use `camel:camel-4.18.1@apache/camel`
 
But now you are having the main repo define an alias in the same place as what this alias points to - it is not that clean and im not seeing how/what we can change without making it harder for everything else.

wfouche:
> the alias itself is not versioned its the "thing" the alias points to that needs to be versioned

Agreed. 

The I don't understand why we don't view CamelJbang.java file as a "template" file to be used to create a version specific file to be compiled by javac. The alias script-ref points to "template" file CamelJBang.java which is parameterized with property jbang.app.version. Just update JBang set set property "jbang.app.version" in the JBang JVM (not the App JVM) and then the property will be available when "template" file CamelJBang.java is processed by JBang.

The locking mechanism that you submitted in another PR I guess will be needed to ensure that apps installed in this way are not updated.

maxandersen:
It already is treated as a "template file" - it's done when you build. 

The thing is that I don't see a consistent way we can be ok applying version to script files without having a version context. 

That might be we need to such thing as default version but given i already showed you multiple ways to do it without it could you tell me why those are not applicable/sufficient ?

wfouche:
My concern is that versioning Camel/JBang is a bit cumbersome, but it works.

To support versions 4.17.0, 4.18.0 and 4.18.1, I need to create three distinct CamelJBang.java files:

image

They are all the same except for the hard-coded version number they each contain.


*camel/4.17.0/CamelJBang.java*
```java
....

//DEPS org.apache.camel:camel-bom:4.17.0@pom
//DEPS org.apache.camel:camel-jbang-core:4.17.0
//DEPS org.apache.camel.kamelets:camel-kamelets:4.17.0

....
```

*camel/4.18.0/CamelJBang.java*
```java
....

//DEPS org.apache.camel:camel-bom:4.18.0@pom
//DEPS org.apache.camel:camel-jbang-core:4.18.0
//DEPS org.apache.camel.kamelets:camel-kamelets:4.18.0

....
```

*camel/4.18.1/CamelJBang.java*
```java
....

//DEPS org.apache.camel:camel-bom:4.18.1@pom
//DEPS org.apache.camel:camel-jbang-core:4.18.1
//DEPS org.apache.camel.kamelets:camel-kamelets:4.18.0

....
```

*jbang-catalog.json*

```json
    "camel": {
      "script-ref": "./camel/${jbang.app.version:4.18.1}/CamelJBang.java"
    }
```

The 4.18.1 CamelJBang.java example shows that one version number is not always sufficient to meet alll //DEPS requirements. In this case camel-kamelets did not have a 4.18.1 version, and had to remain at 4.18.0 although the other camel JARs do use version 4.18.1. This is a convincing counter example, that shows that versioning as it is currently implemented provide the flexibility required to address issues like this.

But it makes me wonder if one should not provide "hints" to the user of what "versions" are available. To that end a "versions" property that can be queried would be really useful. 

```json
    "camel": {
      "script-ref": "./camel/${jbang.app.version:4.18.1}/CamelJBang.java",
      "versions": {
        "4.18.1": "Recommended stable version",
        "4.17.0": "Previous stable version",
        "4.9.0": "Legacy version"
      }
    }
```

> My concern is that versioning Camel/JBang is a bit cumbersome, but it works.

Getting back to my opening sentence. No concerns anymore.

wfouche:
The only remaining issue I'm aware of is fixing:

* [jbang] [ERROR] Unknown catalog 'apps/camel/${jbang.app.version:4.16.0}/jbang-catalog.json'

And then, optionally, adding a "versions" property.  :-)

maxandersen:
> To support versions 4.17.0, 4.18.0 and 4.18.1, I need to create three distinct CamelJBang.java files:

you write as you create these in the same branch - but you don't have to; you already have that file in multiple versions in the tagged versions at https://github.com/apache/camel/commits/main/dsl/camel-jbang/camel-jbang-main/dist/CamelJBang.java

If anything I would just make an alias and skip cameljbang.java completely. It is not doing anything besides calling a main method.

maxandersen:
> ```json
>   "camel": {
>       "script-ref": "./camel/${jbang.app.version:4.18.1}/CamelJBang.java",
>       "versions": {
>         "4.18.1": "Recommended stable version",
>         "4.17.0": "Previous stable version",
>         "4.9.0": "Legacy version"
>       }
>     }
> ```

I'm not a superfan, but I can see how it could be useful for querying/documentation. Open separate issue and lets see.

> To support versions 4.17.0, 4.18.0 and 4.18.1, I need to create three distinct CamelJBang.java files:

No, I don't see why you need to do that. Just use property replacements or different aliases. Your own suggested ways for it required these too. 

But again, as I said multiple times you are trying to do something I consider extremely niche of enabling multiple version support for something (i.e. CamelJBang) by putting a source file inside the same source repo of the thing you try to version. That is always going to be messy.

maxandersen:
> * [jbang] [ERROR] Unknown catalog 'apps/camel/${jbang.app.version:4.16.0}/jbang-catalog.json'

what specifcially generated that error?

wfouche:
> what specifcially generated that error?

```json
    "hw": {
      "script-ref": "hello@./apps/hello/1.0.0/jbang-catalog.json"
    },
    "hw2": {
      "script-ref": "hello@./apps/hello/${jbang.app.version:1.0.0}/jbang-catalog.json"
    }
```

*Works*
- jbang run hw 

*Fails*
- jbang run hw2
`[jbang] [ERROR] Unknown catalog './apps/hello/${jbang.app.version:1.0.0}/jbang-catalog.json'`

wfouche:
> [jbang] [ERROR] Unknown catalog './apps/hello/${jbang.app.version:1.0.0}/jbang-catalog.json'

@maxandersen , this is the only issue that needs to be resolved before this feature can be released. :-)

wfouche:
During the past week I started using `ttcli` (https://github.com/wimdeblauwe/ttcli/)  and of course wanted to run it using JBang.

* https://github.com/wimdeblauwe/ttcli/issues/253

Having to create aliases for the previous stable and current version of ttcli, reminded me of this open PR.

Once alias versioning is supported, the ttcli jbang-catalog.json could be simplified to just:

```
{
  "aliases": {
    "ttcli": {
      "script-ref":  "https://github.com/wimdeblauwe/ttcli/releases/download/${jbang.app.version:1.12.0}/ttcli-${jbang.app.version:1.12.0}.jar",
      "java": "25+",
      "runtime-options": [
        "--enable-native-access=ALL-UNNAMED"
      ]
    }
  }
}
```

maxandersen:
## Updates

Rebased from main and added fixes for the remaining issue:

### 1. Fix: `${jbang.app.version}` expansion in catalog references

The property expansion now happens **before** catalog lookup, fixing the error:
```
[jbang] [ERROR] Unknown catalog './apps/hello/${jbang.app.version:1.0.0}/jbang-catalog.json'
```

With this fix, the following should now work:
```json
{
  "hw": { "script-ref": "hello@./apps/hello/${jbang.app.version:1.0.0}/jbang-catalog.json" }
}
```

- `jbang run hw` → expands to `hello@./apps/hello/1.0.0/jbang-catalog.json`
- `jbang run hw:2.0.0` → expands to `hello@./apps/hello/2.0.0/jbang-catalog.json`

### 2. Security: Version string validation

Added validation to prevent injection attacks via version strings:
- Blocks path separators (`/`, `\`) - prevents `tool:../../etc/passwd@catalog`
- Blocks URL schemes (`://`) - prevents `tool:file:///etc/passwd@catalog`
- Blocks control characters (null bytes, etc.)
- Valid semver versions still work: `1.0.0-SNAPSHOT`, `v1.2.3-alpha+build.456`, `camel-4.18.1`

### 3. Fixed imports after rebase

`ExitException` moved from `dev.jbang.cli` to `dev.jbang` in main - updated imports accordingly.

---

@wfouche Could you verify this fixes the catalog reference expansion issue you reported? 🙏

maxandersen:
Updated: Removed the version string validation I added earlier - after discussion, it was security theater since:
- Catalog authors already control all paths/URLs
- Users can already access any catalog with `tool@evil.com` directly  
- JBang isn't a security sandbox

The fix now just contains the functional change: `${jbang.app.version}` expansion in catalog references before lookup.

wfouche:
> @wfouche Could you verify this fixes the catalog reference expansion issue you reported?

HI @maxandersen ,

I've tested the latest fixes and it is a solid step forward. But I have encountered a new issue.

Unzip the attached archive (jbang-catalog.zip) into an empty folder (test).

set JBC=~/<path to PR 2433 jbang script, created using ./gradlew clean installDist>
* $JBC  version  -> 0.141.0.18
* $JBC run hw ->Hello World! 1.0.0
* $JBC run hw2 ->Hello World! 1.0.1
* $JBC run hw2:1.0.0 ->this run produces the error shown below.

```
$JBC run hw2:1.0.0
[jbang] [ERROR] Cannot apply version '1.0.0' to script-ref 'src/hello.java'. No recognizable version pattern found. Supported patterns: Maven GAV, GitHub/GitLab/Bitbucket URLs, or use ${jbang.app.version:default} in the alias definition.
[jbang] Run with --verbose or -x for more details. The --verbose or -x must be placed before the jbang command. I.e. jbang --verbose run [...]
```

[jbang-catalog.zip](https://github.com/user-attachments/files/31354052/jbang-catalog.zip)

PR #2448, opened 2026-04-19
build: Build with Java 25 toolchain

maxandersen:
This is with intent of having faster builds an potentially use markdown in javadoc; but also to allow use of newer gradle plugins to handle git versioning that does not break with git worktrees.

## Summary

Update the build toolchain from Java 11 to Java 25 while keeping the same bytecode targets (`--release 8` for main code, `--release 9` for multi-release, `--release 11` for tests).

## Changes

- **`build.gradle`** — Toolchain `languageVersion` 11 → 25
- **`.github/actions/shared-build-setup/action.yml`** — Default `java-version` 11 → 25 (avoids Gradle auto-provisioning overhead)
- **`.github/workflows/step-ci-build.yml`** — Unit test job now uses `matrix.java-version` (currently `[11]`) with `-PtestJavaVersion`, matching the integration test pattern

## Notes

- `--release 8` still works on JDK 25 but emits a deprecation warning ("source value 8 is obsolete and will be removed in a future release")
- Test execution versions are unchanged: unit tests run on JDK 11, integration tests on 8/11/17/21/25
- The unit test matrix can easily be expanded to more Java versions in the future

### DIFF
--- a/.github/actions/shared-build-setup/action.yml
+++ b/.github/actions/shared-build-setup/action.yml
@@ -9,7 +9,7 @@ inputs:
   java-version:
     description: The Java version to use
     required: false
-    default: '11'
+    default: '25'
 outputs:
   github-short-sha:
     description: The short github.sha
--- a/.github/workflows/step-ci-build.yml
+++ b/.github/workflows/step-ci-build.yml
@@ -118,26 +118,25 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        java-version: [11]
         os: *os_matrix
     steps:
       - uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e #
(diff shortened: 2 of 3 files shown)


maxandersen:
@quintesse please take a look to see if i've missed something here and if you see any problems with this one.

quintesse (approved):
LGTM. I'd even say that we could make the version used for tests much newer. I mean if we build with 25 anyway, why not test with 25 as well? Is there a reason we shouldn't do that?

maxandersen:
We can definitely add 25 in matrix if missing. 

We test on 11 at runtime to ensure it doesn't accidentally rely on multijars different behaviours. 

For integration tests we can run test with 25 and just run with different external vm.

PR #2449, opened 2026-04-23
fix: cache clear now deletes dep files from disk, not just JSON

### LINKED ISSUE #475: jbang cache clear does not nuke dependency cache
run jbang cache clear and ~/.jbang/cache/depcache* still exist and has content. It should be empty.﻿

Bhumika-SN:
## Problem
`jbang cache clear` was not actually clearing the dependency cache from disk.

The `DependencyCache.clear()` method only set the in-memory `depCache` to null,
but never deleted the actual cached files on disk. This means running
`jbang cache clear` would leave all downloaded JARs and the dependency_cache.json
file still present on disk.

## Fix
Updated `DependencyCache.clear()` to also:
- Delete the `dependency_cache.json` file from disk
- Delete the `depcache` directory containing actual downloaded JARs

## Related Issue
Fixes #475

### DIFF
--- a/src/main/java/dev/jbang/Cache.java
+++ b/src/main/java/dev/jbang/Cache.java
@@ -5,6 +5,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 
 import dev.jbang.cli.ExitException;
+import dev.jbang.dependencies.DependencyCache;
 import dev.jbang.devkitman.Jdk;
 import dev.jbang.devkitman.JdkManager;
 import dev.jbang.util.JavaUtil;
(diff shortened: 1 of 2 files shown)


maxandersen:
this looks right - but would be nice with an actual test to verify the behavour

quintesse (changes requested):
Nope! This is _not_ the place to do this!
The `clear()` method is only used to clean the in-memory representation of the dependencies and is only used by tests.

This should be moved to `Cache.clearCache()`

maxandersen:
Hmm. Not sure what that comment was partially coming from :) 

Thanks for checking @quintesse !

Bhumika-SN:
Hi @quintesse and @maxandersen,

I've moved the fix to `Cache.clearCache()` where it belongs, and 
reverted `DependencyCache.clear()` back to only clearing the 
in-memory cache.

Could you please re-review? Thank you!

Bhumika-SN:
Hi @maxandersen and @quintesse,

I've moved the fix to `Cache.clearCache()` as suggested and reverted 
`DependencyCache.clear()` back to only clearing in-memory cache.

The CI failures appear to be in Java 25 integration tests and smoke 
tests which seem pre-existing and unrelated to this change. Could 
you confirm?

Thank you!

PR #2450, opened 2026-04-23
fix: use USERPROFILE on Windows for non-ASCII username support

### LINKED ISSUE #1110: Powershell installer handles non-ASCII characters in home directory badly
**Describe the bug**
If your Windows home directory contains "special" characters, you need to manually fix your `PATH` variable after the Powershell installer finishes.

**To Reproduce**

1. Create a new Windows user with a username that contains non-ASCII characters. Example: `LászlóvandenHoek`
2. Log in as that user and attempt to install JBang in a PowerShell window: `iex "& { $(iwr https://ps.jbang.dev) } app setup"` (as per https://www.jbang.dev/download/)
3. Close and open new Powershell window as instructed. Attempt to use `jbang`. Doesn't work, command not found.
4. Inspect the new PATH variable, notice it has been prepended with: `C:\Users\Lßszl≤vandenHoek\.jbang\bin` instead of the correct `C:\Users\LászlóvandenHoek\.jbang\bin`.

**Expected behavior**
The path is set correctly by the installer, and jbang works without any manual fixes.

**JBang version**
Not related to jbang version, only to installer.

**Additional context**
After fixing the path manually, jbang does work. Thanks for making such a wonderful tool!

Bhumika-SN:
## Problem
On Windows, users with non-ASCII characters in their username (common
in many countries including India, China, Eastern Europe) experience
broken PATH setup and jbang failing completely.

Root cause: Java's `System.getProperty("user.home")` can return a
corrupted path on Windows when the username contains non-ASCII characters.

## Fix
On Windows, prefer `USERPROFILE` environment variable over `user.home`
as it always returns the correct path regardless of username characters.

## Related Issue
Fixes #1110

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -3,6 +3,7 @@ package dev.jbang.cli;
 import static dev.jbang.util.JavaUtil.defaultJdkManager;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -480,7 +481,18 @@ class AppSetup extends BaseCommand {
 			}
 		}
 		if (!cmd.isEmpty()) {
-			System.out.println(cmd);
+			if (Util.isWindows() && Util.getShell() == Util.Shell.powershell) {
+				try {
+					Path tmpFile = Files.createTempFile("jbang-setup-", ".ps1");
+					tmpFile.toFile().deleteOnExit();
+					Files.write(tmpFile, cmd.getBytes(StandardCharsets.UTF_8));
+					System.out.println(tmpFile.toAbsolutePath().toString());
+				} catch (IOException e) {
+					System.out.println(cmd);
+				}
+			} else {
+				System.out.println(cmd);
+			}
 			return EXIT_EXECUTE;
 		} else {
 			return EXIT_OK;
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -3,7 +3,6 @@ package dev.jbang.cli;
 import static dev.jbang.util.JavaUtil.defaultJdkManager;
 
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -481,18 +480,7 @@ class AppSetup extends BaseCommand {
 			}
 		}
 		if (!cmd.isEmpty()) {
-			if (Util.isWindows() && Util.getShell() == Util.Shell.powershell) {
-				try {
-					Path tmpFile = Files.createTempFile("jbang-setup-", ".ps1");
-					tmpFile.toFile().deleteOnExit();
-					Files.write(tmpFile, cmd.getBytes(StandardCharsets.UTF_8));
-					System.out.println(tmpFile.toAbsolutePath().toString());
-				} catch (IOException e) {
-					System.out.println(cmd);
-				}
-			} else {
-				System.out.println(cmd);
-			}
+			System.out.println(cmd);
 			return EXIT_EXECUTE;
 		} else {
 			return EXIT_OK;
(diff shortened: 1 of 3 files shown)


src/main/java/dev/jbang/dependencies/DependencyCache.java
```diff
@@ -140,6 +140,22 @@ public static List<ArtifactInfo> findDependenciesByHash(String depsHash) {
 
 	public static void clear() {
 		depCache = null;
+		// Also clear the dependency cache JSON file from disk
+		try {
```
maxandersen: This does not seem related to this issue ?

maxandersen (changes requested):

Bhumika-SN:
Hi @maxandersen, thank you for the review!

You're right — the DependencyCache.java changes belong to a separate 
PR for the cache clear issue. I have rebased this branch to only 
include the USERPROFILE fix for the non-ASCII username issue.

maxandersen:
it makes sense afaics; @quintesse do you see any issue using this as default on windows as it seems the jdk is not changing it afaics.

quintesse:
I don't know, this is really hard to test so how do we know this works?

If you look at the original issue I did quite a bit of research all those years ago and it definitely seemed to be more than just a user home issue, it was any path that had weird characters in it.

But right now I can't even reproduce the issue, not with Java 8, nor with 17.

But of course I set my Windows default codepage to UTF8 some time ago (and no, I'm not setting it back just to test :-) too much shit can go wrong on Windows)

Not sure if I still have VM around with Windows on it, I doubt it.

Bhumika-SN:
Hi @quintesse and @maxandersen,

You're right that the issue was deeper than just `user.home`. 

Based on quintesse's research in the original issue, the root cause is 
that Java's stdout encoding gets corrupted when PowerShell captures it.

I've added a second fix: instead of printing the PowerShell setup command 
to stdout, we now write it to a UTF-8 encoded temp file and print the 
file path instead. This avoids the encoding corruption entirely.

Could you please re-review? Thank you!

quintesse:
Now that it isn't only the user home, which is hard to test, it would be great if you were able to add a test that shows the issue is fixed. I'll try to come up with one myself, but I first have to find time, so if you were able to provide tests it would speed things up considerably :-)

src/main/java/dev/jbang/cli/App.java
```diff
@@ -480,7 +481,18 @@ public static int setup(boolean withJava, boolean force, boolean chatty) {
 			}
 		}
 		if (!cmd.isEmpty()) {
-			System.out.println(cmd);
+			if (Util.isWindows() && Util.getShell() == Util.Shell.powershell) {
+				try {
+					Path tmpFile = Files.createTempFile("jbang-setup-", ".ps1");
+					tmpFile.toFile().deleteOnExit();
+					Files.write(tmpFile, cmd.getBytes(StandardCharsets.UTF_8));
```
maxandersen: this does seem right - shouldn't need to write to a tempfile to get UTF_8 string.

quintesse: I guess you mean "does _not_ seem right"? (I agree)

maxandersen: Yes. Not right.

maxandersen:
I pushed a follow-up commit (f0b5c3d6) that attempts to address the full issue. Still needs review.

**What changed:**

1. **Extracted `Util.getUserHomeDir()`** — single shared method that prefers `USERPROFILE` on Windows, falling back to `user.home`. This fixes the non-ASCII corruption in one place instead of scattering the fix.

2. **Applied it to all `user.home` call sites** — `Settings.getConfigDir()`, `Settings.getLocalRootDir()`, and `App.getHome()` all use the new helper now.

3. **Reverted the temp file approach in `App.setup()`** — the temp file broke the EXIT_EXECUTE/eval contract. When `setup()` returns exit code 255, the wrapper scripts (`jbang.ps1` line 120) do `Invoke-Expression "& $output"` on whatever was printed to stdout. Writing a file path instead of a command would break `jbang app setup` entirely.

**Still needed:**
- Tests (as @quintesse requested)
- Rebase onto current main (there are merge conflicts)
- If the PowerShell encoding issue is still a concern, that should be fixed in `jbang.ps1` (e.g. `[Console]::OutputEncoding`) rather than changing the Java/wrapper contract

quintesse:
The test case would also be good to be able to reproduce, because so far I haven't been able to reproduce myself (I already created a user with special characters in their name years ago, just for the exact purpose of testing JBang, but perhaps the characters I chose aren't special enough).

PR #2451, opened 2026-04-24
fix: use full executable path for editors on Windows in Git BashFix/edit intellij windows gitbash

### LINKED ISSUE #1340: edit: IntelliJ found in PATH won't start on Windows
If you use `jbang edit` on Windows and you have IntelliJ's `idea` on your path and you tell JBang that you want to use it then nothing happens.

This is because for Windows JBang is correctly looking for things like `idea.exe`, `idea.cmd` and `idea.bat`, but what it finally tries to execute is simply `idea`. Normally that would work, and it does for CMD and PowerShell, but it seems that when using something like Git Bash it won't actually work unless you specifically tell it to use `idea.bat`.

Bhumika-SN:
## Problem
Fixes #1340

On Windows with Git Bash, `jbang edit` finds IntelliJ's `idea` on PATH
correctly but then tries to execute just `"idea"` instead of the full
path like `"idea.bat"`.

`Util.searchPath()` correctly resolves the full path with extension,
but `findEditorsOnPath()` was discarding that result and returning
just the plain editor name.

## Fix
On Windows, use the full resolved path returned by `Util.searchPath()`
instead of the plain editor name, so Git Bash can execute it correctly.

### DIFF
--- a/src/main/java/dev/jbang/Settings.java
+++ b/src/main/java/dev/jbang/Settings.java
@@ -56,7 +56,12 @@ public class Settings {
 		if (jd != null) {
 			dir = Paths.get(jd);
 		} else {
-			dir = Paths.get(System.getProperty("user.home")).resolve(".jbang");
+			// On Windows, USERPROFILE is more reliable than user.home for non-ASCII
+			// usernames
+			String home = Util.isWindows()
+					? System.getenv().getOrDefault("USERPROFILE", System.getProperty("user.home"))
+					: System.getProperty("user.home");
+			dir = Paths.get(home).resolve(".jbang");
 		}
 
 		if (init)
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -379,7 +379,17 @@ public class Edit extends BaseCommand {
 	}
 
 	private static List<String> findEditorsOnPath() {
-		return Arrays.stream(knownEditors).filter(e -> Util.searchPath(e) != null).collect(Collectors.toList());
+		return Arrays.stream(knownEditors)
+			.filter(e -> Util.searchPath(e) != null)
+			.map(e -> {
+				Path found = Util.searchPath(e);
+				// On Windows, use the full path with extension so Git Bash can execute it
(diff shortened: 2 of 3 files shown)


src/main/java/dev/jbang/cli/Edit.java
```diff
@@ -379,7 +379,17 @@ private static void setupEditor(Path editorBinPath, Path dataPath) throws IOExce
 	}
 
 	private static List<String> findEditorsOnPath() {
-		return Arrays.stream(knownEditors).filter(e -> Util.searchPath(e) != null).collect(Collectors.toList());
+		return Arrays.stream(knownEditors)
+			.filter(e -> Util.searchPath(e) != null)
```
maxandersen, line 383: this is going to 2x the calls to searchPath() that does not seem right.

src/main/java/dev/jbang/Settings.java
```diff
@@ -56,7 +56,12 @@ public static Path getConfigDir(boolean init) {
 		if (jd != null) {
 			dir = Paths.get(jd);
 		} else {
-			dir = Paths.get(System.getProperty("user.home")).resolve(".jbang");
+			// On Windows, USERPROFILE is more reliable than user.home for non-ASCII
+			// usernames
+			String home = Util.isWindows()
```
maxandersen, line 61: this overlaps with #2450

src/main/java/dev/jbang/dependencies/DependencyCache.java
```diff
@@ -140,6 +140,22 @@ public static List<ArtifactInfo> findDependenciesByHash(String depsHash) {
 
 	public static void clear() {
 		depCache = null;
+		// Also clear the dependency cache JSON file from disk
+		try {
+			Path cacheFile = Settings.getCacheDependencyFile();
```
maxandersen, line 145: this does not seem related to this issue?

src/main/java/dev/jbang/cli/Edit.java
maxandersen: can't we simply return the right name ? idea.bat ? or exec it shell?

PR #2455, opened 2026-05-02
fix: prevent duplicate PATH entries when running setup multiple times on WindowsFix/duplicate path entries on windows setup

### LINKED ISSUE #940: When calling setup multiple times in the same shell, multiple path entries are created
I made a mistake when setting up jshell, not closing and re-opening the entire Windows Terminal window. I just closed tabs, instead:
https://github.com/jbangdev/jbang/issues/939

This resulted in multiple identical path entries in my environment variables:

image

It can be reproduced like this:

image

I.e. on a fresh environment (removing all the path entries, and removing `rm -Rf ~/.jbang`), run this  command several times:

```
curl -Ls https://sh.jbang.dev | bash -s - app setup
```

I wonder if a previous setup could be detected even if the terminal window hasn't been closed and re-opened yet...?

Bhumika-SN:
## Problem
Fixes #940

Running `jbang app setup` multiple times on Windows creates duplicate 
PATH entries in the user's permanent PATH (registry).

`needsSetup()` only checks the current session's PATH, so in a new 
shell it always thinks setup is needed and prepends the path again 
without checking if it already exists in the permanent user PATH.

## Fix
Before adding to the permanent user PATH, check if the binDir is 
already present using PowerShell's `-notlike` operator. Only add 
if not already present.

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -427,11 +427,12 @@ class AppSetup extends BaseCommand {
 						"[EnvironmentVariableTarget]::User)";
 			}
 			if (force || needsSetup()) {
-				// Create the command to change the user's PATH
-				String newPath = binDir + ";";
-				env += " ; [Environment]::SetEnvironmentVariable('Path', '" + newPath + "' + " +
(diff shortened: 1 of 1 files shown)


quintesse (approved):
LGTM

PR #2460, opened 2026-05-09
fix: avoid double searchPath call, use full path for editors on Windows

### LINKED ISSUE #1340: edit: IntelliJ found in PATH won't start on Windows

Bhumika-SN:
## Problem
Fixes #1340

On Windows with Git Bash, `jbang edit` finds IntelliJ's `idea` on 
PATH correctly but then tries to execute just `"idea"` instead of 
the full path like `"idea.bat"`.

`findEditorsOnPath()` was calling `searchPath()` twice — once in 
`filter` and once in `map` — which is inefficient.

## Fix
Use a single `map + filter` to call `searchPath()` only once.
On Windows, return the full resolved path with extension so Git 
Bash can execute it correctly.

### DIFF
--- a/src/main/java/dev/jbang/cli/Edit.java
+++ b/src/main/java/dev/jbang/cli/Edit.java
@@ -11,6 +11,7 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.*;
 import java.util.*;
+import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -266,7 +267,7 @@ public class Edit extends BaseCommand {
 					+ System.getenv("GITPOD_WORKSPACE_URL") + "#" + projectPathString + "\n\n");
 		} else {
 			List<String> optionList = new ArrayList<>();
-			optionList.add(editor);
+			optionList.add(resolveEditorCommand(editor));
 			optionList.add(projectPathString);
 			optionList.addAll(additionalFiles);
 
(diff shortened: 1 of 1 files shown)


maxandersen:
Reworked the approach here. The core issue is that Git Bash on Windows can't execute bare command names like `idea` - it needs the actual filename with extension (`idea.bat`).

The original PR fixed this in `findEditorsOnPath()` only, but the same bug hits when someone does `jbang config set edit.open=idea` or `--open idea` - that path bypasses discovery entirely and goes straight to execution.

So instead of a Windows-specific check in one place, this:

1. Makes `findEditorsOnPath()` return the resolved filename from `searchPath` (e.g. `idea.bat` on Windows, `idea` on Linux/Mac) - no platform branching needed since searchPath already does the right thing per OS.

2. Adds `resolveEditorCommand()` which does the same resolution at execution time in `openEditor()`, so both the auto-discovery and the explicit `--open`/config paths are covered. It skips resolution if the editor is already a path (has separators).

One fix in the execution path rather than patching each entry point separately.

quintesse:
> The core issue is that Git Bash on Windows can't execute bare command names like `idea`

I don't understand this comment? You can definitely just type `idea` in Git Bash, so I'm guessing this is about some specific corner case?

maxandersen:
That's because via git bash you run via git bash. It does the magical flip from idea to idea.bat. 

That's basically what this pr mimics because when you run in git bash, call Jbang - bash is no longer around to do that trick. 

Thus we teach jbangs lookup that it should look for .bat/.exe too. 

And since search path looks in PATH running idea.bat should work.

PR #2461, opened 2026-05-09
feat: Add download retry support to startup scripts

### LINKED ISSUE #2459: Set retry options for curl in the JBang startup scripts
Right now a single error will fail the installation of Java which is an issue for CI that depends on JBang.

Simply adding something like `--retry 5` would be enough.

JBang can also use `wget` which by default already uses retries. Supposedly to make curl behave like wget we could use:

```
--retry 20 --retry-all-errors --retry-max-time 0
```

maxandersen:
Fixes #2459

## Problem

A single transient network error during JDK or JBang bootstrap download fails the entire setup, which is particularly painful in CI environments.

## Solution

Adds a `JBANG_DOWNLOAD_RETRY` environment variable (default: `5`) that controls retry attempts for downloads across all platforms:

| Platform | Tool | Retry mechanism |
|----------|------|----------------|
| Linux/macOS (bash) | `curl` | `--retry N --retry-delay 1` |
| Linux/macOS (bash) | `wget` | `--tries=N --waitretry=1` |
| Windows (PowerShell) | `Invoke-WebRequest` | `Invoke-Download` retry loop with 1s backoff |
| Windows (CMD) | delegates to PS1 | Inherits PowerShell retry behavior |

### Usage

```bash
# Increase retries for flaky CI networks
export JBANG_DOWNLOAD_RETRY=10

# Disable retries
export JBANG_DOWNLOAD_RETRY=0
```

## Changes

- **`src/main/scripts/jbang`** — Added retry flags to `curl` and `wget` in the `download()` function
- **`src/main/scripts/jbang.ps1`** — Added `Invoke-Download` function with retry loop; replaced raw `Invoke-WebRequest` calls (PS5-compatible, no dependency on PS7's `-MaximumRetryCount`)
- **`src/test/java/dev/jbang/cli/TestScriptRetry.java`** — 7 unit tests validating retry support in script contents
- **`docs/modules/ROOT/pages/installation.adoc`** — New "Download Retries" section
- **`docs/modules/ROOT/pages/troubleshooting.adoc`** — Updated "Network Issues" section

## Summary by CodeRabbit

* **New Features**
  * Startup scripts now support configurable download retries, retry delays, and customizable download URL via environment variables.

* **Documentation**
  * Installation guide adds "Download Retries" and "Startup Script Environment Variables".
  * Troubleshooting updated to cover transient download failures and retry guidance.

* **Tests**
  * New functional tests verify startup-script retry behavior against flaky download scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbangdev/jbang/pull/2461)


maxandersen:
the tests are bogus - not testing it works; just testing the scripts has retry logic.

maxandersen:
updated tests to be "real" via wiremock test that fails so can see retry work.

maxandersen:
ok, now tests are real, use progressive backoff by default, env vars added for user and test control + documented the various env vars.

@quintesse wdyt? and should all those vars be treated equally supported ? :)

quintesse:
>wdyt?

LGTM 👍 

> and should all those vars be treated equally supported ? :)

Not sure what you mean by this?

maxandersen:
> > wdyt?
> 
> LGTM 👍
> 
> > and should all those vars be treated equally supported ? :)
> 
> Not sure what you mean by this?

Wether all the env vars in the script that this PR add should be considered all equally supported/stable.

quintesse:
Well I think most of them are okay. Perhaps the USE_NATIVE one because it's supposed to be for internal testing?

maxandersen:
@quintesse any idea why powershell specifically would fail here ..just on java 11 ?

quintesse:
No idea really

maxandersen:
## CI failure analysis & fix

The failures are **not Java 11-specific** — unit tests only run on Java 11 in the CI matrix, so that's the only place they surface. The same tests would fail on any Java version.

### Two bugs found and fixed

**Bug 1: `break` exits with code 0 (fixed in 46bbafdd)**

The PowerShell script uses `break` to bail out on download failures. However, `break` outside of a loop/switch terminates the script with **exit code 0** — so download failures were silently reported as success. The two failing tests (`downloadFailsWhenRetriesExhausted` and `downloadFailsWithZeroRetries`) correctly expected a non-zero exit code but the script never provided one.

**Fix:** Replaced `break` with `exit 1` in the two download error paths (JBang archive and JDK archive).

**Bug 2: stale `$err` variable (fixed in same commit)**

The old code captured the exception into `$err` via `catch { $ok=$false; $err=$_ }`. After the refactor to `Invoke-Download`, the `$err` variable is never set, so the line `[Console]::Error.WriteLine($err)` printed nothing useful. Removed the stale reference.

maxandersen:
llms to the rescue again - dumb logic error :) tests passes. lets merge and things should hopefully still work :)

PR #2479, opened 2026-05-27
fix: warn when --sources/--files are used with executable archives

### LINKED ISSUE #2363: Running a jar file silently ignores --sources
If a JAR file is run together with a `--sources` option, then the source files are silently ignored. For example:

```
$ jbang run --sources HttpUser.java jarGAV  config.json
```

Either the source files should be compiled and added to the classpath, or a warning should be displayed informing the JBang user that in this particular context source files are ignored.

maxandersen:
## Summary
- When running a JAR/WAR file with `--sources` or `--files`, these options were silently ignored
- Now emits a `[WARN]` message and skips source/resource resolution for executable archives
- Warning is placed in `ProjectBuilder.updateProject()` so it fires before source file resolution (which would otherwise fail for nonexistent files)

Fixes #2363

## Test plan
- [x] Added `testJarWithSourcesWarning` test in `TestRun.java`
- [x] Existing `testHelloWorldJar` still passes
- [ ] Manual: `jbang run --sources Foo.java some.jar` prints warning to stderr

## Summary by CodeRabbit

* **New Features**
  * Added warnings when users attempt to use `--sources` or `--files` options with executable archives (jar/war files), as these options are not supported for this archive type.

* **Tests**
  * Added test coverage for executable archive handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbangdev/jbang/pull/2479?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


maxandersen:
@quintesse wdyt ? going for warn about files /sources not being honored in this pr - mainly because I can't see a (good) usecase where we want to "reopen" the jar and add the resources....do you know of one or is this good enough

PR #2480, opened 2026-05-28
feat: support release tag names in JBANG_DOWNLOAD_VERSION (early-access)

maxandersen:
## Why

We now have a perpetual early-access release URL: https://github.com/jbangdev/jbang/releases/download/early-access/

This PR makes it reachable via the existing `JBANG_DOWNLOAD_*` variables, so users can opt in with a one-liner instead of pinning a full `JBANG_DOWNLOAD_URL`.

## What

Extend the existing `JBANG_DOWNLOAD_VERSION` semantics rather than introduce a new variable:

- Numeric values (existing behaviour) keep getting the customary `v` prefix — `0.120.0` → `.../download/v0.120.0/jbang.tar`.
- Non-numeric values are used as a release tag name as-is — `early-access` → `.../download/early-access/jbang.tar`.

So this Just Works:

```bash
export JBANG_DOWNLOAD_VERSION=early-access
```

`JBANG_DOWNLOAD_BASEURL` and `JBANG_DOWNLOAD_URL` continue to behave as before for mirrors / air-gapped setups.

## Changes

- `src/main/scripts/jbang` — bash `case` decides whether to prepend `v`.
- `src/main/scripts/jbang.ps1` — same logic via `-match '^[0-9]'`.
- `src/main/scripts/jbang.cmd` — unchanged (delegates downloads to `jbang.ps1` per AGENTS.md).
- `docs/modules/ROOT/pages/installation.adoc` — new subsection "Downloading a specific version or the early-access build" and updated description in the env-var table.
- `src/test/java/dev/jbang/cli/TestScriptDownloadVersion.java` — new WireMock-backed tests running the real bash and pwsh scripts, asserting both `/download/v0.120.0/jbang.{tar,zip}` and `/download/early-access/jbang.{tar,zip}` are requested. Uses `assumeTrue` to skip when bash / pwsh aren't available, matching the existing `TestScriptRetry` style.

## Summary by CodeRabbit

* **New Features**
  * `JBANG_DOWNLOAD_VERSION` now supports non-numeric release tags (e.g., `early-access`) in addition to numeric versions. Numeric versions automatically receive a `v` prefix in download URLs, while non-numeric tags are used as-is.

* **Documentation**
  * Updated installation documentation describing version selection behavior.

* **Tests**
  * Added comprehensive tests validating version download URL construction.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbangdev/jbang/pull/2480?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -444,6 +444,23 @@ By default JBang downloads itself from GitHub. You can override the download URL
 export JBANG_DOWNLOAD_URL=https://internal-mirror.example.com/jbang/jbang.tar
 ----
 
+=== Downloading a specific version or the early-access build
+
+Set `JBANG_DOWNLOAD_VERSION` to pick a specific release. Numeric values are resolved to the matching `vX.Y.Z` GitHub release; any other value is used as a release tag name as-is. This means you can opt-in to the perpetual early-access build with a simple:
+
+[source,bash]
+----
+export JBANG_DOWNLOAD_VERSION=early-access
+----
+
+which resolves to https://github.com/jbangdev/jbang/releases/download/early-access/ . Equivalent forms:
+
+[source,bash]
+----
+export JBANG_DOWNLOAD_VERSION=0.120.0     # -> .../download/v0.120.0/jbang.tar
+export JBANG_DOWNLOAD_VERSION=early-access # -> .../download/early-access/jbang.tar
+----
+
 == Version Check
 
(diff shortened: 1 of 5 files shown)


quintesse (dismissed):
LGTM

maxandersen:
i changed the expression sligthly so it will treat any 1.2.3-adsff as a tag with v  but anything else should just be fetched directly.

+ added text and full changelog to the earlyaccess so all changes since last real release is shown.

Issue #2485, opened 2026-05-29
docs: --manifest option missing description in export fatjar/maven/native man pages


maxandersen:
The `--manifest` option is documented in three `export` subcommand man pages but its description block is blank, leaving users without any guidance on file format or behavior. This also pre-dates the aesh migration (#2453) — same on `main` today.

## Affected pages (on `main`)

- `docs/modules/cli/pages/jbang-export-fatjar.adoc` line 92–93
- `docs/modules/cli/pages/jbang-export-maven.adoc` line 99–100
- `docs/modules/cli/pages/jbang-export-native.adoc` line 92–93

In each, `*--manifest*=_<...>_::` is followed by a blank line where the description should be.

## Root cause

`src/main/java/dev/jbang/cli/BuildMixin.java:25`:

```java
@OptionGroup(name = "manifest")
public Map<String, String> manifestOptions;
```

The `@OptionGroup` has no `description=` attribute, so the man-page generator emits an empty description line. (Same pattern on the picocli-based main and the aesh-migration PR.)

## Fix

Add a `description = "…"` to the `@OptionGroup` in `BuildMixin.java` explaining the expected key/value format (e.g. `--manifest Main-Class=foo.Bar`) and that entries are merged into the generated JAR/POM manifest. Regenerate the docs.

## Originally spotted in PR #2453

- https://github.com/jbangdev/jbang/pull/2453#discussion_r3248980225 (fatjar)
- https://github.com/jbangdev/jbang/pull/2453#discussion_r3248980233 (maven)
- https://github.com/jbangdev/jbang/pull/2453#discussion_r3248980241 (native)

Those review threads will be resolved as the issue exists on `main` and is tracked here.

PR #2506, opened 2026-06-04
docs: add description to --manifest option (#2485)

stalep:
The @OptionGroup for manifest had no description attribute, causing blank description lines in the generated CLI docs for all commands that include build options (export fatjar/maven/native, run, build, etc.).

Add description explaining the --manifestKey=Value syntax and regenerate all CLI reference pages.

### DIFF
--- a/docs/modules/cli/pages/jbang-alias-add.adoc
+++ b/docs/modules/cli/pages/jbang-alias-add.adoc
@@ -8,12 +8,8 @@ jbang alias add -- Add alias for script reference.
 
 [source]
 ----
-jbang alias add [-hnigox] [-s=<sources>] [--files=<files>] [-T=<source-type>] [--jsh] [--catalog=<catalog>] [-j=<java>] [-m=<main>] [--module] [-C=<compile-option>] [--manifest<key>=<value>] [--[no-]integrations] [-D<key>=<value>] [--deps=<deps>] [--repos=<repos>] [--cp=<cp>] [--ignore-transitive-repositories] [-N=<native-option>] [-R=<runtime-option>] [--jfr] [-d=<debug>] [--enableassertions] [--enablesystemassertions] [--javaagent=<javaagent>] [--[no-]cds] [-f=<file>] [--description=<description>] [--name=<name>] [--force] [--enable-preview] [--docs=<docs>] [--config=<config>] [--insecure] [--verbose] [--quiet] [--fresh] [<scriptOrFile>] [<params>]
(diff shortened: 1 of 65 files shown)


src/main/java/dev/jbang/cli/BuildMixin.java
```diff
@@ -26,7 +26,7 @@ public class BuildMixin {
 	@OptionList(shortName = 'C', name = "compile-option", description = "Options to pass to the compiler")
 	public List<String> compileOptions;
 
-	@OptionGroup(name = "manifest")
+	@OptionGroup(name = "manifest", description = "Add entry to the JAR manifest (use --manifestKey=Value, e.g. --manifestFoo=Bar)")
```
maxandersen: this `--manifestKey=Value` does work on main but its not how it works on old jbang main.

its supposed to be `--mainifest=Key=Value`

And yes, realized we dont have tests for this and why it was uncaught :/

Is this expected aesh behavior or bug?

stalep: This is how aesh works currently, let me create an issue.

stalep: fixed in the next version of aesh; https://github.com/aeshell/aesh/issues/496

docs/modules/cli/pages/jbang-alias-add.adoc
```diff
@@ -135,19 +132,19 @@ Path to config file to be used instead of the default
...
-*--quiet*::
+*--[no-]quiet*::
 jbang will be quiet, only print when error occurs.
 
-*-o*, *--offline*::
+*-o*, *--[no-]offline*::
 Work offline. Fail-fast if dependencies are missing.
 
-*--fresh*::
+*--[no-]fresh*::
```
maxandersen, line 144: remind me why it is that we have turned verbose, quiet, offline, fresh, preview and stacktrace as negatable?

They weren't negatable in picocli as they don't read meaningful and always are some you enable.

Is there a downside to mark these as negatable = false (the default)

maxandersen: Though I'm not really hard set on it - having --no-... available for all booleans are at least consistent but it is a change so just  want to grok if we are doing it intentionally or because aesh has a difference/limit here?

also, I notice `jbang --verbose=definitetlynottrue` does not fail with invalid argument like picocli does - is that aeash treating everything but "true" as false? @stalep ?

maxandersen (approved):

Issue #2481, opened 2026-05-29
possible issue with shadow merge of jars


maxandersen:
possible issue spotted in #2453 but issue exist in main already

https://github.com/jbangdev/jbang/pull/2453#discussion_r3248691947

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/main/java/dev/jbang/cli/Export.java` around lines 469 - 477, The merge of
META-INF/services entries in Export (in the block handling
zipEntry.getName().startsWith("META-INF/services/")) uses
FileOutputStream.getChannel().transferFrom with position 0 which can misbehave
on some platforms; change the write to append explicitly by opening the target
with an append-capable channel or by seeking the FileChannel to its size before
transferring, or simpler: read the zip entry input stream and write its bytes
using FileOutputStream (opened with append=true) or Files.write with
StandardOpenOption.APPEND so the service file contents are reliably appended to
outFile instead of risking overwrite; update the code in Export where
readableByteChannel and fileOutputStream are used to perform an explicit append
operation.
```

quintesse:
Simplest fix seems to be:

```
try (ReadableByteChannel readableByteChannel = Channels.newChannel(zipFile.getInputStream(zipEntry));
        FileOutputStream fileOutputStream = new FileOutputStream(outFile.toFile())) {
    fileOutputStream.getChannel().transferFrom(readableByteChannel,  outFile.toFile().length(), Long.MAX_VALUE);
}
```

PR #2508, opened 2026-06-04
fix: use stream copy instead of transferFrom for service file merging.

stalep:
… (#2481)

FileChannel.transferFrom with position 0 could overwrite existing content instead of appending on some platforms, even when the FileOutputStream was opened with append=true. Replace with plain stream copy which correctly appends to the output file.


maxandersen:
seems fine to me but any way to make test that shows what it actually fixes?

stalep:
> seems fine to me but any way to make test that shows what it actually fixes?

Added that now!

maxandersen (approved):
perfect! verified test fails without it.

Issue #559, opened 2020-11-28
//MDEPS for adding things to --module-path ? 


maxandersen:
wanted to try out nashorn 15 and seems it is impossible to run without putting it on the module-path.

Given this code:

```
///usr/bin/env jbang "$0" "$@" ; exit $?
//JAVA 15+
//DEPS info.picocli:picocli:4.5.0
//DEPS org.openjdk.nashorn:nashorn-core:15.0
import picocli.CommandLine;
import picocli.CommandLine.Command;
import picocli.CommandLine.Parameters;

import javax.script.ScriptEngine;
import javax.script.ScriptEngineManager;
import javax.script.ScriptException;
import java.util.concurrent.Callable;

@Command(name = "js", mixinStandardHelpOptions = true, version = "js 0.1",
        description = "js made with jbang")
class js implements Callable<Integer> {

    @Parameters(index = "0", description = "The script to evaluate", defaultValue = "print('Hello, World!');")
    private String script;

    public static void main(String... args) {
        int exitCode = new CommandLine(new js()).execute(args);
        System.exit(exitCode);
    }

    @Override
    public Integer call() throws Exception { // your business logic goes here...
        // create a script engine manager
        ScriptEngineManager factory = new ScriptEngineManager();
        // create a Nashorn script engine
        ScriptEngine engine = factory.getEngineByName("nashorn");
        // evaluate JavaScript statement
        try {
            engine.eval(script);
        } catch (final ScriptException se) { se.printStackTrace(); }
        return CommandLine.ExitCode.OK;
    }
}
```

jbang generates something like:

```
[...]

PR #2511, opened 2026-06-05
fix: JDK26: Put modules required by JavaFX modules on the module-path

koppor:
This adresses https://github.com/jbangdev/jbang/issues/559#issuecomment-4610215903

IMHO required by [JDK-8359760](https://bugs.java.com/bugdatabase/view_bug?bug_id=JDK-8359760). See 

Code 100% authored and tested w/ JabRef by Claude Code. Guided by me.

---

## Problem

JBang only promotes jars whose module name starts with `javafx` to the
`--module-path`; everything else stays on the class-path. That breaks when a
JavaFX module `requires` a non-`javafx` module shipped as a separate artifact:
`javafx.web` requires `jdk.jsobject` (org.openjfx:jdk-jsobject), which then
stays on the class-path and the boot layer fails:

      java.lang.module.FindException: Module jdk.jsobject not found, required by javafx.web

## Fix

After detecting the JavaFX modules, follow their `requires` edges transitively
and put every required module that is part of the resolved artifacts on the
module-path as well. `--add-modules` is unchanged (the JavaFX roots pull in
`jdk.jsobject` transitively once it is on the module-path).

## Verification

- `./gradlew spotlessApply build` green; `TestRun.testJavaFX*` (4) green.
- installDist run of an unmodified JabRef `jabkit` JBang launcher on JDK 26:

### DIFF
--- a/src/main/java/dev/jbang/dependencies/ModularClassPath.java
+++ b/src/main/java/dev/jbang/dependencies/ModularClassPath.java
@@ -5,13 +5,17 @@ import static dev.jbang.Settings.CP_SEPARATOR;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.codehaus.plexus.languages.java.jpms.JavaModuleDescriptor;
@@ -93,18 +97,52 @@ public class ModularClassPath {
 
 				resolvePathsResult.getPathElements().forEach((key, value) -> pathElements.put(key.getPath(), value));
 
-				pathElements.forEach((k, v) -> {
-					if (v != null && v.name() != null && v.name().startsWith(JAVAFX_PREFIX)) {
-						// only JavaFX jars are required in the module-path
-						modulePaths.add(k);
-					} else {
-						// classpathElements.add(k);
+				// Index the resolved modules by name so we can follow `requires` edges.
+				Map<String, String> moduleNameToPath = new HashMap<>();
+				Map<String, JavaModuleDescriptor> moduleNameToDescriptor = new HashMap<>();
+				pathElements.forEach((path, descriptor) -> {
+					if (descriptor != null && descriptor.name() != null) {
+						moduleNameToPath.put(descriptor.name(), path);
+						moduleNameToDescriptor.put(descriptor.name(), descriptor);
 					}
 				});
 
+				// Start from the JavaFX modules and transitively collect everything they
+				// require. This way modules like `jdk.jsobject` (required by `javafx.web`)
+				// also end up on the module-path instead of the class-path, where the boot
+				// layer would not find them. See https://github.com/jbangdev/jbang/issues/559
+				Set<String> requiredModules = new HashSet<>();
+				Deque<String> toVisit = new ArrayDeque<>();
+				moduleNameToDescriptor.keySet()
+					.stream()
+					.filter(name -> name.startsWith(JAVAFX_PREFIX))
+					.forEach(toVisit::add);
+				while (!toVisit.isEmpty()) {
+					String name = toVisit.poll();
+					if (!requiredModules.add(name)) {
+						continue;
+					}
+					JavaModuleDescriptor descriptor = moduleNameToDescriptor.get(name);
+					if (descriptor != null) {
+						descriptor.requires()
+							.stream()
+							.map(JavaModuleDescriptor.JavaRequires::name)
+							.filter(moduleNameToDescriptor::containsKey)
+							.forEach(toVisit::add);
+					}
+				}
+
+				// Put the JavaFX modules and their required modules on the module-path.
+				requiredModules.stream()
+					.map(moduleNameToPath::get)
+					.filter(Objects::nonNull)
+					.forEach(modulePaths::add);
+
 				if (!modulePaths.isEmpty()) {
 					commandArguments.add("--module-path");
-					String modulePath = String.join(File.pathSeparator, modulePaths);
+					String modulePath = modulePaths.stream()
+						.distinct()
+						.collect(Collectors.joining(File.pathSeparator));
 					commandArguments.add(modulePath);
 				}
 
--- a/src/main/java/dev/jbang/dependencies/ModularClassPath.java
+++ b/src/main/java/dev/jbang/dependencies/ModularClassPath.java
@@ -5,17 +5,13 @@ import static dev.jbang.Settings.CP_SEPARATOR;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Deque;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.codehaus.plexus.languages.java.jpms.JavaModuleDescriptor;
@@ -29,6 +25,9 @@ import dev.jbang.util.Util;
 
 public class ModularClassPath {
 	static final String JAVAFX_PREFIX = "javafx";
+	// Required by `javafx.web`; removed from the JDK in 26 and now shipped as
+	// `org.openjfx:jdk-jsobject`.
+	static final String JDK_JSOBJECT_MODULE = "jdk.jsobject";
 
 	private final List<ArtifactInfo> artifacts;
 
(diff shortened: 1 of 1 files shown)


maxandersen:
got a test that fails otherwise ? 

I assume its only on some versions of java and jfx deps?

quintesse:
@koppor can you check if adding the `//MODULE` directive to your script fixes the problem?

Because this is closely related to the [//MDEPS discussion](https://github.com/jbangdev/jbang/issues/559) and I promoted the idea there that instead of trying to patch the module vs classpath we should just be telling users to switch to using modules.

koppor:
> @koppor can you check if adding the `//MODULE` directive to your script fixes the problem?

I pointed Claude first to this. And he didn't succeed. Had sth. to do w/ ordering of CLI arguments.

Will instruct him to provide an MWE with //MODULE as comment(s) here.

koppor:
## Without modules in a non-module project

First, the experiment with an MWE without modules - I think, this should work - or should I activate modules?

MWE - jfxweb.java

```java
//JAVA 26+
//DEPS org.openjfx:javafx-web:26.0.1
//DEPS org.openjfx:jdk-jsobject:26

import javafx.scene.web.WebView;

public class jfxweb {
	  System.out.println("WebView class: " + WebView.class.getName());
  }
}
```

Output

```terminal
$ jbang jfxweb.java
[jbang] Building jar for jfxweb.java...
Error occurred during initialization of boot layer
java.lang.module.FindException: Module jdk.jsobject not found, required by javafx.web
```

### With module

```java
//JAVA 26+
//MODULE
//DEPS org.openjfx:javafx-web:26.0.1
//DEPS org.openjfx:jdk-jsobject:26

package mwe;

import javafx.scene.web.WebView;

public class jfxwebmod {
	  System.out.println("WebView class: " + WebView.class.getName());
  }
}
```

Output:

```terminal
$ jbang jfxwebmod.java
[jbang] Building jar for jfxwebmod.java...
C:\TEMP\jfxwebmod.java:9: error: package javafx.scene.web is not visible
import javafx.scene.web.WebView;
                   ^
  (package javafx.scene.web is declared in module javafx.web, but module jfxwebmod does not read it)
1 error
[jbang] Run with --verbose or -x for more details. The --verbose or -x must be placed before the jbang command. I.e. jbang --verbose run [...]
```

Much more work seems to be requried to get this running.

---

Claude's longer reasoning - but I think the above says it all

**Disclaimer** Maybe, Claude is totally wrong here, but maybe there is a grain of truth there. Let's see...

For JabRef's `jabkit` (a headless CLI that only drags JavaFX in *transitively* via `jablib` — the user never asked for JavaFX), `//MODULE` hits three separate blockers:

1. **Default package + transitive reads.** The launcher is in the default package, so `//MODULE` first fails with `Module code cannot work with the default package, adding a 'package'
statement is required`. After adding a `package`, compilation fails because the `//SOURCES` import classes from modules pulled in transitively — `afterburner.fx`, `jul.to.slf4j`,
`org.tinylog.api` — none of which the module reads, so each has to be re-declared as an explicit `//DEPS`.

2. **`//MODULE` alone still launches from the class-path.** Without also passing `--module`, `runAsModule` stays false (`JarCmdGenerator`), so JavaFX is on the module-path while
`jdk.jsobject` is on the class-path — same `FindException` as MWE 1.

3. **`jbang run --module` puts *all* deps on the module-path** and then dies on non-modular jars in the tree:

 ```
 java.lang.module.FindException: Unable to derive module descriptor for .../fastparse_2.13-2.3.3.jar
 Caused by: java.lang.IllegalArgumentException: fastparse.2.13: Invalid module name: '2' is not a Java identifier
 ```

 (same for `latex2unicode_2.13`). "Just use modules" can't work once the dependency graph contains any non-modularizable jar, and real-world trees usually do.

### Why this PR instead

It doesn't try to move everything to modules. It stays scoped to the JavaFX case jbang already special-cases: it just keeps the modules the auto-detected JavaFX modules *already*
`requires` (like `jdk.jsobject`) on the same module-path jbang is already building for JavaFX. `--add-modules` is left unchanged.

koppor:
Claude and me also tried fixing this purely from the script via `//RUNTIME_OPTIONS` (no patched jbang) — adding the module, add-reads, etc. **None work.** Tested on the MWE (same root cause):

| `//RUNTIME_OPTIONS` | result |
|---|---|
| `--add-reads javafx.web=ALL-UNNAMED` | `FindException: Module jdk.jsobject not found, required by javafx.web` |
| `--add-modules jdk.jsobject` | `Module jdk.jsobject not found` |
| `--add-modules ALL-MODULE-PATH` | `...not found, required by javafx.web` |
| `--module-path <jsobject> --add-modules jdk.jsobject` | `Module jdk.jsobject not found` |
| `--upgrade-module-path <jsobject>` | `...not found, required by javafx.web` |

**Why no `//RUNTIME_OPTIONS` fix exists with stock jbang:**

1. `javafx.web requires jdk.jsobject`, so `jdk.jsobject` must be a **module on the module-path**. This is resolved at boot, before anything else runs.
2. `--add-reads` only affects **readability**, applied *after* resolution. A missing required module makes boot fail first — reads can't conjure a module into existence.
3. `--add-modules` (including `ALL-MODULE-PATH`) pulls roots from the **module-path**, and `jdk.jsobject` isn't on it.
4. `--module-path` set via `//RUNTIME_OPTIONS` is **overridden**: jbang appends its own JavaFX `--module-path` *last*, and Java takes the **last** `-p` only (verified separately — two
`-p` flags = last wins, not merged). There's no way to extend jbang's single `-p` string from the script.
5. `--upgrade-module-path` only *replaces* upgradeable JDK modules; it can't introduce a new module to satisfy a `requires` → `not found`.

The only seam that would work — getting `jdk.jsobject` into jbang's single JavaFX `--module-path` string — isn't reachable from the script. So the realistic fixes are: this PR, or full
module mode (`//MODULE` + `--module`), which is broken here by JabRef's non-modular transitive jars.

maxandersen:
Let me summarize the findings @koppor has - which matches mine when I added jfx support to jbang years back.

The issue is that JDK's module mechanism is *invasive* onto users application. You have a perfectly fine and working script/application that just works with a simple classpath ...but because on of your transitive deps are using modules you are *forced* to deal with encapsulation no matter if you care or not. 

Thats why jbang will detect javafx when present and then enable module path transparently - its mechanically easily to do and user don't have to care.  

Yes, I think we should probably just have an explicit flag to do what implicit add of javafx does - which is to enable "i'm a classpath app but if some dependency is a module - lets add them to module path"

src/main/java/dev/jbang/dependencies/ModularClassPath.java
```diff
@@ -93,18 +97,52 @@ public List<String> getAutoDectectedModuleArguments(@NonNull Jdk jdk) {
 
 				resolvePathsResult.getPathElements().forEach((key, value) -> pathElements.put(key.getPath(), value));
 
-				pathElements.forEach((k, v) -> {
-					if (v != null && v.name() != null && v.name().startsWith(JAVAFX_PREFIX)) {
-						// only JavaFX jars are required in the module-path
-						modulePaths.add(k);
```
koppor, line 99: Other fix could be to check if name is javafx.web and then additionally add `jdk.jsobject` to the module path. No need for self-crafted dependency inference.

koppor:
Nice TL;DR from Claude:

>  Why JDK25 worked, JDK26 broke: javafx.web always requires jdk.jsobject (true since v20). On JDK ≤25, jdk.jsobject shipped inside the JDK → boot layer found it on the system
  module path. JDK26 removed jdk.jsobject from the JDK; openjfx now ships it as the separate org.openjfx:jdk-jsobject artifact (artifact only exists for v26+). jbang drops
  that jar on the class-path → boot layer can't find it → FindException. Your "most simple patch" intuition is correct: the only thing that changed is this one module moving
  out of the JDK.


```
  Non-javafx.* modules required by any JavaFX module:

  ┌───────────────────────────────────────┬─────────────────┬────────────────────────────────────────────┐
  │            Required module            │       By        │                    Kind                    │
  ├───────────────────────────────────────┼─────────────────┼────────────────────────────────────────────┤
  │ java.* (base, desktop, xml, net.http) │ various         │ JDK built-in                               │
  ├───────────────────────────────────────┼─────────────────┼────────────────────────────────────────────┤
  │ jdk.jfr (requires static)             │ javafx.base     │ JDK built-in                               │
  ├───────────────────────────────────────┼─────────────────┼────────────────────────────────────────────┤
  │ jdk.xml.dom                           │ javafx.web      │ JDK built-in                               │
  ├───────────────────────────────────────┼─────────────────┼────────────────────────────────────────────┤
  │ jdk.jsobject                          │ javafx.web only │ external artifact org.openjfx:jdk-jsobject │
  └───────────────────────────────────────┴─────────────────┴────────────────────────────────────────────┘
```

src/main/java/dev/jbang/dependencies/ModularClassPath.java
koppor: Done — switched to special-casing `jdk.jsobject`, dropped the dependency inference (a2ef1773).

Claude checked the `module-info` of every `org.openjfx` JavaFX artifact (v20→v26): `jdk.jsobject` is the **only** non-`javafx.*`, separately-shipped module any JavaFX module requires (and only `javafx.web` requires it). Everything else they pull in (`java.*`, `jdk.jfr`, `jdk.xml.dom`) is bundled in the JDK. That also explains the JDK 25→26 regression: `jdk.jsobject` used to live in the JDK and was removed in 26, so openjfx now ships it as `org.openjfx:jdk-jsobject`. Hard-coding the one module is enough.

koppor:
I now streamlined the fix - just handels the `jsobject` case - and not more. -- Long reasoning at https://github.com/jbangdev/jbang/pull/2511#discussion_r3369806813

quintesse:
It's nice that it works, but I'm really interested in knowing if the "i'm a classpath app but if some dependency is a module - lets add them to module path" option would work for this or not (and work in general, meaning that we'd remove the special casing of JavaFx as well).

We could do it as a proof of concept first, for example by checking a System property, let's say `jbang.hybrid.module.resolve=true` and then scanning the dependencies for modules and moving them (and their transitive dependencies) to the module path,.

If we can show that it works we can think about how to properly expose it to users.

Edit: btw, I'm not against merging this simplified PR in the meantime!

quintesse (approved):
LGTM

maxandersen:
> It's nice that it works, but I'm really interested in knowing if the "i'm a classpath app but if some dependency is a module - lets add them to module path" option would work for this or not (and work in general, meaning that we'd remove the special casing of JavaFx as well).

+1 or just have presence of jfx on classpath triggers that hybrid option so jfx continues to "just work".
 
> We could do it as a proof of concept first, for example by checking a System property, let's say `jbang.hybrid.module.resolve=true` and then scanning the dependencies for modules and moving them (and their transitive dependencies) to the module path,.

+1

quintesse:
> +1 or just have presence of jfx on classpath triggers that hybrid option so jfx continues to "just work".

+1 That's also a good idea, we could have a list of known modules that have to be on the module path (for now only javafx, but could be extended).

Edit: although, on second thought, we should limit that list to things that used to work on the classpath but no longer do, like JavaFx. In the cases that something is "known to work only with modules" we should probably expect that people know that and use `//MODULE`.

quintesse:
@koppor with respect to your tests: I haven't been very clear to be honest. When I said that we should try `//MODULE` I didn't actually explain that this isn't a silver bullet or anything. Right now the module support in JBang is extremely limited, there's just this `//MODULE` directive and the _only_ thing it does is:

- put every dependency on the module path
- creates a module-info file (see [`ModuleUtil.generateModuleInfo()`](https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/util/ModuleUtil.java#L72)) 

And that creates a file with:

- the name of the module
- the name of the package
- a list of dependencies

And those dependencies are:

- the names of all JDK modules
- the names of all direct //DEPS that are modules

And there is currently no way to affect the generated contents of that file (so no way to added `opens` or whatever).

If as a user you need to change something? Then you add your own module-info file!
It will work perfectly fine if you add a `//FILES module-info.java` , but of course then you have to do everything yourself and things can no longer fit in a single source file.

Edit: Of course we should think of ways to JBang-ify things to make them easier, but for now this is what exists.

PS: one of the things I thought of back then was that, instead of adding a bunch of new directives that just copy what module-info does, perhaps we could have "inline files". A bit like bash can do with `cat <<<END ...... END`. Where the contents between two special directives would be included as a file. That way the module-info could be inlined into the source code itself. But I never followed up on that because it seemed to me it would make the IDE plugins much more complex (and they are already lacking sufficient attention as it is).

Edit2: just imagine something like:

```

//DEPS org.hibernate.orm:hibernate-core:jar:7.4.0.Final
//DEPS ... etc

/***FILE module-info.java
open module com.example {
    requires org.hibernate.orm.core;
    requires jakarta.persistence;
    requires org.junit.jupiter.api;
    opens undertow.core/io.undertow to org.jboss.logging;
}
***/

void main() {
   // Do stuff here
}
```

koppor:
> Edit: Of course we should think of ways to JBang-ify things to make them easier, but for now this is what exists.
[...]
> If as a user you need to change something? Then you add your own module-info file!
> It will work perfectly fine if you add a `//FILES module-info.java` , but of course then you have to do everything yourself and things can no longer fit in a single source file.

Yeah, I think, this is the solution. I will try to include the  [existing module-info.java](https://github.com/JabRef/jabref/blob/42aa85998920ebdaa2cf21e53ae7fa0ecfcba6a4/jabkit/src/main/java/module-info.java#L1), too. - I didn't do, because I have had too much issues with the module system.

(Internal link: Current state of the JBang script: https://github.com/JabRef/jabref/blob/42aa85998920ebdaa2cf21e53ae7fa0ecfcba6a4/.jbang/JabKitLauncher.java#L20)

**Edit**

It's more complex it seems

```diff
diff --git a/.jbang/JabKitLauncher.java b/.jbang/JabKitLauncher.java
index 45248c977e..df7adfac68 100755
--- a/.jbang/JabKitLauncher.java
+++ b/.jbang/JabKitLauncher.java
@@ -41,8 +41,9 @@
 //SOURCES ../jabkit/src/main/java/org/jabref/toolkit/commands/Pseudonymize.java
 //SOURCES ../jabkit/src/main/java/org/jabref/toolkit/commands/Search.java
 
+//FILES module-info.java=../jabkit/src/main/java/module-info.java
+
 //FILES tinylog.properties=../jabkit/src/main/resources/tinylog.properties
 
 // This is the main class - directly called by JBang
 //SOURCES ../jabkit/src/main/java/org/jabref/toolkit/JabKitLauncher.java
```

```
$ jbang .jbang/JabKitLauncher.java 
Error occurred during initialization of boot layer
java.lang.module.FindException: Module jdk.jsobject not found, required by javafx.web
```

With `//MODULE` and a `package` statement, I get

```bash
$ jbang .jbang/JabKitLauncher.java 
[jbang] Building jar for JabKitLauncher.java...
/export/home/kopp/.jbang/cache/jars/JabKitLauncher.java.c2a8eccfaec4c2f74a87d7fb28610f4bc0f553629bc108ab4ba5ab6e60231691/generated/module-info.java:39: error: module not found: jdk.random
    requires jdk.random;
                ^
```

PR #2453, opened 2026-04-28
feat: replace picocli with aesh for CLI parsing

stalep:
Migrate jbang's CLI framework from picocli to aesh, leveraging aesh's compile-time annotation processor to eliminate runtime reflection and improve startup performance.

Key changes:
- Replace all picocli annotations (@Command, @Option, @Parameters) with aesh equivalents (@CommandDefinition, @GroupCommandDefinition, @Option, @Argument, @Arguments, @OptionList, @OptionGroup, @Mixin)
- Convert Callable<Integer> commands to Command<CommandInvocation> with CommandResult return type
- Port custom type converters, parameter consumers, and preprocessors to aesh Converter and OptionParser interfaces
- Implement external plugin command discovery in help output
- Add default value provider using aesh's post-parse processing
- Wire mutual exclusion validation via aesh's exclusiveWith
- Restore version checking, dynamic completion, option aliases, negatable options, and grouped help sections
- Regenerate native-image configuration for aesh
- Update tests for aesh CLI parsing

Performance (vs picocli, median):
- Native image: 6ms vs 33ms (5.5x faster) for single commands
- JDK 25: 109ms vs 228ms (2.1x faster)
- JDK 11: 149ms vs 269ms (1.8x faster)

## Summary by CodeRabbit

* **New Features**
  * Added CLI benchmark tests/task, improved dynamic shell completion, async version check, and an “External” help section for discovered plugins.

* **Bug Fixes**
  * Standardized invalid-input exit codes, improved help/option output and unknown/alias handling, and clearer deprecation messaging.

* **Build & Configuration**
  * Updated build/test task behavior (benchmark separation) and adjusted native-image reachability/resources; .gitignore expanded.

* **Documentation**
  * Regenerated CLI manpages to reflect the new command behavior and output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jbangdev/jbang/pull/2453)

PR #2513, opened 2026-06-05
fix: restore multi-JDK features lost in aesh migration

maxandersen:
## Problem

PR #2271 (*feat: install multiple JDKs with same major version*) was merged on May 14, but several of its features were lost when PR #2453 (aesh migration) was squash-merged on June 3. The aesh migration rewrote `Jdk.java` from picocli to aesh annotations but didn't preserve all the behavioral changes from #2271.

### `Jdk.java` — 7 fixes

| Feature | What happened |
|---|---|
| `jdk list --providers/-P` | Entire option + handler deleted |
| `jdk list --distros/-D` | Entire option + handler deleted, `JdkDistroQuery` import dropped |
| Full version comparison (`defVersion` string) | Reverted to `defMajorVersion` int — breaks default marker with multiple JDKs of same major version |
| Separate available/installed mapping | Collapsed to single path; `linkedId` always null, no `LinkedJdk` detection |
| Tag display in list output | Dropped from text output |
| `jdk uninstall` two-pass (`canInstall` → `canUpdate`) | Replaced with single `canUpdate`; info message uses raw input instead of `jdk.id()` |
| `jdk java-env` uses `canInstall` | Reverted to `canUpdate` |

### `TestJdk.java` — 4 restored tests

| Test | What it covers |
|---|---|
| `testJdkInstallWithLinkingAndIntegerId` | Integer ID rejection when linking |
| `testJdkInstallWithLinkingToExistingJBangJdkPath` | JBang-managed path rejection |
| `testJdkInstallSameVersion` | "Already installed" message |
| `testJdkInstallSameVersionForced` | `--force` creates versioned dir + symlink (core multi-JDK feature) |

## Verification

All 30 TestJdk tests pass (including the 4 restored ones), plus TestRun (124) and TestProject (9) — zero failures.

### DIFF
--- a/src/main/java/dev/jbang/cli/Jdk.java
+++ b/src/main/java/dev/jbang/cli/Jdk.java
@@ -22,6 +22,7 @@ import com.google.gson.annotations.SerializedName;
 
 import dev.jbang.Cache;
 import dev.jbang.Settings;
+import dev.jbang.devkitman.JdkDistroQuery;
 import dev.jbang.devkitman.JdkManager;
 import dev.jbang.devkitman.JdkProvider;
 import dev.jbang.util.CommandBuffer;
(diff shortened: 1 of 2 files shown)


maxandersen:
@quintesse this is result after asking opus to investigate what features/fixes from #2271 got lost in aesh port.

does it look right or I can also go back rebasing ash pr but that would be much bigger diff I fear.

quintesse:
I already wnet through it by hand but I'll check it against this to see if either of us missed anything

maxandersen:
related to #2515

maxandersen:
on my own quick look and LLM deeper look #2515 does the same but more so closing this.

#2515 covers everything plus additional fixes (the JdkProvidersMixin visibility change, linkedId display, and more thorough test infrastructure updates that make mock JDKs match the real directory layout).

PR #2517, opened 2026-06-07
test: add ArchUnit rule to prevent primitive types on @Option fields + fix object versions

maxandersen:
## Summary

Add an ArchUnit test (`CliArchitectureTest`) that ensures all Aesh CLI annotation fields (`@Option`, `@Argument`, `@Arguments`, `@OptionList`, `@OptionGroup`) use boxed types instead of Java primitives.

## Problem

Primitive types have implicit defaults (`false`, `0`, etc.) which make it impossible for `JBangDefaultValueProvider` to distinguish "user didn't pass the flag" from "user explicitly set the default value". This breaks config-file overriding — a config value can never be negated from the CLI because the primitive default shadows it.

For example, with `boolean insecure`, setting `insecure=true` in `jbang.properties` works, but passing `--no-insecure` on the CLI can't override it because `false` (the primitive default) is indistinguishable from "not set".

## What this PR adds

- **ArchUnit dependency** (`archunit-junit5:1.4.0`) in `build.gradle`
- **`CliArchitectureTest`** with 5 rules (one per Aesh annotation), checking all 8 Java primitive types

## Current violations detected

The test currently **fails intentionally** to document the 36 existing violations:
- 35 `boolean` fields that should be `Boolean`
- 1 `int` field (`Deps.max`) that should be `Integer`

Example violation message:
```
dev.jbang.cli.BaseCommand.insecure: @Option field must not be primitive (boolean);
use Boolean so JBangDefaultValueProvider can distinguish 'not set' from the default value
```

## Next steps

The actual migration of primitive → boxed types should be done in a follow-up PR, with null checks added where each field is read.

Refs #2504

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -188,6 +188,7 @@ dependencies {
 	testImplementation "com.github.stefanbirkner:system-rules:1.17.2"
 	testImplementation "org.assertj:assertj-core:${assertjVersion}"
 	testImplementation "org.hamcrest:hamcrest-library:2.2"
+	testImplementation "com.tngtech.archunit:archunit-junit5:1.4.0"
 	testImplementation "org.wiremock:wiremock:${wiremockVersion}"
 	testImplementation platform("io.qameta.allure:allure-bom:$allureVersion")
 	testImplementation "io.qameta.allure:allure-junit5"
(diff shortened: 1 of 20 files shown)


quintesse:
Never heard of ArchUnit before, pretty neat 👍

maxandersen:
I have mentioned it before :) but hadn't had a good usecase to find problems and now realize it's output is great for driving LLMs to fix stuff. So perfect storm!

Note: @stalep working on fix in aesh so we hopefully don't need to keep this.

quintesse:
> I have mentioned it before

As you can see I don't always listen very well 😅

maxandersen:
It's more me sucking at speaking well :)

maxandersen:
didnt need it as fixed the causes - archunit might come back :)

Issue #514, opened 2020-11-16
jbang install


maxandersen:
we should have a `jbang install` that would make the app/script available on the users path. kinda like `go install`.

i.e. `jbang install myscript.java`  would put a `myscript` script or `myscript.cmd` in `~/.jbang/bin` 

the `jbang/bin` gets added by some setup in users env/shell files.

quintesse:
We can close this issue, right?

maxandersen:
spring cleaning!

PR #511, opened 2020-11-15
docs: Clarify that .jsh are not cached and no native support (fixes #506)

vorburger:
fixes #506

also mentions no native image support for .jsh (re. #510)

PR #2519, opened 2026-06-08
build: upgrade to aesh 3.13 and fix primitive option types (#2504, #2…

stalep:
Upgrade aesh from 3.12.1 to 3.14.1 and readline-api from 3.11 to 3.14.
Leverage new aesh features to remove manual post-parse boilerplate.

## aesh fixes used
- #511: Fallback value chain runs after custom OptionParser
- #512: `${env:...}` resolution on annotation processor path
- #513: LinkedHashMap for @OptionGroup (preserves insertion order)
- #514: Cascading `${key:-fallback}` variable resolution

## Changes

**DefaultValueProvider:**
- Add `fallbackValue()` to JBangDefaultValueProvider for config-based
  fallback on bare flags (`--debug`, `--jfr`, `--open`)
- Split FALLBACK_ONLY_OPTIONS to skip config defaults for debug/jfr
  when the option is omitted (prevents always-on debug)

**DebugConverter (new):**
- Converts raw `--debug` string directly into `Map<String,String>` via
  `@Option(converter=...)`, replacing separate debugRaw/debugString fields

**DebugOptionParser:**
- No longer sets empty sentinel for bare `--debug`; aesh's fallback
  chain (provider fallbackValue -> annotation fallbackValue) handles it

**@OptionGroup for --javaagent:**
- Replace `@OptionList` + manual string splitting + lazy getter with
  `@OptionGroup(defaultValue="")` which handles key=value parsing,
  key-only entries, and insertion order natively

**Boolean field cleanup:**
- Keep `Boolean` wrapper for negatable inherited flags (`verbose`, `quiet`,
  `offline`, `fresh`, `preview`, `printExceptions`) to support tri-state
  config override semantics (e.g. `jbang config set verbose true`
  overridden with `--no-verbose` on the CLI)
- Convert non-negatable options to primitive `boolean` where null
  semantics aren't needed (`insecure`, `enablePreviewRequested`, etc.)

**Env var defaults via `${env:VAR:-}` (aesh#514):**
- AIOptions: `defaultValue = "${env:JBANG_AI_*:-}"` replaces manual
  `System.getenv()` lookups in `Init.afterParse()`

**Other simplifications:**
- Field initializers for userParams (`List<String> = new ArrayList<>()`)
- Remove `resolveAfterParse()` calls from Run, App, Alias
- Remove redundant config fallback from `Edit.applyEditorDefaults()`
- Remove empty `afterParse()` overrides

Co-authored-by: Max Rydahl Andersen <max@xam.dk>

### DIFF
--- a/src/main/java/dev/jbang/cli/JBangDefaultValueProvider.java
+++ b/src/main/java/dev/jbang/cli/JBangDefaultValueProvider.java
@@ -15,11 +15,11 @@ import dev.jbang.Configuration;
 
 public class JBangDefaultValueProvider implements DefaultValueProvider {
 
-	// Options that use fallbackValue="" or DebugOptionParser for three-state
-	// semantics (null=not specified, ""=bare flag, "value"=explicit). The provider
-	// must not supply defaults for these because the empty-string sentinel triggers
-	// config-file lookup in RunMixin.resolveAfterParse() instead.
-	private static final Set<String> FALLBACK_OPTIONS = new HashSet<>(Arrays.asList("debug", "jfr"));
+	// Options where config defaults should only apply as fallback values (when
+	// the option is explicitly used bare, e.g. --debug), not as default values
+	// (when the option is omitted entirely). This prevents config-set debug/jfr
+	// from activating on every command invocation.
+	private static final Set<String> FALLBACK_ONLY_OPTIONS = new HashSet<>(Arrays.asList("debug", "jfr"));
 
 	private static volatile Map<Class<?>, String> classToPath;
 
@@ -29,7 +29,7 @@ public class JBangDefaultValueProvider implements DefaultValueProvider {
 			return null;
 		}
 
-		if (FALLBACK_OPTIONS.contains(option.name())) {
+		if (FALLBACK_ONLY_OPTIONS.contains(option.name())) {
 			return null;
 		}
 
@@ -52,6 +52,26 @@ public class JBangDefaultValueProvider implements DefaultValueProvider {
 			}
 		}
 
+		return getValue(optName);
+	}
+
+	@Override
+	public String fallbackValue(ProcessedOption option) {
+		if (option.name() == null) {
+			return null;
+		}
+		String optName = option.name().replace("-", "");
+		String fullPath = null;
+		if (option.parent() != null && option.parent().getCommand() != null) {
+			fullPath = getCommandPathForClass(option.parent().getCommand().getClass());
+		}
+		if (fullPath != null) {
+			String key = fullPath + "." + optName;
+			String val = getValue(key);
+			if (val != null) {
+				return val;
+			}
+		}
 		return getValue(optName);
 	}
 
--- a/build.gradle
+++ b/build.gradle
@@ -156,11 +156,11 @@ dependencies {
 	implementation 'org.jspecify:jspecify:1.0.0'
 	implementation 'org.apache.commons:commons-text:1.15.0'
 	implementation 'org.apache.commons:commons-compress:1.27.1'
-	implementation('org.aesh:aesh:3.12.1') {
+	implementation('org.aesh:aesh:3.14.2') {
 		exclude group: 'org.aesh', module: 'readline'
 	}
-	implementation 'org.aesh:readline-api:3.11'
-	annotationProcessor 'org.aesh:aesh-processor:3.12.1'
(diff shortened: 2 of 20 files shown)


src/main/java/dev/jbang/cli/JBangDefaultValueProvider.java
```diff
@@ -55,6 +55,26 @@ public String defaultValue(ProcessedOption option) {
 		return getValue(optName);
 	}
 
+	@Override
+	public String fallbackValue(ProcessedOption option) {
```
maxandersen, line 59: if this kicks in...why do we need to change all to Boolean?

stalep: Yes, the `Boolean` wrappers have been reverted back to primitives in the latest push.

maxandersen:
@stalep — Use aesh 3.14.1's `${env:VAR:-fallback}` variable resolution (aeshell/aesh#514) to declare env var defaults directly in `@Option` annotations, replacing manual `System.getenv()` lookups in `afterParse()`.

Changes:
- **AIOptions**: Add `defaultValue = "${env:JBANG_AI_*:-}"` to `--ai-provider`, `--ai-api-key`, `--ai-endpoint`, `--ai-model`
- **Edit**: Add `defaultValue`/`fallbackValue = "${env:JBANG_EDITOR:-}"` to `--open`, remove `applyEditorDefaults()`
- **Init**: Remove `afterParse()` override that manually checked `JBANG_AI_*` env vars

This brings env var handling in line with how picocli did it (e.g. `defaultValue = "${JBANG_EDITOR:-${default.edit.open:-}}"`) — declarative in annotations rather than imperative post-parse code. Config-based defaults are still handled by `JBangDefaultValueProvider`.

stalep:
Looks good! :+1:

quintesse:
> can now be overridden with '--no-force' on the CLI.

And with `--force=false`, I assume, right?

stalep:
No, `--force=false` and `--no-force` don't work. 
The `force` options are `hasValue = false` and not `negatable`. The commit message example mentioning `--no-force` was wrong — the negatable override pattern only applies to the inherited boolean flags (`verbose`, `quiet`, `offline`, `fresh`, `preview`) which are declared `negatable = true` with `Boolean` wrapper type. 
A correct example would be: `jbang config set verbose true` overridden with `--no-verbose`. I've fixed the commit message.

quintesse:
> the negatable override pattern only applies to the inherited boolean flags (verbose, quiet, offline, fresh, preview) 

~What is your opinion about that @maxandersen ? Because it's definitely a breaking change. When we talked about this before we only mentioned those options because that's what was visible at that point in the code, but that syntax of `booleanOpt=true/false` is applicable to any boolean in Picocli,~

~So are we okay with not being backward compatible? If so (we're not at JBang 1.0.0 afterall) do we even want to special case those 5 options? Or do we go the other way and try to be 100% backward compatible?~

Edit: see next message

quintesse:
> No, --force=false and --no-force don't work.
> the negatable override pattern only applies to the inherited boolean flags

Well the issue is that ALL options have config overrides in JBang, not only those 5 global options. So how can I ever disable anything that was enabled by the user's config?

> A correct example would be: jbang config set verbose true overridden with --no-verbose

And here my question is again: "and with `--verbose=false` right?" Otherwise we're still not backward compatible. (Although we could perhaps have a discussion about breaking compatibility once and go for this new syntax, but that will depend a lot on @maxandersen , personally I'm not a proponent of breaking things unnecessarily)

stalep:
Good news with the new version of aesh (3.14.2): `--flag=false` works in aesh for **all** boolean options, even those with `hasValue = false` and without `negatable = true`. So `--force=false` already works to override a config-set `force = true`.

Verified:
- `--verbose=true` / `--verbose=false` -- works (negatable option)
- `--force=true` / `--force=false` -- works (non-negatable, `hasValue = false`)

So the `=true`/`=false` syntax is backward compatible with picocli. The `--no-*` syntax is only available for options declared `negatable = true` (the inherited global flags), but the `=false` syntax works universally.

maxandersen (dismissed):

maxandersen:
merging in as its doing all the right things. lets fix anything remaining/new in separate prs

maxandersen (approved):

PR #2521, opened 2026-06-08
refactor: simplify CLI post-parsing with aesh 3.13 features

maxandersen:
Building on #2504 (aesh 3.13 upgrade), this PR removes manual post-parse boilerplate by leveraging aesh's native APIs:

### Use aesh Converter for `--debug` (`11b8b600`)
- New `DebugConverter` converts the raw debug string directly into `Map<String,String>` via aesh's `@Option(converter=...)`
- Merges separate `debugRaw` (String) and `debugString` (Map) fields into a single field
- Fixes `DebugOptionParser` to resolve fallback values from `DefaultValueProvider`, matching aesh's `applyOptionalFallback()` behavior
- Removes `resolveDebugArgs()`

### Replace `resolveAfterParse()` with lazy getter for javaAgentSlots (`ef872932`)
- Converts `javaAgentSlots` from a public field populated by `resolveAfterParse()` into a cached lazy getter
- Removes all `resolveAfterParse()` call sites (Run, App, Alias)
- Note: aesh's `@OptionGroup` can't be used here because it uses `HashMap` internally, losing insertion order ([aesh#513](https://github.com/aeshell/aesh/issues/513))

### Simplify `afterParse()` in Run and App (`38ac4985`)
- Replace null-guard for `userParams` with field initializer
- Remove now-empty `afterParse()` override in App

### Remove redundant config fallback from `applyEditorDefaults()` (`f974d0d3`)
- The config lookup for `edit.open` when `--open` is used bare is now handled by `JBangDefaultValueProvider.fallbackValue()` in aesh 3.13
- Only the `JBANG_EDITOR` env var override remains (can't be expressed via `DefaultValueProvider`)

### DIFF
--- a/src/main/java/dev/jbang/cli/JBangDefaultValueProvider.java
+++ b/src/main/java/dev/jbang/cli/JBangDefaultValueProvider.java
@@ -15,11 +15,11 @@ import dev.jbang.Configuration;
 
 public class JBangDefaultValueProvider implements DefaultValueProvider {
 
-	// Options that use fallbackValue="" or DebugOptionParser for three-state
-	// semantics (null=not specified, ""=bare flag, "value"=explicit). The provider
-	// must not supply defaults for these because the empty-string sentinel triggers
-	// config-file lookup in RunMixin.resolveAfterParse() instead.
-	private static final Set<String> FALLBACK_OPTIONS = new HashSet<>(Arrays.asList("debug", "jfr"));
+	// Options where config defaults should only apply as fallback values (when
+	// the option is explicitly used bare, e.g. --debug), not as default values
+	// (when the option is omitted entirely). This prevents config-set debug/jfr
+	// from activating on every command invocation.
+	private static final Set<String> FALLBACK_ONLY_OPTIONS = new HashSet<>(Arrays.asList("debug", "jfr"));
 
 	private static volatile Map<Class<?>, String> classToPath;
 
@@ -29,7 +29,7 @@ public class JBangDefaultValueProvider implements DefaultValueProvider {
 			return null;
 		}
 
-		if (FALLBACK_OPTIONS.contains(option.name())) {
+		if (FALLBACK_ONLY_OPTIONS.contains(option.name())) {
 			return null;
 		}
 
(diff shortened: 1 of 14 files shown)


src/main/java/dev/jbang/cli/RunMixin.java
```diff
@@ -19,9 +17,7 @@ public class RunMixin {
 	@Option(name = "jfr", fallbackValue = "", description = "Launch with Java Flight Recorder enabled.")
 	public String flightRecorderString;
 
-	@Option(shortName = 'd', name = "debug", parser = DebugOptionParser.class, description = "Launch with java debug enabled. Set host/port or provide key/value list of JPDA options (default: ${DEFAULT-VALUE})")
-	public String debugRaw;
-
+	@Option(shortName = 'd', name = "debug", parser = DebugOptionParser.class, converter = DebugConverter.class, fallbackValue = "4004", description = "Launch with java debug enabled. Set host/port or provide key/value list of JPDA options (default: ${DEFAULT-VALUE})")
```
quintesse, line 20: Shouldn't the `fallbackValue` come from the `JBangDefaultValueProvider`? Because this value might be set in the config file and in that case `4004` is most likely not correct.

src/main/java/dev/jbang/cli/JBangDefaultValueProvider.java
```diff
@@ -15,11 +15,11 @@
 
 public class JBangDefaultValueProvider implements DefaultValueProvider {
 
-	// Options that use fallbackValue="" or DebugOptionParser for three-state
-	// semantics (null=not specified, ""=bare flag, "value"=explicit). The provider
-	// must not supply defaults for these because the empty-string sentinel triggers
-	// config-file lookup in RunMixin.resolveAfterParse() instead.
-	private static final Set<String> FALLBACK_OPTIONS = new HashSet<>(Arrays.asList("debug", "jfr"));
+	// Options where config defaults should only apply as fallback values (when
```
quintesse, line 18: I don't understand exactly why we need an explicit list for this? Isn't that the difference between `defaultValue` and `fallbackValue`?

maxandersen: Yeah I'm trying with this draft to get back to handling most of this directly on the annotations and make for cleaner usage. 

But finding some gaps in aesh support - hoping this can help cover them. By improving aesh or adjusting our approach.

stalep:
This can be closed as this pr should have all of the improvements https://github.com/jbangdev/jbang/pull/2519

maxandersen:
Closing — #2519 supersedes this entirely. It upgrades to aesh 3.14.1 which fixes the issues this PR worked around manually (aesh#511 fallback chain, aesh#513 LinkedHashMap for @OptionGroup), and adds 6 tests covering the debug fallback behavior. The only non-overlapping bits (wildcard import cleanup, javadoc about aesh#513) are either cosmetic or now incorrect since the upstream fix landed.

PR #2522, opened 2026-06-09
feat: lock/digest integrity verification (aesh + resolver SPI)

maxandersen:
This PR adds lock-based integrity + reproducibility checks for JBang refs, ported to the aesh CLI framework and wired into the Maven Resolver trusted-checksums SPI.

Supersedes #2410 (which was picocli-based and predates the aesh migration).

 p.s. The IKEA table from last time is holding up great.

## What's included

- New command: `jbang lock <ref>`
- Default lock file behavior:
  - local file refs (`app.java`) → `app.java.lock`
  - alias/GAV/URL refs → `.jbang.lock`
  - override via `--lock-file=...`
- Run policy flag: `--locked=<none|lenient|strict>`
  - `none`: ignore lock checks
  - `lenient` (default): if lock data exists, enforce it; if lock missing, run normally
  - `strict`: require lock entry when lock file exists and enforce strict matching
- Lock structure per ref:
  - `ref=sha256:...` (main resource digest)
  - `ref.sources=...` (resolved source manifest)
  - `ref.deps=...` (resolved transitive dependency coordinates)
  - `ref.dep.<gav>=sha256:...` (per-artifact dependency digests)

## Verification behavior

When lock data exists, JBang validates:

1. Main resource digest
2. Source manifest (`.sources`) consistency
3. Dependency graph (`.deps`) consistency
4. Per-artifact dependency digests (`.dep.<gav>`)

Mismatch → fail with explicit error including resolved file path and fix guidance.

## Maven Resolver integration

Per discussion with @cstamas, this implements the resolver's `TrustedChecksumsSource` SPI:

- `JBangTrustedChecksumsSource` implements `TrustedChecksumsSource` + `Writer` backed by `.jbang.lock`
- `Lock` command uses the `Writer` interface to record per-artifact checksums
- `DigestUtil` delegates to `ChecksumAlgorithmHelper.calculate()` from `maven-resolver-spi` instead of custom `MessageDigest` code
- Algorithm name mapping between lock-file format (`sha256`) and resolver format (`SHA-256`)

Today the TC source is used standalone by `Lock` and `Run`. When MIMA gains an extension point for custom trusted-checksums sources (no such API exists through 2.4.45 or 3.0.0-alpha-3), the same implementation can be injected directly into the resolver pipeline — at which point the manual dep-digest verification loop in `Run` can be deleted.

## Why this design

- Keeps "just run it" workflow intact when no lock is present.
- Provides stronger guarantees as soon as lock data exists.
- Separates mutation (`jbang lock`) from execution (`jbang run ...`).
- Reuses resolver's own checksum infrastructure instead of rolling custom crypto.
- Shared `DigestUtil` eliminates code duplication between `Lock` and `Run`.

## Feedback requested

1. Are `--locked` mode semantics (`none|lenient|strict`) right?
2. Is `<script>.lock` the right default for local file refs?
3. Is properties-based lock format acceptable for v1 with these keys?
4. Should strict mode require complete per-artifact digests for every locked dep (current behavior)?
5. Is the `TrustedChecksumsSource` SPI the right abstraction to build on for the MIMA integration path?

### Related issues

Relates to:
- #1989 (script/alias SHA verification)
- #1979 (pinning behavior for aliases/scripts)

Supersedes / consolidates discussion from:
- #2410 (original picocli-based lock PR)
- #938 (catalog checksums)
- #687 (older checksum/integrity thread)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -172,6 +172,8 @@ dependencies {
 	implementation "org.jboss:jandex:2.2.3.Final"
 
 	implementation "eu.maveniverse.maven.mima:context:2.4.36"
+	implementation "org.apache.maven.resolver:maven-resolver-spi:1.9.24"
+	implementation "org.apache.maven.resolver:maven-resolver-impl:1.9.24"
 	runtimeOnly    "eu.maveniverse.maven.mima.runtime:standalone-static:2.4.36"
 	implementation "org.apache.maven:maven-model:3.9.11"
 
--- a/build.gradle
+++ b/build.gradle
@@ -172,8 +172,10 @@ dependencies {
 	implementation "org.jboss:jandex:2.2.3.Final"
 
 	implementation "eu.maveniverse.maven.mima:context:2.4.36"
-	implementation "org.apache.maven.resolver:maven-resolver-spi:1.9.24"
-	implementation "org.apache.maven.resolver:maven-resolver-impl:1.9.24"
+	compileOnly    "org.apache.maven.resolver:maven-resolver-spi:1.9.24"
+	compileOnly    "org.apache.maven.resolver:maven-resolver-impl:1.9.24"
+	testImplementation "org.apache.maven.resolver:maven-resolver-spi:1.9.24"
+	testImplementation "org.apache.maven.resolver:maven-resolver-impl:1.9.24"
 	runtimeOnly    "eu.maveniverse.maven.mima.runtime:standalone-static:2.4.36"
 	implementation "org.apache.maven:maven-model:3.9.11"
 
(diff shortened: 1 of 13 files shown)


build.gradle
```diff
@@ -172,6 +172,8 @@ dependencies {
 	implementation "org.jboss:jandex:2.2.3.Final"
 
 	implementation "eu.maveniverse.maven.mima:context:2.4.36"
+	implementation "org.apache.maven.resolver:maven-resolver-spi:1.9.24"
+	implementation "org.apache.maven.resolver:maven-resolver-impl:1.9.24"
```
cstamas: compileOnly: as both are included into standalone-static fat jar below, plus, SPI is already transitive dep of context (it pull api, spi and util, the public APIs).

PR #2525, opened 2026-06-09
fix: generate correct module-info requires for JavaFX //MODULE scripts

koppor:
Follow-up to https://github.com/jbangdev/jbang/pull/2520

## Problem

A `//MODULE` script that depends on JavaFX fails to compile:

```java
//MODULE app
//DEPS org.openjfx:javafx-base:21.0.2
package app;
import javafx.collections.FXCollections;
public class app {
        System.out.println(FXCollections.observableArrayList("hi"));
    }
}
```

```
error: package javafx.collections is not visible
```

## Root cause

JavaFX publishes **two** artifacts per module:

| artifact | module |
|---|---|
| `javafx-base.jar` (no classifier, empty placeholder) | `javafx.baseEmpty` |
| `javafx-base-<os>.jar` (platform classifier, real classes) | `javafx.base` |

`ModuleUtil.generateModuleInfo` matched resolved artifacts to root dependencies by the **full management key**, which includes the classifier (`getManagementKey()` → `groupId:artifactId:type[:classifier]`). So the root dep `org.openjfx:javafx-base` (key `…:jar`) matched only the unclassified empty placeholder and emitted:

```
requires javafx.baseEmpty;
```

leaving `javafx.collections` (which lives in the real `javafx.base`) invisible to the generated module.

## Fix

In `generateModuleInfo`:

1. Match root dependencies **ignoring the classifier**, so the platform artifact carrying the real module is considered too.
2. Skip the `*Empty` placeholder modules, so the real module (`javafx.base`) is required.

```
requires javafx.base;
```

This only changes which module names land in the generated `module-info.java`; the launch path (`-p <classpath>` + `-m`) is untouched.

## Verification

- Real CLI run (GraalVM 25): the `//MODULE` + JavaFX script above now compiles and runs.
- Non-JavaFX `//MODULE` scripts (e.g. picocli) are unaffected.
- New `TestModule.testModuleJavaFX` asserts the generated `module-info` requires `javafx.base` and not `javafx.baseEmpty`, and actually compiles the script.
- `TestModule` and `DependencyResolverTest` suites pass; spotless clean.

> **Note on the test:** it pins JavaFX to `11.0.2`. The test toolchain defaults to Java 11 (`build.gradle`), whose `ModuleFinder` cannot read JavaFX 21+ module descriptors (it throws `FindException`, so the module name resolves to `null`). JavaFX 11.0.2's descriptor is Java 11-readable — the same reason the existing `testResolveJavaModules` uses 11.0.2.

> 🤖 _Generated with [Claude Code](https://claude.com/claude-code)_

### DIFF
--- a/src/main/java/dev/jbang/util/ModuleUtil.java
+++ b/src/main/java/dev/jbang/util/ModuleUtil.java
@@ -92,17 +92,18 @@ public class ModuleUtil {
 				.map(MavenCoordinate::fromString)
 				.map(c -> c.getGroupId() + ":" + c.getArtifactId() + ":" + c.getType())
 				.collect(Collectors.toSet());
-			// Now filter the resolved artifacts down to those root dependencies and get
-			// their module names, skipping the empty `*Empty` placeholder modules so we
-			// require the real module (e.g. `javafx.base`, not `javafx.baseEmpty`).
+			// Match the resolved artifacts back to those root dependencies
+			// (again ignoring the classifier) and collect their module names.
+			// An empty placeholder module may end up required next to the real
+			// one (e.g. `javafx.baseEmpty` next to `javafx.base`); that is
+			// harmless, so we do not special-case it by name.
 			Stream<String> depModNames = ctx.resolveClassPath()
 				.getArtifacts()
 				.stream()
 				.filter(a -> depKeys.contains(a.getCoordinate().getGroupId() + ":"
 						+ a.getCoordinate().getArtifactId() + ":" + a.getCoordinate().getType()))
 				.map(ArtifactInfo::getModuleName)
-				.filter(Objects::nonNull)
-				.filter(name -> !name.endsWith("Empty"));
+				.filter(Objects::nonNull);
 			// And join this list of names with the JDK module names
 			List<String> moduleNames = Stream.concat(ModuleUtil.listJdkModules().stream(), depModNames)
 				.collect(Collectors.toList());
--- a/src/main/java/dev/jbang/util/ModuleUtil.java
+++ b/src/main/java/dev/jbang/util/ModuleUtil.java
@@ -6,9 +6,8 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
-import java.util.function.Function;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


src/main/java/dev/jbang/util/ModuleUtil.java
```diff
@@ -81,20 +80,29 @@ public static Path generateModuleInfo(BuildContext ctx) throws IOException {
...
 			Stream<String> depModNames = ctx.resolveClassPath()
 				.getArtifacts()
 				.stream()
-				.filter(a -> deps.containsKey(a.getCoordinate().getManagementKey()))
+				.filter(a -> depKeys.contains(a.getCoordinate().getGroupId() + ":"
+						+ a.getCoordinate().getArtifactId() + ":" + a.getCoordinate().getType()))
 				.map(ArtifactInfo::getModuleName)
-				.filter(Objects::nonNull);
+				.filter(Objects::nonNull)
+				.filter(name -> !name.endsWith("Empty"));
```
quintesse: So everything in this PR seems great, except for this line. This seems extremely specific and fragile. Will things fail if we add empty modules to the classpath? If not we should simply add them, because we can never trust that a) others will follow the same naming convention b) JavaFX decides to use a different name in the future.

quintesse (changes requested):
In general LGTM except for that one remark.

src/main/java/dev/jbang/util/ModuleUtil.java
koppor: Good call — verified, and you're right: nothing breaks if we keep them.

I dropped the `*Empty` name filter in c7fa3ea3. Matching the root dependency while ignoring the classifier already pulls in the artifact that carries the real module, so we now just `require` every module name a root dep resolves to. For JavaFX that means the empty placeholder ends up required next to the real module:

```
requires javafx.base;
requires javafx.baseEmpty;
```

which compiles and runs fine (the placeholder is an empty automatic module — no packages, nothing to clash). So no naming convention is assumed, and a future JavaFX rename / other libraries don't matter.

Test updated accordingly (it now only asserts the real `javafx.base` is required and that the script actually compiles).

koppor: I also wondered somehow, but wanted to push 😅. Fixed in https://github.com/jbangdev/jbang/pull/2525/commits/c7fa3ea3a7b4bdf0c454135e030e558052ecfa66 .

quintesse (approved):
LGTM

koppor:
Thank you for the feedback and patience! Looking forward to a next release, which enables us to test more and unblock using Java 26 in our project.

koppor:
Can confirm that it works with the `early-access` release.

After some workarounds for jbang EA installation (https://github.com/jbangdev/setup-jbang/issues/18 and https://github.com/jbangdev/setup-jbang/issues/19) ...

PR #2527, opened 2026-06-10
fix: optimize test suite performance (2m42s → ~50s)

maxandersen:
## Test Suite Performance Optimization

Three independent improvements that reduce the test suite wall time from **2m 42s to ~50s** (3.2x faster).

### Commits

1. **fix: make askInput timeout configurable to avoid 60s test delay**
   - Two tests (`testHelloWorldGAVWithNoMain`, `testHelloWorldGAVWithModulesButNoManifest`) blocked 30s each waiting on interactive `Util.askInput()` timeout
   - Added `Util.getAskInputTimeout()`/`setAskInputTimeout()` following existing static field pattern
   - `BaseTest.initEnv()` sets timeout to 0 so tests fail immediately instead of waiting

2. **test: replace Thread.sleep(1100) with file backdating in TestUtilDownloads**
   - 6× `Thread.sleep(1100)` waited for filesystem timestamp granularity for cache-evict checks
   - Replaced with `Files.setLastModifiedTime()` to backdate cached files by 2 seconds

3. **build: enable parallel test execution**
   - Uncommented `maxParallelForks = availableProcessors/2` in `build.gradle`
   - Each fork runs in a separate JVM with its own WireMock and `mavenTempDir` — no concurrency issues

### Results

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Full test suite | **2m 42s** | **~50s** | 🟢 **3.2x faster** |
| Two timeout tests | 60s wasted | 0.6s | 60s saved |
| TestUtilDownloads | 6.9s sleeping | 0.1s | 6.3s saved |
| Parallelism | 1 fork | 7 forks | ~2x wall time reduction |

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -214,7 +214,7 @@ buildConfig {
 tasks.withType(Test).configureEach {
 	//environment "JAVA_HOME", "/Users/manderse/.local/share/mise/installs/java/graalvm-community-25.0.1"
 	//executable = new File("/Users/manderse/.local/share/mise/installs/java/graalvm-community-25.0.1/bin/java")
-	//maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+	maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
 }
 
 // to enable reproducible builds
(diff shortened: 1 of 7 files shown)


quintesse (approved):
LGTM

PR #2528, opened 2026-06-13
fix: typo in RequestedVersionComparator comparing v1 against itself

maxandersen:
## Bug

`RequestedVersionComparator.compare()` had a typo on line 361 of `JavaUtil.java`:

```java
int n1 = minRequestedVersion(v1);
int n2 = minRequestedVersion(v1); // should be v2
```

Both values were derived from `v1`, so the comparator was version-blind — it could never distinguish different version numbers, only the `+` suffix.

This affected:
- `Directives.javaVersion()` — `.max()` across multiple `//JAVA` directives would not reliably pick the highest version
- `ProjectBuilder.updateProject(Source, ...)` — merging java versions from source dependencies could let a lower version silently win

## Fix

One-character fix: `v1` → `v2`.

Added `TestRequestedVersionComparator` covering null handling, same/different versions, exact vs open versions, and `max()` selection.

### DIFF
--- a/src/main/java/dev/jbang/util/JavaUtil.java
+++ b/src/main/java/dev/jbang/util/JavaUtil.java
@@ -358,7 +358,7 @@ public class JavaUtil {
 				return -1;
 			}
 			int n1 = minRequestedVersion(v1);
-			int n2 = minRequestedVersion(v1);
+			int n2 = minRequestedVersion(v2);
 			if (n1 < n2) {
 				return -1;
 			} else if (n1 > n2) {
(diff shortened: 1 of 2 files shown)


maxandersen:
@quintesse just spotted this by random - I guess we haven't had many projects with multiple //JAVA statements.

it does reveal that apparently we have code trying to pick the highest version rather than the first one, but this bug makes it actaully pick the first one -  which is what I expected...

so not just merging this in but something we should figure out - should first one wins or highest win?

quintesse:
My intuition says the highest should win, because if we're including code that says it needs a higher Java version then I'd assume it wouldn't work with a lower one. Now You could say "then the main script should update its //JAVA line", and that wouldn't be wrong, but then I'd at least want a waning saying that you probably should update your //JAVA line because of a conflict... and that's more difficult than just picking the highest by default :-)

quintesse:
At least it seems that devkitman doesn't have that particular code (most of that file was copied to devkitman)

PR #2532, opened 2026-06-14
test: add ArchUnit architecture rules and move EXIT_* constants to ExitException

maxandersen:
## Summary

Replaces the old primitive-options ArchUnit rule (from #2517, no longer needed after #2519) with useful architecture rules that enforce codebase conventions.

### Architecture rules added

| Rule | Type | Status |
|------|------|--------|
| CLI commands must extend `BaseCommand` | ArchUnit | ✅ Active |
| Non-CLI packages must not depend on `dev.jbang.cli` (except `ExitException`) | ArchUnit | ✅ Active |
| `ExitException` must use named `EXIT_*` constants, not raw integers | Source scan | ✅ Active |
| No wildcard imports | Source scan | ⏸️ Disabled (fix incrementally) |

### EXIT_* constant refactor

Moved `EXIT_OK`, `EXIT_GENERIC_ERROR`, `EXIT_INVALID_INPUT`, `EXIT_UNEXPECTED_STATE`, `EXIT_INTERNAL_ERROR`, and `EXIT_EXECUTE` from `BaseCommand` to `ExitException`. `BaseCommand` delegates for backward compatibility.

This enables clean package layering — non-CLI code no longer needs to import `BaseCommand` just for exit code constants. All ~16 raw integer literals in `new ExitException(N, ...)` replaced with named constants.

### Why two test approaches?

- **ArchUnit** for structural rules (bytecode-level: inheritance, package dependencies)
- **Source-scanning JUnit tests** for conventions that javac erases (wildcard imports are resolved at compile time; `static final int` constants are inlined into bytecode)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -191,6 +191,7 @@ dependencies {
 	testImplementation "org.wiremock:wiremock:${wiremockVersion}"
 	testImplementation platform("io.qameta.allure:allure-bom:$allureVersion")
 	testImplementation "io.qameta.allure:allure-junit5"
+	testImplementation "com.tngtech.archunit:archunit-junit5:1.4.0"
 	// only for test running
 	agent "org.aspectj:aspectjweaver:$aspectJVersion"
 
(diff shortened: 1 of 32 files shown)


maxandersen:
blah - too figgity.

PR #2534, opened 2026-06-15
fix: use proper versioned download URL for native

quintesse:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced JBang's auto-download mechanism to intelligently select platform-specific distribution bundles based on native mode settings and system architecture, ensuring users receive optimized versions tailored to their operating system.


quintesse:
@coderabbitai review

PR #2537, opened 2026-06-15
test: add native download URL tests and apply JBANG_USE_NATIVE to jbang.ps1

maxandersen:
Follow-up to #2534.

## What changed

- **`jbang.ps1`**: Apply the same `JBANG_USE_NATIVE` bundle name logic from #2534 — when `JBANG_USE_NATIVE=true`, download `jbang-windows-{arch}.zip` instead of `jbang.zip`
- **`TestScriptNativeDownload`** (new): Verifies the download URL for both bash and PowerShell scripts with `JBANG_USE_NATIVE` enabled/disabled, across latest, versioned, and named-tag downloads, plus `JBANG_DOWNLOAD_URL` override
- **`AbstractScriptTest`** (new): Extracts shared test infrastructure (WireMock lifecycle, archive builders, env helpers, process runner) from the three script test classes to reduce duplication
- **`TestScriptRetry` / `TestScriptDownloadVersion`**: Refactored to extend `AbstractScriptTest`

All 23 tests pass.

### DIFF
--- a/src/main/scripts/jbang.ps1
+++ b/src/main/scripts/jbang.ps1
@@ -151,11 +151,16 @@ if (-not $binaryPath) {
 }
 if (-not $binaryPath -and -not $jarPath) {
   if (-not (Test-Path "$JBDIR\bin\jbang.jar") -or -not (Test-Path "$JBDIR\bin\jbang.ps1")) {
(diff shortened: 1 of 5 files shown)


quintesse (approved):
LGTM

PR #2538, opened 2026-06-17
test: add tests for jdk list

quintesse:
Also added a couple of IT tests that should capture tings like this in the future


quintesse:
Btw, we could easily make this work by forgetting about WireMock and move the problem of how to set system properties for JBang to a different issue. The only thing is that it adds new tests that depend on Foojay which might cause failing CI runs just because it's having problems again.

quintesse:
(perhaps we could change those tests and not make them fail when the http error is something like 5xx)

maxandersen:
Ignorering 5xx would ignore actually issues too :/

maxandersen:
**Fixes for the failing JdkIT tests**

Found three distinct issues causing the CI failures:

**1. Native image: `JBANG_JAVA_OPTIONS` passed as CLI args breaks parsing**

The `eef05b02` commit passes `JBANG_JAVA_OPTIONS` (containing `-Dkey=value` JVM flags) as command-line arguments to the native binary. Native images don't understand `-D` flags — they become invalid positional args and aesh rejects them. This broke all 5 JdkIT tests on native across all platforms.

Fix: Instead of passing `JBANG_JAVA_OPTIONS` as CLI args, the native binary now reads the env var directly in `Main` and calls `System.setProperty()` for each `-D` entry — only when `JavaUtil.inNativeImage()` is true. The script changes from `eef05b02` that injected `JBANG_JAVA_OPTIONS` into the native binary's args are reverted.

**2. `outContains("Available JDKs:\n")` fails on Windows**

The literal `\n` in the expected string doesn't match Windows `\r\n` line endings. Dropped the `\n` — the header text match is sufficient.

**3. `testJdkListDistros` uses wrong output format assertion**

The `--distros` flag outputs plain distribution names (`   aoj`), not the numbered format (`   25 (temurin)`) that the regex expected. Changed the test to use `--available` which produces the numbered format the regex already matches.

maxandersen:
**Updated fixes (rebased on your latest push)**

Added two commits on top of your changes:

**1. `d6f85aa0` — Native image: apply `JBANG_JAVA_OPTIONS` `-D` flags as system properties**

Native binaries can't process `-D` flags as CLI args. Instead of passing `JBANG_JAVA_OPTIONS` on the command line, `Main.applyJavaOptionsFromEnv()` now reads the env var directly and calls `System.setProperty()` for each `-Dkey=value` entry — only when `JavaUtil.inNativeImage()` is true. This fixes all 5 JdkIT tests on native image (all platforms).

**2. `ff1763ee` — Windows compatibility + distMock fix**

- Dropped `\n` from `outContains("Available JDKs:\n")` and `outContains("Available JDK Distributions:\n")` — the literal `\n` doesn't match Windows `\r\n` line endings.
- Removed `assertDiscoPackagesRequestsServedByWireMock()` from `testJdkListDistros` — the `--distros` command hits a different disco endpoint (`/distributions` not `/packages`), so the packages assertion always fails.

All 5 JdkIT tests pass on both JVM and native image locally.

quintesse:
AI can still be so dumb at times:

> The --distros flag outputs plain distribution names (   aoj), not the numbered format (   25 (temurin)) that the regex expected. Changed the test to use --available which produces the numbered format the regex already matches.

Yeah of course, just change the thing the test is supposed to test, brilliant :laughing: 

Btw,not saying it's not also very useful at the same time. But one has to really keep paying attention :-)

quintesse:
Ok, this is now ready to merge

maxandersen:
That was actually me changing its decision as the regex was expecting format of available rather than distros :)Sent from my iPhoneOn 19 Jun 2026, at 11:04, Tako Schotanus ***@***.***> wrote:﻿quintesse left a comment (jbangdev/jbang#2538)
AI can still be so dumb at times:

The --distros flag outputs plain distribution names (   aoj), not the numbered format (   25 (temurin)) that the regex expected. Changed the test to use --available which produces the numbered format the regex already matches.

Yeah of course, just change the thing the test is supposed to test, brilliant 😆

maxandersen (approved):

PR #2540, opened 2026-06-18
fix: update GraalVM native image metadata

quintesse:
Without tests for now because getting the tests to work is much harder than fixing the actual issue.


maxandersen (approved):

maxandersen:
ok merging but can you open pr with what tests you were doing and I can take look ? looked yesterday but had issues ;)

quintesse:
It's this one #2538

PR #2543, opened 2026-06-18
fix: update even more GraalVM native image metadata

quintesse:


quintesse:
Sorry for fixing the metadata bit by bit, I'm using the early-access release process to do my IT testing for me (since I can't get the IT tests to work properly yet).

maxandersen:
What part of it tests are not working for you locally?

maxandersen (approved):

quintesse:
The new ones from the PR related to these graalvm changes that still doesn't pass CI

PR #2465, opened 2026-05-16
feat: replace JLine with TamboUI for dependency search TUI

maxandersen:
## Summary

Replace JLine with [TamboUI](https://github.com/tamboui/tamboui) 0.4.0 for the `jbang deps search` interactive TUI.

### Dependencies
- Remove `jline-console-ui`, `jline-terminal-jni`
- Add `tamboui-toolkit:0.4.0`, `tamboui-aesh-backend:0.4.0`
- Net jar size delta: ~+100KB (TamboUI includes unused widgets that offset the JLine removal)

### TUI rewrite (`ArtifactSearchWidget`)
- **Layout**: dock() with search input (top), scrollable results list (center), inline status bar (bottom)
- **Compact UI**: removed the 5-line Details panel — `//DEPS g:a:v` preview shown inline at the right edge of the status bar. ~6 more rows for results.
- **Async Central search**: runs on scheduler thread via `runner.schedule()`, results marshalled back with `runOnRenderThread()`, generation counter discards stale results
- **Match highlighting**: all result rows show matched characters (green+bold) via `SearchFuzzedResult.highlightTarget()`
- **Version picker**: side pane appears on Enter, Esc goes back
- **Deduplication**: results grouped by groupId:artifactId, showing best-scored version
- **Offline support**: respects existing `--offline` / `-o` flag via `Util.isOffline()`

### Search intent classifier (`SearchIntentClassifier`)
Two intent types: **CLASS** (prefers Central) and **KEYWORD** (local-first).

| Input | Type | Central Query |
|-------|------|---------------|
| `jackson databind` | KEYWORD | `jackson databind` |
| `ObjectMapper` | CLASS | `c:ObjectMapper` |
| `com.fasterxml.jackson.databind.ObjectMapper` | CLASS | `fc:com.fasterxml...` |
| `c:StatelessSession` | CLASS | `c:StatelessSession` |
| `import com.fasterxml...ObjectMapper;` | CLASS | `fc:com.fasterxml...` |
| `error: package org.jsoup does not exist` | CLASS | `org.jsoup` |
| `com.fasterxml:jackson-databind:2.20.0` | KEYWORD | pass-through |

- `c:`/`fc:` prefixes stripped from `localQuery` so fuzzy matching works against GAV strings
- Auto-triggers Central search (400ms debounce) for CLASS intents

### Key bindings
- Tab/F5: search Central (semantic `searchCentral` action via `event.matches()`)
- Enter: select artifact → pick version → done
[...]

PR #2544, opened 2026-06-18
fix: strip c:/fc: prefix from local fuzzy query in search classifier

maxandersen:
When typing `c:StatelessSession`, the `SearchIntentClassifier` was passing the raw input (including `c:` prefix) as `localQuery`. The fuzzy matcher then tried to match `c:StatelessSession` against GAV strings like `org.hibernate:hibernate-core:6.4.0`, causing most Central results to fall below the 0.2 similarity cutoff — so you'd see "Central: added 350 artifacts" but only ~15 results in the list.

**Fix:** strip the `c:`/`fc:` prefix from `localQuery` (used for fuzzy matching against GAVs) while keeping it in `centralQuery` (used for the Central API).

Also simplifies `IntentType` from 7 enum values to 2:
- **`CLASS`** — class lookups, imports, compiler errors (`prefersCentral() = true`)
- **`KEYWORD`** — GAV, keywords, general text (`prefersCentral() = false`)

No consumer ever checked specific intent types; only `prefersCentral()` mattered.


maxandersen:
Duplicate — this fix is already part of #2465 (same branch).

Issue #2482, opened 2026-05-29
possible resource leak


maxandersen:
main issue found in #2453 

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@src/main/java/dev/jbang/cli/ExternalCommandsProvider.java` around lines 80 -
86, The listFiles method returns the raw Stream<Path> from Files.list(dir) which
leaks the underlying DirectoryStream when composed (e.g., via flatMap); change
listFiles to fully consume/close the DirectoryStream before returning by
collecting its contents into a List (e.g., try-with-resources over
Files.list(dir) and return collectedList.stream()) or by attaching an onClose
handler that closes the DirectoryStream; update the method using the existing
name listFiles and the Files.list call so callers receive a safe Stream<Path>
that won’t leave file handles open.
```

quintesse:
~But this code only exists in the PR, there is no `ExternalCommandsProvider` in the current code, so does opening a separate issue for it make sense? Wouldn't it be better to simply fix it in the PR?~

Edit: ok, this is because the code has been refactored. The code with the leak _is_ part of `dev/jbang/cli/JBang.java` in main

Just take care to look for the correct place, depending if this is pre or post PR merge

maxandersen:
Yes. It was just I didn't want to keep piling changes into aesh if it wasn't issues related to the port.

PR #2547, opened 2026-06-20
fix: close shell configuration file stream

sinhasebur:
## Summary

  Closed the stream returned by `Files.lines()` when checking shell configuration files for an existing JBang setup.

  The detection logic is extracted into `hasJBangSetup()`, which uses try-with-resources to ensure the underlying file resource is released. This is similar to the resource issue
  addressed in #2482, but occurs independently in `App.AppSetup`.

  ## Changes
  - Added tests for missing, matching, and non-matching configuration files

  ## Testing
  - `./gradlew test`
  - `./gradlew spotlessCheck`
 - ` ./gradlew build`
  All tests pass.


quintesse (approved):
LGTM

PR #2563, opened 2026-06-26
For JSON resources fetched over HTTP(S), explicitly set the `Accept` header

### LINKED ISSUE #2553: JBang doesn't specify the `Accept:` header when fetching a catalog
Whenever trying to install, e.g. [rnett/gradle-mcp](https://github.com/rnett/gradle-mcp) from behind a caching proxy, I receive the following error (rnett/gradle-mcp#208):

```
[jbang] [ERROR] Unable to download catalog: https://github.com/rnett/jbang-catalog/blob/HEAD/jbang-catalog.json via null
dev.jbang.cli.ExitException: Unable to download catalog: https://github.com/rnett/jbang-catalog/blob/HEAD/jbang-catalog.json via null
        at dev.jbang.catalog.Catalog.getByRef(Catalog.java:241)
        at dev.jbang.catalog.Catalog.getByName(Catalog.java:174)
        at dev.jbang.catalog.Alias.fromCatalog(Alias.java:301)
        at dev.jbang.catalog.Alias.merge(Alias.java:223)
        at dev.jbang.catalog.Alias.get(Alias.java:182)
        ...
Caused by: com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was STRING at line 1 column 1 path $
See https://github.com/google/gson/blob/main/Troubleshooting.md#unexpected-json-structure
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:521)
        ...
```

This is because _JBang_ doesn't explicitly set `Accept: application/json`, and the proxy serves cached HTML version by default.

_JBang_ version is

```console
$ jbang --version
0.138.0
```

0x6675636b796f75676974687562:
Fixes jbangdev/jbang#2553 and rnett/gradle-mcp#208.

### DIFF
--- a/src/main/java/dev/jbang/util/NetUtil.java
+++ b/src/main/java/dev/jbang/util/NetUtil.java
@@ -36,7 +36,9 @@ import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Base64;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


quintesse (approved):
LGTM

quintesse:
Although some kind of test for this would be nice :-)

maxandersen:
So just to be sure I grok it - it's because you have some custom proxy in between?

0x6675636b796f75676974687562:
> So just to be sure I grok it - it's because you have some custom proxy in between?

@maxandersen yes, exactly (though I didn't use `jbang` not from behind a proxy).

The reason is that, whenever you download resources (e.g.: a JSON catalog) from `github.com`, depending on your UA, the headers it sends etc., you're either served a human-readable HTML, or redirected to `raw.githubusercontent.com` which delivers the actual resource requested.

The problem is, unless explicit request headers (`Accept`) are set, a proxy server may return the last result it happens to cache. If you're lucky, it will be JSON. If not, you'll be trying to figure out why Gson has encountered a `JSON_STRING` while expecting a `JSON_OBJECT`.

PR #2565, opened 2026-06-26
feat: add .netrc support and GITLAB_TOKEN for authenticated HTTP access

maxandersen:
Until now, authenticated access to private git hosting was limited to `GITHUB_TOKEN` for github.com and the blunt `JBANG_AUTH_BASIC_USERNAME`/`JBANG_AUTH_BASIC_PASSWORD` pair that applied to everything else. If you ran a self-hosted GitLab (or Gitea, or Bitbucket Server), there was no good way to pass credentials for script/catalog fetches.

This adds `.netrc` file support — the same mechanism that curl, wget, git, and Go modules already use for per-host HTTP credentials. Most developers working with self-hosted git forges already have a `~/.netrc` file because git itself uses it. JBang now picks up those credentials automatically.

## What changed

**`.netrc` / `_netrc` parsing** (`NetrcParser.java`)

Reads `~/.netrc` on Unix/macOS and `~/_netrc` on Windows. Handles the standard format:

```
machine gitlab.mycompany.com
login __token__
password glpat-xxxxxxxxxxxx
```

Supports multi-line and single-line entries, `default` fallback entries, `account` fields, `macdef` blocks, and `#` comments. The file is parsed once and cached for the lifetime of the process.

**`GITLAB_TOKEN` env var**

For parity with `GITHUB_TOKEN` — if set, it sends `Authorization: Bearer <token>` for requests to `gitlab.com`.

**Clearer auth priority chain**

The auth logic in `addAuthHeaderIfNeeded()` is now structured as an explicit fallback chain:

1. `GITHUB_TOKEN` env var (github.com, unchanged)
2. `GITLAB_TOKEN` env var (gitlab.com, new)
3. URL userinfo — `https://user:pass@host/...` (unchanged)
4. `.netrc` file lookup by hostname (new)
5. `JBANG_AUTH_BASIC_*` env vars as global fallback (unchanged)

**Verbose logging for all auth methods**

Every auth path now logs which credential source was used when `--verbose` is active, which should help with debugging access issues.

## Typical usage

For a self-hosted GitLab, add an entry to `~/.netrc`:

```
machine gitlab.mycompany.com
login __token__
password glpat-xxxxxxxxxxxx
```

Then JBang picks it up for any fetch from that host — scripts, catalogs, `//DEPS`, `//FILES`, etc. No jbang-specific configuration needed.

For gitlab.com specifically, `export GITLAB_TOKEN=glpat-xxx` also works as a quick alternative.

## Summary by CodeRabbit

* **New Features**
  * Added support for reading credentials from `.netrc` / `_netrc` files when connecting to remote URLs, including `default` entries.
  * Introduced a `--netrc` option (and matching config) to point to a custom `.netrc` location.
* **Bug Fixes**
  * Improved HTTP authentication selection using a broader credential precedence across GitHub/GitLab tokens, URL userinfo, `.netrc`, and global Basic-auth environment variables.
* **Documentation**
  * Added detailed HTTP Authentication and `.netrc` usage, including quoting/escaping and verbose debugging guidance.
* **Tests**
  * Expanded unit tests for `.netrc` parsing and authentication behavior.


maxandersen:
@quintesse what do you think about using netrc as way to allow basic auth setup for any https access?

I realized it's commonly supported by curl and git so quite generic and allow to use ie. Self hosted gitlab access :) 

Ultimately preferred using something like git auth mechanisms but adding Call to git for every url access

quintesse:
I think it's great :+1:

maxandersen:
i added docs, support for --netrc / jbang.netrc.

PR #2567, opened 2026-06-27
fix: auto-generate non-numeric id when linking JDK with numeric version

### LINKED ISSUE #2557: regression?: jbang jdk install of specific jdk
got this from @gsmet:

jbang blocking quarkus release

```
[jbang] [2026-06-24 11:31:36] Using JDK provider(s): CurrentJdkProvider, DefaultJdkProvider, JavaHomeJdkProvider, PathJdkProvider, LinkedJdkProvider, JBangJdkProvider
Error:  [0:219] [ERROR] When providing an existing JDK path, the versionOrId parameter must be a non-integer id
java.lang.IllegalArgumentException: When providing an existing JDK path, the versionOrId parameter must be a non-integer id
    at dev.jbang.cli.Jdk$JdkInstall.doCall(Jdk.java:123)
    at dev.jbang.cli.BaseCommand.execute(BaseCommand.java:134)
    at org.aesh.command.impl.operator.EndOperator.execute(EndOperator.java:41)
    at org.aesh.command.impl.Executions$ExecutionImpl.execute(Executions.java:239)
    at org.aesh.command.impl.AeshCommandRuntime.runExecutor(AeshCommandRuntime.java:222)
    at org.aesh.command.impl.AeshCommandRuntime.executeCommand(AeshCommandRuntime.java:175)
    at org.aesh.AeshRuntimeRunner.execute(AeshRuntimeRunner.java:160)
    at dev.jbang.Main.main(Main.java:46)
```

I was doing:

> jbang --verbose jdk install java-21 ${JAVA_HOME_21_X64}
> jbang --java 21 ...

So that JBang doesn't download Java 21

 I tried to switch to using a non numeric idGuillaume Smet: but --java only supports numeric ids and I don't see any other parameter to push an install id

maxandersen:
## Problem

When a user tries to link an existing JDK using a numeric version, they hit a catch-22:

| Attempt | Result |
|---|---|
| `jbang jdk install 21 $JDK_PATH` | ❌ "id must be non-integer" |
| `jbang jdk install java-21 $JDK_PATH` then `jbang --java java-21` | ❌ "--java only supports numeric" |

The correct workaround (`jdk install java-21` + `--java 21`) was not discoverable.

## Fix

Instead of rejecting numeric IDs, auto-generate a non-numeric id using the pattern `<version>-user`:

```
$ jbang jdk install 21 /path/to/jdk
[jbang] Numeric id detected, using '21-user' as id
[jbang] JDK 21-user-linked has been linked to: /path/to/jdk
```

If `21-user-linked` already exists, it increments: `21-user-1`, `21-user-2`, etc.

The linked JDK is then found by `--java 21` via the version-based lookup in `LinkedJdkProvider`.

## Tests

- Updated `testJdkInstallWithLinkingAndIntegerId` → verifies numeric id generates `11-user-linked`
- Added `testJdkInstallWithLinkingAndIntegerIdGeneratesIncrementedUserId` → verifies counter suffix
- Updated `testJdkInstallWithLinkingToExistingJdkPathWithDifferentVersion` → verifies `13-user-linked`

Fixes #2557

### DIFF
--- a/src/main/java/dev/jbang/cli/Jdk.java
+++ b/src/main/java/dev/jbang/cli/Jdk.java
@@ -120,8 +120,8 @@ public class Jdk extends BaseCommand {
 				}
 				if (!Util.isNullOrBlankString(path)) {
 					if (isValidInteger(versionOrId)) {
-						throw new IllegalArgumentException(
-								"When providing an existing JDK path, the versionOrId parameter must be a non-integer id");
+						versionOrId = generateUserId(versionOrId);
+						Util.infoMsg("Numeric id detected, using '" + versionOrId + "' as id");
 					}
 					Path jdkCacheDir = Settings.getCacheDir(Cache.CacheClass.jdks);
 					Path jdkPath = Paths.get(path);
@@ -150,6 +150,10 @@ public class Jdk extends BaseCommand {
 			} catch (NumberFormatException e) {
 				return false;
 			}
+		}
+
+		private String generateUserId(String version) {
+			return version + "-user";
 		}
 	}
 
--- a/src/test/java/dev/jbang/cli/TestJdk.java
+++ b/src/test/java/dev/jbang/cli/TestJdk.java
@@ -350,10 +350,20 @@ public class TestJdk extends BaseTest {
 	}
 
 	@Test
-	void testJdkInstallWithLinkingAndIntegerId() {
-		assertThrows(IllegalArgumentException.class,
-				() -> checkedRun("install", "--force", "11", "/non-existent-path"),
-				"When providing an existing JDK path, the versionOrId parameter must be a non-integer id");
+	void testJdkInstallWithLinkingAndIntegerIdGeneratesUserId(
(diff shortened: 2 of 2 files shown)


quintesse (dismissed):
It's okay, although personally I wouldn't do the counter thing.
Because now, instead of being backwards compatible, we've introduced a new behaviour.
If before you would have run `jbang jdk install 21 /foo/bar` twice you would have gotten the message that 21 was already installed and to use `--force` if you really want to do that. With a counter you might end up installing more and more versions without really noticing. If you really want to install multiple 21s I think it should be an explicit act. (eg. the user explicitly giving it a unique name)

maxandersen:
Good point. what prefix would you use for the one default ?

quintesse:
> Good point. what prefix would you use for the one default ?

I'm okay with the one you chose in the PR: "user" 👍

quintesse (approved):
LGTM

PR #2568, opened 2026-06-27
fix: resolve jbang deps search hang on Windows #2549

### LINKED ISSUE #2549: `jbang deps search` do not work on Windows
Jbang 0.139.3

`jbang deps search` start but do not display any output.

arnabnandy7:
### Title
`fix: resolve jbang deps search hang on Windows` 

Fixes #2549 

### Description
This PR resolves the issue where `jbang deps search` hangs indefinitely and produces no output when run on Windows.

#### Cause
The terminal backend in JBang for the `deps search` TUI widget is implemented using TamboUI's `tamboui-aesh-backend`. 

By default, the backend's constructor instantiates a new `org.aesh.terminal.tty.TerminalConnection` to connect to the system terminal. However, because JBang is already running within an active Aesh shell session, trying to open a second native system terminal connection to the same console session on Windows results in a conflict/deadlock, causing the process to hang without displaying the UI.

#### Solution
- Modified Deps.java to extract the active `org.aesh.terminal.Connection` instance from Aesh's command invocation context using `commandInvocation.getShell().connection()`.
- Updated ArtifactSearchWidget.java to support an overloaded `search(Connection connection, String initialQuery)` method.
- Reconfigured the TUI initialization to reuse the connection if it's available:
  ```java
  AeshBackend backend = new AeshBackend(connection);
  TuiConfig config = TuiConfig.defaults().toBuilder().backend(backend).build();
  ```
- If no connection is passed (e.g. falls back to default), it defaults back to creating a new terminal connection.

### DIFF
--- a/src/main/java/dev/jbang/search/ArtifactSearchWidget.java
+++ b/src/main/java/dev/jbang/search/ArtifactSearchWidget.java
@@ -13,11 +13,13 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
+import org.aesh.terminal.Connection;
 import org.eclipse.aether.artifact.Artifact;
 
 import dev.jbang.dependencies.ArtifactResolver;
 import dev.jbang.util.Util;
 
+import dev.tamboui.backend.aesh.AeshBackend;
 import dev.tamboui.style.Color;
 import dev.tamboui.style.Modifier;
 import dev.tamboui.style.Style;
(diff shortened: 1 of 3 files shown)


src/main/java/dev/jbang/util/NetUtil.java
```diff
@@ -44,21 +44,22 @@
 import java.util.stream.Collectors;
```
maxandersen, line 44: these changes in NetUtil seems unrelated?

src/main/java/dev/jbang/search/ArtifactSearchWidget.java
```diff
@@ -101,7 +106,23 @@ public Artifact search(String initialQuery) {
 			.bind(KeyTrigger.key(KeyCode.TAB), ACTION_SEARCH_CENTRAL)
 			.bind(KeyTrigger.key(KeyCode.F5), ACTION_SEARCH_CENTRAL)
 			.build();
-		try (ToolkitRunner r = ToolkitRunner.builder().bindings(bindings).build()) {
+
+		dev.tamboui.tui.TuiConfig config = null;
```
maxandersen: please don't use fully qualified classnames unless actually required.

maxandersen:
pushed fixes for the review - will merge when test passes

arnabnandy7:
> pushed fixes for the review - will merge when test passes
> 

@maxandersen thanks for the review and remedial.

PR #2569, opened 2026-06-27
feat: add support for versions property in aliases

### LINKED ISSUE #2443: Proposal: add support for "versions" property.
In the future when versioning support is released, it would be very useful from a user perspective to be able to run a query to see what versions are available.

```json
  "camel": {
      "script-ref": "./camel/${jbang.app.version:4.18.1}/CamelJBang.java",
      "versions": {
        "4.18.1": "Recommended stable version",
        "4.17.0": "Previous stable version",
        "4.9.0": "Legacy version"
      }
    }
```

A versions query command is not necessarily required. If you run a versioned alias command and the version is not valid, then as part of the exception message the availalbe versions could be displayed.

arnabnandy7:
## Description

This PR adds support for a new `versions` property in JBang alias configurations within catalogs, addressing the feature request in #2443. 

### Key Changes:
- **`versions` Field**: Added a `versions` map (`Map<String, String>`) to the Alias.java configuration to store available versions and their description/info.
- **Placeholder Resolution**: Updated alias path resolution to evaluate property replacements dynamically on the script reference (e.g. resolving `${jbang.app.version:default_version}` in the `script-ref`).
- **Versioned Alias Invocation & Validation**:
  - Implemented parsing to recognize versioned alias references with syntaxes like `<alias>@<version>` and `<alias>@<version>@<catalog>`.
  - Validates that the requested version is present in the alias's `versions` property map. If it is invalid, an error is thrown listing the available versions.
  - If the alias does not support versioning but a version is specified, it raises an exception.
  - Sets the `jbang.app.version` system property dynamically to the requested version to enable proper property substitution.
- **Unit Tests**: Added comprehensive test coverage in TestAlias.java verifying default resolution, valid specified versions, invalid version handling, and cases where versioning is not supported.

Fixes #2443

### DIFF
--- a/src/main/java/dev/jbang/catalog/Alias.java
+++ b/src/main/java/dev/jbang/catalog/Alias.java
@@ -16,6 +16,7 @@ import com.google.gson.annotations.SerializedName;
 
 import dev.jbang.cli.ExitException;
 import dev.jbang.dependencies.DependencyUtil;
+import dev.jbang.util.PropertiesValueResolver;
 import dev.jbang.util.Util;
 
 public class Alias extends CatalogItem {
@@ -64,6 +65,7 @@ public class Alias extends CatalogItem {
 	public final List<JavaAgent> javaAgents;
 	@JsonAdapter(Catalog.SkipEmptyListSerializer.class)
 	public final List<String> docs;
+	public final Map<String, String> versions;
 
 	public static class JavaAgent {
 		@SerializedName(value = "agent-ref")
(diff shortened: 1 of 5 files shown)


maxandersen:
thanks for the attempt - this PR though does not seem to take into account the discussions / code around https://github.com/jbangdev/jbang/pull/2433

also my understanding of #2443 is the wish for being able to somhow list the possible versions available ...I don't see how that is done in this pr ?

arnabnandy7:
> thanks for the attempt - this PR though does not seem to take into account the discussions / code around https://github.com/jbangdev/jbang/pull/2433
> 
> also my understanding of #2443 is the wish for being able to somhow list the possible versions available ...I don't see how that is done in this pr ? 
> 
> 

I might have misunderstood the requirement, will close this PR as it's not aligned to the intent of original issue @maxandersen thanks for reviewing.

PR #2585, opened 2026-07-05
fix: PowerShell quoting for special chars and -D property spacing in app install

### LINKED ISSUE #2564: jbang app install -Dkey=value generates wrapper script with space-separated -D that fails at runtime
## Summary

`jbang app install -Dkey=value` generates a wrapper script where `-D` and the property are space-separated. When the wrapper script runs, `jbang run` rejects the `-D` as a standalone flag with the error: `Option -D, must be part of a property`.

## JBang Version

0.138.0

## Steps to Reproduce

```bash
jbang app install -Dcamel.jbang.version=4.21.0-SNAPSHOT camel@apache/camel/main
cat ~/.jbang/bin/camel
camel version
```

## Actual Behavior

The generated wrapper script at `~/.jbang/bin/camel` contains:

```bash
#!/bin/sh
exec jbang run -D camel.jbang.version=4.21.0-SNAPSHOT camel@apache/camel/main "$@"
```

Note the **space** between `-D` and `camel.jbang.version=4.21.0-SNAPSHOT`.

Running `camel version` (or any subcommand) fails with:

```
Option -D, must be part of a property
Builds and runs provided script. (default command)
Usage: jbang run [-hinx] ...
```

## Expected Behavior

The wrapper script should serialize the `-D` flag attached to the property (no space):

```bash
#!/bin/sh
exec jbang run -Dcamel.jbang.version=4.21.0-SNAPSHOT camel@apache/camel/main "$@"
```

## Workaround

Manually patch the wrapper script after installation:

```bash
sed -i 's/ -D / -D/g' ~/.jbang/bin/camel
```

_This issue was created with AI assistance (Claude Code on behalf of Claus Ibsen)._

### LINKED ISSUE #2584: RUNTIME_OPTIONS: -Dkey=value split on = on Windows
## Problem

On Windows, `//RUNTIME_OPTIONS` splits `-Dkey=value` arguments on the `=` sign, causing the value part to be treated as a separate argument (e.g. a class name).

## Reproduction

```java
//RUNTIME_OPTIONS -Djavafx.enablePreview=true

public class Demo {
        System.out.println(System.getProperty("javafx.enablePreview"));
    }
}
```

On macOS/Linux this works fine. On Windows:

```
Error: Could not find or load main class .enablePreview=true
Caused by: java.lang.ClassNotFoundException: /enablePreview=true
```

## Workaround

Quoting the flag preserves it as a single argument:

```java
//RUNTIME_OPTIONS "-Djavafx.enablePreview=true"
```

## Expected behavior

`-Dkey=value` should be passed as a single JVM argument on all platforms, with or without quotes.

## Context

Discovered in [jhostty](https://github.com/maxandersen/jhostty) which uses JavaFX 26 preview features requiring `-Djavafx.enablePreview=true`.

maxandersen:
turns out we had "bad" quoting rules on windows since 2020 :)

-Dwhatever.period=true breaks on powershell, works on cmd.exe 

and since 2024 our install scripts generated `-D x=y` instead of `-Dx=y` causing issues too.

This fixes both.

### 1. PowerShell safe chars too permissive (#2584)

`pwrSafeChars` included `.,;:()` as safe characters, inherited from original best-guess allowlist (47b9f342, 2020). These characters cause issues in PowerShell — args like `-Djavafx.preview=true` were passed through unquoted and got mangled.

**Fix:** Remove `.,;:()` from `pwrSafeChars`. Args containing these chars now get single-quoted in PowerShell (e.g. `'-Djavafx.preview=true'`). CMD is unchanged — these chars are safe there.

### 2. `app install -Dkey=value` splits `-D` from property (#2564)

`DependencyInfoMixin.opts()` emitted `-D` and `key=value` as two separate list elements. The generated wrapper script ended up with `-D camel.jbang.version=4.21.0` (space-separated), which `jbang run` then rejects with *"Option -D, must be part of a property"*.

**Fix:** Join them: `opts.add("-D" + key + "=" + value)`.

## Tests added

- **Unit:** `TestCommandBuffer` — PowerShell escapes `.,;:()`, CMD does not
- **Unit:** `TestApp.testAppInstallWithSystemProperty` — verifies generated scripts contain `-Dkey=value` (no space)
- **Integration:** `AppIT.shouldKeepSystemPropertyAttached` — end-to-end: `app install -Djavafx.preview=true` runs and JVM receives the property intact

Fixes #2584
Fixes #2564

### DIFF
--- a/src/main/java/dev/jbang/util/CommandBuffer.java
+++ b/src/main/java/dev/jbang/util/CommandBuffer.java
@@ -24,7 +24,7 @@ public class CommandBuffer {
 	// cmdSafeChars = Pattern.compile("[^\\Q&()[]{}^=;!'+,`~<>|\\E]*");
 	static Pattern cmdSafeChars = Pattern.compile("[a-zA-Z0-9.,_+=:;@()-\\\\]*");
 	// TODO: Figure out what the real list of safe characters is for PowerShell
-	static Pattern pwrSafeChars = Pattern.compile("[a-zA-Z0-9.,_+=:@()\\\\-]*");
+	static Pattern pwrSafeChars = Pattern.compile("[a-zA-Z0-9_+=@\\\\-]*");
 	static Pattern shellSafeChars = Pattern.compile("[a-zA-Z0-9._+=:@%/-]*");
 
 	static Pattern cmdNeedQuotesChars = Pattern.compile("[\\Q&()[]{}^=;!'+,`~\\E]");
--- a/src/it/java/dev/jbang/it/AppIT.java
+++ b/src/it/java/dev/jbang/it/AppIT.java
@@ -41,6 +41,34 @@ public class AppIT extends AbstractHelpBaseIT {
 					+ lineSeparator());
 	}
 
+	// Scenario: app install with -D should keep -D attached to property name
+	// Regression test for https://github.com/jbangdev/jbang/issues/2564
+	@Test
+	@DisabledOnOs(OS.WINDOWS)
+	@Description("app install with -D should produce -Dkey=value (no space) in wrapper script")
+	public void shouldKeepSystemPropertyAttached() {
+		assertThat(shell("jbang app install --force --name jbang-itest-app-sysprop"
+				+ " -Djavafx.preview=true -Dcamel.jbang.version=4.21.0 printjvmargs.java"))
+			.succeeded();
+		assertThat(shell("$JBANG_DIR/bin/jbang-itest-app-sysprop"))
+			.succeeded()
+			.outContains("JVMARG:-Djavafx.preview=true")
+			.outContains("JVMARG:-Dcamel.jbang.version=4.21.0");
+	}
+
+	@Test
+	@EnabledOnOs(OS.WINDOWS)
+	@Description("app install with -D should produce -Dkey=value (no space) in wrapper scripts on Windows")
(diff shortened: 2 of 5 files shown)


src/main/java/dev/jbang/util/CommandBuffer.java
```diff
@@ -24,7 +24,7 @@ public class CommandBuffer {
 	// cmdSafeChars = Pattern.compile("[^\\Q&()[]{}^=;!'+,`~<>|\\E]*");
 	static Pattern cmdSafeChars = Pattern.compile("[a-zA-Z0-9.,_+=:;@()-\\\\]*");
 	// TODO: Figure out what the real list of safe characters is for PowerShell
-	static Pattern pwrSafeChars = Pattern.compile("[a-zA-Z0-9.,_+=:@()\\\\-]*");
+	static Pattern pwrSafeChars = Pattern.compile("[a-zA-Z0-9_+=@\\\\-]*");
```
maxandersen, line 27: what is tricky is that 

`jbang com.h2database:h2:1.4.200`

is fine, but

`jbang -Dsdff=sdfd.df com.h2database:h2:1.4.200`

is not. the kicker is `-` needs more quoting than regular arguments.

for now let them be quoted (meaning on windows installed scripts will quote the arguments with GAV.

 might want to refine it in future .

quintesse (approved):
LGTM!

PR #2598, opened 2026-07-12
fix: #2548 package native binaries with Homebrew

### LINKED ISSUE #2548: JBang installed via Homebrew prints warning when using native mode
I have JBang installed via Homebrew on my mac.

After I ran this:

```
export JBANG_USE_NATIVE=true
```

Then JBang prints this warning:

```
WARNING: JBang native binary (jbang-mac-aarch64.bin or jbang.bin) not found or not executable in /opt/homebrew/Cellar/jbang/0.139.3/libexec/bin
```

(It does seem to work fine, but probably using a non-native fallback)

The contents of the `bin` directory:

```
ls -al /opt/homebrew/Cellar/jbang/0.139.3/libexec/bin
total 28520
drwxr-xr-x@ 6 wdb  admin       192 Jun 22 11:16 .
drwxr-xr-x@ 5 wdb  admin       160 Jun 22 11:20 ..
-rwxr-xr-x@ 1 wdb  admin     10892 Jun 22 11:16 jbang
-rw-r--r--@ 1 wdb  admin      4997 Feb  1  1980 jbang.cmd
-rw-r--r--@ 1 wdb  admin  14565645 Feb  1  1980 jbang.jar
-rw-r--r--@ 1 wdb  admin     11721 Feb  1  1980 jbang.ps1
```

arnabnandy7:
## Summary

Fixes #2548.

Homebrew installations now use the platform-specific native distribution, ensuring the appropriate native executable is installed alongside the regular JBang launcher and JAR.

Users can enable native execution with:

```
export JBANG_USE_NATIVE=true
```

## Changes

- Added a dedicated Brew-only JReleaser binary distribution.
- Added a multi-platform Homebrew formula for:
  - Apple Silicon (`mac-aarch64`)
  - Intel macOS (`mac-x64`)
- Added an Intel macOS native-image build to the release and CI workflows.
- Updated the Homebrew formula test to verify `JBANG_USE_NATIVE=true`.
- Preserved the existing JAR-based execution as the default.
- Removed the earlier runtime companion-download approach.
- Updated the Homebrew installation documentation.

## Rationale

Native binaries should be supplied by the package manager rather than downloaded dynamically by the startup script. This keeps installed versions aligned, avoids repeated downloads for releases without native artifacts, and preserves consistent launcher behavior.

## Verification

- Generated the multi-platform formula using JReleaser.
- Verified that the generated formula contains architecture-specific URLs and checksums.
- Verified that the formula retains the existing `libexec` installation and launcher rewrite.
- `git diff --check` passes.


maxandersen:
not sure this is the right thing - for one it can't just be fixed in jbang script - also needs updating in the others.

this would force download of binaries even if you are running a jbang version that might not have the binary present. 

This is why we print the warning for now - as users should have installed the native version first.

What would be better i think is we fix the update mechanism to take into consideration the native flag/updates.

Then things would be more consistent and less "force feeding" that could be problematic.

arnabnandy7:
@maxandersen Thanks, that makes sense. I scoped this to the Homebrew/Bash case and pinned the download to the installed version to avoid version drift, but I agree it still creates inconsistent behavior across the startup scripts and may repeatedly request a native artifact that does not exist for older releases.

I’ll step back from this approach. Handling `JBANG_USE_NATIVE` through the existing install/update mechanism would keep native artifact selection consistent and avoid introducing a separate download path in the launcher.

I’ll investigate that flow and either rework this PR around it or close this PR and follow up with a more focused proposal before implementing it.

arnabnandy7:
> I’ll investigate that flow and either rework this PR around it or close this PR and follow up with a more focused proposal before implementing it.

@maxandersen One alternative worth considering is handling this at Homebrew packaging time: install the matching native release archive using on_arm/on_intel, so `JBANG_USE_NATIVE=true` only selects an already-installed binary.

That would avoid runtime downloads and launcher changes, although it would require publishing both macOS architectures and supplying their checksums to the generated formula. Would that direction be preferable to changing the update mechanism?

maxandersen:
yes, we need to go through the various package mechanisms and see what best approach applies.

arnabnandy7:
> yes, we need to go through the various package mechanisms and see what best approach applies.

Should I then refactor and implement this alternate approach in this PR?

quintesse:
IMO this is only a temporary issue. In the near future, we will want to make the native packages the default installation package for all 3rd party installers. This does not mean that native will be the default to run, we might need a little more time for that, but at least you'll be able to use it by setting the `JBANG_USE_NATIVE` env var.

We just need to figure out how to update all the 3rd party installers so they download the correct package.

Edit: @maxandersen this is perhaps something we could do one by one, right? It's perhaps not necessary to update all installers at once? (I know you said some will not be so easy to update)

arnabnandy7:
> We just need to figure out how to update all the 3rd party installers so they download the correct package.

Thanks I'll check the feasibility of implementing it. Let's see how it goes.

arnabnandy7:
@maxandersen @quintesse I've refactored the change as per the discussion, do let me know if you liked this approach or if there are any suggestions that can be implemented.

maxandersen:
@quintesse yes i expect we'll do them bit by bit.

quintesse:
@arnabnandy7 doesn't this make the Homebrew mac-only? Homebrew also works on Linux.

arnabnandy7:
> @arnabnandy7 doesn't this make the Homebrew mac-only? Homebrew also works on Linux.

Yes good catch

arnabnandy7:
@quintesse I've fixed it now the native Homebrew distribution now supports Linux as well as macOS. Added Linux x64 and ARM64 artifacts to the JReleaser multi-platform formula, added an Ubuntu ARM64 native-image build alongside the existing x64 build, and included the corresponding WireMock fixture so the full native integration suite remains enabled on that runner. 

I also updated the installation docs. JReleaser generation now produces Linux ARM64/x64 and macOS ARM64/x64 formula branches, and spotlessCheck passes.

arnabnandy7:
The ARM CI failure was caused by AlpineJavaDownloadIT using the AMD64-only `ghcr.io/home-assistant/amd64-base:3.19` image. 

I replaced it with the multi-architecture `alpine:3.19` image, which preserves the required musl environment while allowing Docker to select the appropriate architecture on both x64 and ARM64 runners.

PR #2600, opened 2026-07-12
fix: avoid resolving artifact dependencies twice

### LINKED ISSUE #2577: Resolving dependencies twice
:wave:

Not sure if it's a bug but laterly, when executing a JBang script that is in an artifact, I get:

```
[jbang]    io.quarkus.bot:conversational-release-action:999-SNAPSHOT
[jbang] Dependencies resolved
[jbang]    io.quarkus.bot:conversational-release-action:999-SNAPSHOT
[jbang] Dependencies resolved
```

as if it was resolving the dependencies twice, which looks quite odd.

arnabnandy7:
## Description

When running a Maven artifact, JBang resolved its dependencies once to locate the artifact JAR and again when constructing the runtime classpath. The dependency cache usually hid the duplicate work, but both resolution passes were visible for snapshots or when using `--fresh`.

Fixes #2577.

This change reuses the classpath obtained during artifact resolution when creating the build context, while preserving the GAV in the project’s dependency metadata.

## Testing

- Added a regression test that runs a Maven artifact with `--fresh` and verifies that dependency resolution occurs exactly once.
- Ran:

  ```
  ./gradlew test --tests "dev.jbang.cli.TestRun.testGAVDependenciesAreResolvedOnceWhenFresh"
  ```

- Ran `./gradlew spotlessApply`.
- Ran `git diff --check`.
- Started the complete `TestRun` test class, but it exceeded the local 120-second command window.


maxandersen:
Looks like the right fix - but would like to hear from @quintesse if this "carry of state" is acceptable as we didnt need it before.

**The problem:** When running a GAV artifact (e.g. `jbang run info.picocli:picocli-codegen:4.6.3`), dependencies resolve twice. The first resolution happens in `ProjectBuilder.resolveDependency()` when the `GavResourceResolver` lazy ref is materialized (via `getFile()`). The second happens in `BuildContext.resolveClassPath()` during the build/run phase — it has its own `mcp` field, starts at `null`, and resolves from scratch.

**The fix correctly** bridges the gap: stash the already-resolved `ModularClassPath` on the `Project`, then seed `BuildContext.mcp` from it in the constructor. Minimal plumbing, no behavior change beyond skipping the redundant resolve.

**How this regressed:** This was introduced by commit [`4aa562a7`](https://github.com/jbangdev/jbang/commit/4aa562a7) ("refactor: extracted build state from `Project`", Feb 2023). Before that refactor, `resolveClassPath()` lived on `Project` and shared the same resolved state. When build state was extracted into `BuildContext`, the `mcp` field got a fresh copy with no way to receive the already-resolved value from `ProjectBuilder`. Ironically, the earlier fix [`7b978e09`](https://github.com/jbangdev/jbang/commit/7b978e09) / [#1553](https://github.com/jbangdev/jbang/pull/1553) ("Fix double resolve") had fixed a similar issue just 11 days before, but the subsequent extraction re-introduced a new variant of it specifically for the GAV artifact path.

quintesse (approved):
👍

PR #2606, opened 2026-07-12
fix: update build.gradle to support Gradle 9.6.1 and Java 25

wfouche:
Gradle 8.14.5 only supports Java 8 to 24. Upgrading to Gradle 9.6.1 allows Gradle to be run using Java 25.


maxandersen:
Was there issue with this? Had on my todo to update to Gradle 9.

wfouche:
The build tests worked on my PC but then the Github pipeline failed.

You're welcome to recopen it, maybe one can get it to work with a bit more fiddling. :-)

PR #2616, opened 2026-07-18
fix: honor JBANG_USE_NATIVE when updating

### LINKED ISSUE #2611: Update should take into account JBANG_USE_NATIVE
Right now the update only downloads the non-native package

arnabnandy7:
## Summary

Fixes #2611.

`jbang version --update` now respects `JBANG_USE_NATIVE`. When native mode is enabled, JBang downloads and installs the release bundle matching the current operating system and architecture instead of always downloading the generic JAR-based package.

On Windows, an in-use native executable is staged as `.exe.new` and replaced by the launcher on the next invocation.

The `JBANG_USE_NATIVE` installation documentation and relevant tests have also been updated.

## Validation

- `./gradlew spotlessCheck compileTestJava`
- PowerShell tests in `TestScriptNativeDownload`

### DIFF
--- a/src/main/java/dev/jbang/cli/App.java
+++ b/src/main/java/dev/jbang/cli/App.java
@@ -63,7 +63,7 @@ public class App extends BaseCommand {
 
 	@CommandDefinition(name = "install", description = "Install a script as a command.", generateHelp = true)
 	public static class AppInstall extends BaseBuildCommand {
-		private static final String jbangUrl = "https://www.jbang.dev/releases/latest/download/jbang.zip";
+		private static final String jbangDownloadBaseUrl = "https://www.jbang.dev/releases/latest/download/";
 
 		@Mixin
 		RunMixin runMixin;
@@ -214,6 +214,7 @@ public class App extends BaseCommand {
 		public static boolean installJBang(boolean force) throws IOException {
 			Path binDir = Settings.getConfigBinDir();
 			boolean managedJBang = Files.exists(binDir.resolve("jbang.jar"));
+			boolean useNative = Boolean.parseBoolean(System.getenv(Settings.ENV_USE_NATIVE));
 
 			if (!force && (managedJBang || Util.searchPath("jbang") != null)) {
 				Util.infoMsg("jbang is already available, re-run with --force to install anyway.");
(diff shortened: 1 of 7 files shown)


quintesse (changes requested):
Very nice!

I would like to see one change though, Right now it only uses the JBANG_USE_NATIVE environment variable to determine if it should install a native bundle or not.

I would suggest that if you're currently running the native version it should _also_ take that as a sign to install the native bundle. (Who knows, perhaps the user simply ran the native executable manually instead of setting JBANGT_USE_NATIVE)

arnabnandy7:
@quintesse the requested behavior is now implemented. Native bundle selection will now triggered by either `JBANG_USE_NATIVE=true` or `JavaUtil.inNativeImage()`.

src/main/java/dev/jbang/cli/App.java
```diff
@@ -214,6 +215,7 @@ private static void installPSScript(Path file, String scriptRef, List<String> ru
 		public static boolean installJBang(boolean force) throws IOException {
 			Path binDir = Settings.getConfigBinDir();
 			boolean managedJBang = Files.exists(binDir.resolve("jbang.jar"));
+			boolean useNative = shouldUseNative(JavaUtil.inNativeImage());
```
quintesse: Is there a reason for passing this as an argument instead of simply adding it in the method itself?

quintesse (changes requested):

src/main/java/dev/jbang/cli/App.java
arnabnandy7: @quintesse The argument is only there to keep the decision logic directly unit-testable on a regular JVM. `JavaUtil.inNativeImage()` is static runtime state, so passing its value lets the test cover both native and non-native execution without mutating global state. 

There’s no production requirement for the parameter, though if you prefer the simpler method signature, I’m happy to inline the call.

koppor: Maybe all of this is AI generated.

arnabnandy7: @koppor thanks for sharing the thoughts but it's not.

I used to generate non-confrontal, polite and professional replies but I guess it's not required for GitHub.

quintesse: @arnabnandy7 that might be, but I would argue that it is a much worse API just to be able to do testing. We have ways in our tests to set environment variables just for those tests that need it.

In many cases being able to inject information into code is perfect to decouple and at the same time improve testing, but in this case the code is so trivial that injecting the information makes the method almost useless.

So I'd prefer to change that and just find another way to perform the test.

arnabnandy7: Sure, I'll think through and find a better solution rather than patching it up.

src/main/scripts/jbang.cmd
```diff
@@ -12,6 +12,11 @@ rem detect architecture for platform-specific binary lookup
 set "jbang_arch=x64"
 if "%PROCESSOR_ARCHITECTURE%"=="ARM64" set "jbang_arch=aarch64"
 
+if exist "%~dp0jbang.bin-windows-%jbang_arch%.exe.new" (
+  copy /y "%~dp0jbang.bin-windows-%jbang_arch%.exe.new" "%~dp0jbang.bin-windows-%jbang_arch%.exe" > nul 2>&1
```
quintesse, line 16: Why not use `move` ?

quintesse: Ah, because that's what we do with jars as well, strike this

quintesse (dismissed):
LGTM, it works , any improvements can always be done at a later time

arnabnandy7:
@quintesse fixed merge conflict in one of the file, please re-review

quintesse (approved):

maxandersen (approved):

PR #2617, opened 2026-07-18
fix: avoid replacing active Windows command launcher

### LINKED ISSUE #2608: Update shows weird error message (on Windows)
Got this when running an update on Windows:

```
>jbang version --update
[jbang] Downloading and installing jbang...
[jbang] There is a new version of jbang available!
[jbang] You have version 0.139.3 and 0.141.0 is the latest.
[jbang] Run 'jbang version --update' to update to the latest version.
'_with_cli' is not recognized as an internal or external command,
operable program or batch file.
'""' is not recognized as an internal or external command,
operable program or batch file.
```

The update worked though because I'm now running 0.141.0 and there doesn't seem to be any problem.

arnabnandy7:
Fixes #2608.

## What changed

- Prevent the updater from overwriting the currently executing `jbang.cmd` launcher on Windows.
- Continue updating the JBang JAR and other startup scripts normally.
- Add regression tests covering active and inactive Windows command launchers.
- Document `jbang version --update` and package-manager upgrade behavior.

## Root cause

`cmd.exe` reads batch files incrementally. During `jbang version --update`, JBang replaced the active `jbang.cmd` file. CMD then continued reading the new file from its previous byte offset, causing fragments such as `_with_cli` and `""` to be interpreted as commands.

## Verification

- `./gradlew spotlessCheck`
- `git diff --check`

### DIFF
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -74,6 +74,17 @@ For example `iex "& { $(iwr -useb https://ps.jbang.dev) } properties@jbangdev"`
 
 Unzip the https://github.com/jbangdev/jbang/releases/latest[latest binary release], add the `jbang-<version>/bin` folder to your `$PATH` and you are set.
 
+=== Updating a JBang-managed installation
+
+If JBang was installed using the universal installation script, update it with:
+
+[source,shell]
+----
+jbang version --update
+----
+
+When JBang was installed using a package manager such as SDKMAN, Homebrew, Chocolatey, or Scoop, use that package manager's upgrade command instead.
+
 === Wrapper install
 
 If you would like to have `jbang` available in a local directory and committed into a source code repository (akin to Maven and Gradle wrappers) you can use the `jbang wrapper` command.
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -74,20 +74,6 @@ For example `iex "& { $(iwr -useb https://ps.jbang.dev) } properties@jbangdev"`
 
 Unzip the https://github.com/jbangdev/jbang/releases/latest[latest binary release], add the `jbang-<version>/bin` folder to your `$PATH` and you are set.
 
-=== Updating a JBang-managed installation
-
-If JBang was installed using the universal installation script, update it with:
-
-[source,shell]
-----
-jbang version --update
-----
-
-When JBang was installed using a package manager such as SDKMAN, Homebrew, Chocolatey, or Scoop, use that package manager's upgrade command instead.
-
-On Windows, updates to `jbang.cmd` are staged as `jbang.cmd.new` before being moved into place.
-When an update is started from Command Prompt, the move happens after the Java process completes so the batch file is not replaced while Command Prompt is still reading it.
-
 === Wrapper install
 
 If you would like to have `jbang` available in a local directory and committed into a source code repository (akin to Maven and Gradle wrappers) you can use the `jbang wrapper` command.
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -85,8 +85,8 @@ jbang version --update
 
 When JBang was installed using a package manager such as SDKMAN, Homebrew, Chocolatey, or Scoop, use that package manager's upgrade command instead.
 
-On Windows, an update started from Command Prompt stages the new `jbang.cmd` launcher and installs it after the Java process completes.
-This avoids replacing a batch file while it is still being read by Command Prompt.
+On Windows, updates to `jbang.cmd` are staged as `jbang.cmd.new` before being moved into place.
+When an update is started from Command Prompt, the move happens after the Java process completes so the batch file is not replaced while Command Prompt is still reading it.
 
 === Wrapper install
 
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -74,6 +74,20 @@ For example `iex "& { $(iwr -useb https://ps.jbang.dev) } properties@jbangdev"`
 
 Unzip the https://github.com/jbangdev/jbang/releases/latest[latest binary release], add the `jbang-<version>/bin` folder to your `$PATH` and you are set.
 
+=== Updating a JBang-managed installation
+
+If JBang was installed using the universal installation script, update it with:
+
+[source,shell]
+----
+jbang version --update
+----
+
+When JBang was installed using a package manager such as SDKMAN, Homebrew, Chocolatey, or Scoop, use that package manager's upgrade command instead.
+
+On Windows, an update started from Command Prompt keeps the currently running `jbang.cmd` launcher to avoid replacing a batch file while it is being executed.
+The JBang JAR and the other startup scripts are still updated.
+
 === Wrapper install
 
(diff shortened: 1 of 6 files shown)


maxandersen:
damn that is nasty :) 

so jbang.cmd will always stay on old version ? no updates possible?

arnabnandy7:
> damn that is nasty :)
> 
> so jbang.cmd will always stay on old version ? no updates possible?

Thanks for pointing it out, with the current patch the active `jbang.cmd` would remain on its old version indefinitely. That avoids corrupting the running batch file but is not a complete update strategy. 

I’ll revise it to stage the new launcher and replace it through the existing post-Java command handoff, ensuring CMD exits from an already parsed command after replacement.

src/test/java/dev/jbang/cli/TestApp.java
```diff
@@ -60,6 +61,58 @@ void testHasJBangSetup(@TempDir Path tempDir) throws IOException {
 		assertThat(App.AppSetup.hasJBangSetup(rcFile), is(true));
 	}
 
+	@Test
+	void testActiveWindowsCmdLauncher() {
+		Assumptions.assumeTrue(Util.isWindows());
```
quintesse: If the test should only run on WIndows, use: `@EnabledOnOs(OS.WINDOWS)`

src/main/java/dev/jbang/cli/App.java
```diff
@@ -276,6 +280,33 @@ private static void copyJBangFiles(Path from, Path to) throws IOException {
 					}
 				});
 		}
+
+		static boolean isActiveCmdLauncher(Path target) {
+			if (!Util.isWindows() || Util.getShell() != Util.Shell.cmd) {
```
quintesse: This check and others below seem to suggest this doesn't work when running Powershell?

src/main/java/dev/jbang/cli/App.java
```diff
@@ -276,6 +280,33 @@ private static void copyJBangFiles(Path from, Path to) throws IOException {
...
+				return false;
+			}
+			String launchCmd = System.getenv(Util.JBANG_LAUNCH_CMD);
+			if (launchCmd == null || launchCmd.isEmpty()) {
+				return false;
+			}
+			return target.toAbsolutePath().normalize().equals(Paths.get(launchCmd).toAbsolutePath().normalize());
```
quintesse: You can safely use `Files.isSameFile()` here

quintesse: Sorry, I meant `Util.isSameFile()` , so you don't need to worry about exceptions

quintesse (changes requested):
So I have added some smaller remarks to the code, but in general I'm not sure it's even necessary to do all the "active launcher" checking. In the case of WIndows we always copy the jar to ".jar.new", without checking if it's the active jar or not. This is because in most normal situations it _will_ be the active launcher. In the rare case it isn't it doesn't really hurt to go through the new-and-move dance. So I'd remove that and _always_ copy the binary on Windows using ".new".

quintesse:
I haven't tested this but won't we have the same issue with `.ps1` files? Or are those read into memory once? And if so will they allow themselves to be overwritten?

src/main/java/dev/jbang/cli/App.java
quintesse: Or do PowerShell .ps1 files not have this issue?

src/test/java/dev/jbang/cli/TestApp.java
arnabnandy7: I’ll replace the runtime assumption with `@EnabledOnOs(OS.WINDOWS)`

src/main/java/dev/jbang/cli/App.java
arnabnandy7: The deferred handoff is CMD-specific because cmd.exe reads the batch file incrementally. 

PowerShell parses the script before executing it, so replacing `jbang.ps1` should not produce the same offset problem. 

I’ll add/adjust coverage using the real PowerShell launcher to verify that it can update itself safely.

src/main/java/dev/jbang/cli/App.java
arnabnandy7: cool, thanks for this pointer

arnabnandy7:
> I haven't tested this but won't we have the same issue with `.ps1` files? Or are those read into memory once? And if so will they allow themselves to be overwritten?

PowerShell should read and parse the script before execution, unlike CMD’s incremental batch-file reading, and the current updater has historically overwritten `jbang.ps1` successfully. 

Still, that deserves an explicit regression test. I’ll verify it using the real script as part of this revision.

docs/modules/ROOT/pages/installation.adoc
```diff
@@ -74,6 +74,20 @@ For example `iex "& { $(iwr -useb https://ps.jbang.dev) } properties@jbangdev"`
...
+If JBang was installed using the universal installation script, update it with:
+
+[source,shell]
+----
+jbang version --update
+----
+
+When JBang was installed using a package manager such as SDKMAN, Homebrew, Chocolatey, or Scoop, use that package manager's upgrade command instead.
+
+On Windows, updates to `jbang.cmd` are staged as `jbang.cmd.new` before being moved into place.
```
quintesse: This is very technical information that I don't think is of any interest to users. We can leave it out.

quintesse (dismissed):
LGTM! Thanks for your work and patience!

arnabnandy7:
> LGTM! Thanks for your work and patience!

@quintesse Should I remove the document from installation file?

quintesse:
> Should I remove the document from installation file?

@arnabnandy7 yes please. I just didn't want to make it a required change. But if you have some time to do it that would be great.

docs/modules/ROOT/pages/installation.adoc
arnabnandy7: Reverted

arnabnandy7:
> > Should I remove the document from installation file?
> 
> @arnabnandy7 yes please. I just didn't want to make it a required change. But if you have some time to do it that would be great.

It's done

docs/modules/ROOT/pages/installation.adoc
```diff
@@ -74,20 +74,6 @@ For example `iex "& { $(iwr -useb https://ps.jbang.dev) } properties@jbangdev"`
...
-If JBang was installed using the universal installation script, update it with:
-
-[source,shell]
-----
-jbang version --update
-----
-
-When JBang was installed using a package manager such as SDKMAN, Homebrew, Chocolatey, or Scoop, use that package manager's upgrade command instead.
-
-On Windows, updates to `jbang.cmd` are staged as `jbang.cmd.new` before being moved into place.
```
quintesse: Sorry! I didn't mean to revert the entire docs section! Just these two lines! THese two lines are just very technical and not interesting for users

quintesse (dismissed):

docs/modules/ROOT/pages/installation.adoc
arnabnandy7: I'll revert the documents and exclude these two from the same.

It'll take sometime. Please bear with me. Thanks for re-review.

quintesse: No , thank _you_ for having the patience!

arnabnandy7: @quintesse it's added back excluding the technical zargons as you mentioned.

quintesse: Great! Thanks!

quintesse (approved):

maxandersen:
@all-contributors add @arnabnandy7 for code

PR #2620, opened 2026-07-19
feat: prefer relative symbolic links

### LINKED ISSUE #2573: symlinks should be relative where possible
Right now all symbolic links are created with absolute paths which can cause issues if you ever want to mount jbang's config folder into a container. Some (at least on Windows) container runtimes disallow absolute symlinks by default because they have no easy way to check if it points outside the attached volume or not.

If we make them relative where possible (eg. when they point to a folder or file in the same subtree) that issue would be somewhat alleviated.

Now on Windows we default to Junctions which cannot be relative, so perhaps we should try to default to Sym Links if the user has enabled that feature in Windows and fall back to Junctions when they have not?

arnabnandy7:
## Summary

- Create relative symbolic links when the target is beneath the link’s parent directory.
- Prefer symbolic links for Windows directories when supported.
- Fall back to directory junctions when Windows symbolic-link creation is unavailable.
- Document the behavior for relocated or container-mounted JBang directories.
- Add unit tests for relative and absolute link-target selection.

Fixes #2573.

## Validation

- `./gradlew test --tests "dev.jbang.util.TestUtil"`
- `./gradlew spotlessCheck`
- Compilation completed successfully.

### DIFF
--- a/docs/modules/ROOT/pages/installation.adoc
+++ b/docs/modules/ROOT/pages/installation.adoc
@@ -28,6 +28,8 @@ Once you have jbang installed, it is recommended you run `jbang app setup`. This
 [NOTE]
 ====
 By default `jbang` uses `~/.jbang` for configurations and build cache. This can be changed using the environment variable `JBANG_DIR`.
+
+Symbolic links use relative targets when the target is beneath the link's parent directory. This allows linked directory trees, such as the JBang directory, to remain usable when they are moved or mounted into a container. On Windows, JBang prefers symbolic links for directories when they are available and falls back to directory junctions otherwise.
 ====
 
 == Installation Methods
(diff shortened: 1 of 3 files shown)


quintesse (changes requested):
Looks good!

But I'd like a small change: the method `getLinkTarget()` is too generic for me, we already have a "target" at that point, we don't need to "get" it. Instead it tries to make a target relative if it can, the name should reflect that. If you could come up with something slightly better everything will be okay.

arnabnandy7:
@quintesse it's done, please check and suggest if any other improvement is needed

quintesse (approved):
It won't win any beauty contests, but I'm fine with it

PR #2623, opened 2026-07-21
fix: jbang build --native from existing/pre-built jar

mukel:
Huge fan of jbang! I also maintain the AUR [jbang](https://aur.archlinux.org/packages/jbang) for Arch.

### Issue description
When installing a native image from an existing/pre-built jar e.g. `jbang app install --native myapp.jar`,  
the `native-image` invocation fails with the following error:

```
Error: Writing image to non-existent directory /home/mukel/.jbang/cache/jars/myapp.jar.e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 is not allowed.  
Create the missing directory if you want the image to be written to that location.
```

Note that this works when building from sources because the folder is created by jbang's when building from sources.

### Implemented fix
Always create the output folder for `native-image` explicitly.  
Regression test included.

### DIFF
--- a/src/main/java/dev/jbang/source/AppBuilder.java
+++ b/src/main/java/dev/jbang/source/AppBuilder.java
@@ -109,6 +109,8 @@ public abstract class AppBuilder implements Builder<CmdGeneratorBuilder> {
 		}
 
 		if (nativeBuildRequired) {
+			Path outputDir = ctx.getNativeImageFile().getParent();
+			outputDir.toFile().mkdirs();
 			if (integrationResult.nativeImagePath != null) {
 				Files.move(integrationResult.nativeImagePath, ctx.getNativeImageFile());
 			} else {
--- a/src/test/java/dev/jbang/source/TestBuilder.java
+++ b/src/test/java/dev/jbang/source/TestBuilder.java
@@ -8,6 +8,7 @@ import static org.hamcrest.Matchers.*;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
@@ -416,6 +417,30 @@ public class TestBuilder extends BaseTest {
 		runBuild(ctx, (ctxx, optionList) -> {
 		}, null, null, (ctxx, optionList) -> callCount.incrementAndGet());
 		assertThat(callCount.get(), equalTo(2));
+	}
+
+	@Test
+	void testNativeFromExecutableArchive() throws IOException {
+		// A native build from a pre-built jar skips the jar step that creates the
+		// output dir.
+		Path jar = examplesTestFolder.resolve("hellojar.jar");
+
+		ProjectBuilder pb = Project.builder().nativeImage(true);
+		Project prj = pb.build(jar.toString());
+		assertThat(prj.isExecutableArchive(), is(true));
+		BuildContext ctx = BuildContext.forProject(prj);
+
+		runBuild(ctx, null, null, null, (ctxx, optionList) -> {
+			try {
+				// Copy the input .jar into the executable output file to simulate something was
+				// "built", note that native-image is not executed here.
+				Files.copy(jar, ctxx.getNativeImageFile());
+			} catch (IOException e) {
+				throw new RuntimeException(e);
+			}
+		});
+
+		assertThat(Files.exists(ctx.getNativeImageFile()), is(true));
 	}
 
 	@Test


quintesse:
Do we even support `jbang --native myapp.jar`? Not saying we shouldn't , in fact I think we should, but I'm just wondering if the issue isn't about `app install` at all, but a more basic issue of JBang not supporting that use-case right now.

Edit: after reading the code in this PR, it's obvious the answer is "no, we don't"

Edit2: edited the PR title to reflect this

quintesse:
Looking at the CI output it seems you need to run `./gradlew spotlessApply` and re-submit @mukel

src/main/java/dev/jbang/source/AppBuilder.java
```diff
@@ -109,6 +109,9 @@ public CmdGeneratorBuilder build() throws IOException {
 		}
 
 		if (nativeBuildRequired) {
+			// The jar build step that would create the output directory is skipped
```
quintesse: No need for the comments

src/main/java/dev/jbang/source/AppBuilder.java
```diff
@@ -109,6 +109,9 @@ public CmdGeneratorBuilder build() throws IOException {
...
+			// for executable archives, so make sure it exists before writing.
+			Files.createDirectories(ctx.getNativeImageFile().getParent());
```
quintesse: Change this to how other places in this code do it (see lines 90 and 93).

src/test/java/dev/jbang/source/TestBuilder.java
```diff
@@ -418,6 +421,35 @@ void testNativeFresh() throws IOException {
...
+		runBuild(ctx, null, null, null, (ctxx, optionList) -> {
+			try {
+				Files.createFile(ctxx.getNativeImageFile());
+			} catch (IOException e) {
+				throw new RuntimeException(e);
+			}
+		});
+	}
+
+	// A minimal, valid executable jar: just a manifest with a main class.
```
quintesse: Instead of creating a JAR just use the `itest/hellojar.jar`

quintesse (changes requested):
In general a good fix, just need to dot the I-s

mukel:
Formatting done.
Indeed it is rather a jbang build issue. For context here's my use-case:

I use `jbang` to install CLI tools, in particular `pj`, a pi-inspired coding agent written in Java.

I can install it as a jar with `jbang` but startup is not great.  

The `pj` fat jar is spiced for Native Image, installing it with `--native` make it fly, but it still runs `jbang` to launch (which spawn a JVM) ... only if `jbang` were also a native image ... AND it turns out it recently shipped as one; so `jbang` is now both simple and FAST for installing/managing local CLI tools written in Java.

src/main/java/dev/jbang/source/AppBuilder.java
```diff
@@ -109,6 +109,9 @@ public CmdGeneratorBuilder build() throws IOException {
 		}
 
 		if (nativeBuildRequired) {
+			Path outputDir = ctx.getNativeImageFile().getParent();
+			Util.deletePath(outputDir, true);
```
quintesse: Sorry, I didn't mean to suggest to also copy the `deletePath()`, that will most likely delete more than you'd want😉

src/test/java/dev/jbang/source/TestBuilder.java
```diff
@@ -418,6 +419,26 @@ void testNativeFresh() throws IOException {
...
+		Path jar = examplesTestFolder.resolve("hellojar.jar");
+
+		ProjectBuilder pb = Project.builder().nativeImage(true);
+		Project prj = pb.build(jar.toString());
+		assertThat(prj.isExecutableArchive(), is(true));
+		BuildContext ctx = BuildContext.forProject(prj);
+
+		runBuild(ctx, null, null, null, (ctxx, optionList) -> {
+			try {
+				Files.createFile(ctxx.getNativeImageFile());
```
quintesse: Instead if blindly creating a file I think it would be a better idea to copy the input file (which will be a jar) to the output file (which is of course then still a jar but with the name of an executable).

Why? Because it would have picked up the fact that with the `deletePath()` you added you might have deleted the input file.

quintesse: It's also a good idea to explicitly test at then end of the test method that the native image file actually exists. Yes, we didn't get an error, but perhaps something else happened and the file was either never created or deleted by accident.

quintesse: Btw, if you have any questions about how to do this, just ask!

mukel: Added the copy with a comment + assert that "built" file exists at the end.

quintesse (approved):
LGTM!

quintesse:
Thanks for your contribution and your patience @mukel !

mukel:
I've been a `jbang` user and fan for a long time, finally got the chance to contribute!  
BTW The native `jbang` is a killer feature.

quintesse:
> BTW The native jbang is a killer feature.

Yeah! We wanted that already for such a long time, but it was hard so we kept postponing it. I'm glad @maxandersen finally found the time to set up the basis at least (which is the hardest).

maxandersen:
Very nice. Where is that pj ? I have a project of same name but just a poor shell of an idea atm :)

mukel:
See [pijota.org](https://pijota.org) and https://github.com/mukel/pijota

PR #2625, opened 2026-07-22
docs: add a jbang agent skill

koppor:
I wanted to tell Claude how to install and use JBang. I didn't saw any skill.

https://www.skills.sh/ is a nice index

Example "rendering" if one installed a skill available https://www.skills.sh/nbbrd/heylogs and https://www.skills.sh/jabref/jabref


docs/modules/ROOT/pages/installation.adoc
```diff
@@ -50,6 +50,73 @@ Windows Powershell:
...
+[cols="2,3", options="header"]
+|===
+| Platform                      | Bundle
+
+| Linux x64 (glibc)             | `jbang-linux-x64.tar` / `jbang-linux-x64.zip`
+| macOS aarch64 (Apple Silicon) | `jbang-mac-aarch64.tar` / `jbang-mac-aarch64.zip`
+| Windows x64                   | `jbang-windows-x64.zip`
+|===
+
+WARNING: On any other platform (macOS x64, Linux aarch64, Windows ARM64, Alpine/musl, AIX) no native bundle is published. The startup script would try to download a non-existing bundle and the installation fails, so leave `JBANG_USE_NATIVE` unset there.
```
quintesse: Is this true? I never tried it but a failing installation would be a bug.

maxandersen: Yes. It should only print it couldn't find the native binary and then run plain Java version.

docs/modules/ROOT/pages/installation.adoc
```diff
@@ -50,6 +50,73 @@ Windows Powershell:
 iex "& { $(iwr -useb https://ps.jbang.dev) } app setup"
 ----
 
+=== Native Binary
+
+Every release also ships a GraalVM native-image build of `jbang` itself.
+It starts faster and needs no JVM to launch `jbang`, which is useful in containers, on CI agents, and on machines without a JDK.
```
maxandersen: so i get why you write it but how much of that is true? :)

it starts faster - definitel
needs no JVM jvm to launch jbang - true, but you need a jvm to use output of most jbang commands ...
on machines without a JDK - jbang fixed that already without being native

...what I'm saying - the only reason to use native is speed (IMO); if your CI is just calling it once - havinga preinstalled jdk is probably faster/more effecient than installing jbang native only.

skills/jbang/SKILL.md
```diff
@@ -0,0 +1,177 @@
+---
+name: jbang
```
maxandersen, line 2: yes we should add a skill - but can you open that as a separate PR from installation.adoc native improvements.

I have poc for how jbang should work with skills better that makes sense to combine with this

docs/modules/ROOT/pages/installation.adoc
koppor: I wanted to promote the compiled native image :) - but while doing it, I learned that jbang startup is kind of a shell-script downloading some JDK.

What is unclear to me (I didn't check), if the JBang-downloaded JDK matches the JDK of the script called. Meaning: If the script requires JDK21, does JBang download that or "just" the latest LTS JDK?

My (envisioned) "target audience" are users not running Java at all - or "just" having a JRE available in a jpackaged-application (which is hard to discover by JBang). And I want to save download bandwidth for them.

Maybe, one could reformulate the instructions to say: Only use native image if no JDK is available?

skills/jbang/SKILL.md
koppor: Yeah, sorry for the merge -- PR is https://github.com/jbangdev/jbang/pull/2626

docs/modules/ROOT/pages/installation.adoc
maxandersen: if a script does not say anything we run with the java available.
if //JAVA 11+ it will run with the java available if 11 or above, if not i.e. you started jbang with java 8 then it will download java 11.
etc.

PR #2629, opened 2026-07-26
fix: deps search hanging on Windows due to stdout capture

### LINKED ISSUE #2621: Windows support for jbang deps search
On my development machine running Windows 11, the `jbang deps search` command never displays the TUI. It quickly blocks, as it its CPU usage drops to zero and then it just hangs forever. The machine dual boots into either Win11 or Linux. The same commands works fine on Linux.

Image

After pressing Ctrl-C twice, the Win11 display changes to:

Image

Bhumika-SN:
Fixes #2621

## Problem

`jbang deps search` hangs indefinitely on Windows 11 when run via
`jbang.cmd` or `jbang.ps1`. The interactive search box renders its header
once, then never responds to keyboard input — no crash, no error, just a
silent freeze at 0% CPU.

## Root cause

Both Windows launcher scripts capture the spawned Java process's stdout
before deciding how to proceed:

- `jbang.cmd` redirects to a temp file: `"%CMD%" > "%tmpfile%" ... -jar "%jarPath%" %*`
- `jbang.ps1` captures into a variable: `$output = & "$javaExec" ... -jar "$jarPath" $args`

This exists to support a legitimate mechanism: some commands (e.g. `run`)
can exit with code `255` and print a generated native command to stdout,
which the launcher then reads from the temp file/variable and re-executes.
This avoids a second JVM startup for the common case.

For `deps search`, which is a live interactive TUI (built on TamboUI +
Aesh), this capture is fatal: with stdout no longer attached to a real
console, Aesh cannot detect a native Windows console and falls back to
`WinExternalTerminal`, an implementation that does plain blocking
`FileInputStream` reads instead of native keystroke-level input (e.g.
`ReadConsoleInput`). The result is a UI that draws once and then never
receives further input.

### Evidence

Thread dump while hung, captured via `jstack`, shows the terminal pump
thread stuck here:

"org.aesh.terminal.tty.impl.WinExternalTerminal@... input pump thread"
at java.io.FileInputStream.readBytes(Native Method)
at org.aesh.terminal.tty.impl.ExternalTerminal.pump(ExternalTerminal.java:88)

Confirmed the mechanism two ways:
- Running the JAR directly (`java -jar jbang.jar deps search`, bypassing
  both launcher scripts entirely) works correctly — full TUI renders and
  responds to keystrokes as expected. This isolates the bug to the
  launchers, not to TamboUI/Aesh/`ArtifactSearchWidget` itself.
- Reproduced the identical hang independently via both `jbang.cmd` and
  `jbang.ps1`, confirming it's the shared capture-then-inspect pattern in
  both scripts, not a bug specific to one shell.

## Fix

Both launcher scripts now detect `deps search` before launching Java, and
for that case only, run Java directly with stdout left attached to the
console — skipping the temp-file/variable-capture path and the
exit-code-255 re-exec logic entirely. The exit code is passed straight
through.

Every other command (`run`, `build`, `edit`, etc.) is unaffected — the
change is gated behind an explicit check for `deps search` as the first
two arguments, so the existing capture/re-exec behavior is untouched for
everything else.

This is intentionally narrow rather than a general "detect interactivity"
mechanism, since that would be a bigger design change. Happy to discuss
a more general approach (e.g. a marker on `BaseCommand` subclasses that
need a live console) if that's preferred long-term — this PR fixes the
immediate bug with minimal surface area.

## Testing

- Manually verified on Windows 11 (JDK 21.0.10): both `jbang.cmd deps
  search` and `jbang.ps1 deps search` now render the full TUI and filter
  results live as expected, matching the working `java -jar` baseline.
- Ran the full test suite (`gradlew test`) before and after this change.
  10 pre-existing failures in `TestEdit`, `TestEditWithPackage`, and
  `TestEditLocate` are present identically with the change stashed out,
  confirming they're unrelated to this fix (environment-specific on this
  machine, not caused by this PR).
- No changes to any `.java` source — this PR only touches
  `src/main/scripts/jbang.cmd` and `src/main/scripts/jbang.ps1`.

### DIFF
--- a/src/it/java/dev/jbang/it/DepsSearchLauncherIT.java
+++ b/src/it/java/dev/jbang/it/DepsSearchLauncherIT.java
@@ -1,0 +1,155 @@
+package dev.jbang.it;
+
+import static dev.jbang.it.CommandResultAssert.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.jar.Attributes;
+import java.util.jar.JarEntry;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
+
+import javax.tools.JavaCompiler;
+import javax.tools.ToolProvider;
+
+import org.junit.jupiter.api.Test;
+
+import dev.jbang.util.Util;
+
+import io.qameta.allure.Description;
+
+/**
+ * Regression tests for #2621: `jbang deps search` hangs on Windows because
+ * both jbang.cmd and jbang.ps1 capture the launched Java process's stdout in
+ * order to support the "exit code 255" convention (used by commands like
+ * `run`, which can print a generated native command for the launcher to
+ * re-exec instead of running the JVM path directly).
+ *
+ * That capture is what breaks the interactive `deps search` TUI: with
+ * stdout no longer attached to a real console, Aesh can't detect a native
+ * Windows terminal and the UI never receives keystrokes after its first
+ * render.
+ *
+ * A real interactive terminal isn't available in headless CI, so instead of
+ * testing the TUI's visual behavior, these tests verify the actual
+ * mechanism that was fixed: for `deps search`, the launcher must skip the
+ * capture-and-possibly-re-exec path entirely.
+ *
+ * We prove this with a tiny stub "jbang.jar" that mimics the "print a
+ * command and exit 255" convention. If a launcher wrongly intercepts this
+ * for `deps search`, it will read the printed line and execute it as a new
+ * command, and the final exit code will be that command's exit code (0),
+ * not 255. If the launcher correctly bypasses the capture, the real exit
+ * code (255) from the stub passes straight through untouched.
+ */
+public class DepsSearchLauncherIT extends BaseIT {
+
+    private static final String MARKER_COMMAND = "echo SHOULD_NOT_RUN";
+
+    @Test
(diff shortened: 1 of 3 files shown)


maxandersen:
Thanks for the analysis! We'll need the find another fix though as deps search isn't the only one that could use this.

Bhumika-SN:
Thanks for the feedback, that makes total sense!

I dug into the code a bit more and found that only 3 commands actually
need that exit-255 / re-exec trick: `run`, `app setup`, and
`jdk java-env`. Everything else, including `deps search`, doesn't use
it at all.

So my plan is to flip the approach — instead of special-casing
`deps search`, treat those 3 as the exceptions that need the
capture behavior, and let everything else pass stdout straight
through by default. That way it should just work for `deps search`
and any other interactive command down the line too, without needing
to hardcode anything.

Let me know if that sounds right or if you were thinking of a
different approach — happy to adjust before I dive in!

quintesse:
I think what you're proposing is going to be really hard, een your current solution doesn't take into account all the kinds of different ways a user could be writing the command (eg ` jbang --offline deps --verbose search`) but the kicker here is the `run` command, where the command name itself is optional (eg. `jbang run myalias` and `jbang myalias` are the same).

In the Java code we're jumping through all kinds of hoops to detect this situation and handle it correctly. We'd not only have to duplicate that logic, making the scripts much more complex, but we'd have to de so in the `.cmd` script which doesn't deal well with string handling.

Not saying you shouldn't try, but just be aware that it might be a lot of work to get it right and be ready for us to be really picky and strict about the result, because if it's not perfect it might cause all kinds of nasty parsing issues.

quintesse:
I'm wondering if @stalep has any ideas? Because I know that in the posix world (Unix/Mac) it doesn't matter that we redirected the output because AEsh just takes over the console. Is there perhap a way we can do the same in Windows? It might even be a special env var or setting where we tell AEsh: yes, there is a console, I promise.

Bhumika-SN:
Ah gotcha, that makes sense now! Thanks for reworking it to cover
the --open/config path too, that's a much more complete fix than
what I had. Let me know if you need anything else from me on this
one.

PR #2632, opened 2026-07-29
feat: add gradle versions plugin to settings.gradle

wfouche:
I've submited a PR for this plugin in the past, and proposed that it be added to the main build.gradle file. The plugin should rather be  added to the settings.gradle file, as implemented by this PR,  and then available dependency updates can be viewed from the command-line by running:

`./gradlew dependencyUpdates`

This is really useful functionality, if you want to quickly check if any dependencies can be updated.


maxandersen:
Wasn't what you can add to your own ~/.gradle/init.d/ ?

I don't use the plugin so not seeing value of keeping uptodate when easily possible to add locally and it works in any Gradle project for you?

wfouche:
No problem.  :-)

PR #2633, opened 2026-07-30
feat: make `app install` fetch fresh resources by default

koppor:
🤖 Copy-pasting `jbang app install --force <app>` did not pick up updated -SNAPSHOT dependencies, because installs reused cached resources. Following the suggestion in the review, `app install` now defaults to fresh resolution while runs stay lazy; `--no-fresh` or `--offline` opts back out.

Test: `jbang app install --force gavsearch@jbangdev` picks up a new snapshot; `--no-fresh` does not.

### DIFF
--- a/docs/modules/ROOT/pages/app-installation.adoc
+++ b/docs/modules/ROOT/pages/app-installation.adoc
@@ -118,8 +118,11 @@ jbang app uninstall mytool
 
 [source,bash]
 ----
-# Reinstall to update
+# Reinstall to update (always fetches fresh sources and dependencies)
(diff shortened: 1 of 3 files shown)


maxandersen:
Not sure I follow - why should we add --fresh here for something that is not using snapshot ?

Always fresh is waste of time 99% of the time...

...and update should happen every 24hrs I believe. 

Wondering if should default to fresh on install ?

Be lazy on runs?

koppor:
> Wondering if should default to fresh on install ?
> 
> Be lazy on runs?

Yeah, good idea :)

PR #2652, opened 2026-08-26
upgrade to aesh 3.17.4

stalep:
Æsh 3.17.2 have several improvements and should also start even faster than 3.16.*


maxandersen:
It's seems it's also better at breaking the build :)

stalep:
> It's seems it's also better at breaking the build :)

Hopefully should be fixed now :)

Issue #2651, opened 2026-08-26
Make installing different distro JDKs easier


quintesse:
Right now if you try to install a JDK with the same major version as an already installed JDK you get a message saying that version X is already installed and that you can use `--force` to install anyway.

This is still a hold-over from the time we could only install a single major version.

But if you do what it says, eg: `jbang jdk install --jdk-distros=mandrel --force 25` then it will actually **replace** the existing installed JDK, which is most likely not what the user intended.

Installing a second JDK of the same version is possible but the steps are not intuitive:

1. `jbang jdk list --jdk-distros=mandrel -a`
2. Choose an ID from the list
3. `jbang jdk install <id>`

We should make this easier.

- Don't require the use of `--force` but simply install the new JDK. But we should print some messages explaining what happened. We should also print the ID and an example command on how to make this new JDK the default. (Perhaps in some future we could use a TUI)
- Make IDs easier to use by allowing partial matches (eg `25-mandrel` will match `25-mandrel-jbang`)
- Allow special syntax for selecting JDKs (eg, `mandrel-25` would automatically apply `--jdk-distros=mandrel`)

We should probably open separate issues for the last two points, but I want to document the ideas at least

maxandersen:
I feel we had issue some discussing how //JAVA would work for this ?

mandrel:25
mandrel:25+

rather than mandrel-25.

I,e. `[distro:]<version>[+]`

quintesse:
> I feel we had issue some discussing how //JAVA would work for this ?

Thought so too, but couldn't find it

> I,e. [distro:][+]

Sure. I don't care about the syntax.

quintesse:
So I got a PR for the first point (#2656) but while looking into implementing the `[distro:]<version>[+]` point I realized this is far from easy to do. Basically because JDKs don't actually officially publicize/know their distribution name.

We could potentially come up with something where we add this information to JDKs installed by us (an extra file? an extra entry in `release`?) but this has two issues: 1) it won't work for JDKs installed by older versions of JBang and 2) it won't work for JDKs installed by other means (sdkman, system installers, manually linked etc).

Another option would be heuristics: look for distribution names in the JDK archive name or in its `release` file. This will probably work for many but definitely not all cases. (eg. JDKs installed by system installers)

[...]

PR #2656, opened 2026-09-02
feat: installing same version JDKs is now easier

quintesse:
The `--force` options is now only necessary when trying to re-install the exact same JDK (when it has the same id).

See #2651


quintesse:
@maxandersen I think this can be merged safely

maxandersen:
something don't feel right:

```
jbang jdk list
Installed JDKs (<=default):
   28 (28-beta+3-ea)
   27 (27+35-2325)
   26 (26.0.1+8) <
   25 (25.0.2+10-jvmci-b01)
   24 (24.0.2+12)
   23 (23.0.2+7)
   22 (22.0.2+9)
   21 (21.0.11+10-LTS)
   20 (20.0.2+9)
   19 (19.0.2)
   18 (18.0.2.1)
   17 (17.0.19+10)
   11 (11.0.31+11)
  ```
  
  now install jdk 11
  
```
jbang jdk install 11
[jbang] Downloading JDK 11. Be patient, this can take several minutes...
[jbang] Installing JDK 11...
[jbang] JDK 11.0.32.1+1 installed successfully
[jbang] This is an additional JDK for major version 11 with id '11.0.32.1+1-temurin-jbang'
[jbang] To use it you supply its id when selecting a Java version, for example: 'jbang run --java 11.0.32.1+1-temurin-jbang ...'
[jbang] To make it the default for its major version run 'jbang jdk default -v 11 11.0.32.1+1-temurin-jbang'
[jbang] To make it the global Java default run 'jbang jdk default 11.0.32.1+1-temurin-jbang'
```

is it expected it will install latest jdk 11 when 11 is already available ? and if it is ...why not make it the default ?

quintesse:
> is it expected it will install latest jdk 11 when 11 is already available ?

Well that was the whole idea about making this easier, right?
That you can install multiple 11s, either because you have selected a different distro/vendor or a different provider or even simply because there's a newer build.
This allows `jbang jdk install --jdk-distros=zulu 11`

> and if it is ...why not make it the default ?

that's just a choice I made. It seemed there would be more cases where you are simply installing alternatives than installing a new default. And giving instructions on how to set this new one as a default seemed less confusing than giving instructions on how to _undo_ the change back to the old default.

PR #2660, opened 2026-09-20
fix: resolve native binaries from local jbang directories

### LINKED ISSUE #2624: Native command handling does not take `./.jbang` folder into account
We don't look for the binary in the `.jbang` sub-folder if it couldn't be found in the current directory.

What needs to be fixed:

- the startup scripts need to follow the same checks as what we do for JARs, first local folder, then `.jbang` folder in current directory then `~/.jbang/bin`.
- The `App.copyJBangFiles()` should check for binaries in the `.jbang` folder too, just like we do for JARs
- Think what we want to do with the `wrapper` command: should it copy binaries? If we copy only the current one then only people with the same arch/os combination would see the benefit. Should we perhaps add it as an explicit flag? Eg `--include-native`? Perhaps with an optional list of arch/os names to copy? Perhaps open a separate issue for this?

arnabnandy7:
Fixes #2624.

Native command handling now uses the same location precedence as JAR handling:

1. The startup script directory
2. The local `.jbang` directory
3. `$JBANG_DIR/bin`

Within each directory, platform-specific native binaries are preferred over generic binaries. When no native binary is available, startup continues to fall back to `jbang.jar`.

This change:

- Updates the Bash, CMD, and PowerShell startup scripts.
- Updates `App.copyJBangFiles()` to copy native binaries from the source directory or its `.jbang` directory.
- Preserves source-directory precedence when the same binary exists in both locations.
- Documents native binary lookup behavior in `installation.adoc`.
- Adds tests for Bash, CMD, PowerShell, and application installation copying.

Verification:

- `./gradlew spotlessApply`
- `./gradlew test --tests "dev.jbang.cli.TestApp" --tests "dev.jbang.cli.TestScriptNativeLookup"`


docs/modules/ROOT/pages/installation.adoc
```diff
@@ -94,6 +94,8 @@ that will run out of that directory using `./jbang`.
 
 The `./.jbang` directory which `jbang wrapper install` creates is just a cache which you typically would not commit to a source code repository, so you can e.g. `echo .jbang/ >>.gitignore`.
 
+When `JBANG_USE_NATIVE=true`, the wrapper searches for a matching native binary beside the wrapper scripts, then in `./.jbang`, and finally in `$JBANG_DIR/bin`. If no native binary is available, it falls back to `jbang.jar`.
+
```
maxandersen, line 98: We need to be careful here. If the user runs ./jbang and that is version 0.123 then it would be bad if it goes and run ~/.Jbang/bin/jbang.bin which is a different version. 

Reverse true too - if you run Jbang from path i wouldn't expect Jbang to pick up native binary in my local dir.

quintesse: > If the user runs ./jbang

But that's what we do now. It will pick the one that's located next to the script, but if that one doesn't exist it will fall back to any installed version of JBang.

> Reverse true too

Indeed, that should not happen. Your install would be broken (it would mean that you have a `~/.jbang/bin/jbang` but no `~/.jbang\bin\jbang.jar` which should not happen)

Btw, unrelated to this PR, but I've never read the line above the changes:

> which you typically would not commit to a source code repository

THis is not true, you _would_ actually commit this! Just like you'd do with gradlew and mvnw wrappers

arnabnandy7: @maxandersen @quintesse with the current lookup order, a local wrapper that has its own `.jbang/jbang.jar` could still pick up a native binary from the global installation, potentially running a different JBang version.

I think the safer order is:

1. Look for a native binary beside the invoked launcher.
2. Look in the launcher’s `.jbang` directory.
3. Fall back to the launcher-local JAR.
4. Only when none of those exist, delegate to `$JBANG_DIR/bin/jbang`, which can resolve its own matching native binary or JAR.

This also means that invoking the globally installed launcher won’t inspect the project’s local `.jbang` directory, since lookup remains relative to the invoked script rather than the current working directory.

I’ll update the implementation and documentation to follow this order.

